PR #1, opened 2022-08-31
Add patching and ablations features

aVariengien:
I formatted easy_transformer in a python package, and I added features to perform ablation and patching experiments.


neelnanda-io (approved):
This looks _so_ much better than my initial shitty port! Thanks so much :) I'm excited for this to now vaguely look like a real library!

PR #5, opened 2022-09-09
Fixed bug in backward hooks

neelnanda-io:


neelnanda-io:
Also fixed bug in generated hook_dict (with a very, but less, janky fix)

PR #8, opened 2022-09-13
More activation functions, including solu

anshradh:


ArthurConmy:
I was just merging this in, and this:
Screenshot from 2022-09-16 14-22-26
makes me a bit sad, as I have to think about 4 different activation functions in the otherwise really simple and clean file. Can all the activation functions be made modules instead? Happy to work on this if there's agreement

PR #9, opened 2022-09-13
Example training script for autoregressive lm task

anshradh:


neelnanda-io:
Fixed a few bugs, and added a tokenization utility. Added a script to demo that trains a tiny GPT on 2000 samples from OpenWebText

PR #10, opened 2022-09-13
Cache and gen

anshradh:


anshradh:
This has working implementations of sampling (with various modifications to logits, i.e. temperature, top-k, top-p, etc.) and greedy decoding, both with and without caching past keys and values (cache offers 3-4x speedup). These are also all batched. Will put up a PR for beam search next

PR #12, opened 2022-09-16
Modify notebook to show loading custom config/model

anshradh:


anshradh:
This is unnecessary, didn't realize it was done at the end of the demo notebook

PR #13, opened 2022-09-16
Cache and gen

anshradh:
This has working implementations of sampling (with various modifications to logits, i.e. temperature, top-k, top-p, etc.) and greedy decoding, both with and without caching past keys and values (cache offers 3-4x speedup). These are also all batched. Will put up a PR for beam search next

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,13 @@
 __pycache__
+Testing_Notebook.ipynb
+\.*
+easy_transformer/scratch.py
+wandb*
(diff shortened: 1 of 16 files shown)


neelnanda-io:
Thanks and sorry for taking so long on this! Did a bunch of refactoring and integration into my new changes, and just merged in :)

anshradh:
No worries and sorry for not keeping it more up-to-date!

PR #15, opened 2022-09-16
TODO Arthur: add the new metric feature, that doesn't require `List[str]`

ArthurConmy:
TODO TODO TODO


neelnanda-io:
Argh there are so many commits in here

ArthurConmy:
@neelnanda-io   sorry, at RR we use draft PRs to prepare features. I imagine you got some notification and it's annoying, so I've closed this and will use different reminders for myself

PR #17, opened 2022-09-26
Fix bug in bwd hooks

seanwentzel:
They were sometimes getting registered as fwd.

I didn't try running this (just noticed it on GH), but seems obviously right.


neelnanda-io:
Lol, nice spot, thanks!

PR #29, opened 2022-10-04
Delete the capitalized version of the module

aslvrstn:
https://github.com/neelnanda-io/Easy-Transformer/pull/28/files introduced a capitalized version of Easy_Transformer/__init__.py which causes big issues on macs (which doesn't have a case-sensitive filesystem). I don't think this second file is necessary for anything.


neelnanda-io:
Thanks! I was playing around with different ways to generate libraries + docs automatically, and think one of them must have done it?

PR #34, opened 2022-10-20
Evals

neelnanda-io:


neelnanda-io:
A PR to both do solu loading and to add basic model evals - sorry fo rbad practice of 1 PR for two things!

PR #41, opened 2022-10-21
add reference for composition scores

ArthurConmy:
AFAICT there are no definitions of composition score even in Transformer Circuits. So this provides a link.


neelnanda-io:
Thanks!

PR #44, opened 2022-10-25
Cleanup markdown

aslvrstn:
Makes this thing a lot more readable

See it at https://github.com/aslvrstn/Easy-Transformer/blob/main/further_comments.md


aslvrstn:
Also should
> The positional embeddings are added to the residual stream in the standard way, and then the queries and keys are calculated as W_Q(res_stream + pos_embed) and W_K(res_stream + pos_embed).

read "the positional embeddings are _not_ added" ?

neelnanda-io:
Thanks a lot for flagging! Yeah, looks much better, and all those corrections are right

PR #57, opened 2022-11-03
Tensortyping! 

smithjessk:
# Overview

Tensortyping is a library that dimension-checks tensors. 

## Reviewing

Probably review/merge add-black-formatter first.

## Progress

I think this PR probably doesn't cover literally every function that takes a tensor, but it's working in the right direction! (And almost done?)


neelnanda-io:
Went through and added a bunch more type annotations. Thanks for taking point on this, I think the code is _way_ more readable now! And I love his library <3

PR #66, opened 2022-11-07
Add plotly

poppingtonic:
I got a `ModuleNotFoundError: No module named 'plotly'` error when first running `pip install`, this fixes it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,5 +19,6 @@ setup(
         "wandb",
         "fancy_einsum",
         "torchtyping",
+        "plotly",
     ],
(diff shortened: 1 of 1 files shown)


neelnanda-io:
Ah, sorry! That was caused by an unneeded import in experiments.py, I've deleted that line in my latest push to main. Do you still get this issue? Closing this pull request for now - I don't currently think the library should rely on Plotly (but I may add this in future)

poppingtonic:
I don't get the issue anymore, thanks

PR #72, opened 2022-11-15
Eliminate sometimes-broken caches

aslvrstn:
These caches cause problems if doing something like inspect W_Q, train, inspect again. I'm not sure how critical they are for performance on large models. If they're mission critical, maybe I'll put in the work to do proper cache invalidation rather than eliminate them entirely.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ core.py
 *.yml
 nbs
 _modidx.py
+.ipynb_checkpoints
+env


neelnanda-io:
Ah, hmm, I did not foresee that problem

neelnanda-io:
The reason I have the caches is that you're stacking the layers for convenience, which basically creates a _new_ weight matrix. Which is terrible if you do it a bunch (both for performance, and because autodiff can easily pollute your GPU memory with leftover matrices). 

And I don't want to actually calculate them, because that will 2x the GPU memory footprint which fucks you for large models

The library is mostly designed for inference of pretrained models rather than training, so I'm OK with this being a bug, but seems not ideal. (This is easy to patch by just using the weights per layer fwiw, this exists solely for convenience)

neelnanda-io:
The main solution I see is to have a function for "reset all caches". Maybe have a "calculate stacked weights" function which is run the first time a weight matrix is queried, and otherwise cached, but the user can manually re-run it?

Is there a way to reset the lru_cache? 

I'm not aware of a way to automatically invalidate the cache if any model parameters change.

neelnanda-io:
The cache on the SVD of FactoredMatrix is less essential. What are you using this for?

aslvrstn:
The SVD one wasn't actually biting me, it was just another potential
correctness issue I figured I'd get at.

I'm finally (finally!) picking back up the composition-during-training work.

neelnanda-io:
What's the right way to cache? lru_cache is the simplest decorator I found (there's cache iun Python 3.8 onwards, but the system I wrote this on used Python 3.7, and I don't generally want to tie the library to later versions if possible)

aslvrstn:
Yea, lru_cache is totally fine for this, I was more commenting that it's not doing _that_ much for you in this case given that you don't even need the dict to store multiple values, so just manually caching is totally viable.

FWIW there's also https://docs.python.org/3/library/functools.html#functools.cached_property which I just learned about today, but it's also 3.8+ and doesn't actually fix this, it's just yet another thing

aslvrstn:
Alright, I tried a bunch of different ways to try to solve this in a way that pleased me, but without a pretty serious overhaul (like making all the Parameters actually @property methods that I hook into to clear the cache), I can't get it to be Actually Correct, and rather than just make it more sneakily wrong, I'm just going to leave it alone.

My inclination would be to just drop the cache, especially if the situation is that you can't use these on big models anyway because of memory constraints, and small models it's fine to just re-run the stacking. But I am obviously a very very weird and minor user of this thing! Happy to hop on a call or chat somewhere if some brainstorming would be useful.

aslvrstn:
Oh, and for context, I ended up working around this by yea, just using the individual layers and mathing it myself: https://github.com/aslvrstn/mlscratchpad/commit/f1537c9ce2e2fe99f531d0738a3431d8e7ea74b4

neelnanda-io:
Fair! My guess is that the most elegant way is to have a "clear all parameter caches" method on EasyTransformer that just runs `self.W_Q.clear_cache()`. Though, hmm, this works on the *function* but doesn't obviously work when it's a property...

To be clear, the issue with rerunning it isn't the time costs of restacking, it's that each new stacked objects occupies a new block of memory! So if it isn't freed up, this overloads your GPU memory rapidly. And if autodiff is on, it's very easy to shoot yourself in the foot and prevent memory being freed up.

aslvrstn:
Yea, manually clearing the cache work, it's just Bad Form to make callers have to know to do that

I wonder if tensor equality on gigabyte-sized tensors is fast enough that we could check whether the cached_W_V[n] == self.block[n].attn.W_V for all n

neelnanda-io:
Lol. That *would *avoid the problem of memory blow up, but feels so
wasteful...

Yeah, I was imagining a exposed method called reset_stacked_params which
clears the caches for the user. And flagging in the docs that you need to
do this if training the model. Which I think 80/20s this issue

Issue #74, opened 2022-11-20
`prepend_bos` inconsistency


joelburget:
Yesterday @epurdy and I were working on an implementation of causal tracing from ROME. We ended up getting stuck for a while because when we ran the model it unexpectedly prepended a BOS token (in contrast to running the tokenizer by itself or with `model.to_str_tokens`, which doesn't). Looking through the codebase I noticed that of the six functions with a `prepend_bos` option, they're evenly split on the default. I think these defaults make sense in isolation but may be confusing when used together. Not sure what the right fix is here (or if it needs to be changed at all) but thought I'd at least raise this for discussion.

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/EasyTransformer.py#L146

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/EasyTransformer.py#L299

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/EasyTransformer.py#L345

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/EasyTransformer.py#L395

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/EasyTransformer.py#L935

https://github.com/neelnanda-io/Easy-Transformer/blob/e5790469df3ebe26547c2016f899389e9f7bec30/easy_transformer/utils.py#L440

neelnanda-io:
Thanks for flagging! You guys being stuck for a while is useful data (and sorry about that!).

I've gone back and forth about the right way to do this. So, roughly, here's my reasoning:
* Some models are trained with a BOS token (OPT and all of my models) and will completely break without this at the start of the input)
* SOme models were not trained with a BOS token (GPT-2 and probs GPT-Neo) but in practice tend to treat the first token weirdly and empirically prepending BOS makes them work better. Eg "John and Mary went to the store..." will tend to get much worse results on the IOI task than "<BOS>John and Mary went to the store..." or "After John and Mary went to the store..."
* I would like the default to be that all *inputs* to the model have BOS prepended
* I would like the default that when the user is analysing any *substring* BOS is not prepended

[...]

PR #75, opened 2022-11-20
Changed prepend_bos to always default to True

neelnanda-io:
Fixing Issue #74 - thanks @joelburget ! prepend_bos now always defaults to True, along with comments trying to explain the gotcha. 

This comes at the cost that if tokenizing a substring (eg trying to tokenize someone's name) a BOS token will be prepended automatically, and needs to be explicitly turned off.

(Also fixes the loss spikes in the No_Position_Experiment notebook, and adds an error if using `get_token_position` when the token is not present)


neelnanda-io:
Also fixes a bug in refactor_factored_attn_matrices so that W_O is actually just a rotation.

PR #78, opened 2022-11-28
Fix opt tokenizer

ArthurConmy:
Previously, loading in any OPT model and doing `model.to_token("Hello")` would give the tensor `[2, 2, 5625]`. This fixes that change

### DIFF
--- a/easy_transformer/EasyTransformer.py
+++ b/easy_transformer/EasyTransformer.py
@@ -82,7 +82,8 @@ class EasyTransformer(HookedRootModule):
                 self.tokenizer.pad_token = self.tokenizer.eos_token
             if self.tokenizer.bos_token is None:
                 self.tokenizer.bos_token = self.tokenizer.eos_token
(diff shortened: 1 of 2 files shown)


easy_transformer/EasyTransformer.py
```diff
@@ -82,6 +82,8 @@ def __init__(
                 self.tokenizer.pad_token = self.tokenizer.eos_token
             if self.tokenizer.bos_token is None:
                 self.tokenizer.bos_token = self.tokenizer.eos_token
+            if isinstance(self.tokenizer, PreTrainedTokenizer):
```
ArthurConmy, line 85: I imagine the error arises since the tokenizer is not a PreTrainedTokenizerFast, which has different syntax

neelnanda-io:
Thanks! I changed the condition to be more OPT specific and added a comment, but otherwise looks good. I appreciate the bug fix!

PR #80, opened 2022-11-29
Fix broken demo notebook

ArthurConmy:
This PR deletes
1) the import statements for the experiments.py file deleted a while back, and 
2) the two sections that relied on this.

The edits below are quite hard to read, probably here https://colab.research.google.com/github/neelnanda-io/Easy-Transformer/blob/fix_broken_notebook/EasyTransformer_Demo.ipynb is more readable

Thanks to @Andy-Zhou-wz337 for finding this when trying to learn the library


ArthurConmy:
Also, this thing could be caught by some tests$^1$, is https://github.com/neelnanda-io/Easy-Transformer/pull/68 being implemented?

$^1$: this is particularly easy when notebooks are generated from `notebook.py` files with `#%%`: then, you can just use `def test_notebook(): import notebook` as the whole test that there are no crashes!

neelnanda-io:
Thanks! I really need to write that new demo...

neelnanda-io:
Interesting re notebooks, any idea how hard it is to set up the auto-generation? Eg could we get it to auto-generate a demo on commit?

ArthurConmy:
This https://code.visualstudio.com/docs/python/jupyter-support-py#:~:text=Jupyter%3A%20Export%20Current%20Python%20File%20as%20Jupyter%20Notebook does it in VSCode @neelnanda-io . I'm sure there's a way to automate - shall we make an issue for this?

neelnanda-io:
Go for it

PR #81, opened 2022-11-29
Use AdamW optimizer by default.

joelburget:
There's a quote in this very file, "Weight decay in Adam is implemented badly, so use AdamW instead", so I figured we ought to change the default.


neelnanda-io:
I'm not even convinced we should still have that file lol (or at least, it should be a Jupyter notebook that's maybe auto generated from a Python file). But given that we have it, seems like a good change.

PR #82, opened 2022-11-30
EasyBERT :) 

smithjessk:
- PR is into `main` but I based it off of `jsmith-misc-refactors-and-documentation`
- Didn't add many hook points, just `LayerNorm` (the one in `components.LayerNorm`). What's the right approach here? Add more now? Probably, but which? Maybe mirror the ones in `components`?

image


jbloomAus:
I will close this request as it's stale and several months old. If someone else tries to provide BERT support, maybe it will be a useful reference.

PR #85, opened 2022-12-01
added print_tokenized flag to utils.test_prompt

kajetanj:
Use case: I have long prompts that I'm testing on two models. I want to see top k predictions for both, but I don't need to see long tokenized prompt twice.

image

### DIFF
--- a/easy_transformer/utils.py
+++ b/easy_transformer/utils.py
@@ -437,6 +437,7 @@ def test_prompt(
     model,
     prepend_space_to_answer: bool = True,
     print_details: bool = True,
+    print_tokenized: bool = True,
     prepend_bos: bool = True,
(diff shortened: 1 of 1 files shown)


easy_transformer/utils.py
```diff
@@ -455,7 +456,7 @@ def test_prompt(
     answer_str_tokens = model.to_str_tokens(answer, prepend_bos=False)
     prompt_length = len(prompt_str_tokens)
     answer_length = len(answer_str_tokens)
-    if print_details:
+    if print_details and print_tokenized:
```
kajetanj, line 459: Leaving print_details check to make it backward compatible

neelnanda-io:
Looks good, thanks!

PR #86, opened 2022-12-01
utils.test_prompt tokenization bugfix

kajetanj:
It was assumed that tokes(str1+str2) == tokens(str1)+tokens(str2), but this is not always the case
image


neelnanda-io:
Argh I hate tokenization. Thanks for the PR!

PR #87, opened 2022-12-02
Move dependency management to Poetry

alan-cooney:
This will reduce the risk of dependency conflicts & add some useful automations. The downside is that it's slightly less well known (albeit well supported).


neelnanda-io:
I don't fully understand what's going on here, but am happy to defer to you re this being a godo idea!

PR #88, opened 2022-12-02
Reduce torch required version

alan-cooney:
Paperspace installs a much older version by default.


alan-cooney:
@kajetanj has confirmed this fixes the Paperspace install

PR #89, opened 2022-12-03
Support Python 3.7

alan-cooney:
This is used on some older Colab notebooks.


alan-cooney:
Note Pandas must be downgraded to do this without clashes - which is fine as it's only used one place in docs

sheikheddy:
@alan-cooney This broke my python 3.10.10 installation T_T

Had to upgrade pandas again to get it to work.

PR #90, opened 2022-12-07
Rename the package transformer_lens

alan-cooney:
Provides legacy fallback


neelnanda-io:
Ah, I've been making a bunch of these breaking changes on the branch new-demo (the plan was roughly to make a big batch of breaking changes, with what backwards compatibility I can, and then merge into main all in one go). How good have you managed to get the backwards compatibility? Currently for me `import easy_transformer` works but `import easy_transformer.utils` breaks. I think I can probably patch this with making a easy_transformer/utils.py file which runs `from transformer_lens.utils import *`

alan-cooney:
> Ah, I've been making a bunch of these breaking changes on the branch new-demo (the plan was roughly to make a big batch of breaking changes, with what backwards compatibility I can, and then merge into main all in one go). How good have you managed to get the backwards compatibility? Currently for me `import easy_transformer` works but `import easy_transformer.utils` breaks. I think I can probably patch this with making a easy_transformer/utils.py file which runs `from transformer_lens.utils import *`

Ah okay probably best to leave it on your branch then - I'll leave this here in case you want to copy stuff across.

Yes this seemed to work for all imports - see this:

https://github.com/neelnanda-io/TransformerLens/blob/916e04be5eda29ad54570fd41cd6335ce338c4e2/easy_transformer/__init__.py

alan-cooney:
By the way, once this is done just let me know and I'll update the CD to use it & publish

PR #117, opened 2022-12-19
Implement permanent hooks

derpyplops:
Addresses https://github.com/neelnanda-io/TransformerLens/issues/113

Unfortunately, I had some trouble setting up my dev environment so I was unable to confirm if the tests actually pass 😢.

If anyone who can run tests can pull this and check if this is working, that would be great. Otherwise I'll try to come back to this later.

### DIFF
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -39,14 +39,14 @@ class HookPoint(nn.Module):
         # Hook format is fn(activation, hook_name)
         # Change it into PyTorch hook format (this includes input and output,
         # which are the same for a HookPoint)
-
+        name = "_perma" if perma else None
         if dir == "fwd":
 
             def full_hook(module, module_input, module_output):
                 return hook(module_output, hook=self)
 
             handle = self.register_forward_hook(full_hook)
-            name = "_perma" if perma else None
+            
             handle = LensHandle(name, handle)
             self.fwd_hooks.append(handle)
         elif dir == "bwd":
@@ -54,7 +54,6 @@ class HookPoint(nn.Module):
             def full_hook(module, module_input, module_output):
                 return hook(module_output[0], hook=self)
 
-            name = "_perma" if perma else None
             handle = self.register_full_backward_hook(full_hook)
             handle = LensHandle(name, handle)
             self.bwd_hooks.append(handle)
@@ -135,14 +134,17 @@ class HookedRootModule(nn.Module):
         self.remove_all_hook_fns(direction)
         self.is_caching = False
 
-    def add_hook(self, name, hook, dir="fwd"):
+    def add_hook(self, name, hook, dir="fwd", perma=False) -> None:
         if type(name) == str:
-            self.mod_dict[name].add_hook(hook, dir=dir)
+            self.mod_dict[name].add_hook(hook, dir=dir, perma=perma)
         else:
             # Otherwise, name is a Boolean function on names
             for hook_name, hp in self.hook_dict.items():
                 if name(hook_name):
-                    hp.add_hook(hook, dir=dir)
+                    hp.add_hook(hook, dir=dir, perma=perma)
+
+    def add_perma_hook(self, hook, dir="fwd") -> None:
+        self.add_hook(hook, dir=dir, perma=True)
 
     def run_with_hooks(
         self,
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -28,7 +28,7 @@ class HookPoint(nn.Module):
         self.bwd_hooks: List[LensHandle] = []
         self.ctx = {}
 
-        # A var`iable giving the hook's name (from the perspective of the root
+        # A variable giving the hook's name (from the perspective of the root
         # module) - this is set by the root module at setup.
         self.name = None
 
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -149,8 +149,8 @@ class HookedRootModule(nn.Module):
     def run_with_hooks(
         self,
         *model_args,
-        fwd_hooks: List[Tuple[str | Callable, Callable]] = [],
-        bwd_hooks: List[Tuple[str | Callable, Callable]] = [],
+        fwd_hooks: List[Tuple[Union[str, Callable], Callable]] = [],
+        bwd_hooks: List[Tuple[Union[str, Callable], Callable]] = [],
         reset_hooks_end=True,
         clear_contexts=False,
         **model_kwargs,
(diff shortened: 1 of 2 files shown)


neelnanda-io:
Thanks!

neelnanda-io:
When I import this, I get the following error:

```
>>> from neel.imports import *
Traceback (most recent call last):
  File "/workspace/neelutils/neel/imports.py", line 63, in <module>
    from transformer_lens import HookedTransformerConfig, HookedTransformer, FactoredMatrix, ActivationCache
ImportError: cannot import name 'HookedTransformerConfig' from 'transformer_lens' (unknown location)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspace/neelutils/neel/__init__.py", line 6, in <module>
    from .fourier import *
  File "/workspace/neelutils/neel/fourier.py", line 2, in <module>
    from neel.imports import *
  File "/workspace/neelutils/neel/imports.py", line 83, in <module>
    from easy_transformer import EasyTransformerConfig, EasyTransformer, FactoredMatrix, ActivationCache
  File "/workspace/TransformerLens/easy_transformer/__init__.py", line 1, in <module>
    from . import hook_points
  File "/workspace/TransformerLens/easy_transformer/hook_points.py", line 93, in <module>
    class HookedRootModule(nn.Module):
  File "/workspace/TransformerLens/easy_transformer/hook_points.py", line 155, in HookedRootModule
    clear_contexts=False,
TypeError: unsupported operand type(s) for |: 'type' and '_VariadicGenericAlias'
```

neelnanda-io:
I think it's because you use | in your type hints, maybe 3.7 only supports Union? 

Also, mind merging this with main? This branch was built on v1, and the library has since had some breaking changes (sorry!)

derpyplops:
No problem. Rebased on top of your changes. I was told to merge to v2 but I couldn't find that branch, sorry 🤷

neelnanda-io:
Sorry, ignore the stuff about v2! v2 used to be ahead of main, but has since been merged in.

neelnanda-io:
Looks like it's working on some cursory playing around, thanks for writing a test! I don't think I like `group_id` - it's always "_perma" or None, but doesn't look that way in the code (and is called `name` in the add_hooks function), which seems likely to be confusing. Can you change is to a boolean `is_permanent`?

neelnanda-io:
Ditto changing `id_to_remove=None` to `including_permanent` which removes both local _and_ permanent (and by default just local). And adding `including_permanent` to `HookedRootModule.reset_hooks`, since I want the user to keep the API `model.reset_hooks()` to clean up things (rather than needing to manually do it to each hook!)

Your setup makes it easier to later add a feature that allows many hook groups, but I think this feels like overkill at the moment?

neelnanda-io:
Re tests, we're using PyTest, which automatically picks up on functions whose name begins with `test_`. So no need to run the functions at the end of the file - if you rename them to begin with `test_`, then running `pytest` on the CLI at the root directory will automatically pick up on them. I'd change the model from `gpt2` to something small, like `gelu-3l`, just for test running speed.

derpyplops:
Ready for another review. Unfort I was unable to change to another model because I got some weird CUDA error when I tried to change to eg. `gelu-3l`, maybe can fix that in a subsequent commit

transformer_lens/hook_points.py
```diff
@@ -18,45 +27,51 @@
...
+        self.bwd_hooks: List[LensHandle] = []
         self.ctx = {}
 
         # A variable giving the hook's name (from the perspective of the root
         # module) - this is set by the root module at setup.
         self.name = None
 
-    def add_hook(self, hook, dir="fwd"):
+    def add_perma_hook(self, hook, dir="fwd") -> None:
+        self.add_hook(hook, dir=dir, perma=True)
```
neelnanda-io, line 39: Should be is_permanent, not perma

transformer_lens/hook_points.py
```diff
@@ -18,45 +27,51 @@
...
-            self.fwd_hooks = []
-        if (dir == "bwd") or (dir == "both"):
-            for hook in self.bwd_hooks:
-                hook.remove()
-            self.bwd_hooks = []
+    def remove_hooks(self, dir="fwd", including_permanent=False) -> None:
+
+        def _remove_hooks(hooks: list[LensHandle]) -> None:
+            for handle in hooks:
+                if handle.is_permanent or including_permanent:
```
neelnanda-io, line 67: if **not** handle.is_permanent

transformer_lens/hook_points.py
```diff
@@ -18,45 +27,51 @@
...
+                    handle.hook.remove()
+            return [] if including_permanent else [hook for hook in hooks if hook.is_permanent]
```
neelnanda-io, line 69: Seems more elegant to create an `output_hooks=[]` at the start of the function, and add an else statement to the for loop to add to the output list? More robust to bugs, and to changes to the function.

neelnanda-io:
Thanks! Left some minor comments, but overall looks pretty good. Sorry for the pickiness!

neelnanda-io:
OK, this fails the Python 3.7 and 3.8 tests but not 3.9 or 3.10. Looks like a bug with typing
```
FAILED transformer_lens/tests/test_hooks.py::test_hook_attaches_normally - TypeError: 'type' object is not subscriptable
FAILED transformer_lens/tests/test_hooks.py::test_perma_hook_attaches_normally - assert 2 == 1
 +  where 2 = len([LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccd7c50>, is_permanent=False), LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccf7c10>, is_permanent=True)])
 +    where [LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccd7c50>, is_permanent=False), LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccf7c10>, is_permanent=True)] = HookPoint().fwd_hooks
FAILED transformer_lens/tests/test_hooks.py::test_remove_hook - assert 3 == 1
 +  where 3 = len([LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccd7c50>, is_permanent=False), LensHandle(hook=<t..., is_permanent=True), LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693c9cf6d0>, is_permanent=True)])
 +    where [LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693ccd7c50>, is_permanent=False), LensHandle(hook=<t..., is_permanent=True), LensHandle(hook=<torch.utils.hooks.RemovableHandle object at 0x7f693c9cf6d0>, is_permanent=True)] = HookPoint().fwd_hooks
 ```

neelnanda-io:
But otherwise this looks great! Thanks a lot

neelnanda-io:
Fixed the bug and merged, thanks again for the help!

PR #119, opened 2022-12-22
docker image for development

derpyplops:
Installing with poetry didn't work. But this did. I think this is more reproducible than using poetry.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:latest
+
+RUN apt-get -yqq update
+RUN apt-get -yqq install git
+
+RUN apt-get -yqq install python3    
+RUN apt-get -yqq install python3-pip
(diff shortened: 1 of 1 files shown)


derpyplops:
Steps to use:

1. Install docker https://docs.docker.com/get-docker/
2. Do 
```
docker build -t transformer-lens .
docker run -it transformer_lens bash
```
3. To use as a dev environment, ssh-ing into the image with vscode has been pretty nice for me. https://code.visualstudio.com/docs/devcontainers/tutorial

alan-cooney:
@derpyplops this won't work with CUDA unfortunately - you need to use a CUDA specific dockerimage to enable CUDA support

I'll add a PR with my docker image and devcontainer

alan-cooney:
Closing in favour of #122

Issue #105, opened 2022-12-19
Add support for model parallelism


neelnanda-io:
Add support for having a model with layers split across several GPUs.

Make sure the layers (and its HookPoints) know what device they're on, so that hooks can ensure that they aren't needlessly moving information between GPUs. ActivationCache is a dictionary and _should_ work by default.

The MVP would be doing 2 GPUs: putting the embed + first half of layers on GPU 1 and the second half + unembed on GPU 2. This is probably the most that's needed to support eg NeoX? 

I'm not sure of the most elegant way of doing this, or how to do this without making the code really messy. I lean towards either adding a method which edits the model to move layers between devices, or making a separate `ParallelHookedTransformer` class

anshradh:
To be clear: you're looking for support of Pipeline Parallelism, rather than e.g. Tensor Parallelism?

neelnanda-io:
Yeah, pipeline (ie each layer on a single GPU). It seems very important
that an activation is on a single GPU, so hooks can kinda function as
before, just ensuring they act on the correct device. Activations split
across GPUs sounds horrifying. I think this is worth taking a decent
performance hit for.

anshradh:
Yeah agreed tensor parallelism seems pretty cursed.

I'd be interested in taking this on, my proposed implementation would likely involve the user passing the number of devices they want to load the model on and then figuring out the distribution of layers from there (e.g num_devices = 4 for gpt-2-small
-> embed and layers 0-2 on gpu 0, layers 3-5 on gpu 1, layers 6-8 on gpu 2, layers 9-11 and unembed on gpu 3.)

Seems like it would be also nice to support something like gpipe to speed up training (basically just split up minibatches further and do gradient accumulation, should be simple enough).

neelnanda-io:
That API seems pretty reasonable to me! Supporting just a forward pass is
probably fine, and supporting efficient training seems like probably
overkill, until and unless anyone wants to start using this feature for
something hardcore and large scale. TransformerLens is *not *designed to
train 20B models!

jbloomAus:
Tasks required to complete this issue:

- [ ] https://github.com/neelnanda-io/TransformerLens/issues/246

May add more if we decide the model parallelism requires other enhancements not yet obvious.

jbloomAus:
Solved due to #250  solving that last issue.

PR #125, opened 2022-12-24
Start working on pipeline parallelism

anshradh:
Addresses #105


anshradh:
This seems to pass some preliminary tests, was able to load gpt-j-6.7B and things ran just fine. I'll probably test more thoroughly later, but would also appreciate input on design!

transformer_lens/HookedTransformer.py
```diff
@@ -133,8 +140,24 @@ def __init__(
             self.init_weights()
 
         if move_to_device:
-            # Move the model to the relevant device, suppress the logging message
-            self.to(self.cfg.device, print_details=False)
+            if self.cfg.n_devices == 1:
+                # Move the model to the relevant device, suppress the logging message
+                self.to(self.cfg.device, print_details=False)
+            else:
+                # We load the devices in a pipeline manner - the first device gets the embed and pos_embed layers and the first n_layers // n_devices blocks,
```
neelnanda-io, line 147: Huh, I *was* going to object that this would overload the first and last device, for for GPT-J and up, d_model is >=4K, so a block is 12 * d_model >= 50K width, vs 50K width for the embedding, so it's kinda chill

neelnanda-io (commented):
Just read through and this looks great! Seems like a good design approach.

I don't have easy access to a multi GPU device - have you been able to test this? Eg, can you load OPT 66B across 8 GPUs and have things just work?

It'd be great to also have mixed precision in here, if you're working with a large model and doing inference you probably want to be using bfloat16/float16 - no worries if that's too much of a pain though. Do you know if this automatically makes taking gradients work?

anshradh:
I tested loading gpt-j-6b on 7 GPUs and things were fine (was able to generate text and it seemed sensible). I'll try OPT-66-B and let you know how that goes.

One thing that makes me a bit sad is that the current implementation enforces that the number of devices be divisible by the number of transformer blocks - idk if I have an elegant way of making this a not strict requirement and still somehow having a nice balance of layers on each device.

I was also considering keeping the token and position embeddings, final layer norm, and unembed all on the same gpu (probably the rank 0 one) and then splitting the layers across the remaining GPUs. This would be cleaner in a few ways since inputs and outputs would all be on the same device. Lmk what you think of that.

I also like adding mixed precision, I'll try working that in.

transformer_lens/HookedTransformer.py
anshradh: I was also considering keeping the token and position embeddings, final layer norm, and unembed all on the same gpu (probably the rank 0 one) and then splitting the layers across the remaining GPUs. This would be cleaner in a few ways since inputs and outputs would all be on the same device. Lmk what you think of that.

alan-cooney:
Hi @anshradh - nice work on this! The simple approach is great.

A few additional comments:

- Please can you remove the unrelated code changes (e.g. formatting changes) so that the final PR is easier to review? We can add extra linting/formatting to the whole codebase (and this would be great) but it would be better to do this separately.
- The `if n_devices == 1 ... else ...` logic isn’t needed and makes testing hard. The code can be written such that it works for any number of devices (including 1) without these if/else statements. In fact it looks like it mostly works if you just delete the  `if n_devices = 1` blocks.
- Can you also split this logic out into it's own util function/method. This can be tested with pytest for the various use cases (including just one device). We're trying to add tests to the codebase going forwards (I've just recently merged in a CD pipeline to help with this).

anshradh:
Thanks @alan-cooney!

Made the suggested changes - I think I'm a little uncertain if the refactor into a util method was in the form that you were looking for. If not, happy to make further changes. I can also write tests once this looks good!

alan-cooney:
Thanks @anshradh - there still appear to be lots of formatting changes (see https://github.com/neelnanda-io/TransformerLens/pull/125/files ) - are you able to remove these?

anshradh:
Argh sorry about that! My editor's saveOnFormat was a bit trigger happy, even after being disabled ... My bad for also not doublechecking that all of the formatting changes were reverted properly. Things look ok now?

alan-cooney:
> Argh sorry about that! My editor's saveOnFormat was a bit trigger happy, even after being disabled ... My bad for also not doublechecking that all of the formatting changes were reverted properly. Things look ok now?

No worries - thanks for doing that! I think this may also represent a wider thing that we should have some linting config committed to keep editors happy.

transformer_lens/HookedTransformer.py
```diff
@@ -133,9 +139,20 @@ def __init__(
...
+            self.embed.to(utils.get_device_for_block_index(0, self.cfg))
+            self.hook_embed.to(utils.get_device_for_block_index(0, self.cfg))
+            if self.cfg.positional_embedding_type != "rotary":
+                self.pos_embed.to(utils.get_device_for_block_index(0, self.cfg))
+                self.hook_pos_embed.to(utils.get_device_for_block_index(0, self.cfg))
+            if hasattr(self, "ln_final"):
+                self.ln_final.to(utils.get_device_for_block_index(self.cfg.n_layers - 1, self.cfg))
+            self.unembed.to(utils.get_device_for_block_index(self.cfg.n_layers - 1, self.cfg))
+            for i, block in enumerate(self.blocks):
+                block.to(utils.get_device_for_block_index(i, self.cfg))
```
alan-cooney, line 154: When testing this, I suspect you'll want to move all of this into its own class method on `HookedTransformer` as well (but up to you - just depends on how you test it). I think a key ways it could break is if we (a) forget to move one of the modules to a device or (b) move a module to an incorrect device with e.g. an off-by-one error.

This could also keep the init method a little more manageable (it's huge).

transformer_lens/HookedTransformer.py
```diff
@@ -664,9 +696,26 @@ def load_and_process_state_dict(
...
+                if (
+                    k.startswith("ln_final")
+                    or k.startswith("unembed")
+                ):
+                    state_dict[k] = v.to(
+                        utils.get_device_for_block_index(self.cfg.n_layers - 1, self.cfg)
+                    )
+                if k.startswith("blocks"):
+                    state_dict[k] = v.to(utils.get_device_for_block_index(int(k.split(".")[1]), self.cfg))
+
```
alan-cooney, line 718: One small thing is that this should be `if... else if... else if...` as it illuminates otherwise hard-to-catch bugs where multiple conditional blocks are called (again we should have a linter that catches this sort of thing - so I'll look into that). Also it should have a final else statement that either throws an error or puts it on a default device (tending towards the former).

transformer_lens/utils.py
```diff
@@ -576,4 +576,13 @@ def get_dataset(dataset_name: str) -> Dataset:
...
+    return dataset
+
+def get_device_for_block_index(index: int, cfg: HookedTransformerConfig, device: Optional[Union[torch.device, str]] = None):
+    assert cfg.device is not None
+    assert cfg.layers_per_device is not None
+    if device is None:
+        device = cfg.device
+    if isinstance(device, torch.device):
+        device = device.type
+    return torch.device(device, index // cfg.layers_per_device)
```
alan-cooney, line 588: Nice! Yeah something like this (although it doesn't exactly have to be this - just whatever makes sense for testing). I think we can just calculate layers_per_device in this function as well. 

Btw - I think we can move this to its own file rather than continuing this giant one. Bit of a shame we have a utils file rather than folder, but maybe put it in say `transformer_lens/utilities/devices.py`? It's then worth adding a docstring/comments as well.

alan-cooney (dismissed):
Thanks for contributing this - in general it looks really great! I've stuck a few comments in so let me know what you think.

I think the key thing remaining is to write some unit tests such that we're confident this doesn't break anything. One approach that may help here (but not necessarily the best) is to mock out the PyTorch `tensor.to` function so that we can check where everything is being moved.

As a heads up, the CD runs all checks on a CPU-only machine, so all tests will have to work there.

Thanks again for the PR!

anshradh:
I think I'm a little confused about the testing strategy we want here, given that the tests need to run on a CPU-only machine. Would you mind explaining a little further/suggesting some tests that you might write for this? I can obviously test the utility function on various inputs, but I'm not sure I see an avenue for something closer to an integration test involving loading and running a forward pass on a full model.

alan-cooney:
I think testing the utility function alone (esp if it includes this extra logic) should get us to a high degree of confidence, so pragmatically this may be enough. 

I think we should also have an integration test mocking out torch.to(device), but I'm not sure exactly how to do this with PyTorch. Happy to pull that out as a later task therefore. As long as we have good unit test coverage + manual tests with multiple GPUs, I think it should be enough for now.

neelnanda-io:
Curious what the state of this PR is?

anshradh:
Ah, sorry about this - been busy with REMIX and interviews. I think I should definitely be able to wrap this up by end of the month (probably shouldn't be much more than a few hours of work), though someone should feel free to finish it off if they'd like to. I think the only remaining thing is to implement tests, though as Alan and I discussed above, I think this is slightly nontrivial, though not super complicated.

neelnanda-io:
No rush! Just wanted to check it wasn't blocked on me lol

alan-cooney:
Also happy to pairs code on writing the tests next week if you want

jbloomAus:
@anshradh @alan-cooney how is this PR going? Anything I can do to help?

anshradh:
Ah this is my fault, completely dropped the ball 😔. I don't think I have bandwidth to get it over the finish line, but anyone should feel free to pick it up - I think basically all that's left is writing the tests that @alan-cooney wanted. 

Sorry afor leaving this unfinished!

jbloomAus:
Tasks to do (discussion between me and @rusheb )
- [ ] test the utility function on various inputs
- [ ] Work out if it's possible to have an integration test mocking out torch.to(device)
- [ ] either way, write a manual test script which can be be run semi-regularly on a machine with multiple GPUs (something closer to an integration test involving loading and running a forward pass on a full model such as GPT2)

rusheb:
I can try to get this over the line!

@jbloomAus are you happy to give me write access so I can push to @anshradh's branch?

If not, I can clone the branch onto my fork and raise a new PR.

Let me know!

jbloomAus:
@rusheb, I don't think this is possible without @anshradh's help, I'll ask them.

anshradh:
I think you'd just have to add @rusheb as a collaborator?

jbloomAus:
I've pinged @neelnanda-io about this since I don't have the ability to do that :)

rusheb:
Hi, I got a last-minute invitation to an ML bootcamp which I will be working on full time for the next 2 weeks. If this PR is still open when I return then I can pick it back up then!

jbloomAus:
I spent a bunch of time trying to "mock" GPUs for testing yesterday to make testing of the multi-GPU code more frequent. That turned out to be hard, and there weren't any apparent solutions online. I will probably spin up some GPUs and write bespoke test scripts that we can run before releases.

transformer_lens/HookedTransformerConfig.py
```diff
@@ -71,13 +71,16 @@ class HookedTransformerConfig:
             & biases) and 'LNPre' (use LayerNorm, but no weights & biases).
             Defaults to LN
         device(str): The device to use for the model. Defaults to 'cuda' if
-            available, else 'cpu
+            available, else 'cpu'. Must be 'cuda' if `n_devices` > 1.
+        n_devices (int): The number of devices to use for the model. Defaults to 1. Layers are loaded
+            to support "pipeline parallelism", where each device is responsible for a subset of the layers.
+        layers_per_device (int, *optional*): The number of layers to load on each device. Defaults to None and is automatically calculated.
```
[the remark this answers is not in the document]
jbloomAus: the requested change was made.

jbloomAus:
Alrighty so an update:
1. I attempted "Work out if it's possible to have an integration test mocking out torch.to(device)" which was HARD. No clear answers online such as on the pytorch forums. I gave it several hours and discovered some limitations of GPT4/maybe just python. It's hard to mock a complex type signature like the method `.to` in pytorch because it sometimes gets passed a tensor and sometime just a device. I couldn't always retrieve the original tensor making it hard to mock the output. 
2. `test the utility function on various inputs` I assume this is `test_get_device_for_block_index` which I have written a test for. I don't think there's a huge amount of complexity to test. Maybe raising an error if the inputs are wrong is something we should add/test but I'm not super concerned by it. 
3. I've written a test for n_devices varying which only activates if the computer has 4 gpu devices. This test checks the following:
  - The code runs for 1 <= n_devices <= 4. 
  - The logits/activation cache matches for 1 <= n_devices <= 4. 
  - The devices are assigned a roughly equal proportion of layers in the model. 
  
  I think these tests are likely sufficient. We can run them pretty easily before releases and people working with multiple GPUs can also run them after they install the package. 
  
@alan-cooney @neelnanda-io I think it should be fine to merge this PR. I'll wait a day or two in case you want to comment/suggest modifications to the tests but otherwise will merge soon :)

neelnanda-io:
> I attempted "Work out if it's possible to have an integration test
mocking out torch.to(device)" which was HARD. No clear answers online such
as on the pytorch forums. I gave it several hours and discovered some
limitations of GPT4/maybe just python. It's hard to mock a complex type
signature like the method .to in pytorch because it sometimes gets passed a
tensor and sometime just a device. I couldn't always retrieve the original
tensor making it hard to mock the output.

Dumb question - why can't you mock it out with a function that checks
whether the input is a string, device or tensor, and deals with it
differently accordingly?

But yeah, looks decent.

My one concern is that, as is, there's not a way to have a cache keep
tensors on device, or have a hook layer know what device it's on (I
think?). IMO this is a fairly important feature to add to run_with_cache,
if it's not too big a headache, because communicating large tensors between
devices can be a major performance overhead. I would implement it by giving
each HookPoint a `self.device` parameter, and having `run_with_cache` use
that unless overridden.

jbloomAus:
> Dumb question - why can't you mock it out with a function that checks
> whether the input is a string, device or tensor, and deals with it
> differently accordingly?

You can try this, but if you are only handed a string and need to return the tensor, that tensor will need to be in scope for you to return it. The "self" object is the mock function, so you can't return that. I thought this sounded solvable and proceeded to read the mocking docs, and I didn't find the solution. If someone knows how to ensure a mocked method on a class has the original class (being mocked in scope), then we could return it (aka, return the tensor still on the CPU as if it had been moved), however without it I think we're stuck. 

> My one concern is that, as is, there's not a way to have a cache keep
> tensors on device, or have a hook layer know what device it's on (I
> think?). IMO this is a fairly important feature to add to run_with_cache,
> if it's not too big a headache, because communicating large tensors between
> devices can be a major performance overhead. I would implement it by giving
> each HookPoint a `self.device` parameter, and having `run_with_cache` use
> that unless overridden.

This seems reasonable to me, and possibly quite doable. Might it be reasonable to merge this and make an issue for that one? Reasons to go that way include:
1. Having this is better than nothing for now. 
2. We can direct people toward that card if they have complaints about the overhead.
3. It sounds like a great first issue to give someone.

neelnanda-io:
Sorry for the delay! Yes this seems fine to merge as is with that as a
follow-up PR

jbloomAus:
Thanks Neel. I'll merge this. I've updated both the original issue and created a new proposal for the cache-device-awareness which when completed may enable us to close the original issue. https://github.com/neelnanda-io/TransformerLens/issues/246

PR #136, opened 2023-01-12
callums typing fix

callummcdougall:
Description of my changes to follow. All tests passed.

## ActivationCache

Based on type signatures, the functions `remove_batch_dim` and `to` should return self.

Added some `assert isinstance(layer, int)` to signal to the typechecker that layer is an int, not None.

The args from `get_neuron_results` have type `Union[Slice, SliceInput]`. Even inside the loop `if type(arg) is not Slice`, the typechecker doesn't assume that the arg is a SliceInput, which gives some errors.
This is fixed by adding `assert isinstance(neuron_slice, SliceInput)` (which in turn requires that we define `SliceInput: Type = ... in utils (where Type is imported from typing)).

## HookedTransformer

Python knows the output of `forward` is None or Tensor or Loss, but if you e.g. get logits from a forward pass then try to index into it, this raises an error because Python thinks logits might be None.

Solution - use `typing.overload`, and `typing_extensions.Literal`

Example of how this works - you put this code before the forward function:

```python
@overload
def forward(self, input, return_type: Literal["logits"], **kwargs) -> TT["batch", "pos", "d_vocab"]:
    ...
@overload
def forward(self, input, return_type: Literal["loss"], **kwargs) -> Loss:
    ...
```

Another cool feature of doing this - the typechecker can see if `return_type` equals a string that ISN'T "logits" or "loss", and it throws a red line.

I did a similar thing with `run_with_cache` (which can either return a Dict or ActivationCache).

Lastly, I needed to change some instances of `self.cfg.d_vocab` to `self.cfg._d_vocab`, both here and in `components.py`.

### HookedTransformerConfig

I implemented the -1 thing for configs. It was a little messy (required using properties, and replacing some instances of `d_vocab`, `d_vocab_out` with `_d_vocab`, `_d_vocab_out`) but seems to work fine.


neelnanda-io:
Thanks, this looks great! Conditional on it passing all the tests... (Python 3.7 is a headache with types, I appreciate you using typing_extensions for Literal!)

The d_vocab stuff is mildly horrifying, but I can't see a better solution... @alan-cooney any takes?

callummcdougall:
Yeah I think that is pretty horrible, would be reasonable if you wanted to not include that and I could either add the `model.d_vocab` thing or just add appropriate `# type: ignore`s for the MLAB exercises.

neelnanda-io:
I think I prefer just having the = None default be -1? And cutting out all of the private variable nonsense.

neelnanda-io:
As far as I can tell, the type checker won't be able to catch errors to do with d_vocab accidentally not being set either way?

neelnanda-io:
Hmm, I've pushed some new commits to the main branch of your repo but it's not auto-updating the PR. Weird!

alan-cooney:
Agreed with your thoughts Neel, and nice work on the other overloaded type definitions @callummcdougall !

Have you run `mypy` on TransformerLens to check if this causes any issues? We have a broader issue open to fix typing everywhere ( https://github.com/neelnanda-io/TransformerLens/issues/130 ), but at least for now it's worth making sure we haven't broken anything else.

callummcdougall:
I did run all four tests in the tests directory, and it seemed to pass.

Agree the `d_vocab` thing is super bug prone, apologies for that!

neelnanda-io:
I don't see any new mypy issues

PR #139, opened 2023-01-14
Add ability to set individual head QKV inputs via hook

epurdy:
Disabled by default to save GPU memory.

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -986,6 +986,12 @@ class HookedTransformer(HookedRootModule):
         """
         self.cfg.use_attn_result = use_attn_result
 
+    def set_use_attn_input(self, use_attn_input):
+        """
+        Toggles whether to explicitly calculate and expose the input for each attention head - useful for interpretability but can easily burn through GPU memory.
+        """
+        self.cfg.use_attn_input = use_attn_input
+
     def process_weights_(
         self,
         fold_ln: bool = True,
--- a/transformer_lens/components.py
+++ b/transformer_lens/components.py
@@ -351,10 +351,13 @@ class Attention(nn.Module):
         past_kv_cache_entry is an optional entry of past keys and values for this layer, only relevant if generating text. Defaults to None
 
         """
+        # If user requests it, make a hookpoint with separate inputs for each head.
+        # This allows the user to change the inputs to different heads independently.
         if self.cfg.use_attn_input:
             resid_pre = self.hook_input(
+                # cloning here is important, otherwise it's just a view
                 resid_pre.unsqueeze(2).expand(-1, -1, self.cfg.n_heads, -1).clone()
-            )
+            )  # [batch, pos, head_index, d_model]
 
         if self.cfg.positional_embedding_type in ["standard", "rotary"]:
             # Normal attention
@@ -561,7 +564,8 @@ class Attention(nn.Module):
         TT["batch", "pos", "head_index", "d_head"],
         TT["batch", "pos", "head_index", "d_head"],
     ]:
-        # expand the positional embeddings to have a head dimension
+        # Expand the positional embeddings to have a head dimension.
+        # Cloning here is important, otherwise it's just a view.
         shortformer_pos_embed = shortformer_pos_embed.unsqueeze(2).expand(-1, -1, self.cfg.n_heads, -1).clone()
 
         # We add on the positional encodings to the residual stream JUST for the keys and queries, it's not added to the normal residual stream.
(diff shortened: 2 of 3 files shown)


transformer_lens/HookedTransformer.py
```diff
@@ -986,6 +986,12 @@ def set_use_attn_result(self, use_attn_result):
         """
         self.cfg.use_attn_result = use_attn_result
 
+    def set_use_attn_input(self, use_attn_input):
```
neelnanda-io, line 989: set_use_attn_input_hooks seems like a better name (I should probably change use_attn_result too lol)

transformer_lens/components.py
```diff
@@ -298,7 +301,9 @@ def __init__(
         # See HookedTransformerConfig for more details.
         if self.cfg.positional_embedding_type == "shortformer":
             # This tracks the input to the keys and queries, which is resid_pre + pos_embeds
-            self.hook_attn_input = HookPoint()  # [batch, pos, d_model]
+            # The dimension of these depends on whether cfg.use_attn_input is False (1st) or True (2nd)
+            self.hook_attn_input_q = HookPoint()  # [batch, pos, d_model] OR [batch, pos, head_index, d_model]
```
neelnanda-io, line 305: Oh man, the shortformer bits of this are really ugly. I weakly lean towards just having this feature only apply to non shortformer positional embeddings...?

transformer_lens/components.py
```diff
@@ -350,38 +355,95 @@ def forward(
         past_kv_cache_entry is an optional entry of past keys and values for this layer, only relevant if generating text. Defaults to None
 
         """
+        # If user requests it, make a hookpoint with separate inputs for each head.
+        # This allows the user to change the inputs to different heads independently.
+        if self.cfg.use_attn_input:
+            input_to_q = self.hook_input_q(
+                resid_pre.unsqueeze(2).expand(-1, -1, self.cfg.n_heads, -1).clone()
```
neelnanda-io, line 362: Please rewrite to einops

`einops.repeat(resid_pre, "batch pos d_model -> batch pos head_index d_model", head_index=self.cfg.n_heads)` should work and IMO is much more readable

transformer_lens/components.py
```diff
@@ -287,6 +287,9 @@ def __init__(
         if self.cfg.scale_attn_by_inverse_layer_idx:
             self.attn_scale *= self.layer_id + 1
 
+        self.hook_input_q = HookPoint()  # [batch, pos, head_index, d_model]
```
neelnanda-io, line 290: Add comment documenting them

transformer_lens/components.py
```diff
@@ -350,38 +355,95 @@ def forward(
...
+            input_to_v = self.hook_input_v(
+                resid_pre.unsqueeze(2).expand(-1, -1, self.cfg.n_heads, -1).clone()
+            )
+        else:
+            input_to_q = input_to_k = input_to_v = resid_pre
+
         if self.cfg.positional_embedding_type in ["standard", "rotary"]:
             # Normal attention
-            q = self.hook_q(
+            if self.cfg.use_attn_input:
```
neelnanda-io, line 377: Add comment documenting why this if statement is there

neelnanda-io (changes requested):
Thanks for this! Overall looks fairly good, I left some comments for cosmetic changes.

A galaxy brained idea - you *can* broadcast a tensor up to a larger, repeated shape without changing the total GPU memory used, since it just changes the striding on the same underslying memory (the bonus axis has stride=0). So maybe we should implement it like that, so we don't need to have the branching statements for q, k and v. OTOH this is too clever and will probably break something rip.

But yeah, I'm overall torn between this being a useful feature and making the code too ugly, though overall I lean towards this being worth it.

ArthurConmy:
I implemented this here https://github.com/ArthurConmy/Automatic-Circuit-Discovery in einops. I'm happy to take this if wanted @epurdy ?

ArthurConmy:
Also, I am in favour of the galaxy brain solution. Note this could apply to `use_attn_result` too: we would just broadcast up `attn_out` to have an extra dimension with `einops.repeat`

(and then if we `set_use_attn_results`, we'd clone the tensor so that there wasn't a stride 0 element in there)

I can explain, if this was unclear

epurdy:
> I implemented this here https://github.com/ArthurConmy/Automatic-Circuit-Discovery in einops. I'm happy to take this if wanted @epurdy ?

Sure, @ArthurConmy, that would be great.

neelnanda-io:
Sounds great Arthur, it would be good to include a test that the various
acts are actually stride zero in the expected way, and that you can't use
the hook without setting use_... in the relevant way

ArthurConmy:
See https://github.com/neelnanda-io/TransformerLens/pull/158

PR #141, opened 2023-01-16
let 'get_dataset' function pass kwargs

afspies:
Not sure if you like this - I wanted to specify data_dirs for dataset whilst using this util. Also refactored a bit.

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -1375,11 +1375,13 @@ class HookedTransformer(HookedRootModule):
             for h in range(self.cfg.n_heads)
         ]
 
-    def load_sample_training_dataset(self):
+    def load_sample_training_dataset(self, **kwargs):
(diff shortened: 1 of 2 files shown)


neelnanda-io:
LGTM! And seems much neater, I like the dictionaries - that's a much better idiom than my nested if statements. Thanks! I'll merge once it's passed the tests

neelnanda-io:
Curious what you were using it for?

afspies:
@neelnanda-io Nothing too exciting - Just preparing to try and run a number of experiments across many models without redownloading the datasets on different machines that have access to a shared drive. 

<details>
<summary>...</summary>
(Doing some probing into the representational drift of residual stream from the original set of embeddings - seems hard to make claims about this with only one model / few examples, as many experiments (layer swapping, looking at activations etc. are coupled to the residual stream representation _and_ specialization of layers)

</details>

Issue #130, opened 2022-12-26
Fix type checking (mypy)


alan-cooney:
Currently mypy type checks are not passing - with a large number of easy to resolve errors. 

```bash
poetry run mypy transformer_lens
```

Resolving these would help make the codebase more robust, and the corresponding checks in `/.github/workflows/checks.yml` can then be enabled (they're commented out).

https://github.com/neelnanda-io/TransformerLens/blob/72b9be00c12b5be37baaa16b57e5df5e2aff3aad/.github/workflows/checks.yml#L58

derpyplops:
I tried a bunch to fix it with stub packages but mainly didn't really get anywhere

neelnanda-io:
Hmm, when I run `mypy`, most of what I see seem to be it eg failing to deal with torchtyping?
Eg:
```
transformer_lens/components.py:16: error: Skipping analyzing "fancy_einsum": module is installed, but missing library stubs or py.typed marker  [import]             
transformer_lens/components.py:23: error: Skipping analyzing "torchtyping": module is installed, but missing library stubs or py.typed marker  [import]              
transformer_lens/components.py:37: error: Name "batch" is not defined  [name-defined]                                                                                
transformer_lens/components.py:37: error: Name "position" is not defined  [name-defined]                                                                             
transformer_lens/components.py:38: error: Name "batch" is not defined  [name-defined]                                                                                
transformer_lens/components.py:38: error: Name "position" is not defined  [name-defined]                                                                             
transformer_lens/components.py:38: error: Name "d_model" is not defined  [name-defined]                                                                              
transformer_lens/components.py:57: error: Name "batch" is not defined  [name-defined]                                                                                
transformer_lens/components.py:57: error: Name "position" is not defined  [name-defined]                                                                             
transformer_lens/components.py:57: error: Name "d_model" is not defined  [name-defined]                                                                              
transformer_lens/components.py:58: error: Name "batch" is not defined  [name-defined]                                                                                
transformer_lens/components.py:58: error: Name "position" is not defined  [name-defined]                                                                             
transformer_lens/components.py:58: error: Name "d_vocab_out" is not defined  [name-defined]                                                                          
transformer_lens/components.py:79: error: Name "batch" is not defined  [name-defined]                                                                                
transformer_lens/components.py:79: error: Name "position" is not defined  [name-defined]                                                                             
transformer_lens/components.py:80: error: Name "batch" is not defined  [name-defined]                                                                                
[...]

PR #145, opened 2023-01-17
Add helper enum for torch typing

dkamm:
Related to #130 

Adds a helper enum to get mypy to work with torch typing

I've basically copied every name into the enum, except I've replaced instances of `position` with `pos`

I also opted for lower case instead of upper because it felt more visually consistent

Some thoughts:

- rename `head_index` -> `head`?
- rename `d_vocab` -> `n_vocab`?
- use `head` in place of `n_heads`?
- I can see potential for a proliferation of names in this enum (can cross that bridge when we get there)


neelnanda-io:
Thanks! As ism this fails all the tests,, because `T.batch` is an enum type, not a string. And I don't really get how the add method is supposed to work. But changing it to this seems to pass the tests (I feel like it should subclass _something_, but idk how best to achieve "a class where every attribute is fixed _and_ is a string rather than smth weird)

```
from enum import Enum

class T():
    """Helper class to get mypy to work with TorchTyping and solidify naming conventions as a byproduct.
    
    Examples: 
    - `TT[T.batch, T.pos, T.d_model]`
    - `TT[T.num_components, T.batch_and_pos_dims:...]`
    - `TT[T.d_vocab + T.n_ctx, T.d_model]`
    """

    batch: str = "batch"
    pos: str = "pos"
    head_index: str = "head_index"
    length: str = "length"
    rotary_dim: str = "rotary_dim"
    new_tokens: str = "new_tokens"
    batch_and_pos_dims: str = "batch_and_pos_dims"
    layers_accumulated_over: str = "layers_accumulated_over"
    layers_covered: str = "layers_covered"
    past_kv_pos_offset: str = "past_kv_pos_offset"
    num_components: str = "num_components"
    num_neurons: str = "num_neurons"
    pos_so_far: str = "pos_so_far"
    n_ctx: str = "n_ctx"
    n_heads: str = "n_heads"
    n_layers: str = "n_layers"
    d_vocab: str = "d_vocab"
    d_vocab_out: str = "d_vocab_out"
    d_head: str = "d_head"
    d_mlp: str = "d_mlp"
    d_model: str = "d_model"

    ldim: str = "ldim"
    rdim: str = "rdim"
    new_rdim: str = "new_rdim"
    mdim: str = "mdim"
    leading_dims: str = "leading_dims"
    leading_dims_left: str = "leading_dims_left"
    leading_dims_right: str = "leading_dims_right"

    a: str = "a"
    b: str = "b"
```

neelnanda-io (changes requested):
See above

dkamm:
Oops sorry about that!  I only checked the mypy command.

I think another way to get something like that is to have a singleton with `@dataclass(frozen=True)`?  But I just tried this and it didn't work with mypy unfortunately :(

I'll update the PR to what you have- think it's the best we can do right now

dkamm:
I think it makes sense to add a `T.sum` method for the shapes with addition like `"d_vocab + n_ctx"`.  Writing `T.d_vocab + "+" + T.n_ctx` seems a bit more awkward and error prone (might forget the `"+"`).  It's possible that torchtyping will support checking arithmetic in tensor shapes in the future (see https://github.com/patrick-kidger/torchtyping/issues/2).  This is also why I had the weird `__add__` thing in the enum version

neelnanda-io (approved):
LGTM pending tests!

neelnanda-io:
Looks good, thanks!

neelnanda-io:
Still failing tests rip `TypeError: type of argument "args"[0] must be a tuple; got str instead` - my guess is that the sum is breaking something? At least in Python 3.7. IMO the sum is not essential, and we can just hardcode the one or two combination strings as a bonus string in the big enum

PR #147, opened 2023-01-18
add hook tokens

callummcdougall:
Added `hook_tokens` option to the config object (default value is `False`). If this is true, then we call:

```python
tokens = self.hook_tokens(tokens)
```

before applying the embedding, in our HookedTransformer's forward pass.

---

The files that were edited were:

* `HookedTransformerConfig` (to add `hook_tokens` as a default argument, and add a docstring to the config object)
* `HookedTransformer` (in 2 places - once to define the hooks, and once to apply them if they exist)

---

This is useful when we combine it with permanent hooks to e.g. implement certain types of model architecture. Here is an example of how we can use this feature to turn a bidirectional transformer into one which applies a mask to the padding tokens at each attention layer, in the style of BERT:

```python
def add_perma_hooks_to_mask_pad_tokens(model: HookedTransformer, pad_token: int) -> model:

    # Hook which operates on the tokens, and stores a mask where tokens equal [pad]
    def cache_padding_tokens_mask(
        tokens: TT["batch", "seq"],
        hook: HookPoint,
    ) -> None:
        hook.ctx["padding_tokens_mask"] = einops.rearrange(tokens == pad_token, "b sK -> b 1 1 sK")

    # Apply masking, by referencing the mask stored in the `hook_tokens` hook context
    # If this attention layer is the last one, then clear the information from the `hook_tokens` context
    def apply_padding_tokens_mask(
        attn_scores: TT["batch", "head", "seq_Q", "seq_K"],
        hook: HookPoint,
    ) -> None:
        attn_scores.masked_fill_(model.hook_dict["hook_tokens"].ctx["padding_tokens_mask"], -1e5)
        if hook.layer() == model.cfg.n_layers - 1:
            del model.hook_dict["hook_tokens"].ctx["padding_tokens_mask"]

    # Add these hooks as permanent hooks (i.e. they aren't removed after functions like run_with_hooks)
    for name, hook in model.hook_dict.items():
        if name == "hook_tokens":
            hook.add_perma_hook(cache_padding_tokens_mask)
        elif name.endswith("attn_scores"):
            hook.add_perma_hook(apply_padding_tokens_mask)

    return model

if MAIN:
        hook_tokens = True,
        ...
    )
    model = add_perma_hooks_to_mask_pad_tokens(model, tokenizer.PAD_TOKEN)
```

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -106,6 +106,9 @@ class HookedTransformer(HookedRootModule):
             self.pos_embed = PosEmbed(self.cfg)
             self.hook_pos_embed = HookPoint()  # [batch, pos, d__dictmodel]
 
+        if self.cfg.hook_tokens:
+            self.hook_tokens = HookPoint() # [batch, pos]
+
         self.blocks = nn.ModuleList(
             [
                 TransformerBlock(self.cfg, block_index)
@@ -254,6 +257,8 @@ class HookedTransformer(HookedRootModule):
                 cache_ctx_length == 0 or ctx_length == 1
             ), "Pass in one token at a time after loading cache"
             pos_offset = cache_ctx_length
+        if self.cfg.hook_tokens:
+            tokens = self.hook_tokens(tokens)
         embed = self.hook_embed(self.embed(tokens))  # [batch, pos, d_model]
         if self.cfg.positional_embedding_type == "standard":
             pos_embed = self.hook_pos_embed(
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -106,7 +106,7 @@ class HookedTransformer(HookedRootModule):
             self.pos_embed = PosEmbed(self.cfg)
             self.hook_pos_embed = HookPoint()  # [batch, pos, d__dictmodel]
 
-        if self.cfg.hook_tokens:
+        if self.cfg.use_hook_tokens:
             self.hook_tokens = HookPoint() # [batch, pos]
 
         self.blocks = nn.ModuleList(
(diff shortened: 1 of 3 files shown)


transformer_lens/HookedTransformer.py
```diff
@@ -106,6 +106,9 @@ def __init__(
             self.pos_embed = PosEmbed(self.cfg)
             self.hook_pos_embed = HookPoint()  # [batch, pos, d__dictmodel]
 
+        if self.cfg.hook_tokens:
```
neelnanda-io, line 109: Rename to use_token_hook or smth

transformer_lens/HookedTransformerConfig.py
```diff
@@ -116,6 +116,8 @@ class HookedTransformerConfig:
             the [scaling laws paper](https://arxiv.org/pdf/2001.08361.pdf) found
             that that was a more meaningful number. Ignoring biases and layer
             norms, for convenience)
+        hook_tokens (bool): If true, will add a hook at tokens (useful for implementing
```
neelnanda-io, line 119: -> "will add a hook point on the token input to HookedTransformer.forward, which lets you cache or intervene on the tokens" maybe?

neelnanda-io:
Thanks! Can you add a test sanity checking that it works? (Eg, load in a tiny model, run check that the loss on Hello World is current, and then patch Hello to be a different token and verify that the loss is correct on the new tokens?). I'd like to accumulate a bunch more tests over time.

neelnanda-io (changes requested):

callummcdougall:
Have now added a `test_hook_tokens` file with a test like you described, that passes. I've also changed it to `use_hook_tokens`, and edited docstring in the config as suggested.

neelnanda-io:
LGTM, thanks!

neelnanda-io (approved):

PR #149, opened 2023-01-22
Fix deps

jas-ho:
When running [Main_Demo.ipynb](Main_Demo.ipynb) for local development, circuitsvis and plotly are not installed.

This PR adds them to the poetry requirements.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -436,6 +436,26 @@ files = [
 unicode-backport = ["unicodedata2"]
 
 [[package]]
+name = "circuitsvis"
+version = "1.38.1"
+description = "Mechanistic Interpretability Visualizations"
+category = "dev"
+optional = false
+python-versions = ">=3.7,<4.0"
+files = [
+    {file = "circuitsvis-1.38.1-py3-none-any.whl", hash = "sha256:33d8494846cbad245491af6af29e6da3a0e87cd680bc14631199b190c19c93ec"},
+    {file = "circuitsvis-1.38.1.tar.gz", hash = "sha256:da9977a3be19519f63fa500ec2196e25cc847723e9ecc548174544c5520d5c78"},
+]
+
+[package.dependencies]
+importlib-metadata = ">=5.1.0,<6.0.0"
+numpy = [
+    {version = ">=1.21,<2.0", markers = "python_version < \"3.10\""},
+    {version = ">=1.23,<2.0", markers = "python_version >= \"3.10\""},
+]
+torch = ">=1.10,<2.0"
+
+[[package]]
 name = "click"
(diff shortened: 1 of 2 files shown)


jas-ho:
note: CI/CD for 3.10 took over 10mins on the depenency installation step
image
not sure why :shrug:

neelnanda-io:
Ah, this is deliberately NOT part of the requirements, and only used in the
tutorials to avoid bloating the package, CircuitsVis in particular comes
bundled with node stuff in a way that's messy.

The demo notebook automatically installs these, but I only want them to be
installed for the demo notebooks not for pip installing the package.

The issue is that the demo notebook checks whether it's in colab (if so,
installs the packages) or in Jupyter (where it assumes it's working locally
and doesn't install things because I already have them installed). I think
that people using the demo locally is a valid use case though, which is a
pain... I could do something stupidly hacky like checking whether a
specific file in my system exists?

Ooh, a better idea would be to try importing them, and use an except
statement to then install and then import?

neelnanda-io:
I'm considering adding some plotly based utils, but want to have a mediumly
high bar for adding new large dependencies unless they seem important.

jas-ho:
yep, you're right. this should definitely not be in the main requirements. 

should it be in the dev dependencies , though?
I think development work is still more likely to happen locally, in which case circuitsvis and plotly are not installed automatically when running Main_Demo.ipynb

jas-ho:
on a side note: I'm actually confused how plotly is pulled in (if it is). does not seem to be installed explicitly anywhere

neelnanda-io:
@alan-cooney Takes?

neelnanda-io:
Ah, Plotly is pre-installed on Colab.

Hmm, I'm not actually sure how dev requirements work, but I agree that
including them there seems reasonable (at least, if there's no convention
to keep *those *simple), happy to take a PR for that.

alan-cooney:
Yeah putting them in dev requirements makes sense. These are installed whilst developing as `poetry install --group dev` or they can be installed by end user as `pip install transformer-lens[dev]`. 

The reason for the slower CD is that where there are changed requirements, the CD checks that all possible combinations of packages in the ranges specified work with each other. This is expensive where packages fail to declare their requirements properly (as it has to download all versions), but is important to do. When there are no changed requirements it uses a cache to speed this up dramatically.

jas-ho:
don't merge yet. they're still in the main requirements. I'll move it to the dev requirements

jas-ho:
done. should be good to merge now

alan-cooney (approved):

alan-cooney:
Thanks for the PR!

PR #154, opened 2023-01-24
Fix OPT BOS Prepending issue

afspies:
This addresses issue https://github.com/neelnanda-io/TransformerLens/issues/153.

For reasons that are unclear to me, the PreTrainedtokenizer for Facebook OPT automatically prepends `</s>` to OPT prompts, unless the tokenizer call sets `add_special_tokens=False`. This has no effect for the GPT tokenizer.

*I Am not sure* if this breaks things for other models, in which case we'd want to do `add_special_tokens='opt' in self.tokenizer.name` but would prefer to avoid that if possible.

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -407,7 +407,7 @@ class HookedTransformer(HookedRootModule):
             padding = True,
             truncation = truncate,
             max_length = self.cfg.n_ctx if truncate else None,
-            add_special_tokens=False  # As we manually add the BOS token
+            add_special_tokens = False  # As we manually add the BOS token
             )["input_ids"]
         if move_to_device:
             tokens = tokens.to(self.cfg.device)
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -406,7 +406,8 @@ class HookedTransformer(HookedRootModule):
             return_tensors = "pt", 
             padding = True,
             truncation = truncate,
-            max_length = self.cfg.n_ctx if truncate else None
+            max_length = self.cfg.n_ctx if truncate else None,
+            add_special_tokens=False  # As we manually add the BOS token
             )["input_ids"]
         if move_to_device:
             tokens = tokens.to(self.cfg.device)
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -407,7 +407,7 @@ class HookedTransformer(HookedRootModule):
             padding = True,
             truncation = truncate,
             max_length = self.cfg.n_ctx if truncate else None,
-            add_special_tokens = False  # As we manually add the BOS token
+            add_special_tokens = False if self.tokenizer.name_or_path.startswith('facebook/opt') else True  # As we manually add the BOS token
             )["input_ids"]
         if move_to_device:
             tokens = tokens.to(self.cfg.device)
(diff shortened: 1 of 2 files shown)


transformer_lens/HookedTransformer.py
```diff
@@ -406,7 +406,8 @@ def to_tokens(
             return_tensors = "pt", 
             padding = True,
             truncation = truncate,
-            max_length = self.cfg.n_ctx if truncate else None
+            max_length = self.cfg.n_ctx if truncate else None,
+            add_special_tokens = False  # As we manually add the BOS token
```
neelnanda-io, line 410: I'm uncomfortable with this applying to _all_ tokenizers in case it results in unexpected bugs, can you replace this with eg `True if self.cfg.original_architecture == "OPTForCausalLM" else False`? (Irritatingly wordy, but this is what `original_architecture` is intended for, and this feels neater than an entire if statement copying the tokenizer run. I think the True, False framing is more readable, even if it is literally equivalent to just `self.cfg.original_architecture == "OPTForCausalLM"`)

neelnanda-io (changes requested):

transformer_lens/HookedTransformer.py
neelnanda-io: `self.tokenizer.name` to OPT actually might be more reasonable, since it allows other archs with the same tokenizer (if this would ever happen, idk)

neelnanda-io:
And can you add a quick test for this? In the style of `test_d_vocab`, initialise a simple transformer with tiny config, and the OPT tokenizer, then run `to_tokens` with prepend_bos = True and with prepend_bos = False and verify that it has the length expected in both cases.

Crib from this:
```
def test_d_vocab_from_tokenizer():
        n_layers=1, 
        d_mlp=10, 
        d_model=10, 
        d_head=5, 
        n_heads=2, 
        n_ctx=20, 
        act_fn="relu"
    )
    model = HookedTransformer(
        cfg=cfg, 
        tokenizer=AutoTokenizer.from_pretrained("gpt2")
    )
    assert model.cfg.d_vocab == 50257
    assert model.cfg.d_vocab_out == 50257
```

Running `pytest TEST_FILE_NAME` tests whether it works

afspies:
Ok these were good suggestions.

I've added a test and noticed some things in the process:
 1. Initializing the `HookedTransformer`  without `FromPretrained` (as in the test) leads to its having no `cfg.original_architecture`. As such I am instead taking the name from the tokenizer, which is a bit less elegant, but sufficiently robust.
 2. I am not sure what (if any) tokenizer is used for your SOLU models, and as such the test doesn't currently include a representative example for one of your models
 3. The token initialization performed when the tokenizer is loaded, along with this faffing, hints that we might want to wrap the tokenizer at some point.
 
Let me know if you think of anything else!

neelnanda-io:
Thanks! My SoLU + toy models use this tokenizer
https://huggingface.co/NeelNanda/gpt-neox-tokenizer-digits/ , the
solu-pile/solu-old models use the NeoX tokenizer.

What do you mean by wrapping the tokenizer?

afspies:
Ok I've included the `gpt-neox-tokenizer-digits` tokenizer in the test. This is a bit ugly - for reasons on which I'm unclear the tokenizer cannot be directly loaded from the huggingface model linked to `solu-l1` (going via `loading.get_official_model_name`. I assume there is some canonical way of linking tokenizers to model cards on huggingface, which you haven't done for your solu models, but which is done for gpt, opt etc.?

By wrapping the tokenizer I just mean creating a e.g. `LensTokenizer(AutoTokenizer)` class which does the token setup as part of the init and can include details about tokenizer names distinct from their models (e.g. for your solu models [maybe]) and the BOS handling we are doing. This would just be so that the HookedTransformer only has to call `lens_tokenizer(text, **kwargs)` and some of that complexity is encapsulated. Probably not worthwhile at present though.

neelnanda-io (approved):
LGTM!

PR #158, opened 2023-01-29
Add QKV split 

ArthurConmy:
This PR 

i) adds the `hook_attn_input` for all `HookedTransformer`s, not just shortformer transformers
ii) adds the `self.cfg.use_split_qkv_input` flag, so inputs to the Q and K and V calculations, per head, can be edited
iii) improves the handling of adding hooks so this is safer
iv) makes the tests for these features (better)

Thanks @ericpurdy for the idea to just PR this, rather than keep it on my branch https://github.com/neelnanda-io/TransformerLens/pull/139


ArthurConmy:
Hmm, thinking more, doesn't it make more sense to hook the Q and K and V inputs before the layer norm into the attention layer?

This is a bit messy to implement. It could be done by either 

- moving the ln1 to the attention block, rather than the transformer block
- moving these q and k and v input changes to the transformer block, not the attention block

Provided that breaking changes are OK, I think the first solution makes more sense

neelnanda-io:
Hmm, I think the library is widely used enough that I'd rather not make breaking changes unless there's a really good argument, which I don't think this meets. _If_ we're assuming linearity of LayerNorm, then having separate Q, K and V after LayerNorm seems totally fine, though I'm open to the argument that this is a dumb assumption.

Moving the LayerNorm to be part of the attention block will break any code using that LayerNorm or hooks on it (though we can patch get_act_name to be backwards compatible which lessens this a bit), which seems bad. And making the attention layer have a separate Q, K and V input also feels ugly and confusing to people reading the code.

Yeah, I think I'll vote against, unless I change my mind on eg being able to patch different inputs before LayerNorm being very important (rather than a nice to have)

ArthurConmy:
Hmm, patching before the LN is the thing I've needed in both applications I've had so far:

This is because the primary use for this is looking at Q-composition or something, and then you want to e.g patch a heads output into this q_input. You don't want to also have to apply the layer norm. 

I think we could i) keep LN in the transformer block and ii) have the QKV inputs before the layer norm by making the attention block's forward method a function of three arguments: q_input, k_input and v_input. How does that sound?

neelnanda-io:
> I think we could i) keep LN in the transformer block and ii) have the QKV
inputs before the layer norm by making the attention block's forward method
a function of three arguments: q_input, k_input and v_input. How does that
sound?

Fair, I think this is the least bad option in that case. (Not *great*, but
I don't see it breaking anything too important). Can you make sure to
comment it well and explain the design decision

transformer_lens/components.py
```diff
@@ -700,11 +673,25 @@ def forward(
...
         resid_pre = self.hook_resid_pre(resid_pre)  # [batch, pos, d_model]
-        normalized_resid_pre = self.ln1(resid_pre)
+
+        query_input = resid_pre
+        key_input = resid_pre
+        value_input = resid_pre
+        
+        if self.cfg.use_split_qkv_input:
+            for qkv_input in [query_input, key_input, value_input]:
+                qkv_input = einops.repeat(qkv_input, "batch pos d_model -> batch pos n_heads d_model", n_heads=self.cfg.n_heads).clone()
```
ArthurConmy, line 683: Clone is vital so individual heads as edited separately

ArthurConmy:
(This can be merged)

neelnanda-io:
Can you only use the hook_q_input etc hooks if split_qkv_input is True? As is, they get added to the cache under run_with_cache, which bloats memory. I should add a test which checks that the list of activations in the cache is correct!

ArthurConmy:
Sure, done

ArthurConmy:
@neelnanda-io is this OK?

neelnanda-io:
LGTM, sorry for dropping this!

PR #160, opened 2023-02-02
Move demos to their own folder

dkamm:
Just a thought for tidying things up

@neelnanda-io I think you'd have to change redirects on your site


neelnanda-io:
I broadly buy that this is a good idea, but changing all the links would be a headache lol, so lean against doing it (changing redirects is easy enough, but there's a bunch of times they've been linked without redirects probably). I've made a demos folder for my latest demo, and will gradually move future things in there.

PR #161, opened 2023-02-02
Implement from_pretrained_no_processing

lukasberglund:
This is my attempt to address [this issue](https://github.com/neelnanda-io/TransformerLens/issues/134) asking for a `from_pretrained_no_processing` function. It's pretty simple, so let me know if I'm missing some nuance here.


lukasberglund:
Looks like this is still failing some tests. Working on fixing them now.

lukasberglund:
When I run the tests locally, they work for 3.8, 3.9, and 3.10. For some reason I haven't been able to get anything to work for 3.7 on my machine. I might look at this tomorrow, but if someone has any guesses as to what's going on it would be appreciated!

ArthurConmy:
@lukasberglund after [this](https://github.com/neelnanda-io/TransformerLens/pull/162) PR is merged, tests will pass (read the PR to know what's up...).

neelnanda-io:
Merged in Arthur's PR

ArthurConmy:
Ok now these tests do pass locally for me. Frustrating that the action provides no signal on what's up too. Maybe repush to try again (...)?

neelnanda-io:
Thanks for the PR! Not sure what's up with the tests lol, their error messages just say "test has been cancelled"?! Not seen this before.

On the actual PR, aesthetically I'd prefer to just have the interface be the model_name and four Boolean flags, and everything else as **from_pretrained_kwargs. And have the docstring say it's a wrapper around from_pretrained and refer people to that. Makes the code more maintainable, rather than all future changes needing to keep the two functions in sync (though maybe at the cost of making the code harder to read/use, esp when looking at just the function docstring? Idk, I mostly don't expect people to use `from_pretrained_no_processing` if they don't know what they're doing)

lukasberglund:
Ok, I've implemented the changes you suggested. The tests are still failing, I wonder if the tests are running out of memory somehow. The test I added loads 4 versions of gpt-2, maybe that's too large. I've got to run now, but I might try if the test passes with smaller models.

neelnanda-io:
Oh! I can totally imagine that breaking things. I'd definitely prefer eg solu-1l, tests running fast is also valuable, and I don't think this loses anything

lukasberglund:
Makes sense. I've changed it to that. The new test takes 16 seconds to run on my machine, but if I remove the second part, it would only be 8 seconds. That part is non-trivial, but not absolutely necessary. I could remove it to save time if you want.

neelnanda-io:
LGTM, thanks!

PR #166, opened 2023-02-04
Slice: doc and tests

### LINKED ISSUE #95: Add Documentation + Tests for utils.Slice
I really need to document this one better lol - basically a wrapper around Python slice, but None -> slice(None) (equivalent to [:], doesn't add a dummy axis), and n -> [n] (for an integer n), which reduces the number of axes, which Python slices don't support.

Xmaster6y:
Main feature:

* Closes #95


jbloomAus:
Looks good, thanks for your contribution!

PR #170, opened 2023-02-09
Minor Readme Changes

jbloomAus:
- I've added a note about Decision Transformer Interpretability to user projects. 
- I've add an instruction for how to run the tests in the readme.


alan-cooney (approved):

alan-cooney:
Noting at some point we should enforce line lengths in markdown :)

PR #171, opened 2023-02-12
Replace torchtyping with jaxtyping

### LINKED ISSUE #164: Convert all TorchTyping to JaxTyping
This project uses Patrick Kidger's torch typing to give tensor types including shapes. He recommends using JAXTyping for newer projects (not JAX specific, better maintained, more compatible with type checkers, etc) And it'd be good to update TL to use it. 

@dkamm might be up your alley? Sorry if this invalidates your enum work, but might be a more elegant solution!

dkamm:
Closes #164


dkamm:
@neelnanda-io ready for review. Had to reformat Attribution Patching Demo because it was saved as one giant line, hence the large diff

neelnanda-io:
Thanks for the thorough PR, nice work! I just merged in #158 which includes some more torchtyping stuff, sorry! Mind fixing those files too? `components.py` and `test_hooks.py`. And is the poetry file supposed to have had so many lines deleted? I'm unsure how this works, @alan-cooney may have takes. 

Otherwise LGTM!

dkamm:
Thanks! Yes, I'll update these.

I removed `torchtyping` so that's probably why the `poetry.lock` has so many deleted lines.  

Thinking about this more, we probably don't want to remove it yet because we could avoid additional merge conflicts on other open PRs

dkamm:
@neelnanda-io ready to go again!

The `poetry.lock` diff is fixed and it was actually because my poetry was out of date.  Good catch on that!

I also changed my mind and decided to keep `torchtyping` removed.  It doesn't look like many PRs will be affected by this change

alan-cooney:
Looks great! Still a few conflicts but otherwise this should be an improvement.

Btw we could alias these types (i.e. export `ResidualStreamType`, `AttentionPatternType` etc somewhere and then use them everywhere). However this could come in a follow-on PR

dkamm:
hey @alan-cooney - are you referring to merge conflicts?  just want to check as it's not showing any for me

alan-cooney:
> hey @alan-cooney - are you referring to merge conflicts? just want to check as it's not showing any for me

Yes GitHub is showing conflicts. Can you try rebasing?

dkamm:
@alan-cooney ah got it! Is there a reason not to squash merge this instead of rebase merge?  Not sure if the commit history is worth preserving here

If not, no worries- I'll go ahead with the rebase.  I think it'd be cleanest if I just redid the commits after [67b2a76](https://github.com/neelnanda-io/TransformerLens/pull/171/commits/67b2a76014f1ec4ca419311b56b86b6f91b4e86d) in that case

alan-cooney:
@dkamm don't worry about preserving history as we'll in any case rebase once more when merging in. Whichever youprefer is fine, as both approaches can be used to fix the conflicts.

dkamm:
hey @alan-cooney - I think we might be talking about slightly different things here. I'm requesting for you to use the "squash and merge" option as shown here 

Screenshot 2023-02-23 at 11 00 26 PM

I think if you switch to that option, you should see that there's no conflicts

Sorry to make a fuss about this, but I thought it might be useful to clarify and it's pretty common practice to use the 
"squash and merge" option AFAIK.  (And I also tried rebasing it and ran into issues lol)

Also FYI, poetry files had minor conflicts with the latest `main` which I just resolved

alan-cooney:
Nice - it was blocking both for me but now fine with the latest push - I'll review now

alan-cooney (approved):
Checked as best as I can, and mypy isn't reporting any issues with these changes specifically.

PR #172, opened 2023-02-12
Test coverage reports

jbloomAus:
- I've added pytest-cov to the dependencies and updated tests to run with a coverage report.

Not a super essential commit now but worth doing sooner/later, especially since building comprehensive coverage is a reasonable goal for Transformer Lens.


jbloomAus:
I've just noticed this built on my previous PR so they're not atomic. My apologies. Probably not worth dealing with since both are fine.

neelnanda-io:
Your previous PR?

PR #173, opened 2023-02-15
Add gotcha to HookedTransformer.to_str_tokens docstring

rusheb:
Hello, this caught me out, so adding docs in the hope they will help somebody.

If you don't think that is helpful then feel free to close.

Many thanks!


neelnanda-io:
Oh huh, thanks! I wasn't even aware of that lol, though I deliberately made to_tokens do that, since the model forward pass can use it implicitly on a string input. Feel free to make a pr to add a param to fix this.

I'm curious what the use case was such that this was an issue?

rusheb:
> I'm curious what the use case was such that this was an issue?

I'm not doing anything serious (yet!) -- I was just playing around to get more familiar and I was confused about why shapes were not matching up.

Happy to add the flag. Since there's no real use case, would it still be helpful?

If yes, I'll take a look tomorrow!

PR #183, opened 2023-02-26
Add a constructor for ActivationCache from a list of Caches

afspies:
I wanted to combine many activation caches into one big one, and created a constructor to that end.

Not sure if this is elegant / something you want - feel free to reject. If you do think this is useful, then perhaps it is worth adjusting the `model.run_with_cache()` function to accept a cache which it writes into (i.e. just calling `return ActivationCache.from_caches([cache_computed, cache_passed_in], in_place=True)`)

### DIFF
--- a/transformer_lens/ActivationCache.py
+++ b/transformer_lens/ActivationCache.py
@@ -45,8 +45,12 @@ class ActivationCache:
 
     @classmethod
     def from_caches(cls, cache_list: List[ActivationCache], in_place: Bool = True) -> ActivationCache:
-        # Constructor that combines a list of caches into a big one
-        
+        """
+        Alternate Constructor. Takes a list of ActivationCaches and concatenates them along the batch dimension. 
+
+        cache_list (List[ActivationCache]): The list of caches to stack.
+        in_place (Bool): Whether to clear the caches after stacking them, or create copies. Defaults to True => don't copy.
+        """
         assert len(cache_list) > 0, "Cannot stack an empty list of caches"
         assert all(
             cache.has_batch_dim for cache in cache_list
@@ -57,17 +61,17 @@ class ActivationCache:
 
         if not in_place:
             cache_dict = deepcopy(cache_dict)
-        model = cache_list[0].model.name
+        model = cache_list[0].model
+        model_name = cache_list[0].model.cfg.model_name
         has_embed = cache_list[0].has_embed
         has_pos_embed = cache_list[0].has_pos_embed
 
         for activation_cache in cache_list[1:]:
-            assert (model == activation_cache.model.name), "Cannot stack caches from different models"
+            assert (model_name == activation_cache.model.cfg.model_name), "Cannot stack caches from different models"
             assert has_embed == activation_cache.has_embed, "Cannot stack caches with different embed hooks"
             assert has_pos_embed == activation_cache.has_pos_embed, "Cannot stack caches with different pos_embed hooks"
             
             for key in cache_dict:
-                # cache_dict[key] = torch.cat((cache_dict[key], activation_cache.cache_dict[key]), dim=0)
                 cache_dict[key].append(activation_cache.cache_dict[key])
 
         cache_dict = {k:torch.cat(v, dim=0) for k,v in cache_dict.items()}
--- a/transformer_lens/ActivationCache.py
+++ b/transformer_lens/ActivationCache.py
@@ -43,6 +43,39 @@ class ActivationCache:
         self.has_embed = "hook_embed" in self.cache_dict
         self.has_pos_embed = "hook_pos_embed" in self.cache_dict
 
+    @classmethod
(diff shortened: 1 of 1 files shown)


transformer_lens/ActivationCache.py
```diff
@@ -43,6 +43,39 @@ def __init__(
         self.has_embed = "hook_embed" in self.cache_dict
         self.has_pos_embed = "hook_pos_embed" in self.cache_dict
 
+    @classmethod
+    def from_caches(cls, cache_list: List[ActivationCache], in_place: Bool = True) -> ActivationCache:
+        # Constructor that combines a list of caches into a big one
```
neelnanda-io, line 48: Can you add a docstring? In particular, what `in_place` does should be documented

transformer_lens/ActivationCache.py
```diff
@@ -43,6 +43,39 @@ def __init__(
...
+        assert all(
+            cache.has_batch_dim for cache in cache_list
+        ), "Cannot stack caches without batch dimensions"
+
+        # Concat a dict of lists once rather than concatting many times
+        cache_dict = {k:[v] for k,v in cache_list[0].cache_dict.items()}
+
+        if not in_place:
+            cache_dict = deepcopy(cache_dict)
+        model = cache_list[0].model.name
```
neelnanda-io, line 60: Rename to `model_name`

transformer_lens/ActivationCache.py
```diff
@@ -43,6 +43,39 @@ def __init__(
...
+            
+            for key in cache_dict:
+                # cache_dict[key] = torch.cat((cache_dict[key], activation_cache.cache_dict[key]), dim=0)
+                cache_dict[key].append(activation_cache.cache_dict[key])
+
+        cache_dict = {k:torch.cat(v, dim=0) for k,v in cache_dict.items()}
+        if in_place:
+            [cache.cache_dict.clear() for cache in cache_list] 
+
+        return cls(cache_dict, model, has_batch_dim=True)
```
neelnanda-io, line 77: You want the actual model, not the model's name

transformer_lens/ActivationCache.py
neelnanda-io: Also this doesn't work. You maybe want `model.cfg.model_name`?

neelnanda-io: Not a big deal either way

transformer_lens/ActivationCache.py
```diff
@@ -43,6 +43,39 @@ def __init__(
...
+        
+        assert len(cache_list) > 0, "Cannot stack an empty list of caches"
+        assert all(
+            cache.has_batch_dim for cache in cache_list
+        ), "Cannot stack caches without batch dimensions"
+
+        # Concat a dict of lists once rather than concatting many times
+        cache_dict = {k:[v] for k,v in cache_list[0].cache_dict.items()}
+
+        if not in_place:
```
neelnanda-io, line 58: Unnecessary, you're concatenating them anyway

transformer_lens/ActivationCache.py
```diff
@@ -43,6 +43,39 @@ def __init__(
...
+            assert has_embed == activation_cache.has_embed, "Cannot stack caches with different embed hooks"
+            assert has_pos_embed == activation_cache.has_pos_embed, "Cannot stack caches with different pos_embed hooks"
+            
+            for key in cache_dict:
+                # cache_dict[key] = torch.cat((cache_dict[key], activation_cache.cache_dict[key]), dim=0)
+                cache_dict[key].append(activation_cache.cache_dict[key])
+
+        cache_dict = {k:torch.cat(v, dim=0) for k,v in cache_dict.items()}
+        if in_place:
+            [cache.cache_dict.clear() for cache in cache_list]
```
neelnanda-io, line 75: Is memory management actually an issue here? If so, you'll probably get the best results by going through each activation type, concatenating all of those across all caches, and then deleting the component activations.

Also, moving things to the CPU! Helps a lot

neelnanda-io: IMO you can just not care though.

neelnanda-io (changes requested):
Thanks! Seems like a reasonable feature, my most important requested changes are a docstring and fixing the `model.name` part.

neelnanda-io:
I don't think this is *that *useful, so I don't want to merge it into
`run_with_cache` (I already have an ugh at how bloated that function is,
and how it's mutated beyond HookedRootModule, but is technically a subclass
of it)

transformer_lens/ActivationCache.py
afspies: Thanks for noticing that!

afspies (commented):
Have made the requested changes

jbloomAus:
Closed based on the above conversation (ActivationCache being very bloated already).

PR #185, opened 2023-03-04
Add `verbose` flag to disable tqdm on `model.generate(...)`

afspies:
There's no easy way to disable tqdm with context managers (afaik!), and patching tqdm around generate calls where one wants to avoid the tqdm output is tedious.

Not sure "verbose" is the best name for this flag as it's really more of a `show_progress`, but might be better to be generic and imprecise.


neelnanda-io:
LGTM!

Issue #98, opened 2022-12-19
Add tests + better docs for FactoredMatrix


neelnanda-io:
Add tests that the FactoredMatrix class works (essentially that each of its methods correctly mimics the result for the actual matrix product).

Go through the documentation and clarify things that are unclear (this is hard for me to do, so even just having someone new to the library flag confusions is helpful!)

inwaves:
Taking a look at this over the next week or so.

rusheb:
Hello, I would like to help with adding unit tests. I may find it hard to make all the changes all at once, and it would be good to work incrementally to make sure I am on the right track!

Are you okay with me breaking down the work into multiple PRs?  This may also enable the work to be parallelised between several developers.

If so, the following table could be used to track progress:
| method        | pull request | assigned | completed | notes                           |
|---------------|--------------|----------|-----------|---------------------------------|
| `__init__`    |  https://github.com/neelnanda-io/TransformerLens/pull/188    |   @rusheb        |           |                                 |
| `__matmul__`  |              |          |           |                                 |
| `__rmatmul__` |              |          |           |                                 |
| `AB`          |              |          |           |                                 |
| `BA`          |              |          |           |                                 |
| `T`           |              |          |           |                                 |
| `svd`         |              |          |           | includes U, S and Vh properties |
| `eigenvalues` |              |          |           |                                 |
| `__getitem__` |              |          |           |                                 |
| `norm`        |              |          |           |                                 |
| `__repr__`    |              |          |           |                                 |
| `make_even`   |              |          |           |                                 |
| `get_corner`  |              |          |           |                                 |
| `collapse_l`  |              |          |           |                                 |
| `collapse_r`  |              |          |           |                                 |
| `unsqueeze`   |              |          |           |                                 |
| `pair`        |              |          |           |                                 |

rusheb:
I think that all the tests for this are actually done as of #218. @jbloomAus maybe we should close this ticket and spin out a new one for docs.

PR #188, opened 2023-03-08
Test constructor of FactoredMatrix

rusheb:
Add unit tests for the FactoredMatrix constructor. 

This is the first of a series of PRs addressing #98.


rusheb:
I managed to fix the TypeError that was been thrown by `test_factored_matrix_inner_mismatch`.

I did this by modifying the type hints in the FactoredMatrix constructor to make `mdim` of the `A` and `B` tensors distinct, in order to disable [multi-argument consistency checks](https://github.com/google/jaxtyping/blob/main/API.md#runtime-type-checking).

Let me know whether you are happy with this change! If you prefer, it would be possible to remove the assertion from the constructor and let typeguard perform the check.

<details>
<summary>Details</summary>

## Problem: typeguard/jaxtyping sometimes intervene with runtime type checking, but not always

(to repro the following steps, checkout commit c091923)

Consider the following test function in the file `maze_transformer/tests/factored_matrix/test_constructor.py`:
```python
def test_factored_matrix_inner_mismatch():
    A = torch.ones((2, 3, 4))
    B = torch.ones((2, 3, 5))
    FactoredMatrix(A, B)
```

For reasons still unclear to me, the output of this test changes depending which directory the tests are run from:

```python
# run from parent directory
poetry run pytest transformer_lens/tests -k "test_factored_matrix_inner_mismatch"
...
TypeError: type of argument "B" must be jaxtyping.Float[Tensor, '... mdim rdim']; got torch.Tensor instead
```

```python
# run from child directory
poetry run pytest transformer_lens/tests/factored_matrix -k "test_factored_matrix_inner_mismatch"
...
AssertionError: Factored matrix must match on inner dimension, shapes were a: torch.Size([2, 3, 4]), b:torch.Size([2, 3, 5])
```


Furthermore, if we run the same test code from a script file (rather than via pytest), the AssertionError is thrown (i.e. no runtime type checking happens):
```python
# scratch.py
A = torch.ones((2, 3, 4))
B = torch.ones((2, 3, 5))
FactoredMatrix(A, B)

# output
AssertionError: Factored matrix must match on inner dimension, shapes were a: torch.Size([2, 3, 4]), b:torch.Size([2, 3, 5])
```

## Solutions considered

1. Modifying the type hints to make `mdim` distinct in `A` and `B` means typeguard will not check that the `mdim` is the same for `A` and `B`. Thus it will never throw a TypeError and we will always reach an AssertionError. This is the option I have gone with.

4. An alternative approach would be to add the `@jaxtyping` and `@typechecked` annotations to the FactoredMatrix constructor. This would mean that typeguard would always run, making the assertion in the constructor redundant. Therefore we could delete the assertion and rely on typeguard. However, I was worried about this causing unintended side-effects. I also think the TypeError message is less helpful to the user than the AssertionError message (see previous section).

</details>

rusheb:
@jbloomAus - as discussed, I have reverted the change to the type hints and disabled the test with a comment linking to a new issue.

jbloomAus:
@rusheb I like big trunk development so happy to merge, I assume that was your intent with this PR despite there being more work to do?

rusheb:
> @rusheb I like big trunk development so happy to merge, I assume that was your intent with this PR despite there being more work to do? 

Yes please! Was planning to do it incrementally in separate PRs

jbloomAus:
awesome, great!

Issue #106, opened 2022-12-19
Add evals for algorithmic-ish tasks like Indirect Object Identification


neelnanda-io:
Add to `evals.py` support for checking how good a model is at tasks like [Indirect Object Identification](https://arxiv.org/abs/2211.00593). Notably, where the eval involves generating a synthetic dataset of names, as they do in the paper (can copy from [their codebase](https://github.com/redwoodresearch/Easy-Transformer)), running the model on them, and returning the average logit diff and accuracy. 

Bonus:
* Add support for prompts of different token length 
* Support multi-token answers.
* Write this in a generic way, where any generated dataset can be subbed in
* Give an option to pair up prompts, so "John and Mary went to the store, John gave the bag to" is followed by "John and Mary went to the store, Mary gave the bag to", to avoid biases where the model just favours common names

jbloomAus:
Closed by #198

PR #198, opened 2023-03-16
IOI Evaluation

slavachalnev:
Implements #106.

Example usage:
```python
from transformer_lens.evals import ioi_eval, IOIDataset
from transformer_lens.HookedTransformer import HookedTransformer

model = HookedTransformer.from_pretrained('gpt2-small')

# Eval like this
print(ioi_eval(model, num_samples=100))
# returns {'Logit Difference': 3.655226745605469, 'Accuracy': 1.0}

# Can use custom dataset
ds = IOIDataset(
    tokenizer=model.tokenizer,
    num_samples=100,
    templates=['[A] met with [B]. [B] gave the [OBJECT] to [A]'],
    names=['Alice', 'Bob', 'Charlie'],
    nouns={'OBJECT': ['ball', 'book']},
    )
print(ioi_eval(model, dataset=ds))
# returns {'Logit Difference': 3.7498160457611083, 'Accuracy': 1.0}

# Handles multi-token names and names which begin with the same token
ds = IOIDataset(
    tokenizer=model.tokenizer,
    num_samples=100,
    names=['John Smith', 'John Doe'],
    )
print(ioi_eval(model, dataset=ds))
# returns {'Logit Difference': 4.861071290969849, 'Accuracy': 1.0}

# Sanity check: flipping the template should give a negative logit difference
ds = IOIDataset(
    tokenizer=model.tokenizer,
    num_samples=100,
    templates=['[B] met with [A]. [A] said hello to [A]'],
    )
print(ioi_eval(model, dataset=ds))
# returns {'Logit Difference': -4.056923713684082, 'Accuracy': 0.0}
```

### DIFF
--- a/tests/acceptance/test_evals.py
+++ b/tests/acceptance/test_evals.py
@@ -1,0 +1,65 @@
+import pytest
+from transformer_lens.evals import ioi_eval, IOIDataset
+from transformer_lens.HookedTransformer import HookedTransformer
+
+@pytest.fixture(scope="module")
+def model():
+    return HookedTransformer.from_pretrained('gpt2-small')
+
+def test_basic_ioi_eval(model):
+    """
+    Test IOI evaluation with default dataset and settings.
+    """
+    results = ioi_eval(model, num_samples=100)
+    assert results['Accuracy'] >= 0.99
+
+def test_symmetric_samples(model):
+    """
+    Test IOI evaluation with symmetric=True so prompts are in symmetric pairs.
+    """
+    ds = IOIDataset(tokenizer=model.tokenizer, num_samples=100, symmetric=True)
+    results = ioi_eval(model, dataset=ds)
+    assert results['Logit Difference'] > 2.0
+    assert results['Accuracy'] > 0.9
+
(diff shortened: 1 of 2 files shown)


slavachalnev:
I've found two failure cases:
1. There are two identical names.
2. A name is a prefix of another name.
```python
ds_1 = IOIDataset(tokenizer=model.tokenizer, names=["Alice", "Alice"])
ds_2 = IOIDataset(tokenizer=model.tokenizer, names=["John", "John Doe"])
```
I'm not sure what the behaviour should be in these cases.

jbloomAus:
@slavachalnev Would you possibly be able to do some basic tests around these changes? Optionally, a tutorial would also be good.

jbloomAus:
Thanks Slava! I've accepted this PR.

In the future, we should add docstring examples I did for yours. 

In terms of the failure cases, I'm not too bothered. Ideally, we should have a tutorial for this, but we can cover that in a review later. Likewise, the test coverage is ok (81%) for the module but that's better than the rest of the repo and playing around with it myself, it seemed good.

PR #204, opened 2023-03-19
Add Redwood 2L model

ArthurConmy:
# Description

Adds Redwood Attn only 2L model from https://www.alignmentforum.org/posts/j6s9H9SHrEhEfuJnq/causal-scrubbing-results-on-induction-heads

Fixes # 180 
https://github.com/neelnanda-io/TransformerLens/issues/180 

## Type of change

## Usage 

Load in with 

```
unprocessed_model = HookedTransformer.from_pretrained(
    "redwood_attn_2l",
    center_writing_weights=False, 
    center_unembed=False,
    fold_ln=False,
)
```

The flag must be (False, False, False) or (False, True, False), because I'm getting bugs when I put these as True. MWE here: https://pastebin.com/mCVNpUWT :( - I would LOVE if someone could look into what's up here!

Putting the PR up so it doesn't get left for longer undone, but I'm frustrated I can't fix the issue. Maybe it's something to do with how this architecture has layer norms but no MLPs? Or shortformer-related?


jbloomAus:
Thanks @ArthurConmy, I'll try to recommend to anyone I know filling PRs :)

neelnanda-io:
Hmm, I'm guessing they need to be false because shortformer? I believe it
adds the pos embed after the layernorm. While center writing weights
assumes pos embed writes to the residual stream. Ditto fold ln fails
because adding pos embed is between the ln weights being used and the W_Q
being used

neelnanda-io:
I think the correct solution is to raise a warning when this is tried if
the config says shortformer is being used, and force it to be false

jbloomAus:
Thanks, looks great!

PR #234, opened 2023-04-04
Add LLaMA support

### LINKED ISSUE #201: Add LLaMA support
From Neel: we should add support for LLaMA and T5 (issue here: https://github.com/neelnanda-io/TransformerLens/issues/94 ).

LLaMA isn't open source so maybe with instructions to download from local weights.

(T5's architecture is a bit different (I think) but it's also somewhat useful as google have lots of models that are pretty powerful and use the same codebase, e.g https://huggingface.co/google/flan-ul2 )

0amp:
# Description

Add support for all 4 LLaMA models using huggingface as a backend. Requires main branch of transformers repo and accelerate (```pip install accelerate git+https://github.com/huggingface/transformers```). 

I added accelerate to ```setup.py```. I'm not sure how to incorporate the main branch of transformers into the package dependencies. 

Fixes #201 

## Type of change

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works (see ```LLaMA.ipynb``` demo)


neelnanda-io:
Seems like a very thorough PR, thanks a lot! Gated MLPs as a separate class
seems correct to me. Do you know what's up with LLaMA being on hugging
face, from someone who is NOT Meta? I wasn't aware of this and am concerned
that's against Meta's terms and will be taken down (and thus would be a bad
idea for TransformerLens to get involved), since my understanding is that
Meta is still acting like they didn't open source the models(?)

0amp:
Yup! I'm not sure what's up with the upload to huggingface. For some reason, the PR was merged into the main transformers branch and there are at least a half dozen other clones (or fine-tuned versions) on huggingface now anyways so I would bet that it probably won't get taken down. My understanding is that Meta has not said anything new about their license. 

Also, is there some easy way to fix the dependencies so that the tests pass?

neelnanda-io:
Hmm, can you raise an issue on the HuggingFace repo and ask about it, and
tag the author of the PR? If it's explicitly not allowed by Meta, then I
would favour not explicitly having LLaMa in from pretrained and instead
having instructions on how to load it from the weights stored locally

0amp:
So there's this https://github.com/huggingface/transformers/issues/21796. Looks like the merged pr is just setting up the LLaMA class and that individual people have uploaded the weights. Seems like the HF team isn't going to do anything about it and Meta hasn't said anything about it yet. Also, there's a script for converting meta weights included in the original PR https://github.com/huggingface/transformers/issues/21796 so I could change this PR to just have people download + convert weights themselves.

neelnanda-io:
Lol. Thanks for the context. Based on that this is definitely *not *legit,
and I'd prefer if TransformerLens didn't explicitly have support to do
something that violates Meta's terms (in particular, having
`HookedTransformer.from_pretrained("llama-7b")` work, implicitly implies to
a normal user that this *is* legitimate use, in my opinion).

But I love the idea of having it be supported for anyone who already has
access to the weights, that all sounds great and I think there's some
interesting questions here. If you could edit the colab/docstring to have
clear instructions on how to convert the legitimate weights, that would be
amazing.

jbloomAus:
@neelnanda-io I think it sounds like @0amp has completed the PR per your requests, and it is documented/expects users to have legal access to the weights. However, due to the nature of the PR, I'll let you merge it, or if you reply here asking me to, I'll merge it.

PR #220, opened 2023-03-23
Add apply_ln param to accumulated_resid and decompose_resid

ckkissane:
# Description

This PR addresses #209 making the following changes:
- add optional apply_ln param to accumulated_resid and decompose_resid to automatically apply layernorm to corresponding stack
- adds 2 basic tests to sanity check these both work as intended
- Corrects a typo in get_full_resid_decomposition doc string

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes
- [x ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

Issue #209, opened 2023-03-20
[Proposal] Support decompose_resid and accumulate_resid having apply_ln parameter


neelnanda-io:
### Proposal 

ActivationCache.get_full_resid_decomposition has the parameter apply_ln which scales the output (the stacked output of each head and neuron/MLP layer) by the LayerNorm scale factor at the relevant layer (normally the final_ln before the unembed). It'd be useful if other similar functions that return a stack of residual streams did the same, namely `decompose_resid` and `accumulate_resid`, since I often want to use this

jbloomAus:
Solved by PR #220

PR #239, opened 2023-04-09
Add apply_ln to stack_head_results and stack_neuron_results

ckkissane:
# Description

Previously I added an apply_ln parameter to accumulated_resid and decompose_resid in PR #220 to address issue #209 . After using them, I'm noticing it's also natural to make an analogous call to stack_head_results and stack_neuron_results (like when doing direct logit attribution). This PR adds the same functionality to stack_head_results and stack_neuron_results so they are all consistent. I also deleted a line from the stack_head_results doc string which doesn't seem true.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

# Checklist:

- [x ] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes
- [x ] I have not rewritten tests relating to key interfaces which would affect backward compatibility


jbloomAus:
Thanks, looks good!

PR #242, opened 2023-04-11
Improve patching docs

ckkissane:
# Description

There are a handful of patching helper functions with similar names, so I sometimes read the code to determine which one I should be using and what the inputs / outputs are. I think this can be a hassle because the partial functions are not documented, so I need to jump between them and generic_activation_patch to figure out how they work. To help solve this, I added docstrings for a bunch of patching helper functions. 

## Type of change

- [x ] This change requires a documentation update

# Checklist:

- [x ] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [x ] New and existing unit tests pass locally with my changes
- [x ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

### DIFF
--- a/transformer_lens/patching.py
+++ b/transformer_lens/patching.py
@@ -253,20 +253,20 @@ get_act_patch_resid_pre = partial(
     activation_name = "resid_pre",
     index_axis_names = ("layer", "pos")
 )
-"""
-Applies activation patching for every resid_pre activation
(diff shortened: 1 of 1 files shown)


jbloomAus:
Thanks, this is really good!

neelnanda-io:
I had no idea you could do that to a partial function! Thanks a lot

PR #245, opened 2023-04-14
Allow to_str_tokens of 1x1 tensors and add HookedTransformer.to_singl…

adamyedidia:
…e_str_token

# Description

-Makes HookedTransformer.to_str_tokens work on 1x1 tensors and numpy arrays
-Creates HookedTransformer.to_single_str_token, which takes in a single int and returns the string version of that token

Before:

In [11]: HookedTransformer(cfg).to_str_tokens(torch.tensor([2]))
Using pad_token, but it is not set yet.
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In [11], line 1
----> 1 HookedTransformer(cfg).to_str_tokens(torch.tensor([2]))

File ~/TransformerLens/transformer_lens/HookedTransformer.py:492, in HookedTransformer.to_str_tokens(self, input, prepend_bos)
    490     # if tokens.dim() > 1:
    491     tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
--> 492     assert (
    493         tokens.dim() == 1
    494     ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
    495 elif isinstance(input, np.ndarray):
    496     tokens = input

AssertionError: Invalid tokens input to to_str_tokens, has shape: torch.Size([])

After:

(Adds new functionality)

In [8]: HookedTransformer(cfg).to_str_tokens(torch.tensor([2]))
Using pad_token, but it is not set yet.
Out[8]: ['#']

In [9]: HookedTransformer(cfg).to_single_str_token(2)
Using pad_token, but it is not set yet.
Out[9]: '#'

In [13]: HookedTransformer(cfg).to_str_tokens(np.array([2]))
Using pad_token, but it is not set yet.
Out[13]: ['#']

(Does not change existing functionality)

In [23]: HookedTransformer(cfg).to_str_tokens(torch.tensor([[[[[1],[2],[3]]]]]))
Using pad_token, but it is not set yet.
Out[23]: ['"', '#', '$']

In [24]: HookedTransformer(cfg).to_str_tokens(np.array([[[[[1],[2],[3]]]]]))
Using pad_token, but it is not set yet.
Out[24]: ['"', '#', '$']

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -487,13 +487,15 @@ class HookedTransformer(HookedRootModule):
             tokens = self.to_tokens(input, prepend_bos=prepend_bos)[0]
         elif isinstance(input, torch.Tensor):
             tokens = input
-            tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
+            if tokens.dim() > 1:
+                tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
             assert (
                 tokens.dim() == 1
             ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
         elif isinstance(input, np.ndarray):
             tokens = input
-            tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
+            if tokens.ndim > 1:
+                tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
             assert (
                 tokens.ndim == 1
             ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
@@ -512,6 +514,12 @@ class HookedTransformer(HookedRootModule):
         # If token shape is non-empty, raise error
         assert not token.shape, f"Input string: {string} is not a single token!"
         return token.item()
+
+    def to_single_str_token(self, int_token: int) -> str:
+        # Gives the single token corresponding to an int, in string form
+        assert isinstance(int_token, int)
+        token = self.to_str_tokens(torch.tensor([int_token]))
+        return token[0]
 
     def get_token_position(
         self,
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -460,7 +460,12 @@ class HookedTransformer(HookedRootModule):
 
     def to_str_tokens(
         self,
-        input: Union[str, Union[Float[torch.Tensor, "pos"], Float[torch.Tensor, "1 pos"]], list, torch.Tensor, np.ndarray],
+        input: Union[str,
+                     Float[torch.Tensor, "pos"],
+                     Float[torch.Tensor, "1 pos"],
+                     Float[np.ndarray, "pos"],
+                     Float[np.ndarray, "1 pos"],
+                     list],
         prepend_bos: bool = True,
     ) -> List[str]:
         """Method to map text, a list of text or tokens to a list of tokens as strings
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -487,15 +487,13 @@ class HookedTransformer(HookedRootModule):
             tokens = self.to_tokens(input, prepend_bos=prepend_bos)[0]
         elif isinstance(input, torch.Tensor):
             tokens = input
-            if tokens.dim() > 1:
-                tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
+            tokens = tokens.squeeze(dim=1)  # Get rid of a trivial batch dimension
             assert (
                 tokens.dim() == 1
             ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
         elif isinstance(input, np.ndarray):
             tokens = input
-            if tokens.ndim > 1:
-                tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
+            tokens = tokens.squeeze(dim=1)  # Get rid of a trivial batch dimension
             assert (
                 tokens.ndim == 1
             ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -525,7 +525,7 @@ class HookedTransformer(HookedRootModule):
         return token.item()
 
     def to_single_str_token(self, int_token: int) -> str:
-        # Gives the single token corresponding to an int, in string form
+        # Gives the single token corresponding to an int in string form
         assert isinstance(int_token, int)
         token = self.to_str_tokens(torch.tensor([int_token]))
         assert len(token) == 1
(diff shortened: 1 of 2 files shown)


jbloomAus:
Thanks Adam!

- [ ] Can you please explain why this is needed? I get "Makes HookedTransformer.to_str_tokens work on 1x1 tensors and numpy arrays", but I don't get why maybe it hasn't come up yet/what this is needed for. 

- [ ] Can you please verify you can't get the same effect by passing in a dim argument to `squeeze`? Chat GPT thinks that would work. 
Screenshot 2023-04-14 at 4 54 43 pm

- [ ] Depending on the previous point, can you please use ternary statements for neatness? (`tokens = tokens.squeeze() if tokens.dim() > 1 else tokens`) rather than have the if statement on another line.

- [ ] Lastly, can you please write add a unit test for this based on your examples above?

adamyedidia:
> Thanks Adam!
> 
> * [ ]  Can you please explain why this is needed? I get "Makes HookedTransformer.to_str_tokens work on 1x1 tensors and numpy arrays", but I don't get why maybe it hasn't come up yet/what this is needed for.

It seemed inelegant that to_str_tokens works on torch.tensor([1,2]), but not torch.tensor([1]); it was causing my code to break, and also making it hard to write a helper function that converts a single integer into its string token representation (I wrote a function that padded the integer passed in with another dummy integer, packaged them into a tensor of shape 2, and then pruned away the extra dummy integer; but that seemed like an ugly way to write to_single_str_token, so I wanted to fix the core issue, which is that to_str_tokens can't take in one-entry tensors)

> * [ ]  Can you please verify you can't get the same effect by passing in a dim argument to `squeeze`? Chat GPT thinks that would work.
> 
> Screenshot 2023-04-14 at 4 54 43 pm
> 
I think ChatGPT is wrong on this one; when I tried that, I got this stack trace:

IndexError                                Traceback (most recent call last)
Cell In [3], line 1
----> 1 tensor_1x1_to_str_tokens = model.to_str_tokens(torch.tensor([1]))

File ~/TransformerLens/transformer_lens/HookedTransformer.py:491, in HookedTransformer.to_str_tokens(self, input, prepend_bos)
    489     tokens = input
    490     print(tokens)
--> 491     tokens = tokens.squeeze(dim=1)  # Get rid of a trivial batch dimension
    492     assert (
    493         tokens.dim() == 1
    494     ), f"Invalid tokens input to to_str_tokens, has shape: {tokens.shape}"
    495 elif isinstance(input, np.ndarray):

IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)

I had a longer conversation with it and converged on the solution you see in the PR currently. It has the advantage that it correctly handles an input like torch.tensor([[[2]]]), which the old solution didn't.

> * [ ]  Depending on the previous point, can you please use ternary statements for neatness? (`tokens = tokens.squeeze() if tokens.dim() > 1 else tokens`) rather than have the if statement on another line.

Done!

> * [ ]  Lastly, can you please write add a unit test for this based on your examples above?

Done! 

Note that in order to get the tests to pass, I had to add torch.Tensor and np.ndarray to the type signature of the input. Given that the downstream code is considering the possibility that the input is a torch.Tensor or np.ndarray with isinstance, I think the old type signature was wrong. Please let me know if my understanding was incorrect.

Thanks for the detailed comment and help!

jbloomAus:
Thanks, that all seems pretty reasonable. 

A couple of small things but otherwise good to go:

- [ ] I think it might be possible to treat np.ndarray's and torch.Tensor's similarly in the type signature like this
```
Union[
    str,
    Float[torch.Tensor, "pos"],
    Float[torch.Tensor, "1 pos"],
    Float[np.ndarray, "pos"],
    Float[np.ndarray, "1 pos"],
    list,
]
```

See documentation here: https://github.com/google/jaxtyping/blob/c92b0d0ab140cb842b6ef93c7d18450c73f06940/API.md#array

- [ ] Optional, but nice to have, if you use pytest fixtures in your tests this may reduce some code duplication. 

I really appreciate you taking feedback, explaining your reasoning and checking for better implementations!

adamyedidia:
> Thanks, that all seems pretty reasonable.
> 
> A couple of small things but otherwise good to go:
> 
> * [ ]  I think it might be possible to treat np.ndarray's and torch.Tensor's similarly in the type signature like this
> 
> ```
> Union[
>     str,
>     Float[torch.Tensor, "pos"],
>     Float[torch.Tensor, "1 pos"],
>     Float[np.ndarray, "pos"],
>     Float[np.ndarray, "1 pos"],
>     list,
> ]
> ```
> 
> See documentation here: https://github.com/google/jaxtyping/blob/c92b0d0ab140cb842b6ef93c7d18450c73f06940/API.md#array
> 
Done!

> * [ ]  Optional, but nice to have, if you use pytest fixtures in your tests this may reduce some code duplication.
> 
Done!

> I really appreciate you taking feedback, explaining your reasoning and checking for better implementations!

Of course! Thank you very much for all your help!

transformer_lens/HookedTransformer.py
```diff
@@ -488,12 +493,14 @@ def to_str_tokens(
         elif isinstance(input, torch.Tensor):
             tokens = input
             tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
+            tokens = tokens.unsqueeze(0) if tokens.dim() == 0 else tokens  # Don't pass dimensionless tensor
```
neelnanda-io, line 496: Can you make this into a proper if statement? Reading the code it's a bit confusing to see `tokens.squeeze()` and then on the next line `tokens.unsqeeze(0)` before I noticed the inline if

transformer_lens/HookedTransformer.py
```diff
@@ -513,6 +520,12 @@ def to_single_token(self, string):
         assert not token.shape, f"Input string: {string} is not a single token!"
         return token.item()
 
+    def to_single_str_token(self, int_token: int) -> str:
+        # Gives the single token corresponding to an int, in string form
+        assert isinstance(int_token, int)
+        token = self.to_str_tokens(torch.tensor([int_token]))
+        return token[0]
```
neelnanda-io, line 527: Nit: Maybe this should be `token.item()`? `item` returns a single int/float if the tensor has a single element, otherwise raises an error. Here they *should* do the same thing, but it's a good sanity check that you aren't silently throwing away some extra stuff (eg, if you have a dumb tokenizer that somehow maps a single element tensor of tokens to multiple strings.

neelnanda-io:
(This seems like a clearly reasonable feature to exist, thanks Adam!)

adamyedidia:
Huh, the test failure is confusing—I could be wrong, but it looks like it's unrelated to the PR? Something about a connection timeout?

adamyedidia:
Yup, looks like the test was just flaky

jbloomAus:
Thanks @adamyedidia great work!

PR #247, opened 2023-04-15
fix typo in main demo

adzcai:
# Description

In the main demo: fix outdated link to model properties file (link had `easy_transformer` instead of `transformer_lens`)


neelnanda-io (approved):
LGTM!

PR #251, opened 2023-04-19
fix outdated link and shape typo in main demo

daspartho:
In the main demo update the outdated link to the exploratory analysis demo in the activation patching section and fix the shape typo in the transformer architecture section.


jbloomAus:
Thanks!

PR #253, opened 2023-04-22
Implemented functionality for loading mingpt-style models off HF (e.g. othello-gpt)

Baidicoot:
# Description

Added `load_mingpt_weights` function in `loading_from_pretrained`, integrated with loading pipeline

## Type of change

# Issues:

Loading seems broken when loading without `ln_fold=False, center_writing_weights=False, center_unembed=False`; achives ~1e-12 MSE on outputs with the above config, ~1e-2 without.


neelnanda-io:
When you say MSE, do you mean on the logits or on the log probs?
cneter_unembed changes the logits but keeps the log probs the same.

Baidicoot:
Oh, logits, it's probably fine then.

jbloomAus:
This is really awesome, thanks for the PR. 

does it make sense to add a test for this? If it's not a test you would want to run every-time, could you write a test that is set to "skip" that could still be run manually to check that the MSE remains low?

jbloomAus:
This is definitely the plan. New models/architectures are probably our fastest growing feature.

jbloomAus:
I've merged this but had to fork it due to not being able to push tests to your main. Best to do PR's from branches in the future :)

PR #255, opened 2023-04-23
Improving head detector

MatthewBaggins:
# Description

Some improvements and refactoring to `head_detector`. Most notably, I changed the error metrics to one harder to Goodhart (based on conversation on the mechinterp Slack). Also, added many more tests and support for passing multiple sequences to `detect_head`.

## Type of change

    - I changed the error metric in `head_detector` (as suggested by Jay), so I needed

# Checklist:

    - Also, refactored the tests quite extensively
    - Since I changed the error metric, I had to rewrite expected outputs in tests (`test_head_detector.py`)


transformer_lens/head_detector.py
```diff
@@ -1,157 +1,237 @@
...
+      exclude_bos: `True` if the beginning-of-sentence (BOS) token should be omitted from comparison. `False` otherwise. 
+      exclude_bcurrent_token: `True` if the current token at each position should be omitted from comparison. `False` otherwise. 
+      error_measure: "abs" for using absolute values of element-wise differences as the error measure. "mul" for using element-wise multiplication (legacy code).
+    """
+    assert is_square(attention_pattern), f"Attention pattern is not square; got shape {attention_pattern.shape}"
+
+    # mul
+    
+    if error_measure == "mul":
+        if exclude_bos:
```
JayBaileyCS, line 219: Since exclude_bos and exclude_current_token are the same here and in line 232, these can just occur once at the start of the function.

transformer_lens/utils.py
```diff
@@ -657,3 +657,54 @@ def get_dataset(dataset_name: str, **kwargs) -> Dataset:
...
+
+def is_square(x: torch.Tensor) -> bool:
+    """Checks if `x` is a square matrix."""
+    return x.ndim == 2 and x.shape[0] == x.shape[1]
+
+def is_lower_triangular(x: torch.Tensor) -> bool:
+    """Checks if `x` is a lower triangular matrix."""
+    if not is_square(x):
+        return False
+    return (x - x.tril()).sum().item() == 0
```
JayBaileyCS, line 669: This should assert x - x.tril() is the zero matrix, rather than just summing to 0.

JayBaileyCS (commented):
Looks good - just added a couple of minor comments. Will still need Neel/Joseph's call to merge if I understand TransformerLens correctly.

transformer_lens/utils.py
JayBaileyCS: Alternatively, could you just assert x = x.tril? I'm not actually sure here.

jbloomAus:
Hey sounds good. A few comments:

- [ ] 1. Can you explain how the metrics are now harder to goodhart? 
- [ ] 2. As annoying as it sounds, can you please write tests for the two functions added to utils? Moving forward I'm trying to be strict about this on principle. 
- [ ] 3. What was the configuration of your pylintrc file? I'd be open to a PR with suggestions for one... :D 

Things I liked about this PR:
- Many separate commits are good practice!

transformer_lens/utils.py
MatthewBaggins: `x.equal(x.tril())` will do it, thanks!

transformer_lens/head_detector.py
MatthewBaggins: They operate on different tensors, depending on whether `error_measure` is `"mul"` or `"abs"`. I don't see a clear way of merging them.
The best I can come up with is to delegate them to some external function `_exclude_bos_or_current_token` or something but it is so specific that it would probably make the code more obscure instead of clarifying it.

MatthewBaggins:
> 1. Can you explain how the metrics are now harder to goodhart?

Until now, the score saying how well the detection pattern matches each attention head is calculated as the fraction of the attention pattern kept. So you can goodhart it by keeping all attention, i.e., by passing a detection pattern of all ones (or at least along the main diagonal and below it). The new metric is 1 - mean absolute difference between the attention pattern and the detection pattern. It ranges from -1 to 1, where 1 is perfect match and -1 total mismatch.

> 2. As annoying as it sounds, can you please write tests for the two functions added to utils? Moving forward I'm trying to be strict about this on principle.

No problem, I'm very sympathetic to that approach. Done.

> 3. What was the configuration of your pylintrc file? I'd be open to a PR with suggestions for one... :D

It's mostly basic settings + disabling some warnings (like `missing-module-docstring`) that are not relevant for the project and making valid names for arguments, variables, and attributes more generic because pylint's default snake_case doesn't include names like `x` or `t1`.

neelnanda-io:
> Until now, the score saying how well the detection pattern matches each
attention head is calculated as the fraction of the attention pattern kept.
So you can goodhart it by keeping all attention, i.e., by passing a
detection pattern of all ones (or at least along the main diagonal and
below it). The new metric is 1 - mean absolute difference between the
attention pattern and the detection pattern. It ranges from -1 to 1, where
1 is perfect match and -1 total mismatch.

Hmm, I think I disagree with this. If you have a head which eg attends to
the subject of a sentence, but that sentence is sometimes a single token
and sometimes two tokens, I want to be able to give a mask which has 1s on
all tokens of the subject, but which does not give a different score for
two vs one token words. And I believe the boolean mask swtup allows this
but your's doesn't?

More generally I don't care about this kind of Good Harting. If people set
a dumb metric that's their problem

MatthewBaggins:
@JayBaileyCS do you think there are cases, where somebody who is not trying to goodhart the detector would rather use the new metric instead of the old one?
(for context: changing the metric was Jay's idea)

(also, the commit as is allows you to use the old metric, you just need to specify it when calling `detect_head`)

neelnanda-io:
For context, I have used the "total attention paid to any token in this
word" trick when eg looking for translation heads. Having the option
available seems good to me though

MatthewBaggins:
So how about we make the old method the default and keep the option to use the new one?

neelnanda-io:
Sounds good to me!

JayBaileyCS:
So, the Goodharting was what caused me to notice there was a problem, not being the problem in and of itself.

The advantage of switching to absolute prediction error is that you can make more precise predictions, and have that measured in the score. You can predict, for instance, 0.2 attention to X, and 0.8 attention to Y, and your score will be better if your prediction is closer - the original "mul" metric does not allow this, you'll get the same score if attention is 0.2, 0.8 or 0.5, 0.5 or 0.8, 0.2.

However, your example from above makes a lot of sense Neel. What you could do with absolute error would be:

- Write code to identify the subject in the sentence.
- Determine how many tokens that subject has.
- Create your prediction based on that (i.e, 1 on subject if one token, 0.5, 0.5 if two tokens)

And this would work with absolute error. However, this is also more complicated and time-intensive than "Okay, I want to see how much attention is paid to this word in particular, and I don't care exactly how much attention is paid to which token yet" in which case the original metric is better.

So, I think that for quick or exploratory investigations, the original metric may be superior. For precise examinations where you're trying to reproduce as much functionality as possible or really test your understanding of the attention head, absolute error is best.

As a result, I agree with Neel that the original metric makes sense as the default option, with absolute prediction error existing for people who want to do more precise investigations. Whichever metric is best to start with, that should be the default option. Absolute error is there for people who want to dive in more.

It'd be nice to adjust the demo to make this clear, but I am happy to do that myself after this PR goes through, since I have the most context about it.

transformer_lens/head_detector.py
```diff
@@ -1,157 +1,237 @@
...
+def detect_head(
+    model: HookedTransformer,
+    seq: Union[str, List[str]],
+    detection_pattern: Union[torch.Tensor, HeadName],
+    heads: Optional[Union[List[LayerHeadTuple], LayerToHead]] = None,
+    cache: Optional[ActivationCache] = None,
+    *,
+    exclude_bos: bool = False,
+    exclude_current_token: bool = False,
+    error_measure: ErrorMeasure = "abs",
```
JayBaileyCS, line 38: As per Neel's suggestion, default should be "mul".

JayBaileyCS (changes requested):
Thanks Matt, much appreciated! Looks good to me once the change you and Neel talked about (i.e, "mul" as default) is made.

transformer_lens/head_detector.py
```diff
@@ -1,158 +1,269 @@
...
+
+    1. `"mul"` (default) multiplies both tensors element-wise and divides the sum of the result by the sum of the attention pattern.
+    Typically, the detection pattern should in this case contain only ones and zeros, which allows a straightforward interpretation of the score:
+    how big fraction of this head's attention is allocated to these specific query-key pairs?
+    Using values other than 0 or 1 is not prohibited but will raise a warning (which can be disabled, of course).
+    2. `"abs"` calculates the mean element-wise absolute difference between the detection pattern and the actual attention pattern.
+    The "raw result" ranges from 0 to 2 where lower score corresponds to greater accuracy. Subtracting it from 1 maps that range to (-1, 1) interval,
+    with 1 being perfect match and -1 perfect mismatch.
+
+    **Which one should you use?** `"abs"` is likely better for quick or exploratory investigations. For precise examinations where you're trying to
```
JayBaileyCS, line 55: This should be `"mul` is likely better for quick or exploratory investigations.

JayBaileyCS (changes requested):
Sorry to be a stickler on this, but there is one issue with the docstring - it's not just cosmetic, it could actually confuse a new user.

transformer_lens/head_detector.py
MatthewBaggins: thanks! done

jbloomAus:
Closing this as it seems done. Good work! 

A note for future work: I think PR's and conversations are best motivated in relation to existing research. The space of features is much larger than the space of features that are directly useful to research. That's not to say this conversation/PR wasn't valuable, but I want to try to make sure we have mental space/room to find work motivated by progress in MI possibly more closely. 

By analogy, companies with software products must be very careful about which features they accept, even going so far as saying things like "if 100 people asked for a feature, don't' building it, wait for 1000" (insert numbers relevant to the scale of the operation.  This is especially important for software because of the capacity / maintenance cost which is often underestimated. 

In the future, I'm going to try to establish the value of things like this more strongly so that people can grade invested effort accordingly. I also think this is a hard thing to do so I'll approach the topic with humility and a collaborative mindset.

PR #257, opened 2023-04-26
Add tensor information to typeguard error messages

### LINKED ISSUE #190: Jaxtyping argument consistency checks are enabled at test time but not runtime
## Problem statement

Jaxtyping [multi-argument consistency checks](https://github.com/google/jaxtyping/blob/main/API.md#runtime-type-checking) are enabled at test time but not at runtime, resulting in different behaviour. This prevents us from testing all runtime behaviour in unit tests.

For example, consider calling the [FactoredMatrix](https://github.com/neelnanda-io/TransformerLens/blob/main/transformer_lens/FactoredMatrix.py) constructor with a mismatched inner dimension:
```python
A = torch.ones((2, 3, 4))
B = torch.ones((2, 3, 5))
FactoredMatrix(A, B)
```

At runtime, this will lead to an AssertionError due to an assertion made in the constructor:
```python
AssertionError: Factored matrix must match on inner dimension, shapes were a: torch.Size([2, 3, 4]), b:torch.Size([2, 3, 5])
```

However, at test-time, when running [`test_factored_matrix_inner_mismatch`](https://github.com/rusheb/TransformerLens/blob/f11c3792d9a99b1f94689279b1756fe4fab9b93d/transformer_lens/tests/factored_matrix/test_constructor.py#L63) (note: this test is not yet merged into main), we see a different error:
```python
TypeError: type of argument "B" must be jaxtyping.Float[Tensor, '... mdim rdim']; got torch.Tensor
```

The TypeError is triggered by jaxtyping/typeguard runtime typechecking on the constructor arguments. This error is thrown before we reach the assertion, meaning the AssertionError is not thrown. 

## Mitigation
It's not clear what the ideal solution is. I think these are the options:

* **Enable multi-argument consistency checks always.** This could be achieved using `install_import_hook` or the `@jaxtyping` and `@typechecked` annotations. However, I'm not sure what unintended sied-effects this would cause. On another note, the error message for the TypeError is actually misleading/confusing. The AssertionError has a much more clear and descriptive message. 

* **Disable multi-argument consistency checks always.** I'm not sure how we would disable it at test time. We need to get to the bottom of why it is occurring. I suspect it is related to calling `hook = install_import_hook("transformer_lens", ("typeguard", "typechecked"))` from the test files e.g. `test_stop_at_layer.py`. However in each of these files, the call to `hook.uninstall()` at the end of the file should prevent this from affecting other files.

* **Suppress runtime typechecking in specific tests only.** Typeguard 3 supports [typecheck suppression](https://typeguard.readthedocs.io/en/latest/userguide.html#suppressing-type-checks) However typeguard 3 is still a release candidate; we would need to upgrade.

---

dkamm:
# Description

Fixes #190

This PR adds tensor argument shape and dtype information to the typeguard error message by providing  a `typeguard_fail_callback` function to typeguard.

For example
```
@typechecked
def tensor_foo(
    x: Float[torch.Tensor, "a b"],
    y: Float[torch.Tensor, "b c"],
    z: Float[torch.Tensor, "c d"],
    ) -> Float[torch.Tensor, "a d"]:
    return x @ y @ z

tensor_foo(torch.randn((1,2)), torch.randn((2,3,5)), torch.randn((3,4)))    
```

will print

```
got error: argument "y" (torch.Tensor) is not an instance of jaxtyping.Float[Tensor, 'b c']
tensor argument info:
        x  shape=(1, 2)     dtype=torch.float32
        y  shape=(2, 3, 5)  dtype=torch.float32
        z  shape=(3, 4)     dtype=torch.float32
```

The callback feature was introduced in typeguard 3, so I upgraded typeguard to 3.0.2, which required changing the python requirement slightly (from >=3.7 to >=3.7.4)

I first considered monkeypatching typeguard, but this solution seems adequate and much less brittle.  I'd like to print something like "Shape mismatch" too, but I'd have to rewrite the logic from jaxtyping to detect that is the case.

One thing to consider- to enable typechecking, we have to add that `install_import_hook("transformer_lens", ("typeguard", "typechecked"))` bit everywhere we use transformer lens.  Not sure if there's any way around that.

Note: I also want to include the type hints in the error message, but for some reason they seem to be removed from typeguard 4.  I'm inquiring why here: https://github.com/agronholm/typeguard/discussions/345 

## Type of change

### Screenshots
N/A

# Checklist:

### DIFF
--- a/tests/unit/test_typecheck.py
+++ b/tests/unit/test_typecheck.py
@@ -1,37 +1,41 @@
 import re
 from textwrap import dedent
+
 import pytest
 import torch
 import typeguard
-from typeguard import typechecked
 from jaxtyping import Float
+from typeguard import typechecked
 
 from transformer_lens.typecheck import typecheck_fail_callback
 
 typeguard.config.typecheck_fail_callback = typecheck_fail_callback
 
+
 @typechecked
 def foo(x: int) -> int:
     return x + 1
+
 
 @typechecked
 def tensor_foo(
     x: Float[torch.Tensor, "a b"],
     y: Float[torch.Tensor, "b c"],
     z: Float[torch.Tensor, "c d"],
-    ) -> Float[torch.Tensor, "a d"]:
+) -> Float[torch.Tensor, "a d"]:
     return x @ y @ z
 
-def test_error_message_is_correct_with_tensor_args():
 
-    ref_err_msg = "argument \"x\" (str) is not an instance of int"
+def test_error_message_is_correct_with_tensor_args():
+    ref_err_msg = 'argument "x" (str) is not an instance of int'
 
     with pytest.raises(typeguard.TypeCheckError, match=re.escape(ref_err_msg)):
         foo("a")
 
-def test_error_message_is_correct_without_tensor_args():
 
-    ref_err_msg = dedent("""argument "y" (torch.Tensor) is not an instance of jaxtyping.Float[Tensor, 'b c']
+def test_error_message_is_correct_without_tensor_args():
+    ref_err_msg = dedent(
+        """argument "y" (torch.Tensor) is not an instance of jaxtyping.Float[Tensor, 'b c']
 tensor argument info:
 \tx  shape=(1, 2)     dtype=torch.float32
 \ty  shape=(2, 3, 5)  dtype=torch.float32
@@ -39,4 +43,4 @@ tensor argument info:
     )
 
     with pytest.raises(typeguard.TypeCheckError, match=re.escape(ref_err_msg)):
-        tensor_foo(torch.randn((1,2)), torch.randn((2,3,5)), torch.randn((3,4)))
+        tensor_foo(torch.randn((1, 2)), torch.randn((2, 3, 5)), torch.randn((3, 4)))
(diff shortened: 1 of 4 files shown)


dkamm:
Hm, this is failing because typeguard doesn't work on dataclasses.  Not sure how this didn't come up before

jbloomAus:
Let me know how you go fixing that issue. Maybe @rusheb  can provide support. If you need specific clarification or suggestions comment here, otherwise I'll assume the PR is just delayed a bit. 

I appreciate the effort you are putting in here!

rusheb:
Hey, likewise happy to pitch in or rubber duck if you'd find that helpful!

dkamm:
@rusheb - let's do it! will ping you to coordinate

JayBaileyCS:
A recent change has been pushed to fix the Github Actions failing issue - if you merge the latest changes, they should work now.

jbloomAus:
@dkamm where are we at with this? If you rebase, tests may pass if the reason they aren't passing was CI env memory which I sort of fixed (but is still something I'm working on).

alan-cooney:
Closing as stale (and happy for this to just be done with just a note in the docs rather than adding the code in). But feel free to open again if you want!

PR #259, opened 2023-04-26
fix outdated link in Exploratory Analysis Demo

daspartho:
In the exploratory analysis demo, update the outdated link to the Main_Demo in the introduction section and Activation_Patching_in_TL_Demo in the activation patching section.


jbloomAus:
thanks, good catch.

PR #260, opened 2023-04-27
Optimize imports in HookedTransformer

rusheb:
# Description

Delete a bunch of unused imports in HookedTransformer.py

No issue.

## Type of change

- Refactor

# Checklist:

- ~I have commented my code, particularly in hard-to-understand areas~
- ~I have made corresponding changes to the documentation~
- ~have added tests that prove my fix is effective or that my feature works~
- ~New and existing unit tests pass locally with my changes~
- ~I have not rewritten tests relating to key interfaces which would affect backward compatibility~

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -1,36 +1,23 @@
-from typing import Callable, Union, List, Tuple, Dict, Optional, NamedTuple, overload
-from typing_extensions import Literal
-import torch
-import torch.nn as nn
-import torch.nn.functional as F
-import numpy as np
-import einops
-import logging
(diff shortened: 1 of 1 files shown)


rusheb:
Hmm...i just realised, maybe all these imports are unused due to the star import `from TransformerLens.components import *`. So actually it might be better to remove the star import. I may do this anyway as part of [splitting each component into a different file](https://github.com/neelnanda-io/TransformerLens/issues/258#issuecomment-1524874813)

jbloomAus:
+ 1 to avoiding import *. If you remove that and then make appropriate imports so everything works then will accept. Based on other chat, I'm not expecting splitting up the components file.

Issue #258, opened 2023-04-26
[Proposal] Support BERT Model Series (and/or T5).


jbloomAus:
<img width="489" alt="Screenshot 2023-04-26 at 8 00 40 pm" src="https://user-images.githubusercontent.com/69127271/234541975-b5f48a88-0354-4e2c-a060-0deb52d60de8.png">

### Proposal 

A glaring gap in models supported by TransformerLens is the lack of support for encoder/decoder models like BERT and T5. In order to rectify this, we need to build support for encoder-decoder models, cross attention and likely a few other details. 

### Motivation

- I've been asked a number of times by people whether we have T5 or BERT support coming. 
- I think this is a reasonable contribution for people skilling up. You can ask for help or collaborate with other people working on building it. 
- I (Joseph) and likely others will be able to give feedback on your work and pitch in if needed! 

### Pitch

In order to complete this card we are going to build a class called HookedEncoderDecoder which can be instantiated and have the BERT or T5 weights from [huggingface](https://huggingface.co/bert-base-uncased) loaded in. This is the same strategy the package uses for other models so there's nothing fundamentally new, only components we haven't yet built with hooks. Once we have a class that has corresponding weights as needed, we will want to ensure all components are "hooked" correctly, enabling methods like .run_with_cache to work. It's not as complicated as it sounds to get it working. 

Possibly the "best" version of this is tricky but we're willing to tolerate some code duplication as long we're confident the implementation is accurate (ie: tests show the hooked activations are correct and inference is identical to what we get without using TransformerLens. 

Tasks:

- [ ] Make a fork for BERT support (keep it synced if building this takes more than a moment)
- [ ] Write some tests to be used during development, these should cover any novel architecture components these will involve during development. I'd start with two placeholder tests. 1. for do you have a model architecture which you can load weights into. and 2. if you had that architecture, would it produce identical results to the hugging-face model. 
- [ ] Load the BERT model, look at each component. Maybe read the BERT paper if needed to get context. Make a list of every nn.module that we don't currently support. Some items that will be on the list. 1. a linear block before the unembed, post-layerNorm (can't be unfolded RIP), sinusoidal positional embeddings. T5 might have other stuff.
- [ ] Keep in mind you might need a new config class to configure the entire architecture. If so, build it. 
- [ ] If you can pass the test for the model doing inference correctly and show that you can use .run_activation_cache and all the other methods that work on hooked transformer (within reason), then we're likely ready to merge. 

Some things to note:
[...]

PR #262, opened 2023-04-28
Introduce HookedEncoderConfig (Issue #258)

rusheb:
# Description

Hi! **I'm raising this PR to get feedback on the design choices I've made in order to support BERT before I proceed with the rest of the implementaiton.**

Ideally I would like to work incrementally and avoid branch drift, so if we can merge this that would be a bonus. However maybe these changes are too half-baked to merge.

This PR introduces the `HookedEncoderConfig` class which will be used to store the configuration of a `HookedEncoder` model. I've added the `TransformerLensConfig` as a common interface for both `HookedTransformerConfig` and `HookedEncoderConfig` classes.

I've updated  `Embed` to support initialization with either a `HookedTransformerConfig` object, a `HookedEncoderConfig` object, or a dictionary representing either configuration. I've added unit tests to verify this behaviour. 

I've also added an acceptance test to ensure that the output of the `Embed` component initialized with a HookedEncoderConfig matches the token embedding of a Hugging Face BERT model.

I'm hoping this is a neat way of supporting different model configs which doesn't break existing behaviour or interfaces. But let me know what you think! 

## Summary of changes:
- Add `HookedEncoderConfig` class
- Add `TransformerLensConfig` ABC
- Update `Embed` component to support initialization with either configuration object
- Add unit tests for `Embed` initialization
- Add acceptance test for custom `Embed` component
- Rename `test_transformer_lens.py` to `test_hooked_transformer.py`

Issue #258 

## Type of change

# Checklist:


luciaquirke:
What's the reason behind adding a second config class? Are there plans to add configuration properties that would be incompatible with the ones in HookedTransformerConfig? 

If you do go down this path, two naming suggestions:
- Hide lineage: `HookedDecoderConfig` and `HookedEncoderConfig` rather than `HookedTransformerConfig` and `HookedEncoderConfig`.
- Don't reference the project name in source code: `TransformerLensConfig` -> `HookedTransformerConfig`.

neelnanda-io:
> Hide lineage: HookedDecoderConfig and HookedEncoderConfig rather than
HookedTransformerConfig and HookedEncoderConfig

I think this is a bad idea, because it breaks backwards compatibility (I
think? Not doing that seems very important!). HookedTransformerBaseConfig
might work

I'm pretty fine with GPT-2 style transformers being the main first class
citizens in TransformerLens, and encoder, and encoder-decoder models being
bonus supported things.

I'm open to HookedEncoderConfig just being part of HookedTransformerConfig,
or adding a HookedBERTConfig field to HookedTransformerConfig (nested
dataclasses are fine, right?), given the massive overlap and desire to have
good backwards compatibility. But I also see the point of avoiding an
enormous, bloated config class.

One option might be adding a "print important config" method to the class,
so users can easily inspect it and see key hyper-params

jbloomAus:
I think backwards compatibility is definitely the top priority based what @neelnanda-io is saying. We have more to lose than to gain if this creates work for people/makes life harder with existing research. 

To that end, it seems like this is a good step in the direction of supporting BERT with minimal interface changes elsewhere (which is a clearly stated requirement of PRs, see the checkbox). I like that you're adding tests for the existing functionality too (was going to ask you to do so if you hadn't but you have).

Why not put the `from_dict` logic in the config base class? Seems identical between the config classes. 

@luciaquirke I understand the appeal of naming things that way but I think for the reasons above we're going to proceed cautiously with respect to touching existing code. I think the PR strikes a good balance. 

@rusheb how far have you gotten into making use of the new config class? I'm wondering what the probability is that there related changes down the line which makes accepting this PR less consequential than it appears. I like the move toward making a PR quickly and the changes do seem absolutely fine but the codebase is quiet enough that you could get further along with little to no cost before we start merging stuff.

jbloomAus:
also @rusheb In general I aspire to give specific feedback which is more detailed than I am able to make right now. Feel free to highlight any choices you were unsure about if there were any.

rusheb:
## Summary
Thanks so much for the comments everybody! I really appreciate you all giving feedback! 

Having read through and thought about all your points, I think I'm for keeping everything roughly the same. There are two things that I think we should *maybe* do
- Reuse/extend `HookedTransformerConfig` rather than creating a separate config class
- Rename `TransformerLensConfig` to `HookedTransformerBaseConfig`

Probably the most important question is whether to use a separate config class. I think it's not clear how the configs and validation will diverge, and that it would be easier to unify them later than separate them later, so I'm for keeping them separate. **If anybody has arguments in favour of unifying the configs please tell me!**

In terms of whether to merge, I don't think there is much risk associated with merging now, and there might be some benefit (trunk development and early feedback if I've broken something). So **I think we should merge this PR once everyone is happy.**

Detailed discussion below. Let me know if you disagree!  
## Details
### Separate config class
- I made a couple of assumptions when deciding to create a new config class, rather than folding it in
	- `HookedTransformerConfig` does a lot of input validation. I assumed that this validation was not guaranteed to apply to future models.
	- `HookedTransformerConfig` has some required properties. I assumed these were not necessarily going to be required in future models.

If either of these are false or don't matter, then it might make sense to merge them. As Neel points out, there is still the argument that one massive config would become bloated. 

I think keeping them separate might be a safer option. I think I'm happy to accept some duplication to keep separate concerns separate. In future, if there's loads of duplication, we might be able to use Mixins or similar to reduce this (I think that might end up being a really neat way to handle the duplication with composition). Failing that, I think it would be easier to combine configs in future than separate them.

### Naming
- I agree that backwards compatibility is most important.
- I agree with GPT-2 style transformers being the first-class citizen
- We could create `HookedEncoderConfig` as an alias to `HookedTransformerConfig`. But, if we did this then we would want to deprecate `HookedTransformerConfig` to avoid violating "there should be one obvious way...". I personally don't think it's worth doing this, but I'm not confident in my view. 
- I'm open to using the name `HookedTransformerBaseConfig`, but it may be confusing if this is implemented by `HookedEncoderConfig`?

### from_dict() on TransformerLensConfig
There were 2 suggestions around this:
1. move `config_from_dict` as a classmethod on TransformerLensConfig
	- I actually tried this but I hit circular import issues. I think it's out of scope to debug these. I could put this logic a factory class but I think it'd be overkill.

2. move `from_dict` and `to_dict` methods of `HookedEncoderConfig` and `HookedTransformerConfig` to `TransformerLensConfig`
	- I think I'd rather avoid inheritance keep the base config abstract. I don't think the duplicated logic is a big issue

### Merge now/merge later
- Seems like nobody is against merging now. 
- As Joseph pointed out, merging now would allow us to test the waters incrementally.
- I haven't tested out the new config class that much yet, but i think it'd be fine to refactor in future PRs

luciaquirke:
Nice write up Rusheb, sounds like you're on the right track 😄

> I think this is a bad idea, because it breaks backwards compatibility (I think? Not doing that seems very important!). HookedTransformerBaseConfig might work

Yep agree, I didn't have my open source hat on there.

> arguments in favour of unifying the configs

- To keep things simple we should only add new components if there's a clear and present reason. Unless the HookedTransformerConfig is incompatible with the BERT implementation, i.e. if the issues around input validation and required properties apply to BERT as well as to future models, I don't see a present reason. 
- I don't think we should write scaffolding for more complex future models because planned features often don't materialize and the associated tech debt tends to be left as is. 
- I have a heuristic in my head that splitting is easier than merging, not sure if it holds in libraries with backwards compatibility requirements but thought I'd mention it because places I've worked have had merging nightmares 😂

rusheb:
Okay, thanks Lucia! I'm slightly updating towards unifying the configs.

I think answers to the following questions (paraphrased from above) may be required to make the decision:
- Does HookedTransformerConfig contain any required properties that HookedEncoderConfig will not require?
- Will the input validation done by HookedTransformerConfig apply to the HookedEncoder?

@neelnanda-io @jbloomAus do you know the answer to these?

If not, probably the way to find out is by implementing more! (Locally I've done the embeddings so far, so next step is the BERT block.)

neelnanda-io:
Er, I think nothing required will clash, and that the validation should
still work?

jbloomAus:
I don't expect unresolvable clashes if there are any but it's the kind of
thing where, as you say, you implement to get the understanding.

jbloomAus:
Closing based on a discussion with @rusheb

PR #266, opened 2023-05-04
Add Jupyter Notebook regression tests

### LINKED ISSUE #213: [Proposal] Find a way to run all the tutorials and check that they don't generate errors
### Proposal 

Find a way to run all the tutorials and check that they don't generate errors.

### Motivation

This is a good way to avoid breaking compatibility with the tutorials which also stand in as an acceptance test. 

### Alternatives

We write more tests and be careful when we edit tutorials. 

### Additional context

We probably have had some silent errors where notebooks stop working and I'd like to avoid this in the future. 

### Checklist

- [x ] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/MiniGrid/issues) in the repo (**required**)

bryce13950:
# Description

Full test suite has been implemented for all ipynb files. Along with that a starter template has been added that will help make adding all future demos easy in the new testing suite. This has also gong through quite a bit of changes as I have gone, and I do plan to continue improving this now that the first version of these tests is passing. e.g. I learned later about the ".to" function, so there are some checks using the `IN_GITHUB` variable that can be simplified. 

These tests do take a really long time to execute though, and you cannot replicate the execution locally without having it run in the GItHub specific environment due to unknown settings they have in place that affect the execution. My plan is that as I have more time to work on this project to make some changes when I start for the day, do something else on the project while the GitHub actions run, and as I improve the individual demos, submit smaller more digestible pull requests. The goal is to figure out a way to remove as many of the `IN_GITHUB` flags as I can. For the time being, everything is passing, and everything is ready for new demos to be submitted with minimal inconvenience on the creator.

Fixes #213

## Type of change

### Screenshots

# Checklist:


jbloomAus:
@bryce13950 Thanks again for this! I think what I'd like is for someone to go through each and see that the output is scientifically as expected. I'll tag some people on the slack. 

- [ ] Activation_Patching_in_TL_Demo.ipynb
- [ ] Attribution_Patching_Demo.ipynb
- [ ] Exploratory_Analysis_Demo.ipynb
- [ ] Grokking_Demo.ipynb
- [ ] Head_Detector_Demo.ipynb
- [ ] Interactive_Neuroscope.ipynb
- [ ] LLaMA.ipynb
- [ ] Main_Demo.ipynb
- [ ] No_Position_Experiment.ipynb
- [ ] Old_Demo.ipynb
- [ ] Othello_GPT.ipynb
- [ ] Tracr_to_Transformer_Lens_Demo.ipynb

jbloomAus:
I'm optimistic I'll get people to look at these eventually :) @bryce13950  sorry it's taking so long.

bryce13950:
No worries! I should have some time to mess around with it a bit next week. The execution of this was basically untouched in anything but GitHub, so there shouldn't be any differences. Some of these did not work before though, so that may be a bit of challenge to track down when they last worked.

bryce13950:
@alan-cooney I would discuss some of this with @jbloomAus. We had discussions about this topic, and decided that separating them on branches was going to make them harder to maintain in the future. As for the length of time, the purpose of separating the notebooks onto their own runner group is to mitigate this issue. The tests for the code checks are still run first, and still fail first in case a change in a pull request breaks an existing test. Within 5-10 minutes the developer will know if they are going to break the tests in 95% of the cases.

If we want to speed up these tests as is, there is a flag in all of these files that we can use to skip parts that take a long time, which are not important to test. As for converting them to py files, and then running them that way, I do not believe that is going to achieve the desired effect of making this code test faster, since the underlying reason these take so long to test is due to how much stuff is actually happening in these sets of code. As is, to run these in an environment like colab, some of these can still take close to an hour to fully complete. One thing that would speed up everything is to check if there are parts of the code that are iterating a lot, and dropping that number down drastically when run in GitHub. I did go through and do some of that, but I did not do it systematically. It is on my list of things to do to improve the second iteration of this. 

For now, I want to get it in the code base, so I don't have to worry about continuously merging large changes into these files. The goal is to make those small incremental improvements to individual examples after this large change is merged, so that those improvements can be more easily managed. In order to mitigate this time need for the time being, we could also set it up, so that this test group only runs when we are ready to publish a release. There were a number of issues with these files when I went through them, but they are going to likely break very infrequently now that they have been fixed, so only running this test group periodically would probably suffice in achieving the point while keeping the day to day pull request workflow unimpeded.

jbloomAus:
@alan-cooney @bryce13950 

> As a general rule, entire CI/CD workflows should take less than 10 minutes to run (ideally less than 5). This is because if someone submits a PR they should be able to grab a tea and then come back and see any issues, so that they can make changes and work synchronously. If the checks take much longer, it means that they have to leave it and come back later, which creates a lot of friction and slows down work a lot.

This is generally true but I wanted to have the tutorials tested to track when we were breaking them. Doesn't have be triggered on every PR, maybe on releases? Would that work @alan-cooney ?

> In addition, notebooks are just not great for version control and it's really hard to see what has changed between revisions. I'm thinking a better approach here may be to store the notebooks as .py files and then first run a tool that converts them to notebooks (and publishes this to e.g. an "examples" branch), and then secondly runs your checks here. It would be great if you're able to make both changes (either here or as separate PRs), and I can review that (and ask Neel to update his links to the notebooks to keep them working as well).

I don't think I'm opposed to python files (or python files that produce jupyter notebooks on a separate branch). What I was advising against was always having all of these files on a separate branch (not sure if that's being suggested).

alan-cooney:
Hey both,

Responses here, and I've also suggested a high level approach to resolve this:

> The tests for the code checks are still run first, and still fail first in case a change in a pull request breaks an existing test. Within 5-10 minutes the developer will know if they are going to break the tests in 95% of the cases.

The problem is that since these notebooks use a lot of the underlying functionality of TransformerLens, it's pretty likely that they will fail if e.g. a bug is introduced (the codebase otherwise has relatively sparse testing). This is of course assuming that we get them to install from the local directory rather than from using the pypi/main branch version.

> If we want to speed up these tests as is, there is a flag in all of these files that we can use to skip parts that take a long time, which are not important to test. 

Great yes I think this is the better approach - to skip things that take too long. Let's do this in this version at a minimum.

> As for converting them to py files, and then running them that way, I do not believe that is going to achieve the desired effect of making this code test faster, since the underlying reason these take so long to test is due to how much stuff is actually happening in these sets of code. 

This part is not to make them run faster, it's just to make it easy for maintainers to review changes to the notebooks (as you can then view the diffs).

> This is generally true but I wanted to have the tutorials tested to track when we were breaking them. Doesn't have be triggered on every PR, maybe on releases? Would that work @alan-cooney ?

I think they'll regularly break with bugs introduced in the underlying lib, so I think it's better to run on every PR.

### Suggested Approach

The way I think we should tackle this is to convert all the notebooks to .py files (using e.g. jupytext) and have those as the checked in versions. These are easy to use and view changes. Then we have 2 GitHub action pipelines:

#### Notebook checks pipeline

This can be added to the existing checks pipeline (runs on all PRs) and does the following:

1. Build the notebooks using jupytext
2. Config pip to use the local version of TransformerLens (not from PyPi or GitHub)
3. Run pytest with nbmake (or alternatively we may be able to test the notebooks directly now they are .py files)

#### Notebook deploy pipeline

This runs after checks on every push to main where the notebooks have changed, and does the following:

1. Build notebooks using jupytext
2. Publish changed notebooks to a different protected branch (e.g. `notebooks`) that only contains built notebooks. This is simply a place to store the deployed notebooks, so that they can be used directly with e.g. Colab.

alan-cooney:
Closing as stale, but feel free to re-open if you want

PR #268, opened 2023-05-04
Fix `from_pretrained` with `redwood_attn_2l`

ArthurConmy:
# Description

Currently it is not possible to load `redwood_attn_2l` with `from_pretrained` because the `fold_ln` parameter needs to be rewritten to `False` in two places, and it is only rewritten in one place. Similarly, centering weights needs to be handled properly. 

Regarding the checklist, I think the warnings describe what is happening, but lmk if this is not clear.

Fixes [# (issue)](https://github.com/neelnanda-io/TransformerLens/issues/267)

## Type of change

- [ x  ] Bug fix (non-breaking change which fixes an issue)
- [ x  ] New feature (non-breaking change which adds functionality)

# Checklist:

- [   ] I have commented my code, particularly in hard-to-understand areas
- [   ] I have made corresponding changes to the documentation
- [   ] My changes generate no new warnings
- [ x ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -749,9 +749,17 @@ class HookedTransformer(HookedRootModule):
             n_devices=n_devices,
         )
 
-        if cfg.positional_embedding_type == "shortformer" and fold_ln:
-            logging.warning("You tried to specify fold_ln=True for a shortformer model, but this can't be done! Setting fold_ln=False instead.")
-            fold_ln = False
+        if cfg.positional_embedding_type == "shortformer":
+            if fold_ln:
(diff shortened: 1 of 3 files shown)


transformer_lens/HookedTransformer.py
```diff
@@ -848,12 +860,6 @@ def load_and_process_state_dict(
             self.cfg.n_devices == 1 or move_state_dict_to_device
         ), "If n_devices > 1, move_state_dict_to_device must be True"
 
-        
-        if self.cfg.positional_embedding_type == "shortformer":
```
ArthurConmy, line 852: Note that this override of fold_ln is moved earlier in the file

jbloomAus:
All good! thanks for this!

PR #269, opened 2023-05-04
Upgrade to typeguard 3

dkamm:
# Description

- Upgrades typeguard from 2 to 3 
- Uses typeguard's pytest plugin to run typeguard on all tests (no need for `install_import_hook` stuff)
- Fixes various unit test failures

Some typeguard notes:

- if you reassign an input variable to a different type, you need to add an annotation. otherwise, typeguard complains
```python
def foo(x: int):
  x = str(x)      # fails
  x: str = str(x) # passes
```
- typeguard 3 doesn't work with `property` tags so you need to add `@typeguard_ignore` annotation. this might get fixed in typeguard 4
```python 
from typeguard import typeguard_ignore
class Foo:
  @property
  @typeguard_ignore
  def foo(self):
    pass
```

## Type of change

### Screenshots
N/A

# Checklist:


dkamm:
@rusheb @jbloomAus I decided to split this out into its own PR to make it more focused/reviewable

jbloomAus:
thanks, I want to look over this properly since so many files are affected, will do when I can.

jbloomAus:
Thanks @dkamm !

PR #274, opened 2023-05-12
Wrap docstrings and comments in HookedTransformer

luciaquirke:
# Description

While using TransformerLens I often read HookedTransformer for documentation, but struggle with the lack of line breaks 
in the comments. I'm updating the line breaks to align with IntelliJ's default right margin to fix this. I've tried to err on the side of leaving things as is when the lack of line breaks seems potentially beneficial, such as to avoid line breaks during an equation. Parts of the code use a much tighter line width than the IntelliJ default (80 chars) and I haven't touched that either. 

I'm not pursuing a long term solution i.e. autoformatting. The reason is that `black` will not add line wraps for docstrings because it could affect structured data stored in the strings, so we would have to add a new formatting tool and assess the performance tradeoffs and edge cases that come with that, and I feel that this update gets us 60% of the value with 5% of the effort.

## Type of change

- [x] Formatting 

# Checklist:

- [N/A] I have commented my code, particularly in hard-to-understand areas
- [N/A] I have made corresponding changes to the documentation
- [N/A] I have added tests that prove my fix is effective or that my feature works


neelnanda-io:
My main concern/annoyance (and the reason lines don't currently have this, and that I expect people to use line wrap for) is that editing prose is a pain in the ass if you need to ensure it remains aligned (every line in a block needs to be re aligned, it's so annoying!). None of the Vs code extensions for doing this is right.

That said, this change seems net good, thanks! This just annoys me a lot, and there doesn't seem to be a good solution

jbloomAus:
Thanks @luciaquirke. I think you committed some unrelated changes that shouldn't be in the CI. Can you please edit the commit to remove:

- [ ] .coverage.Lucias-MacBook-Air.local.67990.863551
- [ ] .idea/workspace.xml

After that, I'll accept.

luciaquirke:
Thanks for the reviews @jbloomAus and @neelnanda-io!

I've removed the erroneous files and added `.idea/` to the gitignore. Also thanks for reminding me about soft wrap, next time I won't have to raise a PR 😄

jbloomAus:
thanks, looks good!

PR #285, opened 2023-05-22
Add SVD interpretability feature to TransformerLens

JayBaileyCS:
# Description

Recently, Conjecture released a post (https://www.lesswrong.com/posts/mkbGjzxD8d8XqKHzA/the-singular-value-decompositions-of-transformer-weight) on interpreting the SVD matrices from the OV circuit and the w_in / w_out layers of MLP's. This PR adds this support to TransformerLens, with much less code required than the original Colab. The code has 100% test coverage and contains a demo.

This PR also includes some enforced formatting, and a change to the indices method in Slice to ensure the np.int64 dtype in order to fix a unit test. I'm not sure why it was broken for me in particular, since I wasn't working on that particular class, or even that file. But it's fixed now.

QUESTIONS:

- Should SVDInterpreter even be a class, or just some functions which take in a model as argument? I pared it down quite a lot as I was writing it and it became simpler than expected.
- In order to fix an issue with test_head_detector not importing TypeCheckError, I upgraded typeguard to 4.0.0 locally. Should we make this change in poetry?
- We have numerical instability with linear dependence that means different devices return different results. If the test_svd_interpeter() test fails, this is likely why, and makes it difficult to test in future. What should we do about this?

# Checklist:


jbloomAus:
Possibly failing due to memory issues. Need to rebase on main.

PR #291, opened 2023-05-31
Add support for left padding to HookedTransformer's to_tokens method.

### LINKED ISSUE #240: [Proposal] Add support for padding/masking to the attention computation (Minor feature request)
### Proposal 

### Motivation

### Alternatives

### Additional context

### Checklist

JayBaileyCS:
# Description

- Added keyword argument left_pad (default False to avoid breaking anything) to HookedTransformer's to_tokens method. This keyword argument shifts the tokens tensor so that padding is on the left instead of the right.
- Added tests to support this change.
- Added this change to the Exploratory Analysis Demo.
- Fixed a test in test_utils that was failing by enforcing dtype=np.int64
- Some automated formatting changes.

Fixes #240 

## Type of change

# Checklist:


neelnanda-io:
You'll need to also add support for custom attention masks as in #240 for this to work. Otherwise the model will just attend to the pad tokens! (This is not an issue for right pad because generative models attend backwards only)

transformer_lens/HookedTransformer.py
jbloomAus, line 1: This seems unnecessary to implement ourselves. I think if we're doing this, we should have it as a setting when you load in the model, set the padding to left-side. 

```
>>> model.tokenizer.padding_side = "left"
>>> prompts = ["Hello", "How are you?", "I am fine, thank you."]
>>> tokens_without_padding = model.to_tokens(prompts, prepend_bos=True, left_pad=False)
>>> tokens_without_padding

tensor([[50256, 50256, 50256, 50256, 50256, 50256, 50256, 15496],
        [50256, 50256, 50256, 50256,  2437,   389,   345,    30],
        [50256,    40,   716,  3734,    11,  5875,   345,    13]])
```

tests/unit/test_tokenization_methods.py
jbloomAus, line 1: Thanks for writing tests. 
Would be good to avoid single letter variable names.

demos/Exploratory_Analysis_Demo.ipynb
jbloomAus, line 1: Last few sentences are unnecessary/confusing. 

"In this case, it's not necessary since all our prompts are of the same length. However, since you get the same result either way when no padding is applied, we include it anyway so you don't have to worry about it."

I think we should just have the warning and a seperate example with different length sentences as an aside.

jbloomAus (changes requested):
Hey Jay, 

Thanks for the PR.

Some things I'd like to address:

- [ ] 1. AFAICT, you haven't fixed the positional embedding which will be wrong if we left pad (see Neel's comments on the original issue).
- [ ] 2. Similarly, we haven't started masking the attention to padded tokens which may could be necessary for coherent output. 
- [ ]  As commented, I don't think we should have our own code for doing the padding, only updating the model's tokenizer. 

Thanks for starting this!

transformer_lens/HookedTransformer.py
JayBaileyCS: I'm a bit confused here. My understanding is, the tokenizer (taken from HuggingFace?) already has a padding_side argument that we can use, so we can just use that?

I assume the attention_mask thing will need to be done for the functions where the model is actually being run for inference, since we want the attention_mask to be used to affect attention patterns in the forward pass. 

That said, with tokens_without_padding, why would we be passing in left_pad=False in this scenario if we're not using a left_pad argument? And it seems the tensor is left padded despite left_pad = False here?

jbloomAus: > I'm a bit confused here. My understanding is, the tokenizer (taken from HuggingFace?) already has a padding_side argument that we can use, so we can just use that?

Correct, but it should instantiated with the class since it's used automatically by various methods.

> I assume the attention_mask thing will need to be done for the functions where the model is actually being run for inference, since we want the attention_mask to be used to affect attention patterns in the forward pass.

Yup!

> That said, with tokens_without_padding, why would we be passing in left_pad=False in this scenario if we're not using a left_pad argument? And it seems the tensor is left padded despite left_pad = False here?

This is because I updated the padding side manually. What I did above is not ideal because I'm working around our to_tokens api. I'd rather the api itself make use of the tokenizer method but the actual control flow still make use of the tokenizer property. One solution might be to instantiate the class with a padding side rather than to ever pass it into to_tokens. This way the model class has a consistent padding property.

JayBaileyCS:
This has gotten too big so it's being replaced with https://github.com/neelnanda-io/TransformerLens/pull/323

PR #298, opened 2023-06-03
from_pretrained custom parameters (#288)

### LINKED ISSUE #288: [Proposal] Adapt Loading From Pretrained to give a custom cache dir for the HuggingFace model
### Proposal 

loading_from_pretrained.py, under the hood, uses HuggingFace's AutoModelForCausalLM.from_pretrained. This caches the model weights loaded, to a folder indicated by `cache_dir`, but it can be hard to expose and edit the folder cached to so it goes to the default folder. This is sometimes annoying (in my case, I haev a 20GB disk with a 1000GB persistent volume attached and I want to cache to the persistent volume). It might also be nice to turn off caching altogether. 

A possible solution would be adding a cache_dir param to `HookedTransformer.from_pretrained` or adding the option to pass in any dict of arguments to the underlying HuggingFace loader, to allow future such use cases.

glerzing:
# Description

The issue is about adding the possibility to have a custom cache directory different from the HuggingFace default directory, with the from_pretrained `cache_dir` argument. The solution is more generic as it is designed to allow the use of Hugging Face's from_pretrained arguments in general. I didn't add a test for the parameter cache_dir, because it would be too long to execute if you have to re-download the model each time, but I tested it manually.
This should be backward-compatible, because it mainly adds new possible arguments.

Fixes #288

## Type of change

# Checklist:

### DIFF
--- a/tests/acceptance/test_hooked_encoder.py
+++ b/tests/acceptance/test_hooked_encoder.py
@@ -136,6 +136,27 @@ def test_run_with_cache(our_bert, huggingface_bert, hello_world_tokens):
     assert "mlm_head.ln.hook_normalized" in cache
 
 
+def test_from_pretrained_dtype():
+    """Check that the parameter `torch_dtype` works"""
+    model = HookedEncoder.from_pretrained(MODEL_NAME, torch_dtype=torch.bfloat16)
+    assert model.W_K.dtype == torch.bfloat16
+
+
+def test_from_pretrained_revision():
+    """
+    Check that the from_pretrained parameter `revision` (= git version) works
+    """
+
+    _ = HookedEncoder.from_pretrained(MODEL_NAME, revision="main")
+
+    try:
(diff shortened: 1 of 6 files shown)


glerzing:
Maybe we should try to ignore some functions in the documentation. Because the documentation is getting big, and it's not clear which functions are useful for end users. Adding an underscore at the beginning like with private methods even if it's outside a class (e.g. _select_compatible_kwargs instead of select_compatible_kwargs) for example causes the function to be ignored by sphinx, so it could be a solution.

jbloomAus:
> Maybe we should try to ignore some functions in the documentation. Because the documentation is getting big, and it's not clear which functions are useful for end users. Adding an underscore at the beginning like with private methods even if it's outside a class (e.g. _select_compatible_kwargs instead of select_compatible_kwargs) for example causes the function to be ignored by sphinx, so it could be a solution.

Seems reasonable. Will encourage this in the future/mention it to contributors. 

Otherwise seems solid.

PR #303, opened 2023-06-04
Add prepend option to `model.add_hook`

ArthurConmy:
# Description

This PR allows users to prepend hooks before all other hooks, and a test of this.

This is implemented in the same way that PyTorch 2.0 does this: search https://github.com/pytorch/pytorch/blob/bf52d570d9be0167501cd6a206b7e262e6d3860c/torch/nn/modules/module.py for "prepend"

I have checked that the test and black pass locally. I think that the main failures are not related.

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

# Checklist:

- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation
- [ x ] My changes generate no new warnings
- [ x ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/tests/unit/test_hooks.py
+++ b/tests/unit/test_hooks.py
@@ -21,7 +21,7 @@ class Counter:
 def test_hook_attaches_normally():
     c = Counter()
     _ = model.run_with_hooks(prompt, fwd_hooks=[(embed, c.inc)])
-    assert all([len(hp.fwd_hooks) == 0 for _, hp in model.hook_dict.items()])
+    assert all([len(hp.fwd_hooks) == 0 for _, hp in model.hook_dict.items()]) 
     assert c.count == 1
     model.remove_all_hook_fns(including_permanent=True)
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_hooks.py
jbloomAus, line 1: Test looks good. Took me a sec to understood what you did here. Possibly best to use a fixture rather than iterate over each case (set to zero first or set to randn). Also a similar test showing the behavior when prepend is not set to True would be good to establish the expected behavior in the absence of the prepend=True argument would be more "complete". 

That being said. This is sufficient for me to accept the PR.

jbloomAus: To be extra clear: 

- [ ] Replace for loop with a fixture that seperate the tests by order of hooks
- [ ] Add an equivalent test when prepend=False for completeness.

jbloomAus:
@ArthurConmy Happy to accept. Would be slightly preferable if we improved the test per my previous comment but I am happy to accept regardless. Let me know which you would prefer.

ArthurConmy:
Thanks Joseph! In [1a153d0](https://github.com/neelnanda-io/TransformerLens/pull/303/commits/1a153d0cf48cad0d33b26811e8588c64faf368df) I implemented a fixture to implement four tests: 
- an option for which way round the hooks are attached
- an option for whether we prepend or not

That was a great idea, and I hope this now looks good

tests/unit/test_hooks.py
ArthurConmy: Hey Joseph, I implemented those things and just re-ran the check (oops, black got mad at a single space...)

Is this OK to merge?

jbloomAus:
Looks great! Thanks!

PR #305, opened 2023-06-04
Fix tiny stories model names

### LINKED ISSUE #299: [Bug Report] Missing TinyStories model
**Describe the bug**

This model does not exist:

https://github.com/neelnanda-io/TransformerLens/blob/a417edd6227ce1d8da42e40debe3a3b5f87ece0f/transformer_lens/loading_from_pretrained.py#L120

I think it should be [roneneldan/TinyStories-Instuct-1Layer-21M](https://huggingface.co/roneneldan/TinyStories-Instuct-1Layer-21M) but I'm not certain: notice the different number of parameters of spelling error from Ronen

The same issue arises with `roneneldan/TinyStories-1Layer-28M`- all the 20-29M parameter models should be checked

**Code example**
I found this while following the `docs/README` instructions for building docs. This bug :bug: should have been tested by a similar method (trying to download ALL configs from :hugs: (which is cheap) and checking that none fail)

Alternatively 

```python
from transformer_lens.HookedTransformer import HookedTransformer as HT
m=HT.from_pretrained("roneneldan/TinyStories-Instruct-1Layers-28M")
```

**System Info**
N/A

**Additional context**
N/A

### Checklist

- [ :white_check_mark: ] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

Felhof:
Closes #299 
Some of the names used to load tiny-story models were misspelled and so we were unable to load them. This PR fixes the spellings and adds a simple unit test that checks if each model's config can be loaded so that we will notice when the names are changed in the future.


tests/unit/test_tinystories.py
```diff
@@ -0,0 +1,25 @@
...
+    "roneneldan/TinyStories-Instruct-28M",
+    "roneneldan/TinyStories-Instruct-33M",
+    "roneneldan/TinyStories-1Layer-21M",
+    "roneneldan/TinyStories-2Layers-33M",
+    "roneneldan/TinyStories-Instuct-1Layer-21M",
+    "roneneldan/TinyStories-Instruct-2Layers-33M",
+]
+
+
+def test_all_models_exist():
```
ArthurConmy, line 17:
```suggestion
def test_all_tinystories_models_exist():
```

Please move this test to `test_hooked_transformer`

tests/unit/test_tinystories.py
```diff
@@ -0,0 +1,25 @@
...
+    "roneneldan/TinyStories-Instruct-1M",
+    "roneneldan/TinyStories-Instruct-3M",
+    "roneneldan/TinyStories-Instruct-8M",
+    "roneneldan/TinyStories-Instruct-28M",
+    "roneneldan/TinyStories-Instruct-33M",
+    "roneneldan/TinyStories-1Layer-21M",
+    "roneneldan/TinyStories-2Layers-33M",
+    "roneneldan/TinyStories-Instuct-1Layer-21M",
+    "roneneldan/TinyStories-Instruct-2Layers-33M",
+]
```
ArthurConmy, line 14:
```suggestion
from transformer_lens.loading_from_pretrained import OFFICIAL_MODEL_NAMES

TINY_STORIES_MODEL_NAMES = [
    name for name in OFFICIAL_MODEL_NAMES if name.startswith("roneneldan/TinyStories")
]
```

ArthurConmy: (of course with this statement in the HookedTransformer file)

ArthurConmy (changes requested):
Thanks so much for the quick turn around! This seems correct, I just suggest two formatting changes. Can you implement them?

jbloomAus (approved):

jbloomAus:
Thanks @Felhof

ArthurConmy:
The acceptance test worked on my local machine, thanks @Felhof :raised_hands: !

PR #311, opened 2023-06-06
Add SVD interpretability feature to TransformerLens

JayBaileyCS:
# Description

Replaces PR #285 - it's cleaner this way after it's been sitting for a while and has caused local import issues that a fresh environment solves - unit and acceptance tests now pass without issue after cloning a fresh repo and merging my code from my own repo. It also gets rid of some unnecessary format changes - now all that I change is the SVD feature and altering one test that was failing locally for me, by enforcing the correct dtype.

Recently, Conjecture released a post (https://www.lesswrong.com/posts/mkbGjzxD8d8XqKHzA/the-singular-value-decompositions-of-transformer-weight) on interpreting the SVD matrices from the OV circuit and the w_in / w_out layers of MLP's. This PR adds this support to TransformerLens, with much less code required than the original Colab. The code has 100% test coverage and contains a demo.

This PR also includes a change to the indices method in Slice to ensure the np.int64 dtype in order to fix a unit test. I'm not sure why it was broken for me in particular, since I wasn't working on that particular class, or even that file. But it's fixed now.

QUESTIONS:

- Should SVDInterpreter even be a class, or just some functions which take in a model as argument? I pared it down quite a lot as I was writing it and it became simpler than expected.
- In order to fix an issue with test_head_detector not importing TypeCheckError, I upgraded typeguard to 4.0.0 locally. Should we make this change in poetry?
- We have numerical instability with linear dependence that means different devices return different results. If the test_svd_interpeter() test fails, this is likely why, and makes it difficult to test in future. What should we do about this?

# Checklist:


jbloomAus:
> Should SVDInterpreter even be a class, or just some functions which take in a model as argument? I pared it down quite a lot as I was writing it and it became simpler than expected.

Seems fine. I think I have comments on various things like defensive coding or possibly unnecessary use of loops.  

> In order to fix an issue with test_head_detector not importing TypeCheckError, I upgraded typeguard to 4.0.0 locally. Should we make this change in poetry?

This is another thing that got complicated and @dkamm  and @rusheb  were working on a strategy for. Can check in with them. 

> We have numerical instability with linear dependence that means different devices return different results. If the test_svd_interpeter() test fails, this is likely why, and makes it difficult to test in future. What should we do about this?

How different are the results? This could be a very big deal. Using FactoredMatrix should help if we aren't (and we should be). I don't think that I want to tolerate anything worse than the kind of precision we use to check forward passes between HookedTransformers/HuggingFace. 

Will come back to this in a day or so. I couldn't run this due to Pysvelte/typeguard bs. Hoping can sort out and then check things more thoroughly.

jbloomAus:
Thanks Jay, I'll reach out to the authors and get their thoughts. It's fine if it's just a property of the method and I can now see why that might be the case. Waiting on typeguard response.

JayBaileyCS:
Any word from the authors?

jbloomAus:
No word. I'm going to accept the PR as I think you've done a true to original implementation.

PR #312, opened 2023-06-06
Ignore some functions in the documentation (#310)

### LINKED ISSUE #310: [Proposal] Reduce the code documentation complexity by declaring internal functions with a leading "_"
### Proposal and motivation

The [code documentation](https://neelnanda-io.github.io/TransformerLens/transformer_lens.html) is pretty big. But some functions are not meant to be called by end users. Adding a "_" in front of functions or methods (e.g. `make_model_alias_map` => `_make_model_alias_map`) causes it not to be displayed on Sphinx. This would reduce the quantity of superfluous information in the documentation (in particular in `loading_from_pretrained.py`), but may slightly break backward compatibility if any end user was actually using one of those functions directly (for demos it's easy to verify). If you think this is useful, I can make the modifications, if you think this is not, don't hesitate to close this issue.

glerzing:
# Description

The aim is to ignore in the documentation some functions that are not useful for end users, to reduce the quantity of superfluous information. At first I was proposing to add an underscore (e.g. `make_model_alias_map` => `_make_model_alias_map`), but I finally adopted the safer approach proposed by @ArthurConmy of using the `field exclude-members` in `autodoc_default_options` to specify functions to ignore. It is then indeed not displayed in Sphinx.

I reverted one of the modifications of my last PR (`_select_compatible_kwargs` => `select_compatible_kwargs`) to be consistent everywhere.

The list of functions to ignore is debatable. I didn't add any class method (although `fill_missing_keys` is a function in `loading_from_pretrained.py`, but also a method of `HookedTransformer` which is then ignored as well)

Fixes #310 

## Type of change
Documentation update

### DIFF
--- a/transformer_lens/utils.py
+++ b/transformer_lens/utils.py
@@ -24,7 +24,7 @@ import json
 from jaxtyping import Float, Int
 
 
-def _select_compatible_kwargs(
+def select_compatible_kwargs(
     kwargs_dict: Dict[str, Any], callable: Callable
 ) -> Dict[str, Any]:
     """Return a dict with the elements kwargs_dict that are parameters of callable"""
@@ -53,7 +53,7 @@ def download_file_from_hf(
         filename=file_name,
         subfolder=subfolder,
         cache_dir=cache_dir,
-        **_select_compatible_kwargs(kwargs, hf_hub_download),
+        **select_compatible_kwargs(kwargs, hf_hub_download),
     )
 
     # Load to the CPU device if CUDA is not available
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,7 +12,7 @@ If you are modifying a non-environment page or an atari environment page, please
 
 Install the required packages:
 
-Need to use python 3.9 (3.10 has an issue with napoleon extension) and below 3.8 doesn't have sphinx suppourt.
+Need to use python 3.9 (3.10 has an issue with napoleon extension) and below 3.8 doesn't have sphinx support.
 ```
 poetry install --extras docs
 ```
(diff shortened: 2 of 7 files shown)


transformer_lens/utils.py
```diff
@@ -24,7 +24,7 @@
 from jaxtyping import Float, Int
 
 
-def _select_compatible_kwargs(
+def select_compatible_kwargs(
```
ArthurConmy, line 27: I don't think this change is needed? I think using `_` on functions only used in a given file is a good idea - I just don't think it is worth breaking backwards compatibility for

ArthurConmy (changes requested):
Great! Can you re-generate the docs with `poetry run sphinx-autobuild -b dirhtml ./source build/html`?

I don't think we have automated this yet

transformer_lens/utils.py
glerzing: Ok I can revert that, although I coded this function a few days ago so it's probably not a worry for backward compatibility. I'd like a quick review from @jbloomAus just to be sure we all agree.

glerzing: The docs are on Neel Nanda's website (https://neelnanda-io.github.io/TransformerLens/), so I don't think I can regenerate it. When I do `poetry run sphinx-autobuild -b dirhtml ./source build/html` it just generates a local server on my computer, that no one else can access. I'm not sure but I guess Neel updates it regularly and manually.

ArthurConmy: And our Github actions should automatically push this to Neel's website : )

ArthurConmy (approved):
Actually, good point that we've only had `select_compatible_kwargs` for a couple of days. I'll merge this for that reason.

PR #313, opened 2023-06-06
Add `hook_mlp_in`

ArthurConmy:
# Problem

Currently, there's no clean way to manipulate the output of one attention head into a future MLP in our transformer model. This is because editing hook_resid_mid affects the residual stream beyond the MLP too, as it is used in the computation of both resid_mid and resid_post.

# Solution

This PR introduces a new `hook_mlp_in` `HookPoint` that allows modification of the input to an MLP. 

It adds a test too. I also commented two places I found some repeated code, I'm happy to improve this too if wanted.

# Details

This HookPoint will be enabled by default. Initially, we were concerned about potential memory bloat and potential backwards compatibility issues with old code that expects a specific set of keys in an ActivationCache. However, upon further investigation, it was found that since the resid_mid and hook_mlp_in(resid_mid) share the same underlying storage, there is no memory bloat. 

As for backwards compatibility, the addition of a new key to the ActivationCache could break any existing code that does not specifically rely on the absence of this key, but we expect this to be unlikely.

Closes https://github.com/neelnanda-io/TransformerLens/issues/309

# Checklist:

### DIFF
--- a/tests/acceptance/test_activation_cache.py
+++ b/tests/acceptance/test_activation_cache.py
@@ -289,3 +289,18 @@ def test_stack_neuron_results_with_apply_ln():
     assert torch.isclose(
         ref_scaled_residual_stack, scaled_residual_stack, atol=1e-7
     ).all()
+
+
+def test_hook_mlp_in_memory():
+    """Test that the new hook_mlp_in is not taking any extra memory"""
+
+    model = load_model("solu-2l")
+
+    # Run the model and cache all activations
+    tokens, _ = get_ioi_tokens_and_answer_tokens(model)
+    _, cache = model.run_with_cache(tokens)
+
+    assert (
+        cache["blocks.0.hook_resid_mid"].data_ptr()
+        == cache["blocks.0.hook_mlp_in"].data_ptr()
+    )
--- a/transformer_lens/components.py
+++ b/transformer_lens/components.py
@@ -864,6 +864,8 @@ class TransformerBlock(nn.Module):
         self.hook_v_input = HookPoint()  # [batch, pos, d_model]
 
         self.hook_attn_out = HookPoint()  # [batch, pos, d_model]
+
+        self.hook_mlp_in = HookPoint()  # [batch, pos, d_model]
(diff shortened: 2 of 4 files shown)


transformer_lens/components.py
```diff
@@ -1011,7 +1014,7 @@ def add_head_dimension(tensor):
         resid_mid = self.hook_resid_mid(resid_pre + attn_out)
```
ArthurConmy, line 1014: In line 995 apply DRY and move `add_head_dimension` to the top level of this file so it is defined once in the code.

tests/acceptance/test_activation_cache.py
```diff
@@ -289,3 +289,18 @@ def test_stack_neuron_results_with_apply_ln():
     assert torch.isclose(
         ref_scaled_residual_stack, scaled_residual_stack, atol=1e-7
     ).all()
+
+
+def test_hook_mlp_in_memory():
+    """Test that the new hook_mlp_in is not taking any extra memory"""
+
+    model = load_model("solu-2l")
```
ArthurConmy, line 297: Could use fixtures here and in the rest of the file for less repetition

jbloomAus:
Seems good. Thanks for this!

neelnanda-io:
> However, upon further investigation, it was found that since the resid_mid and hook_mlp_in(resid_mid) share the same underlying storage, there is no memory bloat.

Wait, does this mean that if I create a hook on mlp_in like this:

def hook_fn(mlp_in, hook):
  mlp_in[:, 5:7, :] = clean_mlp_in[:, 5:7, :]
  return mlp_in
 
 It will silently edit resid_mid too?! That seems terrible

neelnanda-io:
I favour clone plus an actual flag

ArthurConmy:
Okay @neelnanda-io seems reasonable. I will flag that I do think HookedTrasformerConfig is getting pretty big

jbloomAus:
@neelnanda-io Sounds like I missed something important here with the editing of mlp_in also affecting resid_mid. Apologies. Feels like the kind of thing we should be testing for. Never paranoid enough apparently. I'll watch out for similar patterns in the future.

PR #315, opened 2023-06-06
In LayerNorm, fix hook_normalized to hook pre-scaled-and-shifted outp…

jkramar:
…uts rather than post-.

# Checklist:


jbloomAus:
@jkramar can you please give more context? This is a non-trivial change. To the extent the previous code was written incorrectly, what issues has this caused? What unexpected effects might occur if we made this change and people were assuming the previous interpretation? What incorrect analysis or processing would result from the existing code?

neelnanda-io:
Discussed offline with Janos, I don't think this is worth it

PR #316, opened 2023-06-09
Update to `hook_mlp_in`

ArthurConmy:
Updates the addition made in https://github.com/neelnanda-io/TransformerLens/pull/313 based on the comments made after the merge.

Namely, adds an optional flag for `hook_mlp_in` and a test that the conditional flag is working.

I'm not copying the boilerplate as that was covered last PR, I hope that's OK

### DIFF
--- a/tests/unit/test_cache_hook_names.py
+++ b/tests/unit/test_cache_hook_names.py
@@ -19,7 +19,6 @@ act_names_in_cache = [
     "blocks.0.attn.hook_z",
     "blocks.0.hook_attn_out",
     "blocks.0.hook_resid_mid",
-    "blocks.0.hook_mlp_in",
     "blocks.0.ln2.hook_scale",
     "blocks.0.ln2.hook_normalized",
     "blocks.0.mlp.hook_pre",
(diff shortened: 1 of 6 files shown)


tests/unit/test_cache_hook_names.py
jbloomAus, line 1: I think we should have a test that shows if the mlp_in flag is true, we also get the hook_mlp_in name in cache.

transformer_lens/components.py
jbloomAus, line 1: I understand why this is now implemented the way it is but I wonder if we could write it in a more obvious way? 

I think simply starting a block with `if self.cfg.use_hook_mlp_in` might be easier to read and save people time if they are reading source to understand the effect of the flag. 

Comment to explain the clone might also be good.

jbloomAus (commented):
Seems good. Requested changes are all optional, but recommended.

- [ ] Add test for name in cache depending on arg. 
- [ ] Write mlp_in hook logic so it's easier to read/less concise. (somewhat aesthetic but that's my recommendation).

jbloomAus:
@ArthurConmy checking up on this. Are you able to make those changes?

Issue #104, opened 2022-12-19
Add mixed precision inference incl loading


neelnanda-io:
Add the option to load models in bfloat16 and float16. Esp important for large models like GPT-J and GPT-NeoX.

Ideally, load from HuggingFace in this low precision, do weight processing on the CPU, and then move the processed model weights to the GPU. Might be easiest to do the weight processing once and caching to HF (see #103 )

neelnanda-io:
Maybe covered by #125

glerzing:
Solved with #298

tbenthompson:
Before I found this issue, I didn't @glerzing was working on #317 so I was planning to report separately. 

Anyway, despite progress, I thought I'd share a demo where I get nans running in float16:
```
import torch
from transformer_lens import HookedTransformer

torch.set_grad_enabled(False)

# Issue #1: there's no way to use float16 on initialization so we're forced to
# convert to float16.
model32 = HookedTransformer.from_pretrained(f"EleutherAI/pythia-70m-deduped")
print(repr(model32.to_string(model32(" Unable")[0, -1].argmax())))

# Issue #2:
model16 = HookedTransformer.from_pretrained(f"EleutherAI/pythia-70m-deduped").to(
    torch.float16
)
print(model16(" Unable"))
```

Thanks for the progress on #317 !!! Psyched to see it merged.

neelnanda-io:
I believe that Pythia 70m can have attention scores as low as -100,000,
which will get you nans in float16 because those can do max -65,536.
Honestly, my take is that this is not our problem, and you should use
bfloat16 instead, so long as HuggingFace *also *gives you nans. I have no
clue why Pythia is this high lol.

tbenthompson:
Wow, that's fascinating about the giant attention scores!! 

I'm seeing big differences in both `bfloat16` and `float16` between Huggingface and TL on Pythia 410M. I was suspicious that the TL processing (fold LN, center unembed, etc) was causing the differences so I tried `from_pretrained_no_processing` but the differences persist. 

I'm gradually learning more about the internals of TL so if I have time soon, I'll dig in on this and try to figure out what's going on. 

<details>
<summary>Source</summary>

```
import torch
from transformer_lens import HookedTransformer
import transformers

torch.set_grad_enabled(False)

model_name = f"EleutherAI/pythia-410m-deduped"
model32 = HookedTransformer.from_pretrained_no_processing(model_name)
[...]

PR #317, opened 2023-06-09
Support for reduced precision (#104)

glerzing:
# Description
`HookedTransformerKeyValueCacheEntry` wasn't compatible with other dtypes, so the argument `past_kv_cache_entry` wasn't working.
There is also an optimization. Instead of initializing the model in torch.float32 and then converting it to the desired dtype, we now directly initialize the model layers in the desired dtype.
The attribute `dtype` was added to the configuration with the default value torch.float32. I hope this is ok. It's practical to have access to it from anywhere.
Also added a test for 8 bits loading, which is skipped if there is no GPU.

Fixes # 104

## Type of change

# Checklist:

### DIFF
--- a/tests/acceptance/test_hooked_transformer.py
+++ b/tests/acceptance/test_hooked_transformer.py
@@ -3,7 +3,7 @@ import os
 
 import pytest
 import torch
-from transformers import AutoConfig
+from transformers import AutoConfig, AutoModelForCausalLM
 
 from transformer_lens import HookedTransformer
 from transformer_lens.components import LayerNormPre
@@ -136,16 +136,6 @@ def test_from_pretrained_no_processing(name, expected_loss):
     assert (reff_loss.item() - expected_loss) < 4e-5
 
 
-@pytest.mark.skipif(
-    torch.backends.mps.is_available(),
-    reason="bfloat16 unsupported by MPS: https://github.com/pytorch/pytorch/issues/78168",
-)
-def test_from_pretrained_dtype():
-    """Check that the parameter `torch_dtype` works"""
-    model = HookedTransformer.from_pretrained("solu-1l", torch_dtype=torch.bfloat16)
-    assert model.W_K.dtype == torch.bfloat16
-
-
 def test_process_weights_inplace():
     """Check that process_weights_ works"""
     model = HookedTransformer.from_pretrained_no_processing("gpt2-small")
@@ -168,6 +158,82 @@ def test_from_pretrained_revision():
         pass
     else:
         raise AssertionError("Should have raised an error")
+
+
+def check_similarity_with_hf_model(tl_model, hf_model, prompt="Hello, world!"):
+    """
+    Check that the TransformerLens model and the HuggingFace model
+    give approximately the same results.
+
+    The logits typically differ by a constant value, but check only the results
+    after the softmax because this is what matters most.
+    """
+    tokens = tl_model.tokenizer.encode(prompt, return_tensors="pt")
+    logits = tl_model(tokens, prepend_bos=False)
+    hf_logits = hf_model(tokens).logits
+    assert torch.allclose(
+        torch.softmax(logits, dim=-1), torch.softmax(hf_logits, dim=-1), atol=1e-5
+    )
+
+
+def check_performance(tl_model, hf_model, margin=0.01):
+    """
+    Check that the TransformerLens model and the HuggingFace have
+    approximately the same confidence in the expected answer.
+    """
+    prompt = " Unable"
+    tokens = tl_model.tokenizer(prompt, return_tensors="pt")["input_ids"]
+
+    expected_token = tl_model.tokenizer.encode(" to")[
+        0
+    ]  # Assume this is the expected token to predict
+
+    tl_logits = tl_model(tokens, prepend_bos=False)[0, -1].float()
+    hf_logits = hf_model(tokens).logits[0, -1].float()
+    tl_prob = torch.softmax(tl_logits, dim=-1)[expected_token].item()
+    hf_prob = torch.softmax(hf_logits, dim=-1)[expected_token].item()
+    assert tl_prob + margin > hf_prob
+
+
+def check_dtype(dtype, margin=0.01):
+    """Check the loading and inferences for different dtypes."""
+    for model_path in ["gpt2", "roneneldan/TinyStories-33M", "EleutherAI/pythia-70m"]:
+        model = HookedTransformer.from_pretrained(model_path, torch_dtype=dtype)
+        hf_model = AutoModelForCausalLM.from_pretrained(
+            model_path,
+            torch_dtype=dtype,
+        ).to("cuda" if torch.cuda.is_available() else "cpu")
+
+        for layer_name, layer in model.state_dict().items():
+            assert layer.dtype in [dtype, torch.bool] or "IGNORE" in layer_name
+
+        check_performance(model, hf_model, margin)
+
+        # Check that generate doesn't throw an error
+        _ = model.generate("Hello, World!")
+
+        del model
+        del hf_model
+        gc.collect()
+
+
+@pytest.mark.parametrize("dtype", [torch.float64, torch.float32])
+def test_dtypes(dtype):
+    check_dtype(dtype, margin=5e-5)
+
+
+@pytest.mark.skipif(
+    torch.backends.mps.is_available() or not torch.cuda.is_available(),
+    reason="bfloat16 unsupported by MPS: https://github.com/pytorch/pytorch/issues/78168 or no GPU",
+)
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16])
+def test_half_precision(dtype):
+    """Check the 16 bits loading and inferences.
+    Note that bfloat16 is generally preferred to float16 for ML due to numerical instabilities,
+    and some float16 operations require having a GPU.
+    bfloat16 can be used without GPU, but surprisingly it doesn't give the same results in this case.
+    """
+    check_dtype(dtype, margin=0.005)
 
 
 @torch.no_grad()
--- a/tests/acceptance/test_hooked_transformer.py
+++ b/tests/acceptance/test_hooked_transformer.py
@@ -182,7 +182,9 @@ def check_performance(tl_model, hf_model, margin):
     approximately the same confidence in the expected answer.
     """
     prompt = " Unable"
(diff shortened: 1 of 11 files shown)


jbloomAus:
I'd rather not accept this until we have resolved the issue described here: https://github.com/neelnanda-io/TransformerLens/issues/104#issuecomment-1589205548 

Moreover, we should have tests for reduced precision model performance if it's the kind of thing that can be impacted by accident. 

The rest seems fine. Thanks for the contribution!

To do list:
- [ ] write tests for precision model performance
- [ ] pass test for precision model_performance

glerzing:
Torch doesn't support 8-bit floating point models, so there is no corresponding torch dtype and it relies on different libraries like bitsandbytes and accelerate. It is bug-prone, so I should probably make further checks.
And I will also take into account the comment about the test for the model precision performance, sorry for the delay.

glerzing:
Actually my previous answer was a bit inaccurate. Quantization doesn't use 8 bit floats, it uses 8 bit integers with a scaling factor. And there is an int8 dtype in torch, but it doesn't seem to be used by the transformers library, as I understand it the hidden states are typically in half precision and only some operations are quantized. I recommend this article if you are interested by the details : https://huggingface.co/blog/hf-bitsandbytes-integration.
It's a special type of computation, and I'm not sure it's reasonable to try to make it work in TransformerLens, I'll see.

slavachalnev:
That's a great blogpost, thanks!

Yeah, it looks like the HuggingFace bitsandbytes integration only quantises nn.Linear layers and HookedTransformer has no nn.Linear layers.

glerzing:
The fact that we use einsum instead of nn.Linear seems like the main reason for the remaining differences between the predictions of TransformerLens and Hugging Face in half precision.

tests/acceptance/test_hooked_transformer.py
```diff
@@ -156,6 +151,81 @@ def test_from_pretrained_revision():
...
+    """Check the loading and inferences for different dtypes."""
+    for model_path in ["gpt2", "roneneldan/TinyStories-33M", "EleutherAI/pythia-70m"]:
+        model = HookedTransformer.from_pretrained(model_path, torch_dtype=dtype)
+        hf_model = AutoModelForCausalLM.from_pretrained(
+            model_path,
+            torch_dtype=dtype,
+        ).to("cuda" if torch.cuda.is_available() else "cpu")
+
+        for layer_name, layer in model.state_dict().items():
+            assert layer.dtype in [dtype, torch.bool] or "IGNORE" in layer_name
```
glerzing, line 199: It may be constroversial but I modified the value of the layer "IGNORE" from `-1e-5` to `torch.finfo(cfg.dtype).min`. This shouldn't change much in practice, I just thought it was cleaner and closer to what Hugging Face did. I can revert that if needed.

transformer_lens/components.py
```diff
@@ -377,7 +409,7 @@ def __init__(
         else:
             raise ValueError(f"Invalid attention type: {self.attn_type}")
 
-        self.register_buffer("IGNORE", torch.tensor(-1e5))
+        self.register_buffer("IGNORE", torch.tensor(torch.finfo(cfg.dtype).min))
```
glerzing, line 412: It may be constroversial but I modified the value of the layer "IGNORE" from `-1e-5` to `torch.finfo(cfg.dtype).min`. This shouldn't change much in practice, I just thought it was cleaner and closer to what Hugging Face did. I can revert that if needed.

transformer_lens/components.py
```diff
@@ -511,6 +543,11 @@ def forward(
         if self.cfg.positional_embedding_type == "rotary":
             q, k = self.rotary_rotate_qk(q, k, kv_cache_pos_offset)
 
+        if self.cfg.dtype not in [torch.float32, torch.float64]:
+            # If using 16 bits, increase the precision to avoid numerical instabilities
+            q = q.to(torch.float32)
+            k = k.to(torch.float32)
+
```
glerzing, line 550: I chose to convert to `float32` before the attention layer, instead of dividing sooner with self.attn_scale as proposed. Mostly because it is what is done in the GPT NEO architecture in `def _attn` and it looked a bit more reliable.

glerzing:
By the way, it's a problem that we can't self-assign issues, because when we chose an issue, we don't know if someone is already on it. Perhaps we could leave public the authorization to self-assign issues. Or we may have to systematically add a comment to indicate when we start working on an issue. 
By the way, is there some existing discussion channel (e.g. Discord), or does everything happen on discord ?

transformer_lens/components.py
glerzing: I'll revert that since it is done in the PR #319.

glerzing:
(rebased on the main branch to solve conflicts)

glerzing:
@jbloomAus, @slavachalnev, can you review this PR ? Thanks in advance.

jbloomAus:
I'm on holiday ATM but can review once @slavachalnev has taken a look depending on what's needed. Thanks for doing this!

slavachalnev:
Looks good but need to fix a few things:

When I run tests on GPU, I get four failures.
```
FAILED test_hooked_transformer.py::test_dtypes[dtype0] - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking...
FAILED test_hooked_transformer.py::test_dtypes[dtype1] - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking...
FAILED test_hooked_transformer.py::test_half_precision[dtype0] - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking...
FAILED test_hooked_transformer.py::test_half_precision[dtype1] - RuntimeError: "baddbmm_with_gemm" not implemented for 'Half'
```
The first two are easily fixed by moving the tokens to device in the test e.g.
```python

def check_performance(tl_model, hf_model, margin=0.01, device='cpu'):  # Added device here
    """
    Check that the TransformerLens model and the HuggingFace have
    approximately the same confidence in the expected answer.
    """
    prompt = " Unable"
    tokens = tl_model.tokenizer(prompt, return_tensors="pt")["input_ids"].to(device)  # Added device here

    expected_token = tl_model.tokenizer.encode(" to")[
        0
    ]  # Assume this is the expected token to predict

    tl_logits = tl_model(tokens, prepend_bos=False)[0, -1].float()
    hf_logits = hf_model(tokens).logits[0, -1].float()
    tl_prob = torch.softmax(tl_logits, dim=-1)[expected_token].item()
    hf_prob = torch.softmax(hf_logits, dim=-1)[expected_token].item()
    assert tl_prob + margin > hf_prob
```

Then we are left with a bfloat16 precision error being higher than the set margin: `assert (0.15567666292190552 + 0.005) > 0.17949192225933075` Probably acceptable to increase margin but not sure.

The final error is caused by a bunch of operations not being defined for Half precision on CPU. In `load_from_pretrained` we do preprocessing on CPU which causes an error. I think that when loading from pertrained, we need to use float32 and move model to dtype after it’s loaded. What do you think?


I would also modify `move_to_and_update_config` function so that it updates config when you move the model to a different dtype. The reason this matters is you check the dtype in the config during the Attention forward pass so it should be up to date.

neelnanda-io:
My recommendation is to disable LayerNorm folding in float16, and make them
use loading_fron_pretrained_no_processing If you really care about memory
you often want to avoid doing this post processing anyway, and it's not
essential, though it is nice. And a user who cares can manually do it on
the GPU themselves,or do it on the CPU in fp32. But it's important that
people can load large models in memory constrained environments like a
colab notebook

neelnanda-io:
Also thanks so much for making the PR! Real mixed precision support has been on my wishlist for *ages*

glerzing:
Oops, sorry @slavachalnev, I missed the bugs induced by the rebase.

Most of the time the results are pretty similar between TransformerLens and Hugging Face. But it's still a bit weird that for EleutherAI/pythia-70m with bfloat16 transformer_lens gives a probability of 0.836 for the token " to" whereas Hugging Face gives only 0.002. Since " to" is considered the correct token it passes ok here.

slavachalnev:
Now getting an error in `test_half_precision[dtype0]`

```

dtype = torch.bfloat16

    @pytest.mark.skipif(
        torch.backends.mps.is_available() or not torch.cuda.is_available(),
        reason="bfloat16 unsupported by MPS: https://github.com/pytorch/pytorch/issues/78168 or no GPU",
    )
    @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16])
    def test_half_precision(dtype):
        """Check the 16 bits loading and inferences.
        Note that bfloat16 is generally preferred to float16 for ML due to numerical instabilities,
        and some float16 operations require having a GPU.
        bfloat16 can be used without GPU, but surprisingly it doesn't give the same results in this case.
        """
>       check_dtype(dtype, margin=0.05, no_processing=True)

tests/acceptance/test_hooked_transformer.py:245: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/acceptance/test_hooked_transformer.py:222: in check_dtype
    _ = model.generate("Hello, World!")
.venv/lib/python3.8/site-packages/torch/autograd/grad_mode.py:27: in decorate_context
    return func(*args, **kwargs)
transformer_lens/HookedTransformer.py:1516: in generate
    sampled_tokens = utils.sample_logits(
transformer_lens/utils.py:348: in sample_logits
    return torch.distributions.categorical.Categorical(logits=final_logits).sample()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Categorical(probs: torch.Size([1, 50257]), logits: torch.Size([1, 50257])), sample_shape = torch.Size([])

    def sample(self, sample_shape=torch.Size()):
        if not isinstance(sample_shape, torch.Size):
            sample_shape = torch.Size(sample_shape)
        probs_2d = self.probs.reshape(-1, self._num_events)
>       samples_2d = torch.multinomial(probs_2d, sample_shape.numel(), True).T
E       RuntimeError: "multinomial_kernel_cuda" not implemented for 'BFloat16'
```
May want to convert probs to float32 before sampling.

Otherwise, changes look good.

slavachalnev:
@jbloomAus this looks good to me

PR #320, opened 2023-06-11
Add assertion to refactor_factored_attn_matrices

ArthurConmy:
# Description

You can't `refactor_factored_attn_matrices` with rotary models so error here.

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation
- [ x ] My changes generate no new warnings

I don't think we need a test for this because that would require loading in a `pythia-70m` which seems an annoying increase to test speed, so this is ready for review


jbloomAus:
Looks good, thanks!

PR #321, opened 2023-06-12
`names_filter` bug fix

ArthurConmy:
These lines don't work in the intended way when `names_filter` is a string because redefining names_filter and using it in the lambda expression doesn't play nicely (CC GPT-4 or "late binding" in Python docs)

More generally I think the type for `NamesFilter` should also contain `str`s (`Sequence[str]` isn't really accurate) and the documentation of `add_caching_hooks` should mention that `names_filter` can be a string. Happy to merge this fix as-is or implement those two docs changes on top of this

### DIFF
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -380,9 +380,11 @@ class HookedRootModule(nn.Module):
         if names_filter is None:
             names_filter = lambda name: True
         elif type(names_filter) == str:
-            names_filter = lambda name: name == names_filter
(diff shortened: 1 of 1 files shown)


jbloomAus:
@ArthurConmy Makes sense. Can you please make the docs changes and then push? 

Was this causing any issues in existing code or new code you were writing? Generally speaking providing examples would be good :) Tests are even better!

jbloomAus:
@ArthurConmy ping about this. Would you prefer to just merge?

PR #323, opened 2023-06-15
Added support for left_pad to HookedTransformer

### LINKED ISSUE #240: [Proposal] Add support for padding/masking to the attention computation (Minor feature request)

JayBaileyCS:
# Description

Replaces PR #291 because it waited too long and is now too big and messy.

* Added keyword argument left_pad (default False to avoid breaking anything) to HookedTransformer. This keyword argument sets the tokenizer's "padding_side" argument to left if True.
* Added attention mask for when padding_side is left to ensure we don't attend to padded tokens.
* Added test to support this change. (This doesn't need to be tested much because it leverages HuggingFace's implementation - we just have to verify it actually runs in forward)
* Added this change to the Exploratory Analysis Demo.
* Fixed a test in test_utils that was failing by enforcing dtype=np.int64

Fixes #240 

## Type of change

# Checklist:

### DIFF
--- a/demos/Exploratory_Analysis_Demo.ipynb
+++ b/demos/Exploratory_Analysis_Demo.ipynb
@@ -295,10 +295,13 @@
             ]
         },
         {
+            "attachments": {},
             "cell_type": "markdown",
             "metadata": {},
             "source": [
-                "The first step is to load in our model, GPT-2 Small, a 12 layer and 80M parameter transformer with `HookedTransformer.from_pretrained`. The various flags are simplifications that preserve the model's output but simplify its internals, see [here](TODO: link) for details."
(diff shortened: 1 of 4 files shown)


jbloomAus (changes requested):
Hey Jay,

Looks good, can you please write a test for / visualize an attention map with padding tokens to show that it's working. I can see the padding tokens are on the write side but there's not test for the mask being correct. 

Cheers,

JayBaileyCS:
Having looked further into this, it seems like it would be a more major undertaking than is worth for a "nice-to-have". The additive attention mask (where Joseph advised me to add it) is part of the Attention block. These blocks are made on construction of the transformer, so you can't pass in the input tokens to generate the mask. It would require rethinking the whole thing, which could be done if this was crucial, but in this case it isn't.

PR #325, opened 2023-06-16
Switch to beartype

dkamm:
# Description

Switching to beartype from typeguard due to incompatibility with jaxtyping on the latest version

Notes:
- typeguard is still in the lockfile because jaxtyping requires it
- had to use forward references in these spots to get beartype to work
   - devices.py - this is a circular import solely due to typehints
   - utils.py - this is a true circular import that we might want to remove
   - FactoredMatrix.py - had to use it for methods returning `Union["FactoredMatrix", ...]`

## Type of change

### Screenshots
N/A

# Checklist:


dkamm:
@jbloomAus can you rerun the CI? think it failed due to an unrelated issue with connecting to huggingface

dkamm:
@jbloomAus thanks! this is ready for review.  Feel free to change the `composition_scores` typehint as I'm not sure how it works

jbloomAus:
Thanks David, will review in the next couple of days :)

jbloomAus:
@dkamm, looks great! I just want to double check/compare equivalent error messages and also have a summary here of why we're switching IRC, it's because we had testing issues and then install issues with typeguard? Feels like a long story but you have most context. Once that's written, I'll share with Neel just to be sure he's cool with it.

dkamm:
@jbloomAus

We're switching because there have been issues getting the newer versions of typeguard to work on this codebase for various reasons.

Typeguard 3 - have to add `@typeguard_ignore` to class properties, doesn't work with the jaxtyping pytest plugin
Typeguard 4 - incompatible with jaxtyping due to how it treats annotations as forward references in its ast transformation

In contrast, the latest version of beartype works with the small number changes described above.

None of these changes preclude switching to typeguard 4 when that issue is resolved, but it's not a trivial issue to solve.  I figured this solution is the best option right now.

jbloomAus:
Thanks @dkamm!

@alan-cooney This seems like a good thing to me, are you able to verify/double check please?

transformer_lens/FactoredMatrix.py
```diff
@@ -41,9 +40,9 @@ def __matmul__(
         other: Union[
             Float[torch.Tensor, "... rdim new_rdim"],
             Float[torch.Tensor, "rdim"],
-            FactoredMatrix,
+            "FactoredMatrix",
```
alan-cooney, line 43: Is there a reason for this switch? My understanding is Python would treat this as a string literal (i.e. the type is a string that exactly matches "FactoredMatrix".

alan-cooney (commented):
Thanks for this PR - nice work!

Just to check - have you checked that providing an incorrect jaxtyping type causes beartype to throw an error (when running pytest)?

Otherwise just one question to confirm in the review - where I think I'm missing something.

transformer_lens/FactoredMatrix.py
dkamm: The switch was made only because beartype worked with one but not the other. The string typehint is actually treated as a forward reference (see https://peps.python.org/pep-0484/#forward-references)

alan-cooney: Ah cool thanks for the explanation!

transformer_lens/utils.py
```diff
@@ -669,10 +669,10 @@ def transpose(tensor: Float[torch.Tensor, "... a b"]) -> Float[torch.Tensor, "..
 
 
 def composition_scores(
-    left: FactoredMatrix, right: FactoredMatrix, broadcast_dims=True
+    left: "FactoredMatrix", right: "FactoredMatrix", broadcast_dims=True
```
alan-cooney, line 672:
```suggestion
    left: FactoredMatrix, right: FactoredMatrix, broadcast_dims=True
```
I would imagine the forward reference is not needed here therefore?

alan-cooney (changes requested):
One small thing to check - otherwise good to go. 

Thanks again for this!
Alan

jbloomAus:
@dkamm, just pinging you to see if you've had a chance to action @alan-cooney's suggestions.

transformer_lens/utils.py
dkamm: I think it's actually needed to get beartype to not complain. I suspect it's because there's a circular dependency between utils.py and FactoredMatrix

dkamm:
Just added it! Sorry I missed it earlier

alan-cooney (approved):
__

alan-cooney (changes requested):
Sorry - it's out of sync actually with the latest changes to master. Are you able to update and then ping me and I'll get it merged right away?

transformer_lens/components.py
```diff
@@ -640,12 +637,12 @@ def apply_causal_mask(
...
-        k: Float[torch.Tensor, "batch pos head_index d_head"],
+        q: Float[torch.Tensor, "batch q_pos head_index d_head"],
+        k: Float[torch.Tensor, "batch k_pos head_index d_head"],
         past_kv_pos_offset,
     ) -> Tuple[
-        Float[torch.Tensor, "batch pos head_index d_head"],
-        Float[torch.Tensor, "batch pos head_index d_head"],
+        Float[torch.Tensor, "batch q_pos head_index d_head"],
+        Float[torch.Tensor, "batch k_pos head_index d_head"],
     ]:
```
dkamm, line 646: @alan-cooney @jbloomAus I had to change this typehint to get one of the tests to pass (test_hooked_transformer::test_dtypes). It failed because this method got passed `q`, `k` with different `pos` dimensions and jaxtyping checks for that as configured.  I'm not familiar enough with this method and its usage to say whether `q`, `k` should be allowed to have different `pos` dimensions.

dkamm:
@alan-cooney - I've updated the branch with latest changes from main. There's one new comment that requires review

transformer_lens/components.py
alan-cooney: Are you sure it's actually getting passed different sized queries and keys, or are the defined type signatures at this point just different?

I think this may be an issue with Beartype getting confused about the size of one of these tensors at a previous point (perhaps in the `if past_kv_cache_entry is not None:` section)?

I'm not very familiar with rotary embeddings either but as far as I can tell the q and k passed in here should be the same size (and they should also be the same size afterwards as they're immediately dot producted).

dkamm: > Are you sure it's actually getting passed different sized queries and keys, or are the defined type signatures at this point just different?

Yep, I confirmed that they were different sizes.  

> I think this may be an issue with Beartype getting confused about the size of one of these tensors at a previous point (perhaps in the if past_kv_cache_entry is not None: section)?

Hmm, I don't think this is the case.  Beartype/Jaxtyping should only look at the arguments passed to the call in this case

Here's a minimal script to reproduce (you'll have to change the typehints back):
```python
from jaxtyping import install_import_hook
with install_import_hook("transformer_lens", "beartype.beartype"):
    model = HookedTransformer.from_pretrained("EleutherAI/pythia-70m")
    _ = model.generate("Hello, World!")
``` 

Actually, reading the code more, I think my change is correct and we should allow different sizes in the `pos` dim.  The reason is that when we're generating completion, after we feed through the initial input, we're using only the latest token as the query vec and using the kv cache for the key vecs to compute attention scores.  It's ok for `q` and `k` to have different sizes in the `pos` dim because they are getting dotted over the `d_head` dim.  Finally, the `past_kv_pos_offset` is just so we get the right positional embedding for `q`.

alan-cooney (approved):

alan-cooney:
Thanks for the PR - merged now!

PR #326, opened 2023-06-19
[Ready] Enable Pytorch GPU acceleration for M1 chips

luciaquirke:
- Add MPS support
- Disable MPS when torch < 2.0

TODO:
[ ] PyTorch support for bfloat16 on MPS: https://github.com/pytorch/pytorch/issues/78168 causing the two dtype tests to fail
[ ] Remove torch 2.0 requirements from TOML once PR ready

Encountered Poetry bug where `poetry install` doesn't update torch to the latest compatible version (2.0.1) as claimed in the docs, but setting the package to 2.0.1 manually in the toml works. 

Can't use torch < 2.0 due to torch broadcasting bug with MPS, e.g. here where mask is [15,15] and attn_scores is [8,8,15,15],  issue: https://github.com/pytorch/pytorch/issues/86239
image


jbloomAus:
@luciaquirke  failing formatting check?

alan-cooney (commented):
For what it's worth, it would be nice to abstract the device-picking logic, e.g.:

```
def get_default_device() -> torch.device:
    """Get the default device to use.

    Returns:
        torch.device: Device to use.
    """
    if torch.backends.mps.is_built():
        return torch.device("mps")

    if torch.cuda.is_available():
        return torch.device("cuda")

    return torch.device("cpu")
```

This can also let us add XLA device support in the future (e.g. TPUs).

luciaquirke:
Closing because PyTorch doesn't support the bfloat16 data type on MPS yet. I will reopen in another few months or when I figure out how to contribute to PyTorch 😄 

If anyone is here because they want to use TransformerLens with MPS, this branch should work (unless you encounter a model that uses the failing data type, I haven't yet).

alan-cooney:
@luciaquirke I think it's fine to just not support mixed floating point precision for MPS?

Was it preventing other things from working? Otherwise it seems reasonable to merge this in for now, and then just support bfloat16 if/when it's added to PyTorch.

jbloomAus:
Sounds good to me!

alan-cooney:
Yep it's a very well-know issue with MPS as well

luciaquirke:
@jbloomAus please may I have a merge 😛

jbloomAus:
Looks good. Thanks Lucia!

Issue #290, opened 2023-05-30
[Bug Report] GPU Memory is leaked when HookedTransformer goes out of scope.


tbenthompson:
I was iterating some operations over the Pythia model suite and realized that each time I loaded a model, the GPU memory usage would increase. Because the memory usage as reported by torch and nvidia-smi is several GB, I'm guessing there's some arrays that are attached to a cache or class rather than to an instance of HookedTransformer. I would be happy to debug when I get a chance. For now, I'm able to just use transformers without TransformerLens and I was originally only using HookedTransformer out of habit. 

Here's a minimal reproducing example:

```
import os
import time
import transformers
import torch
from transformer_lens import HookedTransformer

model_param_str = ['70m', '160m', '410m']
model_names = {mps: f"EleutherAI/pythia-{mps}-deduped" for mps in model_param_str}

# GPU memory usage is 0 (as checked by nvidia-smi) after each model is loaded and garbage collected.
for param_str, model_name in model_names.items():
    model = transformers.GPTNeoXForCausalLM.from_pretrained(model_name)
    model.cuda()

del model
import gc
gc.collect()
torch.cuda.empty_cache()

print(torch.cuda.memory_summary())
# Give the GPU some time to update its memory usage.
time.sleep(1)
os.system('nvidia-smi')

# GPU memory usage accumulates from past models.
for param_str, model_name in model_names.items():
    model = HookedTransformer.from_pretrained(model_name)

del model
import gc
gc.collect()
torch.cuda.empty_cache()

print(torch.cuda.memory_summary())
time.sleep(1)
os.system('nvidia-smi')
```

and the output is:

```
|===========================================================================|
|                  PyTorch CUDA memory summary, device ID 0                 |
|---------------------------------------------------------------------------|
|            CUDA OOMs: 0            |        cudaMalloc retries: 0         |
|===========================================================================|
|        Metric         | Cur Usage  | Peak Usage | Tot Alloc  | Tot Freed  |
|---------------------------------------------------------------------------|
| Allocated memory      |      0 B   |   1642 MiB |   2608 MiB |   2608 MiB |
|       from large pool |      0 B   |   1641 MiB |   2600 MiB |   2600 MiB |
|       from small pool |      0 B   |      6 MiB |      7 MiB |      7 MiB |
|---------------------------------------------------------------------------|
| Active memory         |      0 B   |   1642 MiB |   2608 MiB |   2608 MiB |
[...]

PR #327, opened 2023-06-20
Reduce memory use when loading model

slavachalnev:
# Description

See discussion in #290

Thanks @rusheb for the testing script and for finding the circular reference.

This PR:
1. Removes circular reference in `HookedRootModule`'s `setup`.
2. Sets `requires_grad=False` before copying weights from HuggingFace models.

Running @rusheb's testing script

```python
from tqdm import tqdm

for _ in tqdm(range(10)):
    model = HookedTransformer.from_pretrained("gpt2-small", device="cpu")
```

```
Profiler reports peak tracked memory usage

No changes:          17551.0 MiB
Only fix reference:  12416.2 MiB
Only fix gradient:   7342.4 MiB
Fix both:            2207.5 MiB
```

This fix does not help with GPU memory use.

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -167,6 +167,8 @@ class HookedRootModule(nn.Module):
         self.mod_dict = {}
         self.hook_dict: Dict[str, HookPoint] = {}
         for name, module in self.named_modules():
+            if name == "":
+                continue
             module.name = name
             self.mod_dict[name] = module
             if "HookPoint" in str(type(module)):
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -940,6 +940,10 @@ def get_pretrained_state_dict(
                 )
 
             # Load model weights, and fold in layer norm weights
+
+        for param in hf_model.parameters():
+            param.requires_grad = False
+
         if cfg.original_architecture == "GPT2LMHeadModel":
             state_dict = convert_gpt2_weights(hf_model, cfg)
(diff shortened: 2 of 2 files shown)


slavachalnev:
It only uses 17GB if you load the model 10 times in a row haha

`model.named_modules` also returns the module itself with name `""`. This circular reference can hurt memory use (even though python's garbage collector shoule handle cycles like this)

neelnanda-io:
Oh great, I missed the for loop. I'm glad I didn't fuck up *that *badly!
And huh, I wasn't aware of that behaviour but I guess it makes sense,
thanks!

jbloomAus:
Crazy how simple the fix was! Thanks a bunch @slavachalnev

good-epic:
Was the GPU memory issue ever solved? I'm using the 3070ti on my laptop that has 8GB of GPU memory. Loading the Pythia-410m model as a HookedTransformer takes up ~1.65GB. Then trying model.run_with_cache() with a single sentence allocates 7GB before running into a CUDA out of memory error.

neelnanda-io:
Fascinating. I can't think of a good explanation for why that happens. Can
you give a minimal code snippet? And does it help to put
torch.set_grad_enabled(False)at the top?

good-epic:
Sorry for the spam here. I figured out as I was testing for the minimal code that would reproduce my error that I had lost an index and had been sending in one sentence at a time in my prior runs and was now trying to send in a list of 200 sentences. Back to one at a time and there's no obvious memory leak.

PR #334, opened 2023-07-02
Fix tokenization in utils.test_prompt

### LINKED ISSUE #271: [Bug Report] Tokenization in test_prompt in utils.py sometimes causes IndexError when prompt+answer is tokenized differently together vs separately
in utils.py, function test_prompt the whole prompt+answer string is tokenized, then we get the tokens in string form from each part separately
```
tokens = model.to_tokens(prompt + answer, prepend_bos=prepend_bos)
prompt_str_tokens = model.to_str_tokens(prompt, prepend_bos=prepend_bos)
answer_str_tokens = model.to_str_tokens(answer, prepend_bos=False)
```
the issue is, later in the code we get lengths of the latter two
```
prompt_length = len(prompt_str_tokens)
answer_length = len(answer_str_tokens)
```
and these lengths are used to index over the `tokens` tensor:
```
for index in range(prompt_length, prompt_length + answer_length):
        answer_token = tokens[0, index]
```

so if prompt_length+answer_length happens to be longer, because the model tokenized them differently than it would the whole prompt+answer string, especially when we set `prepend_space_to_answer` to `False`, we get `IndexError`
e.g.
```
 IndexError: index 15 is out of bounds for dimension 1 with size 15
```

Felhof:
# Description

Fixes #271. Before, test_prompt would throw an index error when the concatenated prompt and answer string had a different tokenization from the prompt and answer strings separately. The function now works for such cases. I also added a unit test checking that the issue is handled correctly.

## Type of change


jbloomAus:
@Felhof it seems like there's too many commits here / lots of files changed making it hard to review, can you branch off main with just the relevant changes and then resubmit please?

jbloomAus (dismissed):
Hey @Felhof,

Testing via console output feels icky. Is there no way to make this test operate on python objects?

Felhof:
Ok I reimplemented the test, so it mocks the print calls and asserts on the calls to make sure the correct values are printed.

tests/unit/test_utils.py
```diff
@@ -207,6 +209,135 @@ def test_fail(self, x: torch.Tensor):
         assert not utils.is_lower_triangular(x)
 
 
+@pytest.mark.parametrize(
+    "prepend_space_to_answer, expected_print_calls, expected_rprint_calls",
```
alan-cooney, line 213: I think you are checking that it tokenises correctly here? 

```suggestion
    "prepend_space_to_answer, tokenized_prompt, tokeized_answer",
```

I'd suggest changing to this ^ and then just putting the array in directly. The `call()` part can be done in the test.

tests/unit/test_utils.py
```diff
@@ -207,6 +209,135 @@ def test_fail(self, x: torch.Tensor):
...
+    )
+
+    for received_call, expected_call in zip(
+        mocked_print.mock_calls, expected_print_calls
+    ):
+        assert received_call == expected_call
+    for received_call, expected_call in zip(
+        mocked_rprint.mock_calls, expected_rprint_calls
+    ):
+        assert received_call == expected_call
```
alan-cooney, line 338: I don't think this is needed (and it's best practice to just test one thing/feature in a test - in this case tokenisation).

```suggestion
```

tests/unit/test_utils.py
```diff
@@ -207,6 +209,135 @@ def test_fail(self, x: torch.Tensor):
...
+        "The circumference is the perimeter of the circ",
+        "le.",
+        model,
+        prepend_space_to_answer=prepend_space_to_answer,
+    )
+
+    for received_call, expected_call in zip(
+        mocked_print.mock_calls, expected_print_calls
+    ):
+        assert received_call == expected_call
```
alan-cooney, line 334: You can now just check the first 2 calls

alan-cooney (changes requested):
Nice work on this - I've added some suggestions to make the tests much more readable.

alan-cooney:
Oh and worth adding a comment to the test explaining why it's a good check - that you might otherwise expect it to tokenize the end of the prompt with the answer as one token otherwise, which is bad.

alan-cooney (approved):
Thanks for the PR!

PR #336, opened 2023-07-08
Add `hook_attn_in`

ArthurConmy:
# Description

This new optional hook allows you to edit the input to each attention head separately. 

Also corrects earlier error in the shape of `hook_q_input` etc.

## Type of change

- [ :white_check_mark:  ] New feature (non-breaking change which adds functionality)

# Checklist:

- [ :white_check_mark: ] I have commented my code, particularly in hard-to-understand areas
- [ :white_check_mark: ] I have made corresponding changes to the documentation
- [ :white_check_mark: ] My changes generate no new warnings
- [ :white_check_mark: ] I have added tests that prove my fix is effective or that my feature works
- [ :white_check_mark: ] I have not rewritten tests relating to key interfaces which would affect backward compatibility


transformer_lens/HookedTransformer.py
```diff
@@ -1323,6 +1327,12 @@ def set_use_hook_mlp_in(self, use_hook_mlp_in: bool):
         """
         self.cfg.use_hook_mlp_in = use_hook_mlp_in
 
+    def set_use_attn_in(self, use_attn_in: bool):
+        """
+        Toggles whether to allow editing of inputs to each attention head.
+        """
+        self.cfg.use_attn_in = use_attn_in
```
Aprillion, line 1334: if I understand it correctly, the hook inside `model.add_hook("blocks.0.hook_q_input", ...)` will receive a different tensor shape before/after `model.set_use_attn_in(True)`?

if so, it would be nice to check in this method that no affected hooks are active before calling this method (and if so, to show a warning saying ``Please use `model.set_use_attn_in()` before adding any hooks, or after calling `model.reset_hooks()` - this method changes the tensor shape received by your hooks and you might experience unexpected errors when reusing old input hooks.``)

maybe even a few tests would be nice 😅

Aprillion (approved):
Python code looks OK, though I am not an expert in TransformerLens to say if this is the best way from usage/maintenance perspective.

tests/unit/test_hooks.py
```diff
@@ -149,6 +150,12 @@ def identity_hook(z, hook):
     model.set_use_split_qkv_input(True)
     model.add_hook("blocks.0.hook_q_input", identity_hook)
 
+    model.reset_hooks()
+    model.set_use_attn_in(True)
+    model.add_hook("blocks.0.hook_attn_in", identity_hook)
+
```
JayBaileyCS, line 156: Since the shape should be the same if split_qkv_input is True OR if set_use_attn_in is True, it might be worth a separate test here - my understanding is that this test would pass even if set_use_attn_in did nothing.

transformer_lens/components.py
```diff
@@ -859,14 +859,15 @@ def __init__(self, cfg: Union[Dict, HookedTransformerConfig], block_index):
             else:
                 self.mlp = MLP(cfg)
 
-        self.hook_q_input = HookPoint()  # [batch, pos, d_model]
-        self.hook_k_input = HookPoint()  # [batch, pos, d_model]
-        self.hook_v_input = HookPoint()  # [batch, pos, d_model]
+        self.hook_attn_in = HookPoint()  # [batch, pos, n_heads, d_model]
```
JayBaileyCS, line 862: Is hook_q_input through hook_v_input always [batch, pos, n_heads, d_model]? My understanding is that it's [batch, pos, n_heads, d_model] if split_qkv or attn_in, else [batch, pos, d_model]. If it might sometimes be [batch, pos, d_model] having [batch, pos, n_heads, d_model] as the comment is a bit misleading.

JayBaileyCS (changes requested):
Looks good overall - I think the tests need to be improved a bit (or I'm misunderstanding them) and I have a question about some of the comments but that's relatively minor.

transformer_lens/HookedTransformer.py
ArthurConmy: No, I don't think that this is correct. I think that `blocks.0.hook_q_input` will receive a 4D tensor whether `hook_attn_in` is on or off (note that the old comment annotations of a 3D shape were wrong)

What happens is that we either add the fourth dimension before the splitting into Q, K and V (if hook_attn_in is on) or we do the dimension adding as we split into Q, K and V (if hook_attn_in is off)

I will add all of this as a comment, thanks for pointing out that it is confusing!

transformer_lens/components.py
ArthurConmy: Isn't this false? As far as I can see `hook_q_input` just isn't used at all is `split_qkv` is False?

To ensure we are not talking through each other, I am talking about[ the implementation on main here](https://github.com/neelnanda-io/TransformerLens/blob/feefcf1/transformer_lens/components.py#L895#L910) and I don't think I'm changing this implementation?

transformer_lens/HookedTransformer.py
Aprillion: ah, thanks for explanation, number of expected dimensions souds like something to be covered by explicit assertion instead of a potentially incorrect comment (or at least some test cases, potentially in the future if someone is waiting for this PR)

ArthurConmy:
Just added 
1. Tests that check dimension sizes
2. Docs that explain that hook_q_input should always have the same shape
3. Merged main

transformer_lens/HookedTransformer.py
ArthurConmy: Yep! After writing this, I then made better tests based on @JayBaileyCS's comments. 

E.g I think this is done in [this PR](https://github.com/ArthurConmy/TransformerLens/blob/fc232b300c9849be32a9b59b6f4e09e15bcc6577/tests/unit/test_hooks.py#L158#L167)

transformer_lens/components.py
```diff
@@ -1016,25 +1017,41 @@ def forward(
...
+            key_input = self.hook_k_input(
+                add_head_dimension(attn_in)
+                if not self.cfg.use_attn_in
+                else attn_in  # we've already added the extra dimension if we're using attn_in!
+            )
+            value_input = self.hook_v_input(
+                add_head_dimension(attn_in)
+                if not self.cfg.use_attn_in
+                else attn_in  # we've already added the extra dimension if we're using attn_in!
+            )
```
jettjaniak, line 1048: This should be equivalent, unless I'm missing something?
```suggestion
            # we've already added the extra dimension if we're using attn_in!
            if not self.cfg.use_attn_in:
                attn_in = add_head_dimension(attn_in)
            query_input = self.hook_q_input(attn_in)
            key_input = self.hook_k_input(attn_in)
            value_input = self.hook_v_input(attn_in)
```

jettjaniak (changes requested):

transformer_lens/components.py
ArthurConmy: No, this specific idea shouldn't be used because inplace operations on `attn_in` will cause problems, similar to [here](https://github.com/neelnanda-io/TransformerLens/issues/157). In short, with your change, naively zero ablating `query_input` would zero ablate `key_input` and `value_input` too.

However, thanks for writing as I just noticed that I should be cloning `attn_in` here for the same reason!

jettjaniak: That makes sense, thanks. My issue was the duplicated code for q/k/v_input so I'm happy now.
Personally, I'd have `add_head_dimension` without cloning and clone explicitly. Also, should we make the same changes to BertBlock?

JayBaileyCS (approved):
Sorry about taking so long with this, had to be reminded - I've since set up my Github email to forward all emails to the one I use regularly, so this shouldn't reoccur.

PR #343, opened 2023-07-14
Introduce Global prepend_bos Attribute to HookedTransformer

### LINKED ISSUE #342: [Proposal] Introduce Global prepend_bos Attribute to HookedTransformer
### Proposal
The proposal is to introduce a global `self.prepend_bos` attribute in the HookedTransformer class. The attribute should be modifiable via a new `set_prepend_bos()` method. Also, all methods in the HookedTransformer class that have a `prepend_bos` parameter will now have a default value of None, and will use the value of the global `self.prepend_bos` attribute if not overridden.

### Motivation
The motivation for this proposal comes from the problem of managing the `prepend_bos` flag on individual method calls in the HookedTransformer class. Currently, methods that tokenize inputs have `prepend_bos=True` as their default setting. This means that whenever a user wants to use `prepend_bos=False` as the default, they need to manually set `prepend_bos=False` on _every method call that uses the parameter_, which is very error-prone and cumbersome.

With this proposal, a user can simply call set_prepend_bos(False) once to change the global behavior of the class, making it much easier to manage the `prepend_bos` setting.

### Pitch
By adding a global `self.prepend_bos` attribute and a corresponding set_prepend_bos() method, users can manage the `prepend_bos` setting at a global level. This way, they won't have to worry about setting `prepend_bos=False` for every method call when they want to disable automatic BOS prepending, improving the usability of the library.

### Alternatives
An alternative could be to simply document the current behavior and recommend users to manually set `prepend_bos=False` in each method call where it's required. However, this is not a user-friendly approach and doesn't improve the library's usability. The proposed change should result in a more intuitive and less error-prone user experience.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/transformerlens/issues) in the repo

soheeyang:
# Description

This PR introduces a new attribute, `self.prepend_bos`, to the `HookedTransformer` class. The `self.prepend_bos` flag determines the default behavior of whether a beginning of sequence (BOS) token should be added to the start of the string when it's tokenized. Previously, the default behavior of `prepend_bos` was controlled only as a parameter of several methods, with the default value set as `True`. This was not an issue for users who wanted to use `prepend_bos=True` as default. However, when users did not want to prepend BOS token as default, it required users to explicitly pass `prepend_bos=False` in every such method call, which was cumbersome and very error-prone.

With this update, default behavior of whether to prepend BOS or not is controlled by a global attribute `self.prepend_bos` of the `HookedTransformer` instances. It defaults to `True`, but can be updated using the new `set_prepend_bos` method. The default parameter value of `prepend_bos` in methods is now set to `None`. When it's `None`, the global `self.prepend_bos` is used. However, users can still locally override the behavior in each method call by passing a value to the parameter, e.g., `prepend_bos=True`. This change allows users to set their preference once at a global level and have it apply to all method calls, or to override it at a local level as required, thereby making the API more user-friendly and less error-prone.

The PR includes the following changes:

1. Add `self.prepend_bos` attribute to `HookedTransformer` and set its default value as `True`.
2. Add `set_prepend_bos()` method to allow users to change the `self.prepend_bos` value at a global level.
3. Modify methods within `HookedTransformer` to use the global `self.prepend_bos` attribute by default, but also to allow for local overrides by passing `prepend_bos=True` or `prepend_bos=False` in the method call.
4. Update comments and relevant unit tests to ensure the new functionality works as expected.

This PR improves the flexibility of tokenization and enhances the user experience by reducing the chances of unintended behavior and simplifying the function calls.

Fixes #342 

## Type of change

# Checklist:


tests/unit/test_prepend_bos.py
```diff
@@ -0,0 +1,77 @@
...
+        assert model.prepend_bos == False, "prepend_bos should be set to False"
+
+    def test_default_prepend_bos_at_method_level(self):
+        model = HookedTransformer.from_pretrained("gpt2")
+
+        logits = model(self.prompt)  # [batch pos d_vocab]
+        str_tokens = model.to_str_tokens(self.prompt)
+        tokens = model.to_tokens(self.prompt)
+
+        assert logits.shape[1] == len(str_tokens) == tokens.shape[1]
```
neelnanda-io, line 28: Can you explicitly test that this equals 3? (Assuming that "Hello world" tokenizes as BOS, 'Hello', ' World'). As is, maybe it _never_ prepends BOS

tests/unit/test_prepend_bos.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_set_prepend_bos_at_method_level(self, model):
+        for global_prepend_bos in [True, False]:
+            model.set_prepend_bos(global_prepend_bos)
+
+            logits = model(self.prompt)  # [batch pos d_vocab]
+            str_tokens = model.to_str_tokens(self.prompt)
+            tokens = model.to_tokens(self.prompt)
+
+            assert logits.shape[1] == len(str_tokens) == tokens.shape[1]
```
neelnanda-io, line 38: Ditto here

tests/unit/test_prepend_bos.py
neelnanda-io: I would create a function for "num_tokens_in_prompt" that takes in a prepend_bos Bool and returns the number of tokens in self.prompt, along with the prepend_bos (you can just hard-code these numbers IMO), and then calls in the equals expressions in the tests

transformer_lens/HookedTransformer.py
```diff
@@ -163,6 +163,11 @@ def __init__(
         # Needed for HookPoints to work
         self.setup()
 
+        # Prepend the BOS token to the input as default when the input is a string.
+        # Even for models not explicitly trained with this, heads often use the first position as a resting position
+        # and accordingly lose information from the first token, so this empirically seems to give better results.
+        self.set_prepend_bos(True)
```
neelnanda-io, line 169: Shouldn't this use whatever is in `self.cfg.prepend_bos`?

tests/unit/test_utils.py
```diff
@@ -205,3 +205,13 @@ def test_pass(self, x: torch.Tensor):
     )
     def test_fail(self, x: torch.Tensor):
         assert not utils.is_lower_triangular(x)
+
+
+def test_override_or_use_global_flag():
+    # Case when override is not None
+    assert utils.override_or_use_global_flag(global_flag=True, override=False) == False
```
neelnanda-io, line 212: Nit: Can you also add the True, True and False, False cases

transformer_lens/HookedTransformer.py
```diff
@@ -460,10 +471,20 @@ def set_tokenizer(self, tokenizer):
         if self.cfg.d_vocab_out == -1:
             self.cfg.d_vocab_out = self.cfg.d_vocab
 
+    def set_prepend_bos(self, prepend_bos: bool):
```
neelnanda-io, line 474: Can you rename to set_default_prepend_bos?

neelnanda-io (changes requested):
Thanks this is great! LGTM other than a few minor comments, this seems like a really useful feature

transformer_lens/HookedTransformer.py
soheeyang: Hi Neel, thank you for the review! I don't think that we have something like `cfg.prepend_bos`. Do you mean that we should have `default_prepend_bos` in the config rather than the model? In other words, that it would be better to add `default_prepend_bos` to `HookedTransformerConfig`?

soheeyang: I added `default_prepend_bos` to `HookedTransformerConfig`,
```
default_prepend_bos: bool = True
```

and updated `HookedTransformer` to use the value in the config to set `self.prepend_bos`.
```
self.set_default_prepend_bos(self.cfg.default_prepend_bos)
```

neelnanda-io:
Ah! Yes, please put `prepend_bos` into HookedTransformerConfig, maybe renamed to `default_prepend_bos`. I think that's neater.

soheeyang:
All done!

neelnanda-io (approved):

neelnanda-io:
Thanks so much, merged!

neelnanda-io:
And thanks for all the clear comments and tests, great PR! I liked the override_or_use_default_flag method

transformer_lens/HookedTransformer.py
soheeyang: FWIW, some design considerations: I wanted to use `model.prepend_bos` to control the default behavior (not `model.cfg.default_prepend_bos`) because the values in `HookedTransformerConfig` are intended to be access-only. In fact, `model.cfg.default_prepend_bos` is like a default value for default prepend bos (default of default). I used `model.prepend_bos` instead of `model.default_prepend_bos` because I think it makes it more explicit and clear that we are overriding the value of `self.prepend_bos` when we pass `prepend_bos` as the method parameter.

soheeyang:
Thank you! Happy to contribute to this awesome project 🥰

transformer_lens/HookedTransformer.py
neelnanda-io: Yeah, I see where you're coming from. I guess I don't see HookedTransformerConfig as access only? There's stuff that was set when the model was pre-trained (eg num layers), but also stuff a user can set (eg what precision to use, how man devices, etc). Possibly it woudl be better to have "fixed, pre-trained model config" and "user set config" class, idk.

soheeyang: Oh I see, by 'access-only', I meant something like 'once the values of the config are initialised, they are not updated throughout the lifetime of the process'. For example, I guess we don't expect the users to change the precision or number of devices in the middle of running the process. On the other hand, I think the users might want to change the default `prepend_bos` without restarting the process just to do some tests etc.

Also, since `set_default_prepend_bos()` is a setter method of `model`, it feels a bit weird to change the value of `model.cfg.default_prepend_bos` with a setter of `model`.

soheeyang: Actually, since we already added `default_prepend_bos` to the config, how about getting rid of the `set_default_prepend_bos()` and let the user to just do `HookedTransformer.from_pretrained(model_name, ..., default_prepend_bos=False)` if they want to turn off the behavior?

PR #344, opened 2023-07-16
Add support for left padding

### LINKED ISSUE #240: [Proposal] Add support for padding/masking to the attention computation (Minor feature request)

soheeyang:
# Description

This PR introduces support for left padding for decoder-only models by _properly adjusting the absolute position embedding and causal/attention masks_. This change enables an easier retrieval of the final outputs, e.g., `logits[:, -1, :]`, getting rid of the need to perform a complicated indexing, even when a batch of varying lengths of sequences are processed.

- **Introduction of `util.get_attention_mask()` Method**: Introduced a new method that is called internally by the model in `forward()` to generate attention masks based on the input `tokens` (works for both left-padded or right-padded `tokens`) and `prepend_bos` flag used to create the token. This mask ensures that the model does not pay attention to padding tokens during processing; while it is not necessary when right padding is used, it becomes necessary when left padding is used to adjust the absolute position embedding and attention. The method also handles a special case where the `BOS` token is identical to the `pad` token and is prepended to the input sequences, so that such a token can be properly attended.
  - Design consideration: One thing to note is that `prepend_bos` to pass to the function should be the same with the one used to create `tokens`. At first, I considered making the function infer the value of `prepend_bos` by investigating `tokens` without the need to pass the value as a parameter, but then it requires several assumptions on how `tokens` is processed, which might not be always true. Therefore, I decided to go with the current design.

- **Adjustment of Positional Embeddings**: Made changes to the positional embeddings such that the first real token in each sequence receives the 0th positional embedding. This is critical to the proper functioning of the model when left padding is used. The positional embeddings of the pad tokens are set to zero vectors.
  - Design consideration: I separated the calculation of the positional embeddings for the left padding from that of the right padding for computational efficiency.

- **Adjustment of Causal Masking**: Made changes to the causal mask such that the causal mask also prevents the model from attending to the pad tokens that appears at the front of the real tokens.
  - Design consideration: Again, I separated the calculation for the left padding from that of the right padding for computational efficiency.

- **Test Cases**: Added comprehensive test cases to ensure the correctness of the left padding feature, attention mask generation, adjusted positional embeddings, adjusted causal mask, and the same output behaviours for all of right padded tokens, left padded tokens, and single batch tokens.

- **Explanation**: Added an explanation of the support for the left padding in the exploratory analysis demo.

Fixes #240 

## Type of change

#### before
```
# Requires tedious indexing.
# If you want to get the last token outputs of not only the logits,
# but also the cached activations, etc., the need to do this kind of indexing every time can become very annoying.

model.tokenizer.padding_side = "right"

num_str_tokens_list = [len(t) for t in model.to_str_tokens(prompts)]
logits, cache = model.run_with_cache(prompts)
last_token_positions = torch.tensor(num_str_tokens_list, device=logits.device) - 1
last_logits = logits[torch.arange(len(prompts)), last_token_positions, :].squeeze(1)
```

#### after
```
# life made easier

model.tokenizer.padding_side = "left"

logits, cache = model.run_with_cache(prompts)
last_logits = left_logits[:, -1, :]
```

# Checklist:


jbloomAus:
Thanks! I'm on holiday but @JayBaileyCS will take a look and I'll also look once he's done that since we've discussed this feature before :)

soheeyang:
Hi Joseph,

Thank you for the comment.

I've gone through the discussions you had with @JayBaileyCS on this feature.

One thing that I want to discuss is about your comment in https://github.com/neelnanda-io/TransformerLens/pull/291#discussion_r1228843439.

I think there are mainly three options on how to enable left padding:

(BTW, I think `padding_side` is better than `left_pad` because it matches with what `transformers.Tokenizer` actually uses, so I will use `padding_side` in the example below.)

1. Add `padding_side="right"` argument to all the methods of `HookedTransformer` that preprocess tokens (these methods are the methods which uses `prepend_bos` as the parameter, so we can easily locate them), so that left padding is used when `padding_side="left"` is passed as the argument.
    - This update alone has a disadvantage that if the user wants to set `padding_side=left` as default, they need to do it every time they call a string processing method. There was a similar issue with `prepend_bos=True`; I raised the issue in https://github.com/neelnanda-io/TransformerLens/issues/342 and fixed it in https://github.com/neelnanda-io/TransformerLens/pull/343.
2. Add `set_padding_side()` method to `HookedTransformer` that the user can call to change the padding side. The default value of the attribute is determined by `default_padding_side = "right"` of `HookedTransformerConfig`.
    ```python3
    # in HookedTransformer class

    @property
    def padding_side(self):
        return self.tokenizer.padding_side

    def set_padding_side(self, padding_side: str):
        assert padding_side in ['left', 'right], f"padding_side must be either 'left' or 'right', but {padding_side} is given"
        self.tokenizer.padding_side = padding_side
    ```
    - This is exactly what I added (and got merged to the main branch by @neelnanda-io yesterday) for https://github.com/neelnanda-io/TransformerLens/pull/343 so that option 1 (the original code) and 2 (newly added code) are used together to solve the issue.
3. Add `padding_side: str = "right"` to `HookedTransformerConfig` and let the model access this value to determine the padding side.
    ```python3
    # in HookedTransformerConfig class

    default_padding_side: str = "right"

    def __post_init__(self):
        assert self.padding_side in ['left', 'right], f"padding_side must be either 'left' or 'right', but {self.padding_side} is given"
    ```
    - This might be better if we want to nudge the user to set the padding side _only once_ and prevent them from the urge to change it in the middle of the lifetime of the process. Why would we want to do this? To prevent the user from shooting their foot by using inconsistent `padding_side`. (I also made this suggestion in the PR for `prepend_bos`: https://github.com/neelnanda-io/TransformerLens/pull/343#discussion_r1264498738.) However, I think it is better to go with option 2 rather than this one, because I can imagine the actually use cases where the users want to change the default padding setting without creating another `HookedTransformer` instance.

While I was writing this, I decided to _add both option 1 and option 2_ as I did for `prepend_bos`. I will add this commit shortly, but I am still open to discussions.

soheeyang:
I initially wrote in the above comment that I'd prefer option 1+2, but I changed my mind to option 1+3. The reason is that as I described in the last part of option 3, it is not a good idea to make the user call a setter for a _default_ value _separately after the instance is created_. For example, this kind of issue can happen when we go with option 2 instead of option 3:

```python3
# (Let's assume that I am using Jupyter notebook.)
[1] model = HookedTransformer.from_pretrained("gpt2")

# (This logits and cache are calculated with right padding.)
[2] movie_review_logits, movie_review_cache = model.run_with_cache(movie_reviews)

# ... (I do a lot of things and forget the fact that movie_review_logits and movie_review_cache used right padding)

# Oh! I just realised that I can use left padding as default.
# It would be much more convenient to use left padding as default.
[203] model.set_default_padding_side("left")

# (This logits and cache are calculated with left padding.)
[204] book_review_logits, book_review_cache = model.run_with_cache(book_reviews)

...

# Okay, now let's investigate the activations of the model...
[246] movie_review_last_pre8 = movie_review_cache['pre8'][:, -1, :]  # This is an unintentional bug!!!!
      book_review_last_pre8 = book_review_cache['pre8'][:, -1, :]
      ...
```

@neelnanda-io For the same reason, I think that it would be better to change `prepend_bos` to also use option 3 rather than option 2. How do you think about this? If you agree with this suggestion, I will make a quick PR for `prepend_bos` as well.

neelnanda-io:
Oh lol I didn't realise you'd done 2 rather than 3 - I agree that 3 is better and would be happy to get a quick PR shifting prepend_bos to option 3

soheeyang:
@neelnanda-io Oh no, I just realised that I didn't even explain the difference between option 3 and option 2 in https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637109914. Sorry for making you (and possibly others as well) confused 😂 The code I put as the description for option 3 is also applied in option 2, which is why you said "I didn't realise you'd done 2 rather than 3".

The thing that I really wanted to change (from option 2) with option 3 was when and where the default value from the user is given to the model and is set. To be more specific, I want to get rid of `model.set_default_prepend_bos()` and change the time of default value setup to instantiation of `HookedTransformer`, with the rational described at https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637116480.

soheeyang:
@neelnanda-io Oh no, I just realised that I didn't even explain the difference between option 3 and option 2 in https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637109914. Sorry for making you (and possibly others as well) confused 😂 The code I put as the description for option 3 is also applied in option 2, which is why you said "I didn't realise you'd done 2 rather than 3".

The thing that I really wanted to change (from option 2) with option 3 was when and where the default value from the user is given to the model and is set. To be more specific, I want to make the following changes:
1. Get rid of `model.set_default_prepend_bos()` and `model.prepend_bos`.
2. If the user wants to change the default `prepend_bos` value to `False`, instead of calling `model.set_default_prepend_bos()`, they should pass the value when they create the `model` instance. Then, the model will set the value to `model.cfg.default_prepend_bos`. If the user does not provide the default value, `True` is used.
3. Instead of having a separate `model.prepend_bos`, the model directly accesses to `self.cfg.default_prepend_bos` to check the default setup (which can be locally overriden by `prepend_bos=...` is given to the method call of string processing methods).

The rationale for this change is described in https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637116480.

neelnanda-io:
Doing it all via the config sounds good to me, I agree having special
methods and attributes is unnecessary. Doing everything via the config
object, including being able to edit it on the fly, seems good by me

soheeyang:
Thank you so much for the feedback! I'll make the quick PR when I wake up (it's time for 🛌 now haha).

@jbloomAus @JayBaileyCS I will [do the same thing](https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637217499) with `padding_side` as well and make an update to this PR afterwards.

soheeyang:
@neelnanda-io I just thought of an idea that can nicely handle all the overriden flags without touching the code of the actual methods. How do you find this? I originally came up with this to guarantee the reset of the default value of `tokenizer.padding_side`, but I think we can just handle both of `prepend_bos` and `padding_side` together here. The advantage of using this for `prepend_bos` is that others who make changes to the code of the methods don't need to worry about things like the position of these overriden flags or so, and it's much more simple with just one decorator, removing all the duplicated code of calling `override_or_use_default_flag(prepend_bos, self.cfg.default_prepend_bos)`!

```python3
# in utils.py

def locally_override_defaults_and_restore(function):
    """
    This decorator is used to override the parameters with default values during a function's execution.
    It guarantees that the default parameters are not changed after the function execution,
    even when an error occurs during the execution.
    """
    def wrapper(self, *args, **kwargs):
        # prepend_bos
        # Prepare the overriden value
        prepend_bos = kwargs.pop("prepend_bos", None)
        assert prepend_bos in [None, True, False], (
            f"prepend_bos must be one of None, True, or False, but got {prepend_bos}."
        )
        kwargs["prepend_bos"] = override_or_use_default_flag(prepend_bos, self.cfg.default_prepend_bos)

        # padding_side
        # Prepare the overriden value
        padding_side = kwargs.pop("padding_side", None)
        assert padding_side in [None, "left", "right"], (
            f"padding_side must be one of None, 'left', or 'right', but got {padding_side}."
        )
        kwargs["padding_side"] = override_or_use_default_flag(padding_side, self.cfg.default_padding_side)

        # Guarantees consistent behavior even if the user accidentally changed
        # `model.tokenizer.padding_side` instead of `self.cfg.default_padding_side`
        if self.tokenizer.padding_side != self.cfg.default_padding_side:
            with warnings.catch_warnings():
                warnings.simplefilter("always")
                warnings.warn(
                    f"model.tokenizer.padding_side ({self.tokenizer.padding_side}) is not the same as \
                    model.cfg.default_padding_side ({self.cfg.default_padding_side}). \
                    model.tokenizer.padding_side = {self.cfg.default_padding_side} is set and \
                    will be used as the default padding side from now on. \
                    Please change model.cfg.default_padding_side if you want to use a different padding side."
                )
            self.tokenizer.padding_side = self.cfg.default_padding_side

        try:
            # This is important because self.tokenizer.padding_side is
            # the actual padding_side used by Transformers Tokenizer.
            self.tokenizer.padding_side = kwargs["padding_side"]
            
            # Execute the original function with the overridden padding_side
            outputs = function(self, *args, **kwargs)
            
            # Reset the padding_side of the tokenizer
            self.tokenizer.padding_side = self.cfg.default_padding_side
            return outputs

        except Exception as e:
            # If an error occurs, reset the padding_side of the tokenizer before propagating the exception
            self.tokenizer.padding_side = self.cfg.default_padding_side
            raise e

    return wrapper

# in HookedTransformer class

    @utils.locally_override_defaults_and_restore  # This is all we need!
    def to_str_tokens(
        self,
        input: Union[
            str,
            Int[torch.Tensor, "pos"],
            Int[torch.Tensor, "1 pos"],
            Int[np.ndarray, "pos"],
            Int[np.ndarray, "1 pos"],
            list,
        ],
        prepend_bos: Optional[bool] = None,
        padding_side: Optional[Literal["left", "right"]] = None,
    ) -> Union[List[str], List[List[str]]]:
```

transformer_lens/utils.py
```diff
@@ -886,3 +886,62 @@ def get_causal_mask_for_left_padding(
...
+    def wrapper(self, *args, **kwargs):
+        sig = inspect.signature(function)
+        arg_names = list(sig.parameters.keys())
+        if arg_names[0] == "self":
+            arg_names = arg_names[1:]
+
+        arg_values = dict(zip(arg_names, args))
+        arg_values.update(kwargs)
+
+        # prepend_bos
```
JayBaileyCS, line 906: I assume this comment section should be uncommented or removed?

transformer_lens/utils.py
```diff
@@ -807,3 +807,141 @@ def override_or_use_default_flag(
...
+) -> Int[torch.Tensor, "batch pos"]:
+    """
+    This method calculates an attention mask for a given sequence of tokens when left padding is used. The attention
+    mask is a binary tensor where each element is 1 if the corresponding token is not a pad token and 0 otherwise. In
+    the special case where the BOS token is the same as the pad token and bos token is prepended to process the tokens,
+    this method ensures that the prepended BOS (pad) token also gets a 1 in the attention mask.
+    Gotcha: Make sure that the passed prepend_bos is the same as the one used to generate the tokens!
+    """
+    # Initialize the attention mask to be 1 wherever the token is not a pad token
+    attetnion_mask = tokens.ne(tokenizer.pad_token_id)
```
JayBaileyCS, line 825: Typo

JayBaileyCS (commented):
Overall this looks quite good! I'm humbled by the skill that went into this, having tried and failed to create this feature myself. There's quite a few things in here that I'm going to try to learn from, like the use of .signature and creating a custom decorator to solve this one.

I've made a couple small comments, but overall I am very impressed!

transformer_lens/utils.py
```diff
@@ -807,3 +807,141 @@ def override_or_use_default_flag(
...
+    if tokenizer.bos_token_id == tokenizer.pad_token_id and prepend_bos:
+        if tokenizer.padding_side == "right":
+            # Set the first token to 1 in the attention mask
+            attetnion_mask[:, 0] = True
+
+        else:
+            is_pad_token = 1 - attetnion_mask.int()
+
+            # Find the position of the pad token used as the BOS token and thus should get attended
+            pad_bos_positions = is_pad_token.cumsum(dim=-1).argmax(dim=-1)
```
neelnanda-io, line 837: I believe this will break when the BOS EOS *and* PAD tokens are the same, as is the case in eg GPT-2. It would be good to add a test for this case. EOS is often used to separate documents in the same context, eg "<PAD><PAD><BOS>The cat sat on the mat<EOS>The dog sat on the log". I would add a separate check that the pad tokens are contiguous (eg I believe ((is_pad_token.cumsum(-1) * ((1-is_pad_token).cumsum(-1)>0)).argmax(dim=-1) should work?) I think you should also check for the case where there's no PAD/BOS/EOS in the context, and add a test for correct handling. And this can be different by row, ugh.

This is very fiddly, sorry for all the nitpicks!

transformer_lens/utils.py
```diff
@@ -807,3 +807,141 @@ def override_or_use_default_flag(
...
+    mask is a binary tensor where each element is 1 if the corresponding token is not a pad token and 0 otherwise. In
+    the special case where the BOS token is the same as the pad token and bos token is prepended to process the tokens,
+    this method ensures that the prepended BOS (pad) token also gets a 1 in the attention mask.
+    Gotcha: Make sure that the passed prepend_bos is the same as the one used to generate the tokens!
+    """
+    # Initialize the attention mask to be 1 wherever the token is not a pad token
+    attetnion_mask = tokens.ne(tokenizer.pad_token_id)
+
+    # Handle the special case where the BOS token is the same as the pad token
+    if tokenizer.bos_token_id == tokenizer.pad_token_id and prepend_bos:
```
soheeyang, line 828: @neelnanda-io Hi Neel, the special case that you just pointed out is handled here!

transformer_lens/utils.py
```diff
@@ -807,3 +807,141 @@ def override_or_use_default_flag(
...
+    return mask
+
+
+def locally_override_and_restore_defaults(function):
+    """
+    This decorator is used to override the parameters with default values during a function's execution.
+    It guarantees that the default parameters are not changed after the function execution,
+    even when an error occurs during the execution.
+    """
+    def wrapper(self, *args, **kwargs):
```
neelnanda-io, line 897: Maybe rename function to method and wrapper to wrapped_method? The presence of `self` makes this clearly a method specific.

transformer_lens/utils.py
```diff
@@ -807,3 +807,141 @@ def override_or_use_default_flag(
...
+        )
+
+    return mask
+
+
+def locally_override_and_restore_defaults(function):
+    """
+    This decorator is used to override the parameters with default values during a function's execution.
+    It guarantees that the default parameters are not changed after the function execution,
+    even when an error occurs during the execution.
```
neelnanda-io, line 895: Nit: This is a fairly complex construction (that I've found it hard to immediately parse on reading!). It would be good to add a more detailed comment - eg on where the decorator is intended to be used, roughly how it works, etc.

transformer_lens/utils.py
soheeyang: Right, the assumption here was that all the PAD tokens are contiguously at the beginning and no PAD token appears in the middle. I didn't consider the case of PAD token being an EOS token and appearing at the middle of the text. The suggestion of checking if they are contiguous is great. Thank you so much for pointing out the exceptional case!

transformer_lens/utils.py
soheeyang: Hi Jay, you are right! I was actually considering hiding a refactoring commit of the overriding logic of `prepend_bos`, which was the reason why I just commented it out temporarily, but it will be better to separate the PRs.

transformer_lens/utils.py
soheeyang: Oh right, thank you for pointing it out!

soheeyang:
> Overall this looks quite good! I'm humbled by the skill that went into this, having tried and failed to create this feature myself. There's quite a few things in here that I'm going to try to learn from, like the use of .signature and creating a custom decorator to solve this one.
> 
> I've made a couple small comments, but overall I am very impressed!

Thank you so much for the compliment! Actually, the use of .signature and custom decorator makes the code complicated and harder to read, which increases the risk of potential bug. I used the approach just because I couldn't think of a simpler one, but Neel just suggested me a brilliant idea of using a context manager instead, which makes the code simpler and less bug-prone, getting rid of the need of complicated function signature parting but can do the exactly the same thing. I'll make this update and commit it soon!

transformer_lens/HookedTransformer.py
```diff
@@ -815,6 +870,7 @@ def from_pretrained(
         n_devices=1,
         tokenizer=None,
         move_to_device=True,
+        padding_side="right",
```
soheeyang, line 873: This is a rather minor point, but I'm also pondering about which is better between `padding_side` and `default_padding_side`. It's `default_prepend_bos` for `prepend_bos`, but since `padding_side` is an actual attribute of `transformers.Tokenizer`, `padding_side` might be more intuitive for those who know the actual attribute.

Advantages of each:
- `default_padding_side`: It's better for consistency with `default_prepend_bos` and it explicitly tells that it's a default value
- `padding_side`: It's more clear that it will be set as the value of `model.tokenizer.padding_side`

neelnanda-io:
Hey! I wanted to check up on the progress of this @soheeyang ?

soheeyang:
Hi @neelnanda-io, thank you so much for the reminder! I've made all the necessary updates except for some minor things, but my schedule has been so packed these days and I have completely forgotten about this 😂 I will make the necessary commits today or tomorrow!

transformer_lens/HookedTransformer.py
```diff
@@ -271,126 +285,154 @@ def forward(
...
+                    assert (
+                        past_kv_cache is not None
+                    ), "If past_left_attention_mask is not None, past_kv_cache must not be None"
+                    assert (
+                        tokens.shape[1] == 1
+                    ), "If past_left_attention_mask is not None, tokens must be a single token along the sequence dimension"
+                    # past_kv_cache is not None, so we're doing caching.
+                    # We need to extend the past_left_attention_mask.
+                    # Append '1' to the right of the past_left_attention_mask to account for the new tokens
+                    left_attention_mask = utils.extend_tensor_with_ones(
```
soheeyang, line 326: I couldn't think of a way that avoids introducing `past_left_attention_mask`. Any idea?

transformer_lens/HookedTransformer.py
```diff
@@ -271,126 +285,154 @@ def forward(
...
-        elif self.cfg.positional_embedding_type == "rotary":
-            # Rotary doesn't use positional embeddings, instead they're applied when dot producting keys and queries.
-            # See HookedTransformerConfig for details
-            residual = embed
-            shortformer_pos_embed = None
-        else:
-            raise ValueError(
-                f"Invalid positional_embedding_type passed in {self.cfg.positional_embedding_type}"
-            )
+        with utils.LocallyOverridenDefaults(
```
soheeyang, line 288: @neelnanda-io I changed the decorator to a context manager; a downside of using a context manager is that it is more cumbersome to add than a decorator.

transformer_lens/utils.py
```diff
@@ -799,13 +801,241 @@ def get_device():
...
+    # Set the nested attribute's value
+    setattr(obj, attrs[-1], value)
+
+
+class LocallyOverridenDefaults:
+    """
+    Context manager that allows temporary overriding of default values within a model.
+    Once the context is exited, the default values are restored.
+
+    WARNING: This context manager must be used for any function/method that directly accesses
```
soheeyang, line 955: I'm a bit worried about the cases where people forget to add a context manager, but it's hard to enforce it... One possible way to do this is to do something like the following code, but since `padding_side` is a property of `Tokenizer`, how to deal with it is not very straightforward.

@dataclass
class HookedTransformerConfig:
    ...
    _default_prepend_bos: bool = field(default=True, repr=False)
    _inside_context: bool = field(default=False, init=False, repr=False)

    @property
    def default_prepend_bos(self):
        if not self._inside_context:
            raise ValueError("Direct access to default_prepend_bos without context manager is prohibited.")
        return self._default_prepend_bos

    def set_inside_context(self, state: bool):
        self._inside_context = state

transformer_lens/utils.py
soheeyang: Resolved by commit 0591924a63b2591a02b84cfa6f138d0c4303a1b5

soheeyang: Note: 0591924a63b2591a02b84cfa6f138d0c4303a1b5 always considers (regardless of whether pad == bos or pad == sep) _only_ the leftmost leading pads or rightmost trailing pads as _real_ pad tokens that should not be attended.
Rational: Sep tokens should be attended, and there might be the cases where users use sep tokens multiple times.

neelnanda-io (commented):
This looks good to me on a skim, but I don't have the time to read in detail. @JayBaileyCS do you have the time to look at this? If not, I'll just approve.

JayBaileyCS:
@neelnanda-io The change has now gotten too big and involved for me to be confident in my own judgment of it, in all honesty. I don't have a good enough understanding of the HookedTransformer in order to track all the variables that have shifted. If it looks good to you on a skim, that's probably better than what I could do for this one, so if you want to approve it I'd say go ahead.

neelnanda-io:
Fair enough! Thanks Jay. @soheeyang is it ready to be merged in?

soheeyang:
> Fair enough! Thanks Jay. @soheeyang is it ready to be merged in?

@neelnanda-io Yes, it's ready!

soheeyang:
> Fair enough! Thanks Jay. @soheeyang is it ready to be merged in?

@neelnanda-io Yes, it's ready!

Now that we can use left padding, I want to make an enhancement in `generate` so that it is possible to generate
- multiple strings -> strings
- batched tensors -> strings.
 
However, I will make it as another PR rather than making the change in this PR.

transformer_lens/HookedTransformer.py
```diff
@@ -84,15 +87,16 @@ def __init__(
         self.cfg = cfg
 
         if tokenizer is not None:
-            self.set_tokenizer(tokenizer)
+            self.set_tokenizer(tokenizer, default_padding_side=default_padding_side)
```
mivanit, line 90: What is the reason for overwriting the tokenizer padding direction like this? This caused some existing trained (left-padded) models to break in a very confusing way, since the tokenizer property we were setting was being overridden without us noticing.

(in reference to #344 )

soheeyang: The reason was that left-padded models were technically not supported previous to this commit, which means that only right-padded models were supported as default. Left-padded models would not have worked correctly previous to this PR.

Could you give me more details on what model you used and how the model was breaking?

mivanit: We didn't seem to have problems training and running left-padded models, but perhaps there were underlying issues which we did not notice. 

I managed to fix the issue by modifying the tokenizer in our model's `__init__` after it gets modified by the new `HookedTransformer.__init__`
https://github.com/understanding-search/maze-transformer/pull/195

PR #345, opened 2023-07-17
Enhance the API for default_prepend_bos

soheeyang:
# Description

This PR enhances the API for `default_prepend_bos` which is introduced in PR #343. The PR includes the following changes:

1. member attribute `prepend_bos` and method `set_default_prepend_bos()` are removed from `HookedTransformer`.
2. Instead of calling `model.set_default_prepend_bos()` to change the default value of whether to prepend BOS or not, users can change the default value when they create the model instance e.g., `model = Transformer.from_pretrained('gpt2', default_prepend_bos=False)`.
3. If users can change the default value _after_ the instantiation of the model is done, they can use `model.cfg.default_prepend_bos = ...` (hopefully with enough understanding of the risk).
4. Docstrings, tests, and demo files are updated accordingly.

Enhances #343, of which the API design for setting the default value for `prepend_bos` was not optimal because:

1. It is not a good idea to make the user call a setter for a _default_ value _separately after the instance is initiated_. For example, users can make this kind of mistake when they are not careful enough:

```python3
# (Let's assume that the user is using Jupyter notebook.)
[1] model = HookedTransformer.from_pretrained("gpt2")
    movie_review = 'This movie is awesome!'

# (This logits and cache are calculated with `default_prepend_bos=True`.
# Let's assume that the user was however unaware of this default behavior.)
[2] movie_review_logits, movie_review_cache = model.run_with_cache(movie_review)

[3] str_movie_review_tokens = model.to_str_tokens(movie_review)

# 👤: Oh, what? What is this '<|endoftext|>'?
[4] str_movie_review_tokens
    >>> ['<|endoftext|>', 'This', ' movie', ' is', ' awesome', '!']

# (The user learns that `default_prepend_bos` is set to `True` as default
# and they can change the default value using `set_default_prepend_bos()`)

# 👤: I don't want to prepend a BOS token to the inputs.
[5] model.set_default_prepend_bos(False)

# 👤: Okay, now this is what I wanted.
[6] str_movie_review_tokens = model.to_str_tokens(movie_review)
    str_movie_review_tokens
    >>> ['This', ' movie', ' is', ' awesome', '!']

# 👤: I want to investigate the output logits at 'is' token.
[7] is_logits = movie_review_logits[0, str_movie_review_tokens.index('is')]

# There is a bug in `is_logits`, since the logits were calculated with a BOS,
# thus the actual logits for 'is' is `movie_review_logits[0, str_movie_review_tokens.index('is') + 1]`.
```

2. It might be confusing for the code readers because it is not clear how `model.cfg.default_prepend_bos` and `model.prepend_bos` are used differently without reading the code thoroughly.

3. Having a setter (`set_default_prepend_bos()`) and member attribute (`model.prepend_bos`) is unnecessary when we can just directly access (and change) the value of `model.cfg.default_prepend_bos`.

## Type of change

#### before
```
# to change the default value
model = HookedTransformer.from_pretrained("gpt2")
model.set_default_prepend_bos(False)

# to change the default value again (although not recommended)
model.set_default_prepend_bos(True)
```

#### after
```
# to change the default value
model = HookedTransformer.from_pretrained("gpt2", default_prepend_bos=False)

# to change the default value again (although not recommended)
model.cfg.default_prepend_bos = True
```

# Checklist:


soheeyang:
@neelnanda-io I just thought of an idea that can nicely handle all the overriden flags without touching the code of the actual methods that they the local overriding should be applied. How do you find this? I originally came up with this to guarantee the reset of the default value of `tokenizer.padding_side`, but I think we can just handle both of `prepend_bos` and `padding_side` like this. The advantage of using this for `prepend_bos` is that others who make changes to the code of the methods don't need to worry about things like the position of these overriden flags or so.

```python3
# in utils.py

def locally_override_defaults_and_restore(function):
    """
    This decorator is used to override the parameters with default values during a function's execution.
    It guarantees that the default parameters are not changed after the function execution,
    even when an error occurs during the execution.
    """
    def wrapper(self, *args, **kwargs):
        # prepend_bos
        # Prepare the overriden value
        prepend_bos = kwargs.pop("prepend_bos", None)
        assert prepend_bos in [None, True, False], (
            f"prepend_bos must be one of None, True, or False, but got {prepend_bos}."
        )
        kwargs["prepend_bos"] = override_or_use_default_flag(prepend_bos, self.cfg.default_prepend_bos)

        # padding_side
        # Prepare the overriden value
        padding_side = kwargs.pop("padding_side", None)
        assert padding_side in [None, "left", "right"], (
            f"padding_side must be one of None, 'left', or 'right', but got {padding_side}."
        )
        kwargs["padding_side"] = override_or_use_default_flag(padding_side, self.cfg.default_padding_side)

        # Guarantees consistent behavior even if the user accidentally changed
        # `model.tokenizer.padding_side` instead of `self.cfg.default_padding_side`
        if self.tokenizer.padding_side != self.cfg.default_padding_side:
            with warnings.catch_warnings():
                warnings.simplefilter("always")
                warnings.warn(
                    f"model.tokenizer.padding_side ({self.tokenizer.padding_side}) is not the same as \
                    model.cfg.default_padding_side ({self.cfg.default_padding_side}). \
                    model.tokenizer.padding_side = {self.cfg.default_padding_side} is set and \
                    will be used as the default padding side from now on. \
                    Please change model.cfg.default_padding_side if you want to use a different padding side."
                )
            self.tokenizer.padding_side = self.cfg.default_padding_side

        try:
            # This is important because self.tokenizer.padding_side is
            # the actual padding_side used by Transformers Tokenizer.
            self.tokenizer.padding_side = kwargs["padding_side"]
            
            # Execute the original function with the overridden padding_side
            outputs = function(self, *args, **kwargs)
            
            # Reset the padding_side of the tokenizer
            self.tokenizer.padding_side = self.cfg.default_padding_side
            return outputs

        except Exception as e:
            # If an error occurs, reset the padding_side of the tokenizer before propagating the exception
            self.tokenizer.padding_side = self.cfg.default_padding_side
            raise e

    return wrapper

# in HookedTransformer class

    @utils.adjust_padding_side  # This is all we need!
    def to_str_tokens(
        self,
        input: Union[
            str,
            Int[torch.Tensor, "pos"],
            Int[torch.Tensor, "1 pos"],
            Int[np.ndarray, "pos"],
            Int[np.ndarray, "1 pos"],
            list,
        ],
        prepend_bos: Optional[bool] = None,
        padding_side: Optional[Literal["left", "right"]] = None,
    ) -> Union[List[str], List[List[str]]]:
```

soheeyang:
@neelnanda-io I just thought of an idea that can nicely handle all the overriden flags without touching the code of the actual methods. How do you find this? I originally came up with this to guarantee the reset of the default value of `tokenizer.padding_side`, but I think we can just handle both of `prepend_bos` and `padding_side` together here. The advantage of using this for `prepend_bos` is that others who make changes to the code of the methods don't need to worry about things like the position of these overriden flags or so, and it's much more simple with just one decorator, removing all the duplicated code of calling `override_or_use_default_flag(prepend_bos, self.cfg.default_prepend_bos)`!

```python3
# in utils.py

def locally_override_defaults_and_restore(function):
    """
    This decorator is used to override the parameters with default values during a function's execution.
    It guarantees that the default parameters are not changed after the function execution,
    even when an error occurs during the execution.
    """
    def wrapper(self, *args, **kwargs):
        # prepend_bos
        # Prepare the overriden value
        prepend_bos = kwargs.pop("prepend_bos", None)
        assert prepend_bos in [None, True, False], (
            f"prepend_bos must be one of None, True, or False, but got {prepend_bos}."
        )
        kwargs["prepend_bos"] = override_or_use_default_flag(prepend_bos, self.cfg.default_prepend_bos)

        # padding_side
        # Prepare the overriden value
        padding_side = kwargs.pop("padding_side", None)
        assert padding_side in [None, "left", "right"], (
            f"padding_side must be one of None, 'left', or 'right', but got {padding_side}."
        )
        kwargs["padding_side"] = override_or_use_default_flag(padding_side, self.cfg.default_padding_side)

        # Guarantees consistent behavior even if the user accidentally changed
        # `model.tokenizer.padding_side` instead of `self.cfg.default_padding_side`
        if self.tokenizer.padding_side != self.cfg.default_padding_side:
            with warnings.catch_warnings():
                warnings.simplefilter("always")
                warnings.warn(
                    f"model.tokenizer.padding_side ({self.tokenizer.padding_side}) is not the same as \
                    model.cfg.default_padding_side ({self.cfg.default_padding_side}). \
                    model.tokenizer.padding_side = {self.cfg.default_padding_side} is set and \
                    will be used as the default padding side from now on. \
                    Please change model.cfg.default_padding_side if you want to use a different padding side."
                )
            self.tokenizer.padding_side = self.cfg.default_padding_side

        try:
            # This is important because self.tokenizer.padding_side is
            # the actual padding_side used by Transformers Tokenizer.
            self.tokenizer.padding_side = kwargs["padding_side"]
            
            # Execute the original function with the overridden padding_side
            outputs = function(self, *args, **kwargs)
            
            # Reset the padding_side of the tokenizer
            self.tokenizer.padding_side = self.cfg.default_padding_side
            return outputs

        except Exception as e:
            # If an error occurs, reset the padding_side of the tokenizer before propagating the exception
            self.tokenizer.padding_side = self.cfg.default_padding_side
            raise e

    return wrapper

# in HookedTransformer class

    @utils.locally_override_defaults_and_restore  # This is all we need!
    def to_str_tokens(
        self,
        input: Union[
            str,
            Int[torch.Tensor, "pos"],
            Int[torch.Tensor, "1 pos"],
            Int[np.ndarray, "pos"],
            Int[np.ndarray, "1 pos"],
            list,
        ],
        prepend_bos: Optional[bool] = None,
        padding_side: Optional[Literal["left", "right"]] = None,
    ) -> Union[List[str], List[List[str]]]:
```

soheeyang:
@neelnanda-io Could you please review this PR? This is the quick fix we've discussed in https://github.com/neelnanda-io/TransformerLens/pull/344#issuecomment-1637221178

neelnanda-io (approved):
LGTM!

PR #352, opened 2023-07-22
Add Llama-2 7B and 13B models

ArthurConmy:
# Description

This PR adds support for the smaller Llama-2 models in TransformerLens. It includes:

- Instructions for downloading and converting the Llama-2 weights
- An updated demo notebook showing usage of Llama-2-7B-chat model 
- Debugging help for convert_llama_weights_to_hf.py to handle Llama-2 weight conversion

## Type of change

Would close the first half of https://github.com/neelnanda-io/TransformerLens/issues/353

## Check list

## Future work

Future work could add Grouped Query Attention so that we can add the two 70B models


ArthurConmy:
Hmm, on a second run, the Llama-2-7B-chat model is not producing great completions. I'd advise not working from here until a sanity check

transformer_lens/components.py
```diff
@@ -673,10 +673,7 @@ def calculate_sin_cos_rotary(
 
         # A set of frequencies evenly spaced in log space
         freq = base ** (dim / (rotary_dim / 2))
-        if (
-            self.cfg.original_architecture == "GPTNeoXForCausalLM"
-            or self.cfg.original_architecture == "LLaMAForCausalLM"
```
ArthurConmy, line 678: `LLaMAForCausalLM` has been superseded

jbloomAus:
Thanks Arthur

Issue #349, opened 2023-07-22
[Proposal] Create a button at the top of demos to open them in colab


neelnanda-io:
### Proposal 

It's easy to find the demos from the TL github, but most people don't know how to open them in colab, which makes them much less user friendly. It would be nice to add a button at the top saying "open in colab" that links to the colab version.

Here's a video tutorial: https://www.youtube.com/watch?v=RoGZIbwzG5w

URLs are easily to just manually convert

Example of github link: https://github.com/neelnanda-io/TransformerLens/blob/main/demos/Main_Demo.ipynb

Example of colab link: https://colab.research.google.com/github/neelnanda-io/TransformerLens/blob/main/demos/Main_Demo.ipynb

PR #359, opened 2023-07-25
add colab buttons to demos

ckkissane:
# Description

Addresses #349 by adding colab link buttons to the top of each demo. Also removes Old_demo since it seems to be empty. I manually checked that all buttons seem to work properly. Feel free to double check here: https://github.com/ckkissane/TransformerLens/tree/add-colab-button/demos


neelnanda-io:
Thanks!

PR #365, opened 2023-07-30
Fix bug in rotary embedding for models other than llama and gpt-neo

### LINKED ISSUE #364: [Bug Report] Bug in rotary embedding for models other than llama and gpt-neo
**Describe the bug**
The following code creates a bug in the calculation of rotary embedding, which affects _the performance of the models that use rotary embedding other than llama and gpt-neo, e.g., gpt-j_.

https://github.com/neelnanda-io/TransformerLens/blob/bf5ed27cef2a566377151228acd5c7eb1b641195/transformer_lens/components.py#L697

```python3
        if (
            self.cfg.original_architecture
            in ["GPTNeoXForCausalLM", "LlamaForCausalLM"],  # the , at the end causes the bug
        ):
            n = x.size(-1) // 2
            rot_x[..., :n] = -x[..., n:]
            rot_x[..., n:] = x[..., :n]
        else:
            rot_x[..., ::2] = -x[..., 1::2]
            rot_x[..., 1::2] = x[..., ::2]
```

Let's say that `self.cfg.original_architecture = 'GPTJForCausalLM'`. Then, the desired behaviour is to execute the `else` branch because `self.cfg.original_architecture` is not in `["GPTNeoXForCausalLM", "LlamaForCausalLM"]`.
However, because of the "," at the end of L697, the if clause becomes `if (False,)`, which is interpreted as `if True` and executes the `if` branch.

I could find out this bug because I've observed _a significant drop in the in-context learning performance of GPT-J-6B_.

**Code example**
## Bug
```python3
>>> original_architecture = "GPTJForCausalLM"
>>> if (
...     original_architecture
...     in ["GPTNeoXForCausalLM", "LlamaForCausalLM"],  # the , at the end causes the bug
... ):
...     print('GPTNeoX or Llama')
... else:
...     print('others')
... 
GPTNeoX or Llama
```

## Bug Fixed
```python3
>>> original_architecture = "GPTJForCausalLM"
>>> if (
...     original_architecture
...     in ["GPTNeoXForCausalLM", "LlamaForCausalLM"]
... ):
...     print('GPTNeoX or Llama')
... else:
...     print('others')
... 
others
```

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

soheeyang:
# Description

This PR fixes the bug in the calculation of rotary embedding, which affects the performance of the models that use rotary embedding other than llama and gpt-neo, e.g., gpt-j.

https://github.com/neelnanda-io/TransformerLens/blob/bf5ed27cef2a566377151228acd5c7eb1b641195/transformer_lens/components.py#L697

```
        if (
            self.cfg.original_architecture
            in ["GPTNeoXForCausalLM", "LlamaForCausalLM"],  # the , at the end causes the bug
        ):
            n = x.size(-1) // 2
            rot_x[..., :n] = -x[..., n:]
            rot_x[..., n:] = x[..., :n]
        else:
            rot_x[..., ::2] = -x[..., 1::2]
            rot_x[..., 1::2] = x[..., ::2]
```

Let's say that `self.cfg.original_architecture = 'GPTJForCausalLM'`. Then, the desired behaviour is to execute the `else` branch because `self.cfg.original_architecture` is not in `["GPTNeoXForCausalLM", "LlamaForCausalLM"]`.
However, because of the "," at the end of L697, the if clause becomes `if (False,)`, which is interpreted as `if True` and executes the `if` branch.

Fixes #364 

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/components.py
+++ b/transformer_lens/components.py
@@ -692,10 +692,7 @@ class Attention(nn.Module):
         GPT-NeoX and GPT-J do rotary subtly differently, see calculate_sin_cos_rotary for details.
         """
         rot_x = x.clone()
-        if (
-            self.cfg.original_architecture
-            in ["GPTNeoXForCausalLM", "LlamaForCausalLM"],
-        ):
+        if self.cfg.original_architecture in ["GPTNeoXForCausalLM", "LlamaForCausalLM"]:
             n = x.size(-1) // 2
             rot_x[..., :n] = -x[..., n:]
             rot_x[..., n:] = x[..., :n]


ArthurConmy:
Yikes, totally for my bad for introducing this when adding Llama 2. Very subtle bug, thanks for documenting it!

It's pretty hard for us to catch bugs with GPT-J's implementation since the only `GPTJForCausalLM` is the 6B model which is way too chunky to run in tests.

By default, I'll implement a test [like this](https://github.com/neelnanda-io/TransformerLens/blame/main/tests/acceptance/test_hooked_transformer.py#L67) for GPT-J, but with the "pytest mark as skipped" fixture so someone can quickly test that something hasn't gone wrong in future. What do you think?

soheeyang:
@ArthurConmy Right, it was indeed very subtle bug! I agree that it's hard to catch bugs for GPT-J because the loading time takes too much time. Adding a skippable test is a great idea!

PR #375, opened 2023-08-26
Fix to include ln_final.w in RMSNorm hook

clarenceluo78:
# Description

Before fix, `model_cache["ln_final.hook_normalized"]` will only return the RMS normalized hidden states without multiplying the `final_ln` weight. This might contractict with the design of this hook. I followed the implementation in LayerNorm hook to include `self.w` into the `hook_normalized` hookpoint.

I have currently found no issues related to this. Hope this helps :)

## Before Fix
```python
def forward(
        self, x: Float[torch.Tensor, "batch pos length"]
    ) -> Float[torch.Tensor, "batch pos length"]:
        if self.cfg.dtype not in [torch.float32, torch.float64]:
            x = x.to(torch.float32)

        scale: Float[torch.Tensor, "batch pos 1"] = self.hook_scale(
            (x.pow(2).mean(-1, keepdim=True) + self.eps).sqrt()
        )
        x = self.hook_normalized(x / scale).to(self.cfg.dtype)  # [batch, pos, length]
        return x * self.w
```

## After Fix
```python
def forward(
        self, x: Float[torch.Tensor, "batch pos length"]
    ) -> Float[torch.Tensor, "batch pos length"]:
        if self.cfg.dtype not in [torch.float32, torch.float64]:
            x = x.to(torch.float32)

        scale: Float[torch.Tensor, "batch pos 1"] = self.hook_scale(
            (x.pow(2).mean(-1, keepdim=True) + self.eps).sqrt()
        )
        x = x / scale  # [batch, pos, length]
        return self.hook_normalized(x * self.w).to(self.cfg.dtype)
```

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/components.py
+++ b/transformer_lens/components.py
@@ -336,8 +336,8 @@ class RMSNorm(nn.Module):
         scale: Float[torch.Tensor, "batch pos 1"] = self.hook_scale(
             (x.pow(2).mean(-1, keepdim=True) + self.eps).sqrt()
         )
-        x = self.hook_normalized(x / scale).to(self.cfg.dtype)  # [batch, pos, length]
-        return x * self.w
+        x = x / scale  # [batch, pos, length]
+        return self.hook_normalized(x * self.w).to(self.cfg.dtype)
 
 
 # Attention


neelnanda-io:
I'm to pushback, but currently think that hook_normalized is working as
intended, because it's invariant between folding layer norm

alan-cooney:
@clarenceluo78 I think the key point here is that a common thing done in TransformerLens is folding the layer norm weights into the next linear layer. See https://github.com/neelnanda-io/TransformerLens/blob/main/further_comments.md#what-is-layernorm-folding-fold_ln for details. 

Therefore excluding it from `hook_normalized` seems to make sense as we want to be able to extract something that is invariant to whether or not `fold_ln` is enabled.

alan-cooney:
@clarenceluo78 I think the key point here is that a common thing done in TransformerLens is folding the layer norm weights into the next linear layer. See https://github.com/neelnanda-io/TransformerLens/blob/main/further_comments.md#what-is-layernorm-folding-fold_ln for details.

Therefore excluding it from hook_normalized seems to make sense as we want to be able to extract something that is invariant to whether or not fold_ln is enabled.

Let me know if you disagree however.

ArthurConmy:
I think that this should not be merged, but we should reopen and merge [this PR](https://github.com/neelnanda-io/TransformerLens/pull/315) that fixes the inconsistency that @clarenceluo78 noticed. Unless @neelnanda-io can explain why the behavior of the the hook_normalized caching in LayerNorm is correct?

bryce13950:
I am going to go ahead and close this one. If anything further needs to be done with this, let's discuss it on slack.

PR #376, opened 2023-09-07
Fix llama tokenizer issue with tokenizer-based control of bos prepending

soheeyang:
# Description

This PR changes the mechanism of `prepend_bos` to dynamically switch between two tokenizers (one that prepends bos and one that does not prepend bos) following the philosophy of huggingface tokenizers that "once a tokenizer is saved, we should not change it again, but rather re-convert it" described in https://github.com/huggingface/transformers/issues/25886#issuecomment-1701013725.

Previously, if we wanted to prepend a bos token to the input, we manually added it to the input, e.g., `self.tokenizer.bos_token + input`. However, this resulted in a bug for LLaMA tokenizers because they behave differently when the bos token is manually prepended and it is automatically prepended by the tokenizer. (https://github.com/neelnanda-io/TransformerLens/issues/373#issuecomment-1701017935)

While the interface for the user remains the same, this PR changes the internal mechanism of prepending bos to holding two tokenizers in `self.tokenizer_dict`, one that prepends bos to the input and one that does not prepend bos. The user can just provide the model one tokenizer; the code detects whether the given tokenizer uses bos by default and then creates the counterpart. Which tokenizer to use is controlled by the value set in `model.cfg.default_prepend_bos`. If `model.cfg.default_prepend_bos=True`, `self.tokenizer` forwards to `self.tokenizer_dict[True]` which prepends pos. Local overriding by calling string preprocessing methods like `to_tokens(text, prepend_bos=False)` is also supported. The code also handles the local/global changes in `padding_side` so that `padding_side` of the two tokenizers remains synced even if the user changes it.

One caveat in the current state of PR is that I monkey-patched the tokenizer when the tokenizer does not originally use a bos token and we need to add it manually. I did this in order to use the same interface for the two tokenizers, but I am aware that this approach is potentially dangerous. I just couldn't think of a better way by myself. I am open to better solutions/suggestions.

I checked that things work as expected with `test_tokenizer_dict.py`. I didn't add models like llama-7b or gpt-j-6b to the tests because the models are too large, but I've manually run the test code using those models as well and checked that there's no issue.

Fixes https://github.com/neelnanda-io/TransformerLens/issues/373#issuecomment-1701017935

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_prepend_bos.py
+++ b/tests/unit/test_prepend_bos.py
@@ -1,5 +1,4 @@
 import pytest
-import torch
 from transformers import AutoTokenizer
 
 from transformer_lens import HookedTransformer
--- a/tests/unit/test_prepend_bos.py
+++ b/tests/unit/test_prepend_bos.py
@@ -1,4 +1,6 @@
 import pytest
+import torch
+from transformers import AutoTokenizer
 
 from transformer_lens import HookedTransformer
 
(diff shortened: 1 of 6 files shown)


soheeyang:
I just realised that we can maybe do it by holding only one tokenizer that prepends bos by default (automatically or manually) and cut the first token off when prepend_bos=False. I think this PR might be overengineering haha. I'll test if that approach works, and if it does, I'll close this one and make that into a new PR.

PR #379, opened 2023-09-09
Fix llama tokenization issue by using tokenizer initialized with add_bos_token=True

soheeyang:
# Description

This PR ensures that the `tokenizer` used by the model is initialized with `add_bos_token=True`.

Previously, the tokenizer that  does not prepend bos was used as default. When we wanted to prepend a bos token to the input, we manually added it to the input, e.g., `tokenizer.bos_token + input`. However, this resulted in a bug for LLaMA tokenizers because they behave differently when the bos token is manually prepended and when it is automatically prepended by the tokenizer (https://github.com/neelnanda-io/TransformerLens/issues/373#issuecomment-1701017935).

It is necessary to ensure that the tokenizer is initialized with `from_pretrained(add_bos_token=True)` because `add_bos_token` cannot be dynamically controlled after initialization (https://github.com/huggingface/transformers/issues/25886).

Because some tokenizers doesn't automatically prepend bos token even when they are initialized with `add_bos_token=True`, `cfg.tokenizer_prepends_bos` is introduced to check if the tokenizer _does_ prepend bos token or not when initialized with `add_bos_token=True`.

The test for tokenizers are in `test_only_tokenizer.py`. (I also manually tested llama tokenizer using the same test.)

Fixes https://github.com/neelnanda-io/TransformerLens/issues/373

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_only_tokenizer.py
+++ b/tests/unit/test_only_tokenizer.py
@@ -1,0 +1,269 @@
+import logging
+from typing import Dict
+
+import pytest
+from transformers import AutoTokenizer
+
+import transformer_lens.loading_from_pretrained as loading
+from transformer_lens import HookedTransformer, HookedTransformerConfig
+
+
+class TokenizerOnlyHookedTransformer(HookedTransformer):
+    def __init__(
+        self,
+        cfg,
+        tokenizer=None,
+        move_to_device=True,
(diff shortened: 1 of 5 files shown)


jbloomAus:
Thanks! Sorry for the delayed merge.

PR #382, opened 2023-09-12
Add start_at_layer parameter to HookedTransformer

UFO-101:
# Description

  - Add `start_at_layer` to HookedTransformer forward function.
  - Pull out the embedding step into a seperate function to enable this.
 - Add parameters `tokens`, `shortformer_pos_embed` and `left_attention_mask` to forward function as these outputs of the embed step are sometimes required.
 - Add tests for `start_at_layer`.

# Motivation

Sometimes I want to run patching experiments where I know all the patches are in the last few layers. In this case, I can cache the result of the earlier layers for improved performance.

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_start_at_layer.py
+++ b/tests/unit/test_start_at_layer.py
@@ -1,0 +1,204 @@
+"""
+Tests for the start_at_layer parameter in HookedTransformer
+"""
+
+from typing import Any, Dict
+
+import pytest
+import torch
+
+from transformer_lens import HookedTransformer, HookedTransformerConfig
+
+
+@pytest.fixture
+def setup_data() -> Dict[str, Any]:
+    cfg = HookedTransformerConfig(
+        n_layers=3,
+        d_mlp=8,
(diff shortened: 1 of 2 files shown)


UFO-101:
(I'm assuming documentation gets generated automatically from docstrings?)

UFO-101:
>  Can you make the case for this change being worth it? 

My current use case is for implementing ACDC. We patch every edge in reverse topological order. My implementation gets a >2x speed up by only computing the layers after the most recently patched destination node.

Excluding test code, almost all the new code is from pulling out the embedding step into a new function. IMO this is probably a good thing to do anyway because the existing forward function is very long and it's generally useful to modularise code and compose functions.

neelnanda-io (approved):
LGTM

neelnanda-io:
Good arguments, I'm convinced. 2x is a great speed-up!

neelnanda-io:
Thanks for the PR!

PR #383, opened 2023-09-18
Improve Key-Value Caching

UFO-101:
# Description

Commit 1:
- Support passing in multiple tokens when using `past_kv_cache`.
- Add tests for `past_kv_cache`.
- Add documentation for `past_kv_cache`.
- Fix type hints for some components that assume left_attention_mask has same number of tokens as input. This was previously unnoticed because there were no tests that covered `past_kv_cache`.

Commit 2:
- Support freezing key-value caches.

Commit 3:
- Integrate `past_left_attention_mask` into `HookedTransformerKeyValueCache` so that it doesn't need to managed manually. Remove from `HookedTransformer.forward()`.

## Motivation for allowing multiple tokens to run with key value cache

In ACDC we run the same prompt many times. Patching only affects token positions after the point where the clean and corrupt prompts differ. We want to run the first part of the prompt that is identical between clean and corrupt, freeze the cache, then pass in only the tokens after the point of divergence for our patched runs.

## Breaking change

Commit 3 removes `past_left_attention_mask` from `HookedTransformer.forward()` because the `left_attention_mask` of previous inputs is stored automatically by `HookedTransformerKeyValueCache`. I think this won't break many people's code as this argument was only added 2 weeks ago in #344. And the fix should be quite trivial as it only requires deleting this input. (See changes to tests in commit 3 for an example).

Overall I think the benefits are worth the cost as this makes caching easier to do and generally reduces complexity. I can't think of any case where someone would want to pass in a `past_left_attention_mask` that doesn't match `past_kv_cache`.

# Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_kv_cache.py
+++ b/tests/unit/test_kv_cache.py
@@ -80,3 +80,98 @@ def test_multiple_new_tokens():
     assert t.allclose(
         no_cache_logits[:, -new_tokens_len:], with_cache_logits, atol=1e-3
     )
+
+
+def test_freeze_cache():
+    past_left_attention_mask = utils.get_attention_mask(
+        model.tokenizer,
+        pre_prompt_tokens,
+        model.cfg.default_prepend_bos,
+    )
+
+    post_prompt_1 = " I'm headed to the church to play bingo."
+    new_tokens_1 = model.to_tokens(post_prompt_1, prepend_bos=False)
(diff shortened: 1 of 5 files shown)


UFO-101:
Sorry actually this isn't ready. I just realized there's an edge case when passing in a left padded input while using the cache.

UFO-101:
Superseded by #386

PR #386, opened 2023-09-26
Improve attention masking and key-value caching.

UFO-101:
# Description

- Support passing in multiple tokens when using `past_kv_cache`.
- Change `left_attention_mask` to `attention_mask`. Update `apply_causal_mask` and `apply_rotary` to support arbitrary masking (and do it faster). This means that caching 'just works' regardless of padding. So you can, for example, cache keys and values for a batch of tokens and then run a left padded batch of tokens, and the padding tokens in the middle of the sequence will be ignored.
- Add tests for `past_kv_cache`.
- Add `test_pos_embed_with_cache` in `test_left_padding`.
- Add documentation for `past_kv_cache`.
- Fix type hints for some components that assume `attention_mask` has same number of tokens as input. This was previously unnoticed because there were no tests that covered `past_kv_cache`.
- Support freezing key-value caches.
- Integrate `past_attention_mask` into HookedTransformerKeyValueCache so that it doesn't need to managed manually.
- Fix bug caused by previous PR #382. start_at_layer would cause enumerate to produce incorrect block indices in `HookedTransformer.forward()`.

## Motivation for arbitrary padding

1. This implementation is simpler and faster than the previous left-padding implementation.
2. It allows us to use caching while running batches with arbitrary padding. For example, I can cache a batch of prompts and then run another batch of left-padded prompts which makes it simpler to index the final token in each prompt. 

## Motivation for allowing multiple tokens to run with key value cache

In ACDC we run the same prompt many times. Patching only affects token positions after the point where the clean and corrupt prompts differ. We want to run the first part of the prompt that is identical between clean and corrupt, freeze the cache, then pass in only the tokens after the point of divergence for our patched runs.

## Breaking changes

Removes `past_left_attention_mask` from `HookedTransformer.forward()` because the `left_attention_mask` (now just `attention_mask`) of previous inputs is stored automatically by `HookedTransformerKeyValueCache`. I think this won't break many people's code as this argument was only added 4 weeks ago in #344. And the fix should be quite trivial as it only requires deleting this input.

Overall I think the benefits are worth the cost as this makes caching easier to do and generally reduces complexity. I can't think of any case where someone would want to pass in a `past_attention_mask` that doesn't match `past_kv_cache`.

Renames `left_attention_mask` to `attention_mask`. This was introduced 2 weeks ago in #382, so is probably also fine.

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_kv_cache.py
+++ b/tests/unit/test_kv_cache.py
@@ -1,0 +1,265 @@
+import pytest
+import torch as t
+
+from transformer_lens import HookedTransformer
+from transformer_lens.past_key_value_caching import HookedTransformerKeyValueCache
+
+
+# Pythia models seem to have some kind of numerical stability issue.
+# See: https://github.com/neelnanda-io/TransformerLens/issues/385
+@pytest.fixture(scope="session", params=[("gpt2-small", 1e-4), ("pythia-14m", 1e-2)])
+def model_and_atol(request):
+    return request.param
+
+
+@pytest.fixture(scope="session")
+def pretrained(model_and_atol):
+    name, atol = model_and_atol
+    model = HookedTransformer.from_pretrained(name, default_padding_side="left")
+    return model, atol
+
+
+def test_single_new_token(pretrained):
+    model, atol = pretrained
+    pre_prompt = "I went to Staten Island,"
+    pre_prompt_tokens = model.to_tokens(pre_prompt)
+    pre_prompt_tokens_len = pre_prompt_tokens.shape[-1]
+    single_token_post_prompt = " Sharon"
+    single_new_token = model.to_tokens(single_token_post_prompt, prepend_bos=False)
+    full_prompt_tokens = t.cat([pre_prompt_tokens, single_new_token], dim=-1)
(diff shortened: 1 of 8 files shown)


UFO-101:
This is ready for review 👍

jbloomAus:
@UFO-101 , rebased but formatter is annoying. Don't worry about this, I'll fix before I merge (should be tomorrow morning).

alan-cooney:
> @UFO-101 , rebased but formatter is annoying. Don't worry about this, I'll fix before I merge (should be tomorrow morning).

Just to check - is this due to the changes I made to VSCode settings? It doesn't do this on my end, but if it is let me know and I'll take a look/open an issue.

alan-cooney (changes requested):
@UFO-101 if you're happy to rebase onto the latest repo please do that and then give me a nudge, and I'll do the final review if @jbloomAus is busy.

UFO-101:
@alan-cooney Done rebase

jbloomAus:
I'm happy with this. Can't accept since @alan-cooney you have requested changes.

alan-cooney (approved):

PR #388, opened 2023-10-01
Added explicit dtype option

neelnanda-io:
# Description

A cosmetic change to add a dtype option on from_pretrained. Previously you could set torch_dtype and it would work, but this was not documented (and I didn't realise it until implementing a PR for mixed precision lol)

### DIFF
--- a/transformer_lens/HookedEncoder.py
+++ b/transformer_lens/HookedEncoder.py
@@ -201,6 +201,7 @@ class HookedEncoder(HookedRootModule):
         device: Optional[str] = None,
         tokenizer=None,
         move_to_device=True,
+        dtype=torch.float32,
         **from_pretrained_kwargs,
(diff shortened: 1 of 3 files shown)


neelnanda-io:
For backwards compatibility, setting torch_dtype overrides dtype

neelnanda-io:
Huh, I am failing some of the left padding tests. I can't replicate this locally, and can't see how anything I've changed affects that test? I'm confused and probably don't care enough to debug this right now, since the feature _is_ usable with the keyword torch_dtype, just badly documented

jbloomAus:
@neelnanda-io not your fault. Something fell through the cracks with automatic testing. Bryce and I almost solved it but then  didn't. I've pinged Connor (who's change caused it) but if he doesn't do in in the next day or so I'll either solve it or revert the changes.

PR #389, opened 2023-10-02
bug fix attention inf

jbloomAus:


ArthurConmy:
Shouldn't we be hooking the pattern after removing NaNs?

transformer_lens/components.py
```diff
@@ -612,6 +612,7 @@ def forward(
         pattern = self.hook_pattern(
             F.softmax(attn_scores, dim=-1)
         )  # [batch, head_index, query_pos, key_pos]
+        pattern = torch.where(torch.isnan(pattern), torch.zeros_like(pattern), pattern)
         pattern = pattern.to(self.cfg.dtype)
```
ArthurConmy, line 616:
```python
        raw_pattern = F.softmax(attn_scores, dim=-1) # [batch, head_index, query_pos, key_pos]
        clean_pattern = torch.where(torch.isnan(raw_pattern), torch.zeros_like(raw_pattern), raw_pattern)
        pattern = self.hook_pattern(
            pattern.to(self.cfg.dtype)
        )  # [batch, head_index, query_pos, key_pos]
```

jbloomAus: good catch, I'll amend this.

PR #391, opened 2023-10-04
Feature auto rebase

bryce13950:
# Description

This adds a GitHub action, which will rebase all pull requests when anything is pushed to main to avoid issues where CI actions have been run on an old base while reviewing pull requests.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/.github/workflows/rebase-pull-requests.yml
+++ b/.github/workflows/rebase-pull-requests.yml
@@ -1,0 +1,9 @@
+name: Rebase Pull Requests
+on:
+  push:
+    branches: [main]
+jobs:
+  rebase:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: linhbn123/rebase-pull-requests@v1.0.1
--- a/.github/workflows/rebase-pull-requests.yml
+++ b/.github/workflows/rebase-pull-requests.yml
@@ -1,4 +1,4 @@
-name: Rebase Pull Requests
+name: Rebase
 on:
   push:
     branches: [main]
@@ -6,4 +6,8 @@ jobs:
   rebase:
     runs-on: ubuntu-latest
     steps:
-      - uses: linhbn123/rebase-pull-requests@v1.0.1
+      - uses: peter-evans/rebase@v2
+        with:
+          base: main
+          exclude-labels: |
+            no-rebase


jbloomAus:
Thanks Bryce.

alan-cooney (changes requested):
My understanding is this also won't trigger a re-running of the checks, as it defaults to using `GITHUB_TOKEN` which doesn't trigger further workflows (to prevent circular workflow calls). To over-ride this we'd need Neel to set a PAT as an env variable and then this could be set using the token property.

In general I'm not sure what is best practice here (i.e. to auto-rebase or for reviewers to just click the rebase button where necessary).  But unless we're fixing a problem that has popped up in real life, I'd be tempted to not worry about this for now.

bryce13950:
I tested it on my fork, and it did trigger the workflow there without issue. The purpose of this is to fix situations where a workflow succeeded previously, but would now fail without the reviewer realizing, as what happened last week when the main branch ended up in a state of failure.

jbloomAus:
> I tested it on my fork, and it did trigger the workflow there without issue. The purpose of this is to fix situations where a workflow succeeded previously, but would now fail without the reviewer realizing, as what happened last week when the main branch ended up in a state of failure.

@alan-cooney we had this issue in real life where the tests were run on main and on PR but it was the rebased PR that fails.  I don't think this should create circular checks but if we try this out and have that issue, we could roll it back easily.

alan-cooney (approved):
Ah cool in which case I think this makes sense from my perspective (still subject to @jbloomAus approving of course).

As a heads up it may still not work on PRs on the main repo. I'm not exactly sure how GitHub's circular checks work (and I agree it's not actually circular so its a false negative). However I've seen it cause confusion a few times with other auto-commit actions. Details here - https://github.com/marketplace/actions/add-commit#the-commit-from-the-action-is-not-triggering-ci as to the workaround if it still doesn't work. In any case we can just see what happens so no harm.

.github/workflows/rebase-pull-requests.yml
```diff
@@ -0,0 +1,9 @@
+name: Rebase Pull Requests
+on:
+  push:
+    branches: [main]
+jobs:
+  rebase:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: linhbn123/rebase-pull-requests@v1.0.1
```
alan-cooney, line 9: Looked into this in more detail and the GitHub action you're using seems unmaintained.

How about we switch to this, and also add a tag that prevents a PR from being rebased (so we can stop very old ones from being rebased).

https://github.com/peter-evans/rebase

```suggestion
name: Rebase
on:
  push:
    branches: [main]
jobs:
  rebase:
    runs-on: ubuntu-latest
    steps:
      - uses: peter-evans/rebase@v2
        with:
          base: main
          exclude-labels: |
            no-rebase

```

alan-cooney (dismissed):
Suggested switching to a different action which should also resolve Joseph's concerns

bryce13950:
That seems to work fine. I just tested with and without the label on my fork, and it all seems to be working as expected.

alan-cooney:
Hey Bryce I had a bit more of a think about this. I think the key problem we still have is that because git rebase is a destructive action (it re-writes history), it has the potential to get quite annoying for contributors as it will keep re-writing their history whilst a PR is in draft (and so they'll have to keep reverting, stashing, pulling down changes and then popping the stash). On top of this for reviewers it'll also prevent reviewing just changes since the last commit that was previously reviewed (as with all rewritten history GitHub can't provide this functionality as it doesn't know what changed).

I'm tempted to say the problem is a sign of a bigger issue in that we have PRs dangling for a long time, and I suspect with your ability to work on this a bit more we can just tackle that better by getting PRs reviewed faster and into main. We can also just git revert if the occasional PR does make it through that breaks the tests.

Keen to get your thoughts on this - what do you think?

bryce13950:
I think there is a bit of confusion as to what this is doing. When you rebase while merging, what you are describing is correct. However, on a pull request, it is not affecting the branch at all. All this does is change the pull request to say that we are now calculating merge as if the pull request was going to be merged into the most recent commit. When a pull request is opened, GitHub calculates the changes to the most recent commit, and runs any related actions as if the opened pull request was merged into the branch the pull request is targeting, providing that there are no merge conflicts. 
This action we are talking about implementing does not write to any branch, all it does is rerun the actions as if the pull request had just been opened. If you open a pull request, and your branch is 20 commits behind the branch you are merging into, the branch you are merging is not updated to the main branch. This is not making any changes to the history or structure of the repository, it only simulates what would happen if the requested change was made.

alan-cooney:
Makes sense - thanks for the clarification.

alan-cooney (approved):

alan-cooney:
Reverted as the run is failing - details at https://github.com/neelnanda-io/TransformerLens/actions/runs/6540632494/job/17760882677 

Relevant logs here:

```
2023-10-17T00:00:36.4374951Z ##[group]Pushing changes to head ref 'small-improvements'
2023-10-17T00:00:36.4379365Z [command]/usr/bin/git push --force-with-lease 5676b6f6-dc29-4eb1-a403-969ad11238a7 HEAD:small-improvements
2023-10-17T00:00:37.3835809Z To https://github.com/UFO-101/TransformerLens
2023-10-17T00:00:37.3838690Z  ! [remote rejected] HEAD -> small-improvements (refusing to allow a GitHub App to create or update workflow `.github/workflows/rebase-pull-requests.yml` without `workflows` permission)
2023-10-17T00:00:37.3841563Z error: failed to push some refs to 'https://github.com/UFO-101/TransformerLens'
2023-10-17T00:00:37.3890112Z ##[error]The process '/usr/bin/git' failed with exit code 1
2023-10-17T00:00:37.4363694Z Cleaning up orphan processes
```

bryce13950:
I'll do some more testing with this using forks off of my own fork, and get back to this. I found a few comments in regards to what happened here that seem to indicate a combination of different actions resolve the issue.

alan-cooney:
Okay cool - just to flag it did seem to also rebase the underlying branches so it was destructive.

bryce13950:
I did some digging into the actual code of the action you suggested, and it was running the rebase as you had thought. I assumed that it was doing something slightly differently, but that was incorrect on my part. I did do some digging, and I found at least a combination of actions that can achieve the desired action as I described.

PR #395, opened 2023-10-10
Move towards consistent commenting/docstring style

alan-cooney:
# Description

Convert ActivationCache and HookedTransformer docstrings/comments to the consistent Google Python Style Guide.

Note the only changes made are:
- Make sure docstrings have titles (and they end in a full stop).
- Put all args under "Args:" (some were missing this section title so they didn't show on the Sphinx docs)
- Some very minor spelling mistakes/moving titles to the imperative tense.
- Use rewrap to make all the comments the same length (100, as per Pylint config in the project)

There are still other things missing (e.g. sometimes args are not documented) but it's a good start.

## Type of change

### DIFF
--- a/transformer_lens/ActivationCache.py
+++ b/transformer_lens/ActivationCache.py
@@ -553,7 +553,7 @@ class ActivationCache:
         pos_slice: Union[Slice, SliceInput] = None,
     ) -> Float[torch.Tensor, "*batch_and_pos_dims num_neurons d_model"]:
         """Get Neuron Results.
-        
+
         Get the results of for neurons in a specific layer (i.e, how much each neuron contributes to
         the residual stream). Does it for the subset of neurons specified by neuron_slice, defaults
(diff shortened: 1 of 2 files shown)


alan-cooney:
Usual checklist didn't seem to make sense (as it's only changing comments/docstrings), but I've manually gone through and also made sure all wording changes are minor (so it's just formatting).

jbloomAus:
Seems good! Surely there's a linter or something that might help with this stuff? We could make it default in the dev container?

alan-cooney:
Yeah pylint will flag stuff like this - but I don't think we can make it mandatory yet as it'll cause headaches for ongoing PRs (as quite a lot of the codebase has squiggly lines :) ). At least re-wrap did most of the hard work for now.

I've just re-checked and will merge now, as leaving a big PR around like this will create merge conflicts otherwise.

PR #406, opened 2023-10-13
Remove unused root directory files

alan-cooney:
These files aren't used, with the exception of setup.py .

Setup.py appears to be being used sometimes, which has created hard-to-reproduce bugs. I've moved the dependencies it had into pyproject and removed it as well.


alan-cooney:
Please squash when merging! Lots of trys to get Python 3.7 working.

PR #409, opened 2023-10-13
Replace `NeelNanda/gpt-neox-tokenizer-digits` with new tokenizer

ArthurConmy:
# Description

Breaking changes in HF tokenizers described [here](https://github.com/neelnanda-io/TransformerLens/issues/404) mean that the `NeelNanda/gpt-neox-tokenizer-digits` tokenizer doesn't work as intended. 

This PR replaces this tokenizer with my fork that isn't affected by the HF transformers changes, described here: https://huggingface.co/ArthurConmy/alternative-neel-tokenizer

Fixes: https://github.com/neelnanda-io/TransformerLens/issues/404

I haven't tested that every possible tokenization is identical, but did add a test of a large amount of python code so hopefully that's sufficient.

## Type of change

# Checklist:


tests/unit/test_tokenization_methods.py
```diff
@@ -163,3 +163,113 @@ def test_get_token_position_int_1_pos():
...
+for placeholder in placeholder_list:
+    # This is placeholder code to test tokenization
+    if callable(placeholder):
+        # This is placeholder code to test tokenization
+        placeholder(10, 5)  # assuming this input is valid for all placeholder functions
+
+# This is placeholder code to test tokenization"""
+
+
+def test_correct_tokenization():
```
ArthurConmy, line 205: This test fails on main currently!

tests/unit/test_tokenization_methods.py
```diff
@@ -163,3 +163,113 @@ def test_get_token_position_int_1_pos():
...
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2,     2,     2,     2,     2,     2,     2,
+             2,     2,     2,     2]]).tolist()
+    # fmt: on
```
alan-cooney, line 270: I'd suggest using https://pypi.org/project/snapshottest/ to avoid having a huge tensor in the test file (the snapshot can be added manually to make sure it's correct v. 1.6.1). 

Also in this case you could just test the entire vocab space (as file length is no longer a problem)?

alan-cooney (commented):
Added a small suggestion

alan-cooney (changes requested):
Marking as request changes based on the above

tests/unit/test_tokenization_methods.py
ArthurConmy: Done. However 

1. I don't know how to setup manually snapshotting 1.6.1
2. I had to `pip install snapshottest`. Should this library be in our `pyproject.toml`?

alan-cooney: Yep `poetry add --group dev snapshottest`

If you run it once it should create a snapshot directory in the tests directory with the snapshot in it. It'll just put the tensor in there so you can overwrite it if you want to manually change it

ArthurConmy:
@neelnanda-io I've fixed this bug (HF tokenizers not backwards compatible...) and [have asked whether this will be changed](https://github.com/huggingface/tokenizers/issues/1358#issuecomment-1761697696), but the fix is a big ugly (replacing your tokenizer with a fork). You could also merge this PR to your tokenizer, or propose another solution if you don't like this

tests/unit/test_tokenization_methods.py
ArthurConmy: Oh @alan-cooney you want me to commit the file with the tensor?

ArthurConmy:
Resolved with https://huggingface.co/NeelNanda/gpt-neox-tokenizer-digits/discussions/1 instead, probably we should add tests but I am occupied at the moment

PR #417, opened 2023-10-15
Add recursive DLA

alan-cooney:


alan-cooney:
To note - I'm investigating a simpler approach than this now (that scales better)

PR #418, opened 2023-10-16
Update docs to show prepend_bos and padding_side are optional

UFO-101:
# Description

 - Improve documentation for prepend_bos and padding_side in several functions in HookedTransformer and utils. It is currently unclear that they are just intended to be used as overrides of the default values set in the initial config.
 - Improve the syntax of some type hints.

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -239,6 +239,20 @@ class HookedTransformer(HookedRootModule):
         Optional[Float[torch.Tensor, "batch pos d_model"]],  # shortformer_pos_embed
         Optional[torch.Tensor],  # attention_mask [batch pos]
     ]:
+        """Convert input to first residual stream.
+
+        Args:
+            input (Union[str, List[str], Int[torch.Tensor, "batch pos"]]): The input to the model.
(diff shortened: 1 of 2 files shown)


alan-cooney:
I think a better approach may be to just make this clear in the docstring, so that we don't have to release a breaking change. As long as the title is e.g. "Override Default for Prepend BOS" it's fairly clear.

The other changes seem good but mind splitting into separate PRs so we can keep the commit history clean?

alan-cooney (changes requested):
I think a better approach may be to just make this clear in the docstring, so that we don't have to release a breaking change. As long as the title is e.g. "Override Default for Prepend BOS" it's fairly clear.

The other changes seem good but mind splitting into separate PRs so we can keep the commit history clean?

UFO-101:
@alan-cooney Undid parameter renaming and added / updated docstrings instead.

Moved other changes into new PRs #420 and #421

neelnanda-io:
+1 to not rename the parameters, and just update the docs

alan-cooney (approved):
We're going to add some more standards for docstrings shortly, but approving now and will make some tweaks once those are added.

PR #420, opened 2023-10-17
Return the current residual when stop_at_layer is not None.

UFO-101:
# Description

Return the current residual when stop_at_layer is not None.

## Type of change

# Checklist:


UFO-101:
Writing tests doesn't feel necessary here 😬

UFO-101:
Strictly speaking this could be breaking change, but I doubt it will be in practice.

alan-cooney (changes requested):
For the tests, can we instead assert that the output is the residual stream?

PR #421, opened 2023-10-17
Remove unnecessary uses of prepend_bos and padding_side

UFO-101:
# Description

Remove unnecessary uses of prepend_bos and padding_side arguments in HookedTransformer.

# Checklist:

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -504,12 +504,7 @@ class HookedTransformer(HookedRootModule):
                     tokens,
                     shortformer_pos_embed,
                     attention_mask,
-                ) = self.input_to_embed(
-                    input,
-                    prepend_bos=prepend_bos,
-                    padding_side=padding_side,
-                    past_kv_cache=past_kv_cache,
-                )
+                ) = self.input_to_embed(input, past_kv_cache=past_kv_cache)
             else:
                 assert type(input) == torch.Tensor
                 residual = input
@@ -828,9 +823,7 @@ class HookedTransformer(HookedRootModule):
                     )
                 )  # type: ignore
             elif isinstance(input, str):
-                tokens = self.to_tokens(
-                    input, prepend_bos=prepend_bos, padding_side=padding_side
-                )[0]
+                tokens = self.to_tokens(input)[0]
             elif isinstance(input, torch.Tensor):
                 tokens = input
                 tokens = tokens.squeeze()  # Get rid of a trivial batch dimension
@@ -1857,9 +1850,7 @@ class HookedTransformer(HookedRootModule):
                 assert (
                     self.tokenizer is not None
                 ), "Must provide a tokenizer if passing a string to the model"
-                tokens = self.to_tokens(
-                    input, prepend_bos=prepend_bos, padding_side=padding_side
-                )
+                tokens = self.to_tokens(input)
             else:
                 tokens = input
 
@@ -1924,16 +1915,12 @@ class HookedTransformer(HookedRootModule):
                         logits = self.forward(
                             tokens[:, -1:],
                             return_type="logits",
-                            prepend_bos=prepend_bos,
-                            padding_side=padding_side,
                             past_kv_cache=past_kv_cache,
                         )
                     else:
                         logits = self.forward(
                             tokens,
                             return_type="logits",
-                            prepend_bos=prepend_bos,
-                            padding_side=padding_side,
                             past_kv_cache=past_kv_cache,
                         )
                 else:
@@ -1942,8 +1929,6 @@ class HookedTransformer(HookedRootModule):
                     logits = self.forward(
                         tokens,
                         return_type="logits",
-                        prepend_bos=prepend_bos,
-                        padding_side=padding_side,
                     )
                 final_logits = logits[:, -1, :]
 


alan-cooney:
Please can you provide details on why you think they're not needed?

UFO-101:
> Please can you provide details on why you think they're not needed?

They are not required because the implementation of `generate` is wrapped in the context manager:

```
        with utils.LocallyOverridenDefaults(
            self, prepend_bos=prepend_bos, padding_side=padding_side
        ):
```
and confusingly `prepend_bos` and `padding_side` are only intended to be used as for overriding the default.

alan-cooney (changes requested):
This is a fun one. Technically this is working as desired already (as far as I can tell).

Specifically, if `forward()` is called with an override (e.g. `prepend_bos=something` this is set by the `utils.LocallyOverridenDefaults` context manager in the forward method. However, if the value is not passed into the `self.to_tokens()` call then the default of `None` will apply. The context manager in `to_tokens` will then in turn receive `prepend_bos=None`, which instructs it to apply the override of "None". Since the `default_value_to_restore` has not been set yet in this inner context manager, it will just set the model cfg value to None (rather than left or right), which will break it. So to sum up, 
 you actually do need to pass the variable through with this nested context manager setup.

I think the better solution here is to remove the context manager. It looks like it was a very smart solution, but unfortunately very smart solutions are almost always a bad idea in software engineering (hard to read & likely to cause bugs).

It would be good to understand why it was added, but I suspect that the simplest solution would have been to:

1. Set padding_side as just a method arg, and not a tokenizer config default setting (it's not clear that we need the default setting at all).
2. For prepend_bos just pass the override all the way through from e.g. forward -> to_tokens .

What do you think? @soheeyang you may also have views?

neelnanda-io:
I agree that the context manager is an overly complex solution, and finding a simpler one would be better. I do think that having the ability to disable prepending bos by default is valuable? (I care less about padding side) I'd feel sad if we remove this functionality without having a way to do that.

transformer_lens/HookedTransformer.py
```diff
@@ -504,12 +504,7 @@ def forward(
                     tokens,
                     shortformer_pos_embed,
                     attention_mask,
-                ) = self.input_to_embed(
-                    input,
-                    prepend_bos=prepend_bos,
-                    padding_side=padding_side,
-                    past_kv_cache=past_kv_cache,
-                )
+                ) = self.input_to_embed(input, past_kv_cache=past_kv_cache)
```
soheeyang, line 507: `prepend_bos=prepend_bos` and `padding_side=padding_side` must not be removed from this method call to enable overriding the values.

soheeyang: The body of `input_to_embed` must be wrapped with
```python3
        with utils.LocallyOverridenDefaults(
            self, prepend_bos=prepend_bos, padding_side=padding_side
        ):
```
in order to enable the default value overriding.

I've stated this in the comment below:
```
class LocallyOverridenDefaults:
    """
    Context manager that allows temporary overriding of default values within a model.
    Once the context is exited, the default values are restored.

    WARNING: This context manager must be used for any function/method that directly accesses
    default values which may be overridden by the user using the function/method's arguments,
    e.g., `model.cfg.default_prepend_bos` and `model.tokenizer.padding_side` which can be
    overriden by `prepend_bos` and `padding_side` arguments, respectively, in the `to_tokens`.
    """
```

soheeyang (changes requested):
I'm sorry that my feature brought a confusion.

In short, removing the argument passing of `prepend_bos=prepend_bos` and `padding_side=padding_side` is against my original intention.

According to what I intended, I wouldn't want to make the changes in this PR, and rather make the following two changes to the current code of `input_to_embed`:
1. wrap the body of the method with `with utils.LocallyOverridenDefaults(self, prepend_bos=prepend_bos, padding_side=padding_side)`. Otherwise, https://github.com/neelnanda-io/TransformerLens/blob/bf10751f2e293d05d5116d36539ad34d91a025ae/transformer_lens/HookedTransformer.py#L263C62-L263C74 and https://github.com/neelnanda-io/TransformerLens/blob/bf10751f2e293d05d5116d36539ad34d91a025ae/transformer_lens/HookedTransformer.py#L274 will cause bugs because the values are not controlled properly.
2. replace https://github.com/neelnanda-io/TransformerLens/blob/bf10751f2e293d05d5116d36539ad34d91a025ae/transformer_lens/HookedTransformer.py#L280C46-L284C46 simply with `attention_mask = utils.get_attention_mask(self.tokenizer, tokens, cfg.default_prepend_bos)`

The relevant warning was already stated in the comment of `LocallyOverridenDefaults`, but I understand that it would have been easy to miss this warning.

class LocallyOverridenDefaults:
    """
    Context manager that allows temporary overriding of default values within a model.
    Once the context is exited, the default values are restored.

    WARNING: This context manager must be used for any function/method that directly accesses
    default values which may be overridden by the user using the function/method's arguments,
    e.g., `model.cfg.default_prepend_bos` and `model.tokenizer.padding_side` which can be
    overriden by `prepend_bos` and `padding_side` arguments, respectively, in the `to_tokens`.
    """

soheeyang:
@alan-cooney I agree, this is an example of a smart solution turning out to be a bad idea in an open-source code 😂 I agree that the context manager is an overly complex solution. However, like @neelnanda-io, I would be sad to lose the feature of changing the default value of `prepend_bos` (and also `padding_side` for me personally) as I am myself using these features all the time.

If we don't want to use the context manager, the alternative will be changing the code to use `utils.override_or_use_default_value(cfg.default_prepend_bos, prepend_bos)` for every use of `cfg.default_prepend_bos` and `utils.override_or_use_default_value(tokenizer.padding_side, padding_side)` for every use of `tokenizer.padding_side`. However, bugs can still occur when other contributors miss calling this function.

How do you both think about it?

soheeyang:
(On a related note, I can serve as a consistent reviewer for the PRs related to the usage/control of `prepend_bos` or `padding_side` to prevent a similar issue.)

alan-cooney:
> @alan-cooney I agree, this is an example of a smart solution turning out to be a bad idea in an open-source code 😂 I agree that the context manager is an overly complex solution. However, like @neelnanda-io, I would be sad to lose the feature of changing the default value of `prepend_bos` (and also `padding_side` for me personally) as I am myself using these features all the time.
> 
> If we don't want to use the context manager, the alternative will be changing the code to use `prepend_bos = utils.override_or_use_default_value(cfg.default_prepend_bos, prepend_bos)` for every use of `cfg.default_prepend_bos` and `padding_side = utils.override_or_use_default_value(tokenizer.padding_side, padding_side)` for every use of `tokenizer.padding_side`. However, bugs can still occur when other contributors miss calling this function.
> 
> How do you both think about it?

It feels like there might be a simpler solution than this. Can we just pass the override down (e.g. forward -> to_tokens), wherever it's applied through to the functions where it is used? I don't know why it needs to override cfg temporarily as opposed to just being passed around the different functions?

soheeyang:
> > @alan-cooney I agree, this is an example of a smart solution turning out to be a bad idea in an open-source code 😂 I agree that the context manager is an overly complex solution. However, like @neelnanda-io, I would be sad to lose the feature of changing the default value of `prepend_bos` (and also `padding_side` for me personally) as I am myself using these features all the time.
> > If we don't want to use the context manager, the alternative will be changing the code to use `prepend_bos = utils.override_or_use_default_value(cfg.default_prepend_bos, prepend_bos)` for every use of `cfg.default_prepend_bos` and `padding_side = utils.override_or_use_default_value(tokenizer.padding_side, padding_side)` for every use of `tokenizer.padding_side`. However, bugs can still occur when other contributors miss calling this function.
> > How do you both think about it?
> 
> It feels like there might be a simpler solution than this. Can we just pass the override down (e.g. forward -> to_tokens), wherever it's applied through to the functions where it is used? I don't know why it needs to override cfg temporarily as opposed to just being passed around the different functions?

To clarify, if prepend_bos is True as default but the user wants to use False as default, does your suggestion require the user to pass `False` every time they call a related method? Or does passing prepend_bos=False to one method changes the default value of prepend_bos?

alan-cooney:
Closing for now - we can pick this up when it needs editing again

PR #423, opened 2023-10-19
Remove Python 3.7 support

alan-cooney:
Python 3.7 causes a lot of issues with dependency conflicts, and is due to reach end of life soon (2024).

Dropping so that we can simplify the install process and also use pylint.


neelnanda-io:
Huh, I was going to object to this, since we supported Python 3.7 as Colab uses it. But it turns out Colab upgraded several times and is now at 3.10. No objections from me!

https://github.com/googlecolab/colabtools/issues/3572

PR #428, opened 2023-10-19
Sync readme with docs

alan-cooney:
Edit: Changed to just sync the docs with the readme (will move readme to another pr). An automated solution may be better here, but seems fine for now to just update the docs pages.


neelnanda-io:
Hmm, I'm generally pro putting things in the README, since I don't think people will actually go and read the docs. The reason I have a gallery is to try to create credibility for "this library is actually used by real researchers", and the reason I have the tutorials is to nerd-snipe people who want to learn about mech interp, and IMO the README is something such people are likely to read. I'd be pro keeping them in

alan-cooney:
Makes sense - for now I'll add a quick solution in so that the readme sections and their corresponding pages on the docs site are kept in sync.

In general my approach for starting to use a new lib is this:

1. Skim the readme intro to understand what it does
2. Do a few quick checks (e.g. check it's still maintained with recent commits, see if it seems robust)
3. Skim the API docs to understand what functionality there is
4. Use the Quick Start to get going and then mostly use code hints (from the docstrings) + the API docs pages.

My thinking is if we can support this flow, and encourage people to follow it, it will help people get to grips with the lib in a faster way. It's also nice for the user if all the important docs are in one place, and then tutorials can supplement this for specific use cases.

Btw I've been adding a bunch of PRs recently and merging (been discussing with @jbloomAus and @bryce13950 a bit behind the scenes as well). If you want to have more oversight into the direction of the docs let me know and I can mark more stuff for review instead.

alan-cooney:
> Hmm, I'm generally pro putting things in the README, since I don't think people will actually go and read the docs. The reason I have a gallery is to try to create credibility for "this library is actually used by real researchers", and the reason I have the tutorials is to nerd-snipe people who want to learn about mech interp, and IMO the README is something such people are likely to read. I'd be pro keeping them in

Switched for now to just sync the docs with the readme (which was more up to date). Will pull the other stuff into a separate PR.

PR #434, opened 2023-10-21
Added support for bloom-560m model

SeuperHakkerJa:
# Description

I integrated support for the bloom-560m model, which uses alibi instead of positional encoding. Consequently, the positional_embedding_type flag can be set to 'alibi'.

## Type of change

# Checklist:


transformer_lens/HookedTransformer.py
```diff
@@ -334,6 +334,10 @@ def input_to_embed(
             # keys and queries. See HookedTransformerConfig for details
             residual = embed
             shortformer_pos_embed = None
+        #TODO: alibi embedding doesnt do anything
```
neelnanda-io, line 337: Why is this a TODO? Should Alibi do something here?

transformer_lens/components.py
```diff
@@ -32,12 +33,17 @@ def __init__(self, cfg: Union[Dict, HookedTransformerConfig]):
         self.W_E: Float[torch.Tensor, "d_vocab d_model"] = nn.Parameter(
             torch.empty(self.cfg.d_vocab, self.cfg.d_model, dtype=cfg.dtype)
         )
+        # bloom needs post embedding layer norm
+        if cfg.post_embedding_layer_norm: 
+            self.ln = LayerNorm(cfg)
```
neelnanda-io, line 38: Can we call this post_embedding_ln?

transformer_lens/components.py
```diff
@@ -593,6 +600,27 @@ def forward(
             )
             / self.attn_scale
         )  # [batch, head_index, query_pos, key_pos]
+
+        # alibi encoding before applying causal mask
+        if self.cfg.positional_embedding_type == 'alibi':
+            #TODO: not sure about the side effect of not using standard, double check
```
neelnanda-io, line 606: What does this mean?

transformer_lens/components.py
```diff
@@ -593,6 +600,27 @@ def forward(
...
+            batch_size = attn_scores.size(0)
+            seq_len = attn_scores.size(-2)
```
neelnanda-io, line 608: Should this be -1? Note that when generating text the attention scores are *not* square

neelnanda-io:
Thanks for the PR! I left some minor comments, but it overall looks pretty good. Have you tested that this gives (approx) the same logits as Bloom on HuggingFace?

And would you be able to add a test for it to test_hooked_transformer? https://github.com/neelnanda-io/TransformerLens/blob/main/tests/acceptance/test_hooked_transformer.py

My only hesitation with the test is that 560M is large enough that it might slow things down - thoughts @alan-cooney ?

alan-cooney:
> My only hesitation with the test is that 560M is large enough that it might slow things down - thoughts @alan-cooney ?

Seems likely fine still. In general the tests are starting to take a bit too long however so I'll split the acceptance and unit tests up into parallel workflows.

@SeuperHakkerJa nice work on this! One other thing is the docs are failing to build (probably a formatting error in docstrings). To fix it run `poetry run docs-hot-reload` and then the warning with details about what went wrong will show up (hot reload should let you fix it and then instantly see the warning go away).

SeuperHakkerJa:
Thank you for your feedback and the comments. I'll begin fixing the issue soon.

transformer_lens/components.py
SeuperHakkerJa: Changed

transformer_lens/HookedTransformer.py
SeuperHakkerJa: It's no longer needed, deleted.

transformer_lens/components.py
SeuperHakkerJa: A reminder for myself to double check any potential side effect of setting embedding type to something other than `standard`, no longer needed, deleted!

transformer_lens/components.py
SeuperHakkerJa: Yes it should be set to key_length, changed to -1. Thanks!

alan-cooney:
Btw I switched to draft whilst you're working on this - but feel free to switch back when ready. 

Also happy to review the changes if you want Neel.

transformer_lens/components.py
```diff
@@ -593,6 +598,33 @@ def forward(
             )
             / self.attn_scale
         )  # [batch, head_index, query_pos, key_pos]
+
```
alan-cooney, line 601: Might be worth adding a comment that the original paper calculates this once and applies to the mask
instead when training (which is more efficient)? Not sure about this one, because for our purposes
your approach seems to be the key one.

transformer_lens/components.py
```diff
@@ -593,6 +598,33 @@ def forward(
...
+            # Huggingface impl uses torch.Tensor.baddbmm, with alpha = 1/sqrt(d_head), and beta=1
+            # and alibi.baddbmm(q,k) = beta * alibi + alpha * (q@k),
+            # here the `attn_scores` is already scaled by a factor of self.attn_scale,
+            # we only need to add alibi matrix to the result
+            assert alibi.shape == (
+                attn_scores.size(0),
+                attn_scores.size(1),
+                1,
+                attn_scores.size(-1),
+            ), f"alibi shape {alibi.shape}, expecting {attn_scores.shape}"
```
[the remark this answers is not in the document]
alan-cooney: Also is this right? 

If the query shape is size 1 surelly this only works for predicting the next token, but for the
logits for all
previous tokens it would then give an incorrect answer? Would mean that training wouldn't work with
this code right (or any analysis that looks at logits other than from the last token )

transformer_lens/components.py
```diff
@@ -757,6 +789,49 @@ def apply_rotary(
 
         return torch.cat([x_rotated, x_pass], dim=-1)
 
+    def build_alibi_tensor(
+        self,
+        attention_mask: torch.Tensor,  # batch pos
+        num_heads: int,
+        dtype: torch.dtype,
+    ) -> Float[torch.Tensor, "batch head_index 1 pos"]:
```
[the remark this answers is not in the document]
alan-cooney: Would also be good to have a unit test if possible? Feels like we can check some basic things like:

- For each specific head, the diagonal QK values are the same (e.g. the middle diagonol is 0s)
- For each specific head, the difference between each diagonol scalar value is constant (i.e. m is
  constant for any specific head)

transformer_lens/components.py
```diff
@@ -593,6 +598,33 @@ def forward(
             )
             / self.attn_scale
         )  # [batch, head_index, query_pos, key_pos]
+
+        # alibi encoding before applying causal mask
+        if self.cfg.positional_embedding_type == "alibi":
+            batch_size = attn_scores.size(0)
+            seq_len = attn_scores.size(-1)
+            additive_mask = torch.ones(batch_size, seq_len)
```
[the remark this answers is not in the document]
alan-cooney: Also it needs to have its device set if not (so that it's on the same device as QK)

transformer_lens/components.py
```diff
@@ -757,6 +789,49 @@ def apply_rotary(
...
+                2,
+                device=attention_mask.device,
+                dtype=torch.int32,
+            )
+            slopes = torch.cat([slopes, torch.pow(extra_base, extra_powers)], dim=0)
+
+        arange_tensor = ((attention_mask.cumsum(dim=-1) - 1) * attention_mask)[
+            :, None, :
+        ]
+        alibi = slopes[..., None] * arange_tensor
```
alan-cooney, line 831: Makes sense - worth adding a comment using p5 of https://arxiv.org/pdf/2108.12409.pdf to explain a
bit more about how this gets to the head-specific slope (m).

But can we use their general formula to simplify this a bit ("In general, for n heads, our set of slopes is the geometric sequence that starts
at 2(-8/n) and uses that same value as its ratio.")

alan-cooney (changes requested):
Thanks so much for doing this!

It looks good but I've added some suggestions to improve the readability a bit, and also left a few questions (particuarly regarding the broadcasting onto QK). Let me know what you think.

alan-cooney (changes requested):
Thanks so much for doing this!

It looks good but I've added some suggestions to improve the readability a bit, and also left a few questions (particuarly regarding the broadcasting onto QK). Let me know what you think.

transformer_lens/components.py
```diff
@@ -757,6 +789,49 @@ def apply_rotary(
 
         return torch.cat([x_rotated, x_pass], dim=-1)
 
+    def build_alibi_tensor(
```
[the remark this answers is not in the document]
SeuperHakkerJa: `create_attention_linear_bias` sounds great to me. (I was naming it `build_alibi_tensor` only because it was named so in HF code)

transformer_lens/components.py
SeuperHakkerJa: For this function, I was also using [HF's implementation](https://github.com/huggingface/transformers/blob/41496b95da5d34368a13538548c85527377972a6/src/transformers/models/bloom/modeling_bloom.py#L86). If you prefer, I can switch it back to the general (original?) formula.

SeuperHakkerJa:
**Summary:**

1. **Comment Cleanup & Documentation**: 
   - Enhanced code readability by cleaning up comments.
   - Added detailed documentation for several functions, notably `compute_attention_linear_bias` and `expand_alibi_on_query_dim`.

2. **Reimplementation of Alibi Tensor Function**:
   - This new implementation ensures that the diagonal values remain consistent.
   - To achieve this, the original Alibi tensor has been expanded on the query dimension (previously set to 1) and values have been filled as suggested in the original paper.

**Side Note**:
I examined the original code referenced in the Alibi paper ([source](https://github.com/ofirpress/attention_with_linear_biases/blob/07e8a0eabfa87236395b3e4aebc790ca49f456f0/fairseq/models/transformer.py#L757)). It contains the following remark:
```
#In the next line, the part after the * is what constructs the diagonal matrix (right matrix in Figure 3 in the paper). 
#If you run it you'll see that it doesn't exactly print out the same matrix as we have in Figure 3, but one where all rows are identical.
#This works because the softmax operation is invariant to translation, and our bias functions are always linear. 
```
Thus, the 'original' code leveraged broadcasting and softmax's translation invariant properties to derive the `pattern`. My implementation, as explained above, will in addition ensure that `hook_attn_scores` correctly captures the attention scores.

tests/unit/test_compute_linear_bias.py
```diff
@@ -0,0 +1,45 @@
...
+        expected_remaining_values = torch.tensor(
+            [7] * (query_pos - len(expected_first_values)), device=x.device
+        )
+
+        # Concatenate to form the expected row
+        expected_row = torch.cat((expected_first_values, expected_remaining_values))
+
+        # Check if the result matches the expected row
+        if not torch.equal(result[0, 0, i], expected_row):
+            return False
```
alan-cooney, line 26: I've put some suggestions for how to simplify the underlying functions in the components file.

If you agree with that approach, the functions are now small enough to make simple unit tests that follow a different approach to the main code (i.e. the tests can use loops rather than vectorised tensor operations

An example for this could be testing create_alibi_slope:

```
n_ctx = 100

# Calculate the expected slope, with a non-vectorised approach
slope_expected = torch.zeros((n_ctx, n_ctx))
for row in range(n_ctx):
    for col in range(n_ctx):
        # Create the slope value
        slope_expected[row, col] = float(min(col - row, 0))

# Check against the method's vectorised version
result = Attention.create_alibi_slope(n_ctx)
assert torch.allclose(slope_expected, result)
```

alan-cooney (changes requested):
I've added some suggestions to show how to make this clearer - I think it's getting there but this should make it much more readable.

SeuperHakkerJa:
## Summary of Changes

1. Converted all three ALiBi-related functions to `staticmethod`s.
2. Implemented unit tests for the ALiBi functions.
3. Since constructing a tensor with dimensions `(n_heads, n_ctx, n_ctx)` of `float32` was causing failures in the GitHub acceptance tests. Modified the ALiBi bias tensor to initialize during the first forward pass and to only reconstruct when the new query/key context length exceeds the longest context length encountered in the past. This change addresses issues with GitHub acceptance tests and optimizes for memory and computational efficiency.

alan-cooney (changes requested):
Looks good - one small question but otherwise good to go.

transformer_lens/loading_from_pretrained.py
```diff
@@ -721,6 +723,22 @@ def convert_hf_model_config(model_name: str, **kwargs):
...
             "attention_dir": "bidirectional",
         }
+    elif architecture == "BloomForCausalLM":
+        cfg_dict = {
+            "d_model": hf_config.hidden_size,
+            "d_head": hf_config.hidden_size // hf_config.n_head,
+            "n_heads": hf_config.n_head,
+            "d_mlp": hf_config.hidden_size * 4,
+            "n_layers": hf_config.n_layer,
+            "n_ctx": 2048,
```
alan-cooney, line 733:
```suggestion
            "n_ctx": 2048, # Capped due to HF Tokenizer Constraints
```

alan-cooney (approved):

alan-cooney (approved):

gegallego:
Hi! Is there any reason to not adding the other BLOOM models in this PR? Thanks!

alan-cooney:
> Hi! Is there any reason to not adding the other BLOOM models in this PR? Thanks!

Not aware of any reason. Happy to review a pr if you want to add the other ones

SeuperHakkerJa:
> > Hi! Is there any reason to not adding the other BLOOM models in this PR? Thanks!
> 
> Not aware of any reason. Happy to review a pr if you want to add the other ones

I was only concerned that the other models might be too large, potentially causing the unit tests to take too much time:
- [bloom-1b1](https://huggingface.co/bigscience/bloom-1b1)
- [bloom-1b7](https://huggingface.co/bigscience/bloom-1b7)
- [bloom-3b](https://huggingface.co/bigscience/bloom-3b)
- [bloom-7b1](https://huggingface.co/bigscience/bloom-7b1)

gegallego:
Thanks for the quick response! I don't know exactly how your unit tests work... so I won't do a PR to avoid issues in that sense. But basically if I want to load a larger model I just need to list it in [`OFFICIAL_MODEL_NAMES`](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/loading_from_pretrained.py#L18) and [`MODEL_ALIASES`](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/loading_from_pretrained.py#L145), right?

SeuperHakkerJa:
> Thanks for the quick response! I don't know exactly how your unit tests work... so I won't do a PR to avoid issues in that sense. But basically if I want to load a larger model I just need to list it in [`OFFICIAL_MODEL_NAMES`](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/loading_from_pretrained.py#L18) and [`MODEL_ALIASES`](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/loading_from_pretrained.py#L145), right?

Yes, that's right! 

@alan-cooney  I will then initiate this pull request to integrate the remaining bloom models (up to 7b), potentially this weekend or at some point next week

alan-cooney:
> > > Hi! Is there any reason to not adding the other BLOOM models in this PR? Thanks!
> > 
> > 
> > Not aware of any reason. Happy to review a pr if you want to add the other ones
> 
> I was only concerned that the other models might be too large, potentially causing the unit tests to take too much time:
> 
> * [bloom-1b1](https://huggingface.co/bigscience/bloom-1b1)
> * [bloom-1b7](https://huggingface.co/bigscience/bloom-1b7)
> * [bloom-3b](https://huggingface.co/bigscience/bloom-3b)
> * [bloom-7b1](https://huggingface.co/bigscience/bloom-7b1)

Regarding the tests - I suspect they are actually fine but the easy way to run with the CPU and check they take at most a few seconds.

PR #441, opened 2023-10-23
Test main demo

bryce13950:
# Description

The main demo has been tested completely now. There are two reasons this may fail in the execution here. The first is that cell 41 sometimes has a value that rounds slightly differently. When I was working through this it failed once, but just now it failed twice in a row. We may need to skip that one as well, but we will see if it becomes an issue. The value in the file is `286.4421`, but GitHub seems to sometimes generate it as `286.4420`

Another thing is with the docs generation. Currently it is configured to turn warnings into failures, and I think it is warning that the main demo does not exist on the docs site. The message is as so...

```
Warning, treated as error:
/home/runner/work/TransformerLens/TransformerLens/docs/source/generated/demos/Main_Demo.ipynb:333:File not found: 'generated/demos/neelnanda.io'
```

I ran the CI with the -W removed from make_docs, and it passed. Not sure if my assumption here is correct or now with it being an issue of the file not currently existing, but temporarily turning that off may be the solution to get the PR to be accepted. If it is something else then let me know, as I may have missed a configuration step without realizing it.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


alan-cooney (commented):
Nice work on this! I've added some comments inline, and also have some general suggestions:

- Lots of the imports in the notebook aren't needed (they should be highlighted light gray) - suggest we use this opportunity to get rid of them. That includes the weird matplotlib line which isn't actually used.
- For the `device`, the lib now lets you do `device = utils.get_device()` which also supports mps. Think this is a bit better (however note that we may need to regex out the device name wherever it's shown (as the CD will still be CPU).
- Rounding errors are common across different hardware (particularly GPU vs CPU, but also between different CPUs). The solution in unit tests is usually to use `torch.allclose()` rather than checking exact equality. However here that's not possible so I'd suggest just using `.round(decimals=3)` or similar.
- For random numbers it's best practice to use a random seed. See https://pytorch.org/docs/stable/notes/randomness.html
- I'd just skip the for loop that loads different models at the end entirelly for the tests - its far too slow. Everything else is rapid on the CPU
- Circuitsvis -> I reverted the change and agree we should just also skip it from the tests. HTML just changes too much.

I think that's everything but worth doing another skim post-changes (happy to do this and just check all the outputs/charts are the same as before).

Let me know if you have any questions.

Alan

bryce13950:
All set! Thanks for the tip on the random seed with torch. All tests are now passing again, with the only thing left to look at being the issue with the building of docs not running, which I mentioned earlier with the relation to the flag.

alan-cooney:
I think the docs issue is from the introduction markdown cell btw. It has this:

```markdown
This is a demo notebook for [TransformerLens](https://github.com/neelnanda-io/TransformerLens), **a library I ([Neel Nanda](neelnanda.io))
```

The last url needs https:// before it so that it's detected as an absolute rather than relative link.

alan-cooney (changes requested):
Note above regarding how to fix the doc building, but otherwise I think it's good to go.

alan-cooney (approved):

PR #443, opened 2023-10-27
Support mistral 7 b

### LINKED ISSUE #387: [Proposal] Support Mistral 7B model
### Proposal 

Mistral AI have released a 7B model which they claim is better than LLaMA 13B, it'd be nice if TransformerLens supports it. This seems like a good 7B model, and 7B is about the largest model that's easy enough to work with without needing very high end GPUs, so I'd like TL to support the best 7Bs around. Unlike LLaMA it's actually open source under an Apache 2.0 license

The [blog post](https://mistral.ai/news/announcing-mistral-7b/), and [HuggingFace page](https://huggingface.co/docs/transformers/main/model_doc/mistral)

### Details

New architecture that'd need to be supported:
* Grouped Query Attention (which will also unblock LLaMA 2 70B!)
* Sliding Window attention (I believe this may already be supported for GPT-Neo?)
* Their tokenizer (should be easy)

Felhof:
# Description

This PR closes #387 by adding support for [Mistral](https://mistral.ai/news/announcing-mistral-7b/) and implementing [Grouped Query Attention](https://arxiv.org/abs/2305.13245v2).

### Mistral 7B
The weight's from Huggingface's [Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1) can now be loaded into a HookedTransformer using 
```
# these parameters are necessary as Mistral uses RMS norm and not layernorm.
tl_mistral = HookedTransformer.from_pretrained(
    "mistral-7B",
    fold_ln=False,
    center_writing_weights=False,
    center_unembed=False
)
```
Note that Mistral is only supported in transformers >= 4.34 and hence Python >= 3.8 is required to use it.
The demo notebook `Mistral.ipynb` features a comparison of Huggingface's Mistral with the HookedTransformer implementation. I tested it on the same prompts that were used in the Llama demo. The differences in the resulting logits were around 0.01 which is slightly higher than what Llama2 gets. This may be due to a known issue with [rotary embeddings](https://github.com/neelnanda-io/TransformerLens/issues/385) which also affects LLama2 and Pythia.

### Grouped Query Attention
Mistral utilizes [Grouped Query Attention](https://arxiv.org/abs/2305.13245v2) (GPA) which is not used by any other model supported by TransformerLens and had to be implemented. I added a new class `GroupedQueryAttention` and an abstract base class `AbstractAttention` which now features the common functionality of both attention classes. The difference between `Attention` and `GroupedQueryAttention` is how they handle the key and value projections, as in GPA groups of queries share the same keys and values (see image below). This mostly affects the internal workings of the class. To avoid confusion, to not break existing code interacting with attention, and to make the design of future code easier, public attributes such as `W_K` and `W_V` have the same shape for both classes. This is because in the case of GPA, the underlying parameters are hidden behind a property that expands them using `torch.repeat_interleave`. A GPA block should behave the same as a regular Attention block whose weights are the result of applying `torch.repeat_interleave` to the GPA block's weight. There is a unit test that confirms this.

grouped_query_attention

## Type of change

# Checklist:


transformer_lens/loading_from_pretrained.py
```diff
@@ -721,6 +727,27 @@ def convert_hf_model_config(model_name: str, **kwargs):
...
+            "n_ctx": 32768,
+            "d_vocab": 32000,
+            "act_fn": "silu",
+            "normalization_type": "RMS",
+            "positional_embedding_type": "rotary",
+            "window_size": 4096,
+            "attn_types": ["local"] * 32,
+            "eps": 1e-05,
+            "n_key_value_heads": 8,
+            "final_rms": True,
```
neelnanda-io, line 746: IMO this should be false, final_rms is an override to set _just_ the final normalization to be RMS while the rest is LayerNorm, and isn't needed when it's all RMS

neelnanda-io (commented):
Thanks a lot for doing this! I left some small comments, but this isn't a full review (I haven't looked at exactly how you implemented grouped query attention), so I'll leave that for someone else to finish off.

alan-cooney:
Just to flag I'll review this once the previous PR goes through on attention, as we'll need to adjust for a few conflicts in it

alan-cooney:
Hey @Felhof nice work on this, and the abstract attention class approach is great (we're planning on doing this for a bunch of the components). 

Would you be able to fix the merge conflicts first and then I can do a full review?

pyproject.toml
```diff
@@ -50,6 +50,7 @@ nvidia-nccl-cu12 = { version = ">=2.18.1", markers = "platform_system == 'Linux'
 nvidia-nvtx-cu12 = { version = ">=12.1.105", markers = "platform_system == 'Linux' and platform_machine == 'x86_64'" }
 triton = { version = ">=2.1.0", markers = "platform_system == 'Linux' and platform_machine == 'x86_64'" }
 # End PyTorch 2.1.0 Bug Fix
+better-abc = "^0.0.3"
```
ojh31, line 53: Should require transformers>4.34 https://huggingface.co/mistralai/Mistral-7B-v0.1 "Ensure you are utilizing a stable version of Transformers, 4.34.0 or newer."

neelnanda-io:
Hey! I'm curious what the status of this PR is? A few of my MATS scholars want to use Mistral. Can they just check out this PR?

pyproject.toml
Felhof: Done :)

abdurraheemali:
@alan-cooney @neelnanda-io @Felhof  I want to run an activation patching experiment on llama-70b, so I'm going to check out the support-mistral-7b and report (in an hour-ish?) whether the grouped query attention implementation works for me or not

neelnanda-io:
@alan-cooney @Felhof What's the status of this PR? I'm not sure which of you it's blocked on. Either way, I hear that people have been able to checkout this branch and get Mistral working, so thanks a lot for the work up to that point!

Felhof:
The branch has been working for a while but needs approval to be merged into main :) I made sure it's up-to-date with main again.

wesg52:
@alan-cooney Ping on review :)

tests/unit/test_grouped_query_attention.py
```diff
@@ -0,0 +1,78 @@
+import torch
+
+from transformer_lens.components import Attention, GroupedQueryAttention
+from transformer_lens.HookedTransformerConfig import HookedTransformerConfig
+
+
+def test_grouped_query_attention():
```
alan-cooney, line 7: Adding a docstring or more descriptive test name would be good! General idea makes a lot of sense though - thanks for adding this!

alan-cooney (changes requested):
Nice! Thanks for adding this and the approach is great.

A general point (don't feel the need to change, but to be aware) is that we're trying to type all methods and also add docstrings to them (Google style). If you have a chance it would be good to do this as well to e.g. explain the interleave approach that is used.

Felhof:
Thanks for the review Alan! I have added better typing and documentation and removed the demo

alan-cooney (approved):

alan-cooney:
Thanks! And thanks for adding this!

neelnanda-io:
Woot! Really glad this got merged in! Thanks for adding it @Felhof and sorry for the long delay

PR #451, opened 2023-11-22
Fixed current CI issues with accuracy failing for Pythia model

bryce13950:
# Description

Lowers the margin of error in the test `tests/acceptance/test_hooked_transformer.py` to account for recent changes to the model EleutherAI/pythia-70m, which has caused old margin of error to fail in the CI. I am not sure if there should be a more complicated change to the code base to account for this recent change, but this is an option to fix the test at least temporarily, or permanently if it is deemed accurate enough with a margin of 5e-4 instead of 5e-5.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


alan-cooney (approved):
Seems like a sensible solution for now

PR #452, opened 2023-11-22
Refactor components

bryce13950:
# Description

Everything that was previously in `transformer_lens/components` has been refactored so that every class has a separate file. I also simplified and improved a couple things here and there in some of the classes. e.g. changed some of the setup in some of the init functions to use ternary statements instead of large if/else block & removing modification of global variables in the loop for the Attention component to instead have a local variable from a wrapper that manages the state of the variable.

There are a lot of various improvements that can be made here that would enable the creation of unit tests. I am refraining from doing too much of it at the moment, so that the changes can more easily be tracked individually from the base of these separate new files as opposed to making a bunch of changes at the same time as splitting up one giant file.

No changes are needed for this in other parts of the code base, so people should be able to continue using these components as is. The only thing that I changed as far as naming is the variable previously mentioned in the Attention component. However, I don't really think that it is ever referenced outside of the Attention component itself. If it does seem like it is going to be an issue, then I can change the name back to the original. 

I do still have to go through the document headers and update the copy and pasting I did, but thought I would get this up for review.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


bryce13950:
The file headings have all been updated now, and it should be ready to merge once the bug is fixed that is currently affecting the whole repo.

transformer_lens/components/attention.py
```diff
@@ -0,0 +1,569 @@
+"""Hooked Transformer Attention Component.
```
alan-cooney, line 1: This is missing the latest alibi code - I think it was probably a change since your PR

alan-cooney (changes requested):
Couple of things highlighted but otherwise looks good

alan-cooney:
Note: Will return to after Mistral PR is merged

Issue #385, opened 2023-09-26
[Bug Report] Pythia models / Rotary Embeddings don't match Huggingface.


UFO-101:
**Describe the bug**
Pythia model outputs don't exactly match the Huggingface Transformers implementation.

**Code example**
```
def check_similarity_with_hf_model(tl_model, hf_model, atol, prompt="Hello, world!"):
    tokens = tl_model.tokenizer.encode(prompt, return_tensors="pt")
    logits = tl_model(tokens, prepend_bos=False)
    hf_logits = hf_model(tokens).logits
    assert torch.allclose(t.softmax(logits, dim=-1), t.softmax(hf_logits, dim=-1), atol=atol)

model_name = "EleutherAI/pythia-70m"
tl_model = HookedTransformer.from_pretrained(model_name)
hf_model = AutoModelForCausalLM.from_pretrained(model_name)
check_similarity_with_hf_model(tl_model, hf_model, atol=1e-5)
```
This fails with `model_name = "EleutherAI/pythia-70m"`, but passes with every other model I tried.
It passes with `pythia-70m` if I set `atol=0.1`. Arthur [says](https://opensourcemechanistic.slack.com/archives/C04SRRE96UV/p1695595694931749?thread_ts=1695593544.494209&cid=C04SRRE96UV) it works for him with `atol=1e-3`.

**System Info**
Describe the characteristic of your environment:
 - `transformer_lens` main branch.
 - MacOS
 - Python 3.10.12
 - Pytorch 1.13.1

**Additional context**
See discussion in Open Source Mechanistic Interpretability Slack here: https://opensourcemechanistic.slack.com/archives/C04SRRE96UV/p1695593544494209

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

jettjaniak:
I can confirm the finding on MacOS, TL 1.6.1, Python 3.9.6, PyTorch 2.0.1 - both on cpu and mps and for all small pythia models (14, 31, 70m).

The relevant HF line is https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L182
I went there with a debugger and compared the values to hook_rot_k / hook_rot_q for a few positions and a few heads in layer 0 on pythia-14m. There is no difference, I don't think RoPE are the issue.

UFO-101:
Oh ok good spot, thank you!

[...]

PR #456, opened 2023-11-29
Fixing Llama2 numerical errors

obalcells:
# Description

The TransformerLens implementation of [Llama2-7B-chat](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf) doesn't match the one by Huggingface. This probably also occurs with the other Llama2 models but I haven't tested it.

In the following code the `torch.allclose` assert is triggered if we use an error tolerance of `atol=1.0`. It doesn't get triggered if we use `atol=1.5` instead.

```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer

def check_similarity_with_hf_model(
    tl_model: HookedTransformer,
    hf_model: AutoModelForCausalLM,
    atol: float,
    prompt="Hello world!",
):
    tokens = tl_model.tokenizer.encode(prompt, return_tensors="pt").cuda()
    tl_logits = tl_model(tokens, prepend_bos=False)
    hf_logits = hf_model(tokens).logits
    assert torch.allclose(tl_logits, hf_logits, atol=atol)


MODEL_NAME = "meta-llama/Llama-2-7b-chat-hf"
    MODEL_NAME,
    low_cpu_mem_usage=True,
    torch_dtype=torch.float32,
).cuda()
tl_model = HookedTransformer.from_pretrained(
    MODEL_NAME,
    hf_model=AutoModelForCausalLM.from_pretrained(
        MODEL_NAME,
        low_cpu_mem_usage=True,
        torch_dtype=torch.float32,
    ),
    tokenizer=AutoTokenizer.from_pretrained(MODEL_NAME),
    device="cuda",
    n_devices=1,
    move_to_device=True,
    fold_ln=False,
    fold_value_biases=False,
    center_writing_weights=False,
    center_unembed=False,
    torch_dtype=torch.float32,
)

with torch.no_grad():
    check_similarity_with_hf_model(tl_model, hf_model, atol=1)
```

This issue has already been reported (#385) and I have been able to reproduce it using two RTX A6000 (48GB) GPUs (one for each model).

I believe the issue lies in the value of the constant in the layer norm module (called `eps`) that's added to avoid division by zero ([see implementation here](https://github.com/neelnanda-io/TransformerLens/blob/main/transformer_lens/components.py#L347)).

The TransformerLens Llama2-7B-chat model uses the value `1e-6` but the Huggingface Llama2-7B-chat model uses `1e-5` by default.

Confusingly, both the `LlamaRMSNorm` [class](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L96C8-L96C8) and the [default Llama config](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/configuration_llama.py#L128) use the value `1e-6` by default. **However** this gets overriden by the [model's config in Huggingface](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/blob/main/config.json). The Huggingface config for the other Llama2 models (chat and base, as well as the larger models) also uses the correct value `1e-5`. This was probably the reason why the error was introduced.

### After Fix

I've calculated the mean absolute error and the mean relative error between the residual stream vectors of both models. The plots with `eps=1e-6` are those before the fix, and the ones with `eps=1e-5` are the ones afterwards:

mae_bug

relative_error

mae_fixed

fixed_relative_error

The relative error was calculated with `abs((hf_resid_post-tl_resid_post) / (abs(hf_resid_post) + 1e-5))`.
**Notice that the y-axis is scaled in the last two plots.**

After the fix we're able to run
```python
with torch.no_grad():
    check_similarity_with_hf_model(tl_model, hf_model, atol=1e-4)
```

without triggering the assert. The assert is triggered at atol=`1e-5`.

## Type of change

The only change is setting `eps` to `1e-5` instead of `1e-6` in the default config for all Llama2 models.

**[Here's](https://github.com/obalcells/supermech/blob/main/experiments/llama2_tl_bug/clean_repro.ipynb) the notebook where I reproduce and check the fix for the error.**

I'd also like to thank @andyrdt for helping me find the bug and for helping me create my first PR in GitHub.


ArthurConmy:
Thank god someone found this -- fantastic!

I'm pretty sure we want to keep Llama-1 models with `1e-6` though, from reading [this](https://github.com/ggerganov/llama.cpp/issues/173) back from March. Can you implement this?

obalcells:
Done! The change should only apply to the new llama2 models now.

Btw, there's an acceptance test [here](https://github.com/neelnanda-io/TransformerLens/blob/main/tests/acceptance/test_hooked_transformer.py#L170C5-L170C35) which is made for finding these kinds of bugs, but the test doesn't seem to be turned on? (neither for llama2, nor for any other model)

neelnanda-io:
How did a tiny detail cause that big an atol difference?! Thanks a lot for catching this :)

ArthurConmy (approved):

PR #458, opened 2023-11-30
Make LLaMA 2 loadable directly from HF

### LINKED ISSUE #455: [Proposal] Make LLaMA 2 loadable directly from from_pretrained with just the name
### Proposal 

Currently, to load LLaMA 2 models, you need to explicitly pass in the huggingface model and tokenizer to from_pretrained, unlike eg GPT-2 where passing in just the name suffices. But LLaMA 2 is officially on HuggingFace (you just need to be logged in on HuggingFace on your remote machine and to have been approved to access the weights) so it should be fine to have just the name work.

It was setup this way for LLaMA 1 which is *not* officially accessible on HuggingFace, but there's no reason for the extra faff for LLaMA 2. 

If you make this change, it would be good to edit the [LLaMA demo](https://colab.research.google.com/github/neelnanda-io/TransformerLens/blob/main/demos/LLaMA.ipynb#scrollTo=L-ZJfkBfhJzc) to note this way of loading LLaMA 2.

andyrdt:
# Description

Makes LLaMA-2 models loadable directly from `from_pretrained`.

Since LLaMA-2 downloads are gated by a HF login, this requires logging in to HF prior to downloading the model via `transformers-cli login`.

I updated the LLaMA demo notebook as well.

Fixes #455 

## Type of change

### Screenshots

Example:

Screenshot 2023-11-29 at 9 01 17 PM

# Checklist:


alan-cooney (approved):
Looks great - thanks!

PR #464, opened 2023-12-11
Draft PR: add Pyright static typing to hook_points.py

starship006:
# Description

This draft PR adds Pyright static typing to hook_points.py. See [Alan's message on the slack](https://opensourcemechanistic.slack.com/archives/C04T79RAW8Z/p1699673111412959) for motivation/background. Specifically, this PR:
- modifies the HookPoint and HookedRootModule to support a baseline amount of pyright typing
- adds a [tool.pyright] configuration to pyproject.toml to configure the typing for this project

It's not entirely complete, but is in a state to where it would benefit from some review/assistance. Before being PR ready, we should:
- Address some TODO's I've marked on lines of code that I'm ucnertain about
- Clean up remaining type warnings on hook.py 
- If possible, adjust the pyproject.toml file so the type warnings quit applying to all files

## Type of change

# Checklist:

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~


starship006:
I was also using this draft PR as a way to see if the 50+ tests I was failing locally was due to my machine or not; it seems like this may be the case, but I'm also failing some formatting cases. Will fix.

transformer_lens/hook_points.py
```diff
@@ -158,7 +200,12 @@ class HookedRootModule(nn.Module):
     loss.backward() (and so need to disable the reset_hooks_end flag on run_with_hooks)
     """
 
-    def __init__(self, *args):
+    name: Union[str, None]
+    mod_dict: Dict[str, nn.Module]
+    hook_dict: Dict[str, HookPoint]
+
+    # TODO: check if adding 'Any' is okay here
```
derpyplops, line 207: probably fine, this is also how varargs are typed in the `typing` module

transformer_lens/hook_points.py
```diff
@@ -45,63 +71,77 @@ def __init__(self):
...
                 self.bwd_hooks.append(handle)
         else:
             raise ValueError(f"Invalid direction {dir}")
 
-    def remove_hooks(self, dir="fwd", including_permanent=False, level=None) -> None:
+    def remove_hooks(
+        self,
+        dir: str = "fwd",
+        including_permanent: bool = False,
+        level: Union[int, None] = None,
```
alan-cooney, line 143:
```suggestion
        level: Optional[int] = None,
```

alan-cooney: nit: This is the same, but more standard. Will help us find-and-replace all optional types in the codebase once Optional and Union both get removed (in Python 3.13 I think).

transformer_lens/hook_points.py
```diff
@@ -126,13 +166,15 @@ def clear_context(self):
...
-    def forward(self, x):
+    # TODO: can we assume this will always be a torch.Tensor?
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
         return x
 
     def layer(self):
         # Returns the layer index if the name has the form 'blocks.{layer}.{...}'
         # Helper function that's mainly useful on HookedTransformer
         # If it doesn't have this form, raises an error -
+        assert self.name is not None
```
alan-cooney, line 177: Assertions are just for testing (they are completely ignored by Python if optimization is used (-O)).

The better thing to do here is an if statement, and then raise an error.

transformer_lens/hook_points.py
```diff
@@ -126,13 +166,15 @@ def clear_context(self):
         del self.ctx
         self.ctx = {}
 
-    def forward(self, x):
+    # TODO: can we assume this will always be a torch.Tensor?
```
alan-cooney, line 169: I think so - can't imagine a time when it wouldn't be

```suggestion
```

transformer_lens/hook_points.py
```diff
@@ -158,7 +200,12 @@ class HookedRootModule(nn.Module):
     loss.backward() (and so need to disable the reset_hooks_end flag on run_with_hooks)
     """
 
-    def __init__(self, *args):
+    name: Union[str, None]
```
alan-cooney, line 203:
```suggestion
    name: Optional[str]
```

nit

transformer_lens/hook_points.py
alan-cooney, line 207:
```suggestion
```

transformer_lens/hook_points.py
```diff
@@ -173,20 +220,24 @@ def setup(self):
...
+            if isinstance(module, HookPoint):
                 self.hook_dict[name] = module
 
     def hook_points(self):
         return self.hook_dict.values()
 
     def remove_all_hook_fns(
-        self, direction="both", including_permanent=False, level=None
+        self,
+        direction: str = "both",
```
alan-cooney, line 238: Could also use Literal[] here?

transformer_lens/hook_points.py
```diff
@@ -389,22 +462,25 @@ def add_caching_hooks(
...
-        elif type(names_filter) == str:
+        elif isinstance(names_filter, str):
             filter_str = names_filter
             names_filter = lambda name: name == filter_str
-        elif type(names_filter) == list:
+        elif isinstance(names_filter, list):
             filter_list = names_filter
             names_filter = lambda name: name in filter_list
 
+        assert isinstance(names_filter, Callable)
```
alan-cooney, line 472: Should use if/else and then raise a ValueError

transformer_lens/hook_points.py
```diff
@@ -389,22 +462,25 @@ def add_caching_hooks(
...
-        def save_hook(tensor, hook):
+        def save_hook(tensor: torch.Tensor, hook: HookPoint):
             if remove_batch_dim:
                 cache[hook.name] = tensor.detach().to(device)[0]
             else:
                 cache[hook.name] = tensor.detach().to(device)
 
-        def save_hook_back(tensor, hook):
+        def save_hook_back(tensor: torch.Tensor, hook: HookPoint):
+            assert hook.name is not None
```
alan-cooney, line 483: Similar assert issue

alan-cooney (changes requested):
Thanks for adding this and apologies for the delay over the holiday period!

I've put some comments - mostly just a few small things. Worth doing a search for the key things raised (e.g. use of assert) as there are a few repeated places in the file.

starship006:
Sorry for the long turnaround on this. I've addressed all of the above comments, they were very helpful, thanks. @alan-cooney is this better now?

bryce13950:
@starship006 I just opened a PR on your fork for the most recent changes to TransformerLens. @alan-cooney has become incredibly busy in the last couple months, but I am going to poke him to approve this, so that I can get it merged when it is done. There have been quite a few changes to the `hook_points` since you originally put this together. There is at least one param that needs to be typed out, and then we will see if the CI passes.

Also, for the future, if you name your branch on your fork something different than "main", maintainers can then do things like merge the current head into your PR without needing to open a separate PR on your fork. Not a big deal, but thought I would put that info out there in case you didn't know, so that you don't need to be bothered with it in the future.

starship006:
Thanks for taking care of the merge. I think I've taken care of everything. Two notes:
- both of the `add_caching_hooks` functions in the merge received code which contained an `is_backward` variable, but the function header wasn't updated to reflect that. ive added it back in for both instances
- linter was complaining about the code accessing`hook.name`while assuming that it isn't None, even though it is typed with `Optional`. i didn't quite know how to deal with this: I imagine you could put in `#type: ignore` on the parts where its accessing `hook.name` or just do a check to ensure it is not None. I've opted for the later, but let me know if there's a better way to do this.

starship006:
Ah, okay, nevermind it looks like there are some type checks that need to pass. One thing that I'm noticing is the recurring theme of `NamesFilter`, which is an Optional type of either a sequence or a callable, is getting called but the linter doesn't know that it is a callable and not a sequence. Should I also go ahead and throw in isinstance checks for all of these too? (The reason I'm asking is that I don't know if doing too many of these slows down the code too much / is unprincipled and there's a better way of doing it)

bryce13950:
Go ahead and assert is instance on that param. The hit on performance for doing something like that is incredibly small, especially in the grand scheme of things while dealing with massive files like LLM models. 

Also, I haven't been able to reach Alan, so why don't you reopen a new PR for this, and just reference this PR number in the new one. We can't close this one without him approving it, and I don't know if it is going to happen.

PR #465, opened 2023-12-15
Add pos_slice to run_with_cache

VasilGeorgiev39:
# Description

Very often we care about the activations only at the last token. This change adds a pos_slice param to run_with_cache so that only the activations at the requested positions are returned. This can save memory, especially when running large models

Let me know if this makes sense and I'll add tests and stuff

## Type of change

# Checklist:


VasilGeorgiev39:
Notes from a short conversation with Neel:

- Currently if the pos_slice is an int it will collapse the pos dimension, while many things assume it always exists. A better behaviour would be to keep it as a dummy dimension with size 1
- Currently the code kinda ignores attention heads and assumes that the pos dimension is the second from the end, I should handle attention heads properly. It is not clear if the proper behaviour is to slice on the source or destination, probably would be more reasonable to slice on the dest token

Note: I'll deal with those after Christmas, but if you have more feedback before that please leave it here and I'll respond

VasilGeorgiev39:
I think I addressed the comments - the slicing now keeps the pos dimension and also correctly handles attention heads. I have also added documentation and unit tests

PR #466, opened 2023-12-16
(Draft) Add DLA function to utils

VasilGeorgiev39:
# Description

DLA is usually the first step we do in a new exploration. I think it would be nice to have a common function that does it in a single step.

Let me know if you think this does not generalize well enough or if you have other concerns.

Not sure if Utils is the right place for it tho, maybe we can create a new module that will hold the mech interp toolkit?

If it looks good I'll write tests and stuff.

## Type of change

# Checklist:


alan-cooney:
Thanks for starting on this - it seems useful and I agree that it should be it's own file (probably just for DLA as it'll become quite large once it's fully documented etc).

In general I agree as well that it's probably worth expanding this a bit to work more generally. Specifically you can break DLA down recursively e.g. by attention layer -> attention head -> source layer -> source component... It would be nice to hae this as well.

Hope that makes sense and if you are unsure about how to abstract more I'm happy to have a chat about it!

VasilGeorgiev39:
Hi @alan-cooney, thanks for the comment. I have a couple questions:

I can get the attention head contributions (or even the mlp neurons) with get_full_resid_decomposition(), however I can get the correct and incorrect directions only for the residual stream with tokens_to_residual_directions(). How can I get the directions for the individual heads (or even neurons) ?

Also, what do you mean by break down by 'source layer' and 'source component' ?

bryce13950:
@VasilGeorgiev39 Are you still available to wrap this up?

VasilGeorgiev39:
@bryce13950 Yes, I will be available after the 9th of May. What do you think would be the best approach for this?

jlarson4:
Unfortunately, this PR is much too far behind the current state of the repository. I am going to open an issue for this feature as something that can potentially be added in the future as a tool under the `tools` directory

PR #473, opened 2023-12-28
Make `tokenize_and_concatenate` work with more datasets

ArthurConmy:
# Description

**TL;DR I tried to use `tokenize_and_concatenate` with a general HF Dataset. `tokenize_and_concatenate` was the function I wanted for my use case [1] but I had problems** (two blocked by an error message [2], one that passed through without my knowledge [3])

So this PR fixes this and keeps backward compatibility. The PR:

[1] Makes the function work with more than just arrow datasets. Maybe this was a feature in past HF, but now most datasets are not arrow datasets
[2] a) Only passes `num_proc` if we aren't streaming -- this causes a bug when `IterableDataset` are passed to this function
[2] b) Similarly, skips the final formatting as this also causes a bug with  `IterableDataset`s
[3] Makes remove padding optional. E.g Pythia's training data had pad tokens, sometimes we want this

## Type of change

# Checklist:


neelnanda-io (changes requested):

jlarson4:
Closed in favor of #1287, which implements the concepts here into the modern TransformerLens system

PR #477, opened 2024-01-10
Closes #478: Adding the Qwen family of models

### LINKED ISSUE #478: [Proposal] Adding Qwen to transformer lens
### Proposal 

A proposal to add the Qwen family of models to transformer lens.

### Motivation
 Qwen is a family of [very competitive](https://opencompass.org.cn/leaderboard-llm) large language models built by Alibaba Cloud. There are also a wide variety of model sizes which would allow for research to compare across scales from 1.8b -> 72b. Additionally, there are also fine-tuned chat versions that are effective at refusing harmful requests, which would help research exploring model refusal behaviors. Finally, the models are strongly bilingual, which isn't the case for existing models on transformer lens.

### Checklist

Aaquib111:
# Description

Adds support for Qwen, a family of peformant strongly bilingual LLMs to aid our MATS research on exploring model refusal circuits. Adds:
- Qwen 1.8b, 7b and 14b.
- Qwen-Chat 1.8b, 7b, 14b.

The models above have been tested for numerical accuracy between the transformer lens and huggingface implementations.

TODO:
- [ ] Qwen 72b: was not tested since it was too big.
- [ ] Implement Qwen's log scaling windowed attention. Currently turned off.
- [ ] Implement Qwen's dynamic NTK-aware interpolation. Currently turned off.
- [ ] Test all quantized versions of Qwen: This could be out of scope for this PR.

## Type of change
# Checklist:


transformer_lens/utils.py
```diff
@@ -1083,7 +1083,10 @@ def get_tokenizer_with_bos(tokenizer):
         tokenizer_with_bos = tokenizer
     else:
         tokenizer_with_bos = AutoTokenizer.from_pretrained(
-            pretrained_model_name_or_path, add_bos_token=True, **init_kwargs
+            pretrained_model_name_or_path,
+            add_bos_token=True,
+            trust_remote_code=True,  # TODO: trust_remote_code shouldn't always be true
```
neelnanda-io, line 1088: Can you fix this?

transformer_lens/loading_from_pretrained.py
```diff
@@ -758,6 +798,28 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "scale_attn_by_inverse_layer_idx": hf_config.scale_attn_by_inverse_layer_idx,
```
neelnanda-io, line 798: Can you add trust_remote_code: True if model_name contains santacoder else False or something here?

neelnanda-io (changes requested):
Thanks a lot for adding this! This looks good to me, I'm happy to approve once you've resolved the two small comments.

Would you like this merged in now, or once you've finished the TODOs?

transformer_lens/loading_from_pretrained.py
Aaquib111: done

transformer_lens/utils.py
Aaquib111: done

transformer_lens/loading_from_pretrained.py
```diff
@@ -138,6 +138,22 @@
...
     "bigcode/santacoder",
+    "Qwen/Qwen-1_8B",
+    "Qwen/Qwen-7B",
+    "Qwen/Qwen-14B",
+    # "Qwen/Qwen-72B",
+    "Qwen/Qwen-1_8B-Chat",
+    "Qwen/Qwen-7B-Chat",
+    "Qwen/Qwen-14B-Chat",
+    # "Qwen/Qwen-72B-Chat",
+    # "Qwen/Qwen-1_8B-Chat-Int4",
```
andyrdt, line 149: nit: can we clean up these quantized versions

Aaquib111: done

neelnanda-io (approved):

neelnanda-io:
Thanks a lot for this!

wesg52:
Is there a reason you guys use float16 in your demo?

According to HF the model uses bfloat16?

https://huggingface.co/Qwen/Qwen-7B-Chat?show_tensors=true

andyrdt:
> Is there a reason you guys use float16 in your demo?
> 
> According to HF the model uses bfloat16?
> 
> https://huggingface.co/Qwen/Qwen-7B-Chat?show_tensors=true

Good point, I don't think we thought about it too hard.

When `dtype` is not specified, Qwen has [logic](https://huggingface.co/Qwen/Qwen-7B/blob/main/modeling_qwen.py#L941) to decide which `dtype` to use:
- Use `bf16` if available
- Otherwise use `fp16` if available
- Otherwise use `fp32`

`bf16` isn't available on all devices - on the particular device I was just checking this on (RTX 8000), `torch.cuda.is_bf16_supported()` returns `False`, and so Qwen converts to `fp16` automatically:
```
The model is automatically converting to fp16 for faster inference. If you want to disable the automatic precision, please manually add bf16/fp16/fp32=True to "AutoModelForCausalLM.from_pretrained".
```

I'd probably recommend using `bf16` if available, else `fp16` (in accordance with the Qwen logic).

wesg52:
Makes sense. So long as you aren't getting nan's it shouldn't matter.

Issue #472, opened 2023-12-23
[Bug Report] `ActivationCache.apply_ln_to_stack` silently returns the resid stack for unsupported normalization types


jmsdao:
https://github.com/neelnanda-io/TransformerLens/blob/ce82675a8e89b6d5e6229a89620c843c794f3b04/transformer_lens/ActivationCache.py#L991-L993

`Meta and the RMSnorm gang has entered the chat`

gaabrielfranco:
I am having the same issue. It should be supported by #489 (adding RMSPre in the list ["LN", "LNPre"])

Issue #372, opened 2023-08-26
[Proposal] Implement LayerNorm folding for RMS Norm, to cover LLaMA


neelnanda-io:
### Proposal 

There are two types of LayerNorm: classic LayerNorm that centers to mean zero, scales to standard deviation 1, and then applies learned translation and scaling weights, and RMS norm, which just scales to std 1 and then applies learned scaling weights. Currently from_pretrained only implements classic LayerNorm folding. LLaMA uses RMS Norm which is not covered by this, and it would be nice to do the parts of LayerNorm folding that work for RMS Norm (ie mapping the weights W afterwards to ln.w[:, None] * W)

neelnanda-io:
Note that this should be easy, you basically just need to change some of the flags in `HookedTransformer.load_and_process_state_dict` to allow folding weights for RMS Norm (but not centering!) and edit `HookedTransformer.fold_layer_norm` to only do the bias folding if classic layernorm is used. The model config parameter normalization_type tells you what to use.

neelnanda-io:
Note: Because LLaMA has gated neurons, you'll also need to add support for folding ln2.w into W_gate. This is a bug in the current LN Folding set up, but hasn't triggered because the only models with gated neurons have RMS Norm.

neelnanda-io:
For people who want a temporary work-around, the following code is sufficient

```
for layer in range(n_layers):
    model.blocks[layer].attn.W_K[:] = model.blocks[layer].attn.W_K * model.blocks[layer].ln1.w[None, :, None]
    model.blocks[layer].attn.W_Q[:] = model.blocks[layer].attn.W_Q * model.blocks[layer].ln1.w[None, :, None]
    model.blocks[layer].attn.W_V[:] = model.blocks[layer].attn.W_V * model.blocks[layer].ln1.w[None, :, None]
    model.blocks[layer].ln1.w[:] = torch.ones_like(model.blocks[layer].ln1.w)
    model.blocks[layer].mlp.W_in[:] = model.blocks[layer].mlp.W_in * model.blocks[layer].ln2.w[:, None]
    model.blocks[layer].mlp.W_gate[:] = model.blocks[layer].mlp.W_gate * model.blocks[layer].ln2.w[:, None]
    model.blocks[layer].ln2.w[:] = torch.ones_like(model.blocks[layer].ln2.w)

model.unembed.W_U[:] = model.unembed.W_U * model.ln_final.w[:, None]
[...]

PR #489, opened 2024-01-22
Implement RMS Layer Norm folding

collingray:
# Description

Adds support for folding RMS Norm layers, currently only used by Llama and Mistral. 

Includes a test for checking that RMS/Layer norm folding does not substantially change outputs. 

However, I haven't integrated this test with CI, since (afaik) the smallest models using RMS norm are both 7B, and so it would slow things down quite a bit. (If I'm underestimating the machine running CI, let me know and I'll plug in the test)

See #372 for discussion

## Type of change

# Checklist:


tests/acceptance/test_hooked_transformer.py
```diff
@@ -167,6 +167,66 @@ def test_from_pretrained_revision():
...
+            device = hf_model.device
+        else:
+            device = "cuda" if torch.cuda.is_available() else "cpu"
+    if dtype is None:
+        if hf_model:
+            dtype = hf_model.dtype
+        else:
+            dtype = "float32"
+
+    folded = HookedTransformer.from_pretrained(
```
neelnanda-io, line 197: Nit: Can you call this folded_model?

tests/acceptance/test_hooked_transformer.py
```diff
@@ -167,6 +167,66 @@ def test_from_pretrained_revision():
...
+        model_name=model_name,
+        hf_model=hf_model,
+        device=device,
+        tokenizer=tokenizer,
+        dtype=dtype,
+        fold_ln=True,
+    )
+    tokens = folded.tokenizer.encode(prompt, return_tensors="pt")
```
neelnanda-io, line 205: Nit: Can you use folded.to_tokens, not folded.tokenizer.encode? It's the same thing, but I prefer tests to illustrate library best practices

tests/acceptance/test_hooked_transformer.py
```diff
@@ -167,6 +167,66 @@ def test_from_pretrained_revision():
...
+        device=device,
+        tokenizer=tokenizer,
+        dtype=dtype,
+        fold_ln=False,
+    )
+    unfolded_logits = unfolded(tokens).detach()
+    del unfolded
+    torch.cuda.empty_cache()
+
+    return torch.max(
```
neelnanda-io, line 222: Nit: Can you instead use torch.allclose? As in `check_similarity_with_hf_model`

transformer_lens/HookedTransformer.py
```diff
@@ -1462,25 +1469,41 @@ def fold_layer_norm(self, state_dict: Dict[str, torch.Tensor]):
 
         Args:
             state_dict (Dict[str, torch.Tensor]): State dict of pretrained model.
+            :param fold_biases: Enables folding of LN biases. Should be disabled when RMS Norm is used.
```
neelnanda-io, line 1472: Nit: Can you remove the :param, for consistency?

transformer_lens/HookedTransformer.py
```diff
@@ -1462,25 +1469,41 @@ def fold_layer_norm(self, state_dict: Dict[str, torch.Tensor]):
...
-                * state_dict[f"blocks.{l}.ln1.b"][None, :, None]
-            ).sum(-2)
+            if fold_biases:
+                state_dict[f"blocks.{l}.attn.b_Q"] = state_dict[
+                    f"blocks.{l}.attn.b_Q"
+                ] + (
+                    state_dict[f"blocks.{l}.attn.W_Q"]
+                    * state_dict[f"blocks.{l}.ln1.b"][None, :, None]
+                ).sum(
+                    -2
```
neelnanda-io, line 1488: Nit: In my opinion putting the -2 and ) on separate lines is excessive and harms readability

transformer_lens/HookedTransformer.py
```diff
@@ -1462,25 +1469,41 @@ def fold_layer_norm(self, state_dict: Dict[str, torch.Tensor]):
...
-            ).sum(-2)
-            state_dict[f"blocks.{l}.attn.b_K"] = state_dict[f"blocks.{l}.attn.b_K"] + (
-                state_dict[f"blocks.{l}.attn.W_K"]
-                * state_dict[f"blocks.{l}.ln1.b"][None, :, None]
-            ).sum(-2)
-            state_dict[f"blocks.{l}.attn.b_V"] = state_dict[f"blocks.{l}.attn.b_V"] + (
-                state_dict[f"blocks.{l}.attn.W_V"]
-                * state_dict[f"blocks.{l}.ln1.b"][None, :, None]
-            ).sum(-2)
+            if fold_biases:
```
neelnanda-io, line 1481: Can you add a comment saying this is only used with RMS norm?

transformer_lens/HookedTransformer.py
```diff
@@ -1442,18 +1445,22 @@ def load_and_process_state_dict(
...
     def fill_missing_keys(self, state_dict):
         return loading.fill_missing_keys(self, state_dict)
 
-    def fold_layer_norm(self, state_dict: Dict[str, torch.Tensor]):
+    def fold_layer_norm(
+        self, state_dict: Dict[str, torch.Tensor], fold_biases=True, center_weights=True
+    ):
         """Fold Layer Norm.
 
         Takes in a state dict from a pretrained model, formatted to be consistent with
```
neelnanda-io, line 1466: Can you add a comment saying that this is both used for LayerNorm and RMSNorm?

neelnanda-io (changes requested):
LGTM, apart from some minor requested changes! Thanks for doing this :)

neelnanda-io:
> However, I haven't integrated this test with CI, since (afaik) the smallest models using RMS norm are both 7B, and so it would slow things down quite a bit. (If I'm underestimating the machine running CI, let me know and I'll plug in the test)

Our CI cannot deal with 7B lol, good call

tests/acceptance/test_hooked_transformer.py
collingray: I went with the max diff instead of an allclose check since the tolerance varies greatly with the dtype used, float32 is <1e-5 while float16 is ~1e-2, so it seemed useful to be able to check how the tolerance is affected. I can switch it back/add an allclose assertion if this isn't needed though

transformer_lens/HookedTransformer.py
collingray: Agreed, I think this was the formatter, will fix

tests/acceptance/test_hooked_transformer.py
neelnanda-io: Fair enough, I'm happy for you to leave as is then.

neelnanda-io: Can you add a comment explaining?

collingray: Sure, I may also add the allclose check for some tolerance that passes even on fp16, but continue returning the tolerance

transformer_lens/HookedTransformer.py
collingray: added

transformer_lens/HookedTransformer.py
collingray: Ok it looks like the formatter ci check enforces this style, so I'm going to put it back

transformer_lens/HookedTransformer.py
collingray: added to docstring

neelnanda-io (approved):
LGTM, once you get the formatter passing I am happy to merge it in

collingray:
Just discovered some gqa related issues when using it with mistral, should have a fix soon though

collingray:
Layer norm folding now working on mistral, although it requires the change in #490 or ~48gb vram to test

@neelnanda-io good to merge?

neelnanda-io (approved):
LGTM

wesg52:
We should also add RMS, RMSNorm into
```
 if self.model.cfg.normalization_type not in ["LN", "LNPre"]:
        # The model does not use LayerNorm, so we don't need to do anything.
        return residual_stack
```

to fix #472

Issue #479, opened 2024-01-12
[Proposal] Memory efficient causal mask implementation


andyrdt:
### Proposal 

[Relatively minor proposal - considered making it a bug, but it's not *really* a bug.]

In the initialization of each `Attention` module, we [register](https://github.com/neelnanda-io/TransformerLens/blob/ce82675a8e89b6d5e6229a89620c843c794f3b04/transformer_lens/components.py#L440C9-L440C20) a `causal_mask` buffer. This buffer is a boolean tensor of shape `(self.cfg.n_ctx, self.cfg.n_ctx)`.

This is quite inefficient for 2 reasons:
1. Most times, the prompt context length is much smaller than `self.cfg.n_ctx` (which represents the maximum context length).
2. The same buffer is stored for every layer.

This hasn't really been a visible issue so far with models with smallish context lengths. But consider a model like Qwen 72B, which [has](https://huggingface.co/Qwen/Qwen-72B/blob/main/config.json) max context length of 32768 and 80 layers. With the current implementation, there will be a boolean tensor of shape `(32768, 32768)` initialized for each layer, resulting in 32768 * 32768 * 1 byte * 80 layers ~= 86 GB of overhead.

As a temporary fix, we can just cap `n_ctx` on models to be less than some reasonable value (`2048` or `4096`). But I think the ideal solution is just to compute the attention mask on the fly, and have it be the size of the particular context length.

Note that the same inefficiency exists with rotary embeddings (we precompute sin and cos tensors of length `n_ctx`). But it's not nearly as bad since they grow `O(n_ctx)`, whereas the mask grows `O(n_ctx^2)`.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/transformerlens/issues) in the repo (**required**)

bryce13950:
The PR that was originally created for this issue can be picked up again, and we can probably wrap up where @andyrdt was.

arabel1a:
Is that a reason why `n_ctx` is limited (e.g. for llama3.1-8B model to 2048)? And is it why this particular 8B model doesn't fit into 32G in bfloat16?

PR #491, opened 2024-01-24
Add `max_context_length` parameter to `HookedTransformer`

### LINKED ISSUE #490: [Proposal] Change Mistral's config to reduce context size from 32k to 4k
### Proposal 

The current TL config for Mistral sets its context at 32k, which results in ~34gb worth of attention masks being allocated (32768\*32768\*32 layers), putting it out of range for all consumer-grade single-gpu setups. As described in #479, a more memory efficient implementation of attention masks would be ideal, but in the short term it can just be capped at 2048/4096. If the longer context is needed, it could instead be parameterized, but I suspect that 4k is enough, at least for the short term.

I can put up a pr if needed.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/neelnanda-io/transformerlens/issues) in the repo (**required**)

collingray:
# Description

Adds a `max_context_length` parameter to `HookedTransformer.from_pretrained`,  `HookedTransformer.from_pretrained_no_processing` and `loading_from_pretrained.get_pretrained_model_config`, which caps the context length. By default, it is set to `2048` for `from_pretrained`, and `None` for the others (which leaves it uncapped).

Due to attention masks being stored ahead of time, memory usage grows with the square of context length, leading to some models (e.g. mistral) being unable to fit on consumer GPUs. 

Could potentially be breaking if the longer context was relied on.

Fixes #490 

Related to #479 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -1089,6 +1089,7 @@ class HookedTransformer(HookedRootModule):
         fold_value_biases: Optional[bool] = True,
         default_prepend_bos: Optional[bool] = True,
         default_padding_side: Optional[Literal["left", "right"]] = "right",
+        max_context_length: Optional[int] = 2048,
         dtype="float32",
         **from_pretrained_kwargs,
     ) -> "HookedTransformer":
@@ -1219,6 +1220,9 @@ class HookedTransformer(HookedRootModule):
                 other HuggingFace functions when compatible. For some models or arguments it doesn't
                 work, especially for models that are not internally loaded with HuggingFace's
                 from_pretrained (e.g. SoLU models).
+            max_context_length: The maximum context length to use for the model. Defaults to 2048. Can be set to
+                None to use the full context length of the model. Unless a larger context length is needed, it is
+                recommended to use the default value, as longer context lengths are highly memory intensive.
             dtype: What data type to load the model in (also sets the dtype of
                 the HuggingFace model). Set to bfloat16 or float16 if you get out of memory errors when loading
                 the model.
@@ -1260,6 +1264,7 @@ class HookedTransformer(HookedRootModule):
             device=device,
             n_devices=n_devices,
             default_prepend_bos=default_prepend_bos,
+            max_context_length=max_context_length,
             dtype=dtype,
             **from_pretrained_kwargs,
         )
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -1331,6 +1331,7 @@ class HookedTransformer(HookedRootModule):
         dtype=torch.float32,
         default_prepend_bos=True,
         default_padding_side="right",
+        max_context_length=None,
         **from_pretrained_kwargs,
     ):
         """Wrapper for from_pretrained.
@@ -1348,6 +1349,7 @@ class HookedTransformer(HookedRootModule):
             dtype=dtype,
             default_prepend_bos=default_prepend_bos,
             default_padding_side=default_padding_side,
+            max_context_length=max_context_length,
             **from_pretrained_kwargs,
         )
 
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -869,7 +869,7 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "n_heads": hf_config.num_attention_heads,
             "d_mlp": hf_config.intermediate_size // 2,
             "n_layers": hf_config.num_hidden_layers,
-            "n_ctx": 2048,  # Capped bc the actual ctx length is 30k and the attn mask would be too big
+            "n_ctx": hf_config.seq_length,
             "eps": hf_config.layer_norm_epsilon,
             "d_vocab": hf_config.vocab_size,
             "act_fn": "silu",
@@ -942,6 +942,7 @@ def get_pretrained_model_config(
     device: Optional[str] = None,
     n_devices: int = 1,
     default_prepend_bos: bool = True,
+    max_context_length: Optional[int] = None,
     dtype: torch.dtype = torch.float32,
     **kwargs,
 ):
@@ -974,6 +975,8 @@ def get_pretrained_model_config(
             so this empirically seems to give better results. To change the default behavior to False, pass in
             default_prepend_bos=False. Note that you can also locally override the default behavior by passing
             in prepend_bos=True/False when you call a method that processes the input string.
+        max_context_length (int, optional): The maximum value of `n_ctx` to use for the model.
+            If None, `n_ctx` is left uncapped. Defaults to None.
         dtype (torch.dtype, optional): The dtype to load the TransformerLens model in.
         kwargs: Other optional arguments passed to HuggingFace's from_pretrained.
             Also given to other HuggingFace functions when compatible.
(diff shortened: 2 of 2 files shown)


transformer_lens/loading_from_pretrained.py
```diff
@@ -942,6 +942,7 @@ def get_pretrained_model_config(
     device: Optional[str] = None,
     n_devices: int = 1,
     default_prepend_bos: bool = True,
+    max_context_length: Optional[int] = None,
```
collingray, line 945: I'm not positive if this is the correct default to use here.

My thinking is that "get_pretrained_model_config" implies that it will return the config mostly unaltered

neelnanda-io: I would call it `override_default_context_length` maybe? That has the problem of sounding like a Boolean, but combined with the type being Optional[int] I think it's pretty obvious what's going on

collingray: My reasoning for max vs. override was that max makes it clear that the context length will only be changed if its too big, whereas there is some ambiguity about whether override would always set it to a fixed value.

transformer_lens/HookedTransformer.py
```diff
@@ -1219,6 +1220,9 @@ def from_pretrained(
                 other HuggingFace functions when compatible. For some models or arguments it doesn't
                 work, especially for models that are not internally loaded with HuggingFace's
                 from_pretrained (e.g. SoLU models).
+            max_context_length: The maximum context length to use for the model. Defaults to 2048. Can be set to
```
neelnanda-io, line 1223: This is a complex parameter, so I think the docstring should be more detailed. I would rephrase the comment to something like:
> This allows us to override the TransformerLens default max context length for the model. TransformerLens may default to a smaller max context length than a model was trained with, you can use this parameter to restore a higher max if needed for your use case. Note that this will only work with positional encodings like rotary and sinusoidal that support multiple max context lengths, it will not work with absolute position encodings. If you set the max context to be larger than what the model was trained with, it is unlikely that it can make full use of the context.
For example, Mistral 7B was trained with a 32K context length, but TransformerLens defaults to 2K, because each attention layer has a n_ctx x n_ctx attention mask attached, which can get very memory intensive. If you need the full context, you can set this to a higher value.

transformer_lens/loading_from_pretrained.py
```diff
@@ -869,7 +869,7 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "n_heads": hf_config.num_attention_heads,
             "d_mlp": hf_config.intermediate_size // 2,
             "n_layers": hf_config.num_hidden_layers,
-            "n_ctx": 2048,  # Capped bc the actual ctx length is 30k and the attn mask would be too big
+            "n_ctx": hf_config.seq_length,
```
neelnanda-io, line 872: Why did you change this?

transformer_lens/loading_from_pretrained.py
collingray: perhaps `override_max_context_length` would actually be best

transformer_lens/loading_from_pretrained.py
```diff
@@ -1016,6 +1019,10 @@ def get_pretrained_model_config(
 
     cfg_dict["dtype"] = dtype
 
+    # If max_context_length is specified, use it to cap n_ctx
+    if max_context_length and "n_ctx" in cfg_dict:
+        cfg_dict["n_ctx"] = min(cfg_dict["n_ctx"], max_context_length)
```
neelnanda-io, line 1024: IMO if the user sets max_context_length higher than n_ctx, we should allow it to increase n_ctx

neelnanda-io (changes requested):
Overall looks good, I added some minor comments

transformer_lens/loading_from_pretrained.py
neelnanda-io: And why didn't you change the Mistral entry?

neelnanda-io:
BTW, thanks a lot for all the PRs recently!

transformer_lens/loading_from_pretrained.py
collingray: My thinking was to separate the parsing of the configs from the filtering of them. Its functionally the same, the filter is just applied at a different stage

transformer_lens/loading_from_pretrained.py
collingray: Wolud this be possible, assuming that n_ctx is already the maximum context length of the model?

neelnanda-io: Yeah, rotary models allow you to increase n_ctx beyond what a model was trained with, it may just go completely off the rails. Generally what happens, I think, is that it works fine, but doesn't get any better at predicting the 64Kth token than the 32Kth token (but is better at predicting 32Kth than 16Kth)

transformer_lens/loading_from_pretrained.py
neelnanda-io: Hmm. I lean towards doing it here? My reasoning is that, if someone wants to understand the Qwen config, they can just come here and read this block of code. If there's later code that adjusts it, that's hard to notice, and may be actively misleading, because they'll see 32K here and expect that to end up in the model.

A counter-argument is that adding an "if n_ctx > 4K and override_max_ctx is None: n_ctx=4K" statement can be done once and work for all present and future models, rather than needing people to notice each time. But on net I still prefer all the "understanding Qwen config" code to be in one place

transformer_lens/loading_from_pretrained.py
neelnanda-io: `override_max_context_length` sounds good to me

transformer_lens/loading_from_pretrained.py
collingray: It may actually make sense to split it into multiple parameters for this (e.g. `override_max_context_length`, `override_min_context_length`) in order to still use a default maximum n_ctx, since if it always sets it directly to the provided value then smaller models will get raised up to that n_ctx without this being clear to the user

neelnanda-io: I don't think it's worth having two parameters for such a subtle distinction. IMO if you use this parameter you should know what you're doing. Maybe have it output a warning?

collingray:
> BTW, thanks a lot for all the PRs recently!

I'm glad to help, its a very useful library

Also lmk if there are any other particular issues that would be good to solve

transformer_lens/loading_from_pretrained.py
collingray: I had set the default for it at 2048 in HookedTransformer.from_pretrained, the thinking being that the people using larger contexts likely know the library better, but new users may get tripped up trying to use mistral and having to change a flag to get it to load

A warning for when n_ctx is being extended is a good idea though

transformer_lens/loading_from_pretrained.py
collingray: Thats a fair point, another option would be to add a comment mentioning that it may be overridden by get_pretrained_model_config

I'm fine with either, lmk which you prefer

collingray: Oh I forgot the other reason I changed it, which is that someone using the package not from source would have to directly modify the config in order to change n_ctx

collingray:
Closing this in favor of #493, its perhaps worth adding the option to extend the context length, but there seems to be no need to shorten it given how little memory it should use

PR #495, opened 2024-01-25
Cap Mistral's context length at 2k

### LINKED ISSUE #490: [Proposal] Change Mistral's config to reduce context size from 32k to 4k

collingray:
# Description

Caps Mistral's context length at 2k, as its default length of 32k uses a very large amount of memory.

Discussed in #490 and #491

This change will not be needed after #493, and is just to allow for Mistral to be easily used in the short term while #493 is in review. 

Fixes #490 

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -815,7 +815,7 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "n_heads": 32,
             "d_mlp": 14336,
             "n_layers": 32,
-            "n_ctx": 2048,
+            "n_ctx": 2048, # Capped due to memory issues
             "d_vocab": 32000,
             "act_fn": "silu",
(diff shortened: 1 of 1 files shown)


alan-cooney (approved):
Yep thanks makes sense - approving now and we'll revert once we have the full fix

PR #501, opened 2024-01-30
[Draft] Support Flash Attention

### LINKED ISSUE #378: [Proposal] Optionally use flash attention.
It would be nice to have a flag to enable flash attention in models where that would make sense. This is helpful for performance and memory usage in larger models. In my case working with Pythia 12B, I get ~50% better performance and ~4x larger batch sizes when using flash attention. I also find numerical stability in float16 to be better using flash attention, probably because the model was trained using flash attention.

The downside of using flash attention in TransformerLens is that the we would not have access to intermediate quantities in the attention calculation like the attention matrix itself. This is why I would suggest having a default-off flag so that users can choose whether they need those intermediate values to be available. In addition, when only a small subset of attention intermediates are needed, it's much faster to just cache the input to the attention layer and then recompute those intermediates when needed. 

Thanks!

cmathw:
# Description

Resolves #378 by adding support for `torch.nn.functional.scaled_dot_product_attention` as found [here](https://pytorch.org/docs/master/generated/torch.nn.functional.scaled_dot_product_attention.html). This implementation includes FlashAttention-2, as well as, two other alternative (potentially faster) attention implementations. PyTorch attempts to automatically select the most optimal implementation based on inputs. Thank you @alan-cooney for recommending this implementation!

Currently still in draft because the tolerances between model that uses a fast attention implementation and one that does not are a bit high. This is likely due to the fact `scaled_dot_product_attention` requires casting to float16 (which we then cast back to float32 after doing the fused attention). I will look into this further though and see if there is an improvement to be made here.

## Type of change

# Checklist:


bryce13950:
Is there any update on this, and any idea on what can be done to get the tolerance to an acceptable level?

cmathw:
> Is there any update on this, and any idea on what can be done to get the tolerance to an acceptable level?

I haven't taken a further look yet, is this something currently blocking other features?

bryce13950:
Nope, I have just been going through PRs and closing out anything that can be closed, and helping get anything else closed out. If you need some help with this, let me know. If I have time to help you out, I would be happy to.

winglian:
fails with llama-3

```
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/transformer_lens/components.py", line 591, in forward
    z = self.calculate_z_with_sdpa(q, k, v)  # [batch, pos, head_index, d_head]
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/transformer_lens/components.py", line 785, in calculate_z_with_sdpa
    z = F.scaled_dot_product_attention(query, key, value, is_causal=True)
RuntimeError: The size of tensor a (32) must match the size of tensor b (8) at non-singleton dimension 1
```

jlarson4:
The bridge handles this through HF's attn_implementation="sdpa" or "flash_attention_2". For users wanting fast attention specifically, the path forward is migrating to `TransformerBridge`. Closing

PR #505, opened 2024-02-05
Refactor hook_points

VasilGeorgiev39:
# Description

As I was extending get_caching_hooks for #465 I had to keep adding duplicate code to save_hook() and save_hook_back() so I combined them into the same function. Similarly I also removed some duplicate code in add_hook()

## Type of change

- [x] Refactor (does not change functionality)

# Checklist:


bryce13950:
@VasilGeorgiev39 There has been quite a bit of action in this file since you submitted your PR. I am in the middle of getting it up to date with the main branch. The actual merge has turned into a very manual process due to this having been touched so much, but when I am done with that, if you have time to give it a once over, I can then get this in next.

bryce13950:
@VasilGeorgiev39 Alright, so it should be all up to date. After going through it in detail, it looks like the changes you did on the three different blocks are now only relevant to the first two. The third one looks to already have what you did there. I am not sure if you did this or not in another PR, but what is on that third block is very similar to what you had, with some changes made to the logic. The two blocks you refactored have been restored, and I think they are all ready to merge, but I will wait for you to take a look to double check that I didn't miss anything.

bryce13950 (approved):
I went ahead and double checked it, and I think we are all good to merge. Thanks for making this change!

VasilGeorgiev39:
Thanks for merging! I also just checked it and seems good to me too. Actually it seemed to me that initially I made a mistake in add_hook where in the 'prepend' case I was always adding the hook to the forward hooks but you have fixed this, thanks. I am actually surprised that no tests caught this?

bryce13950:
@VasilGeorgiev39 One of my pending task on getting this project up to date is to run a code coverage report, and add publishing it to the GitHub actions flow. I have not run a code coverage report on this project in probably 8 months, so I cannot tell you definitively what is going on in that respect. I would guess that either the block is not covered at all, or the coverage is suboptimal. If you want to write a test that would improve that coverage, it would be greatly appreciated. In either case, I will open an issue to note this missing coverage.

PR #507, opened 2024-02-08
add qwen1.5 models

### LINKED ISSUE #506: [Proposal] Add support for Qwen1.5 models
### Proposal 

Add support for the recently released Qwen1.5 models ([link](https://qwenlm.github.io/blog/qwen1.5/)).

Should include base and chat models for sizes 0.5B, 1.8B, 4B, 7B, 14B, and 72B.

Their announcement does not mention any architectural changes, so I think this change should be quick/straightforward, but is worth double checking.

### Additional context

See #477 for original Qwen PR.

### Checklist

andyrdt:
# Description

Adding support for 10 new [Qwen1.5 models](https://huggingface.co/collections/Qwen/qwen15-65c0a2f577b1ecb76d786524):
- Qwen/Qwen1.5-0.5B
- Qwen/Qwen1.5-0.5B-Chat
- Qwen/Qwen1.5-1.8B
- Qwen/Qwen1.5-1.8B-Chat
- Qwen/Qwen1.5-4B
- Qwen/Qwen1.5-4B-Chat
- Qwen/Qwen1.5-7B
- Qwen/Qwen1.5-7B-Chat
- Qwen/Qwen1.5-14B
- Qwen/Qwen1.5-14B-Chat

Fixes #506 

I updated the existing `Qwen.ipynb` to demonstrate loading the original generation and loading the new generation.

I also verified that the following holds for all models (except for 14B where I had to run in half precision, and there it holds for `atol=1e-2`):
```
torch.allclose(torch.softmax(hf_logits, dim=-1), torch.softmax(tl_logits, dim=-1), atol=1e-3)
```

I didn't include the 72B models in this PR, as I was not able to test them.


transformer_lens/loading_from_pretrained.py
```diff
@@ -896,6 +916,28 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "final_rms": True,
             "gated_mlp": True,
         }
+    elif architecture == "Qwen2ForCausalLM":
```
wesg52, line 919: Presumably there will be a Qwen 2.0 as well as we should maybe reserve this namespace for those models.

andyrdt: Oh this isn't a namespace that I up with - this is actually the name of the HuggingFace model class. See the following:
- [`config.json`](https://huggingface.co/Qwen/Qwen1.5-0.5B/blob/main/config.json#L3)
- [`Qwen2ForCausalLM`](https://github.com/huggingface/transformers/blob/d123e661e44225bc30cfaf6691daba9691d04153/src/transformers/models/qwen2/modeling_qwen2.py#L1100C1-L1100C46) class definition in the [`transformers`](https://github.com/huggingface/transformers) library

I was confused to see the Qwen1.5 models as instances of `Qwen2ForCausalLM`, but I guess they implemented `Qwen2ForCausalLM` with a Qwen2.0 release in mind. Given this, I anticipate Qwen2.0 to use the same architecture class. If there are subtle edge cases (something like introducing grouped query attention), I think this can be handled in the future PR that adds support for Qwen2.0 models.

neelnanda-io:
It wouldn't surprise me if these were going to be called Qwen 2 but they
changed their mind at the last minute, and couldn't be bothered to edit the
code lol

transformer_lens/loading_from_pretrained.py
```diff
@@ -1665,6 +1709,79 @@ def convert_qwen_weights(qwen, cfg: HookedTransformerConfig):
     return state_dict
 
 
+def convert_qwen2_weights(qwen, cfg: HookedTransformerConfig):
```
neelnanda-io, line 1712: Can you rename this to qwen1_5? I know Qwen2 this is consistent with HuggingFace, but IMO it's best to be consistent with the actual model names a user would recognise

transformer_lens/loading_from_pretrained.py
neelnanda-io: Can you add a comment clarifying?

neelnanda-io:
Added some minor comments, otherwise LGTM! Happy to merge once those are addressed

transformer_lens/loading_from_pretrained.py
andyrdt: done

wesg52:
Can I bump this? (Would be nice to have a really small Qwen model for dev purposes)

PR #512, opened 2024-02-22
Add `@overload` to `FactoredMatrix.__{,r}matmul__`

JasonGross:
# Description

This should `FactoredMatrix.__matmul__` to be used in `functools.reduce` without triggering type errors.

## Type of change

# Checklist:

- [ ] My changes generate no new warnings (does CI test this?)
- [ ] I have added tests that prove my fix is effective or that my feature works (is this necessary for a small change like this?)
- [ ] New and existing unit tests pass locally with my changes (have not tried this, I'm relying on CI, but also, this is just a typing annotation change)


jbloomAus:
Thanks!

PR #516, opened 2024-02-29
chore: fixing type errors and enabling mypy

### LINKED ISSUE #130: Fix type checking (mypy)

chanind:
# Description

This PR enables MyPy checking for the codebase in CI, and fixes all type errors preventing this from being enabled. https://github.com/neelnanda-io/TransformerLens/issues/130 implies that MyPy checking has been turned on, but this is not the case. In Github actions, there is a commented-out mypy check. It appears this has been commented out from the start: https://github.com/neelnanda-io/TransformerLens/pull/123/files#diff-4af11422a4987e947e5a47adead7a30d32cdb2db82e2d3fe36f8e6cbe84d5ac5R40.

This PR should not change any existing behavior. There were some places where the types are broken where I didn't attempt to fix them, as this could require code changes which I wasn't comfortable making, or were simply too involved. I left `TODO` notes for those places where we might want to investigate the typing in the future to see if there's an underlying bug being hidden by the typing, or where a larger refactor is required.

Some issues with the code style which make typing difficult, and should be fixed later:
- methods that return different types of things depending on a flag parameter (this should be avoided IMO)
- configs listing things as optional that the code later assumes are present
- redefining the type of a variable multiple times in a block of code (e.g. x starts as a list, then is a tensor, then an int, etc..., this makes it hard to follow what's going on even for a human)

Fixes #130

## Type of change

# Checklist:


bryce13950:
I just merged the most recent main branch into this PR to try and get it merged. It looks like since you made your changes, some of the newer changes are now causing this to fail https://github.com/neelnanda-io/TransformerLens/actions/runs/8576552819/job/23507706369. If you have time to resolve these issues, then I think this is a no brainer to get merged as soon as these errors are resolved. Some of the changes could be caused by the merge, but I am not quite sure. There were quite a bit of changes to a few of the files you had touched, so there could be a number of reasons why there are new typing errors.

I am in the process of reviewing all pending PRs to see what the current status is. If you don't find time to address this by the time I am done checking out existing PRs, then when I get done with all of the other reviews, I will come back to this and resolve the typing errors in order to get this into the main branch. If you do have time, then that would help out a lot to get this into main. If you do have time, make sure you merge main into your branch. Right now it is up to date, but there may be more changes put into main relatively quickly.

chanind:
Amazing, I'll take another pass at it this weekend. Thanks for looking at the PR!

chanind:
@bryce13950 should be good to go

bryce13950:
Thanks very much for this!

PR #525, opened 2024-03-25
Add support for Llama-2-70b-chat-hf

sheikheddy:
The main changes are:

1. We check if GQA is being used by checking if `cfg.n_key_value_heads` is not None.
2. We introduce a variable `gqa_uscore` in `convert_llama_weights()` which is set to "_" if GQA is used, and an empty string otherwise. This is used to modify the state dict keys for the key and value weights and biases.
3. When rearranging the key and value projection weights (`W_K` and `W_V`), we use `cfg.n_key_value_heads` if GQA is used, otherwise we use `cfg.n_heads`.
4. We modify the state dict keys for the key and value weights and biases by appending `gqa_uscore` to the key names.
5. When creating the key and value biases (`b_K` and `b_V`), we use `cfg.n_key_value_heads` if GQA is used, otherwise we use `cfg.n_heads`.

These modifications ensure that the key and value projection weights and biases are loaded correctly when GQA is used in the Llama-2-70b model.

### DIFF
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -513,6 +513,7 @@ MODEL_ALIASES = {
         "CodeLlama-7b-instruct",
         "codellama/CodeLlama-7b-Instruct-hf",
     ],
+    "Baidicoot/Othello-GPT-Transformer-Lens": ["othello-gpt"],
     "roneneldan/TinyStories-1M": ["tiny-stories-1M"],
     "roneneldan/TinyStories-3M": ["tiny-stories-3M"],
     "roneneldan/TinyStories-8M": ["tiny-stories-8M"],
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -115,6 +115,7 @@ OFFICIAL_MODEL_NAMES = [
     "meta-llama/Llama-2-7b-chat-hf",
     "meta-llama/Llama-2-13b-hf",
     "meta-llama/Llama-2-13b-chat-hf",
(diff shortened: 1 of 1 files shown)


jbloomAus:
Looks good! Thanks!

transformer_lens/loading_from_pretrained.py
```diff
@@ -1652,20 +1675,30 @@ def convert_llama_weights(llama, cfg: HookedTransformerConfig):
         W_K = llama.model.layers[l].self_attn.k_proj.weight
         W_V = llama.model.layers[l].self_attn.v_proj.weight
         W_Q = einops.rearrange(W_Q, "(n h) m->n m h", n=cfg.n_heads)
-        W_K = einops.rearrange(W_K, "(n h) m->n m h", n=cfg.n_heads)
-        W_V = einops.rearrange(W_V, "(n h) m->n m h", n=cfg.n_heads)
+        W_K = einops.rearrange(
+            W_K, "(n h) m->n m h", n=cfg.n_key_value_heads if using_gqa else cfg.n_heads
+        )
+        W_V = einops.rearrange(
+            W_V, "(n h) m-n m h", n=cfg.n_key_value_heads if using_gqa else cfg.n_heads
```
collingray, line 1682: This einops pattern is missing the arrow, it should be "(n h) m->n m h"

collingray:
@jbloomAus I think this should be reverted so that the einops error can be fixed

transformer_lens/loading_from_pretrained.py
jbloomAus: I'll just fix it with a PR. thanks for catching!

PR #494, opened 2024-01-25
Add Support for Yi-6B and Yi-34B

collingray:
# Description

Adds support for 01 AI's Yi-6B and Yi-34B. Their architecture is almost identical to Llama's with the exception of using grouped query attention, but since that has been implemented in #443, Yi is a pretty straightforward addition.

Yi-6B, Yi-6B-Chat and Yi-34B are all tested and working, with the exception of some outlier logits with strangely high error rates (especially the chat model, as shown in the chart below). As far as I can tell these don't seem to be a particular error with my implementation, as the standard deviation is average compared to other models I've tested, but I can't say for sure.

Addresses #449 

Absolute error of logits as compared to the outputs produced with the default HF model. All were calculated from the same set of four sequences, and had all processing options enabled (ln/value bias folding and writing weights/unembed centering). I also experimented with all processing options disabled as well as each processing option individually enabled, but no configuration produced any meaningful differences.

| model   | dtype    |    max |        mean |      median |         std |
|:--------|:---------------------|----------:|------------:|------------:|------------:|
| yi-6b   | bfloat16 | 0.0530524 | 1.1068e-06  | 1.60625e-09 | 9.91611e-05 |
| yi-6b   | float16 | 0.0548902 | 6.77286e-07 | 1.0312e-08  | 7.42205e-05 |
| yi-6b-chat | bfloat16 | 0.0875687 | 1.29017e-06 | 1.24604e-09 | 0.000172055 |
| yi-6b-chat | float16 | 0.0699415 | 1.00371e-06 | 9.40543e-09 | 0.000155543 |
| yi-34B  | bfloat16  | 0.0307519 | 8.3753e-07  | 1.35014e-09 | 7.06597e-05 |

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #529, opened 2024-03-28
Update loading_from_pretrained.py

jbloomAus:
fix missing arrow in PR #525 Thanks @collingray


sheikheddy:
Sorry about missing the arrow. @collingray is there a way to catch that in tests?

collingray:
> Sorry about missing the arrow. @collingray is there a way to catch that in tests?

It happens. A test which loaded the model would have caught it, but since its only the 70B model which uses GQA, such a test couldn't be run on most machines/ci.

A more general piece of advice that would have prevented it: it looks like that code was pulled from my yi pr (#494), but somehow the `>` didn't get copied. When possible you should use git checkout/cherry-pick for things like that, since it prevents transcription errors

PR #534, opened 2024-04-03
Bugfix: remove redundant assert checks

tkukurin:
# Description

The `isinstance` check fails on non-`Slice` input: `TypeError: Subscripted generics cannot be used with class and instance checks`.
Since the check seems redundant (same thing is repeated in `Slice` constructor), removing should be fine.

## Type of change


neelnanda-io:
LGTM, thanks!

PR #536, opened 2024-04-08
HookedSAETransformer

ckkissane:
# Description

Implements HookedSAETransformer - an extension of HookedTransformer that enables attachment of SAEs to activations. I implemented this for MATS research, and found it useful, so I'm excited to open source it. Feedback is welcome!

Here's a demo notebook: https://colab.research.google.com/github/ckkissane/TransformerLens/blob/hooked-sae-transformer/demos/HookedSAETransformerDemo.ipynb 

## Type of change

- [ x] New feature (non-breaking change which adds functionality)
- [ x] This change requires a documentation update

# Checklist:

- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] My changes generate no new warnings
- [ x] I have not rewritten tests relating to key interfaces which would affect backward compatibility


transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,95 @@
...
+
+from typing import Union, Dict
+
+
+class HookedSAE(HookedRootModule):
+    """Hooked AutoEncoder.
+    
+    Implements a standard SAE with a TransformerLens hooks for SAE activations
+    
+    Should probably just be used for inference / analysis, not training.
```
ArthurConmy, line 19: Add link to Joseph's codebase for training.

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    This is just a template, feel free to copy this into a notebook and change it!
```
ArthurConmy, line 21: What does this mean? What would you edit? The SAE architecture? I suggest removing the comment

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,95 @@
...
+        """SAE Forward Pass.
+        
+        Args:
+            input: The input tensor of activations to the SAE. Shape [..., d_in]
+        """
+        self.hook_sae_input(input)
+        if input.shape[-1] == self.cfg.d_in:
+            x = input
+        else:
+            x = einops.rearrange(input, "... n_heads d_head -> ... (n_heads d_head)")
```
ArthurConmy, line 58:
```suggestion
                        # Assume this this is an attention output SAE
                        x = einops.rearrange(input, "... n_heads d_head -> ... (n_heads d_head)")
```

ArthurConmy: Also, you should assert that the hook name `endswith` "_z" if you're going to reshape things

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,95 @@
...
+            ) + self.b_enc # [..., d_sae]
+            sae_acts_post_clean = F.relu(sae_acts_pre_clean)
+            x_reconstruct_clean = (einops.einsum(
+                sae_acts_post_clean, self.W_dec,
+                "... d_sae, d_sae d_in -> ... d_in"
+            ) + self.b_dec).reshape(input.shape)
+            
+            error = self.hook_sae_error(input - x_reconstruct_clean)
+  
+        output = self.hook_sae_output(x_reconstruct + error)
```
ArthurConmy, line 90: I think output is a bad name here

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        sae_acts_post = self.hook_sae_acts_post(F.relu(sae_acts_pre)) # [..., d_sae]
+        x_reconstruct = self.hook_sae_recons(
+            (einops.einsum(
+                sae_acts_post, self.W_dec,
+                "... d_sae, d_sae d_in -> ... d_in"
+            ) + self.b_dec).reshape(input.shape)
+        )
+        
+        with torch.no_grad():
```
ArthurConmy, line 76: Can't you just do `x_reconstruct.detach()`?

transformer_lens/HookedSAETransformer.py
```diff
@@ -0,0 +1,221 @@
...
+from transformer_lens import HookedTransformer
+import torch
+from transformer_lens.ActivationCache import ActivationCache
+from transformer_lens.HookedSAE import HookedSAE
+
+SingleLoss = Float[torch.Tensor, ""]  # Type alias for a single element tensor
+LossPerToken = Float[torch.Tensor, "batch pos-1"]
+Loss = Union[SingleLoss, LossPerToken]
+
+def get_deep_attr(obj, path):
```
ArthurConmy, line 16: Add docstring

ArthurConmy (changes requested):
Thanks for getting this PR up!

Will comment more detailed things in a bit but here are first thoughts!

transformer_lens/HookedSAE.py
ckkissane: I removed it. But i do think it's somewhat important that people know they can hack the architecture and still work with HookedSAETransformer. It would be sad if someone couldn't use the library because their SAEs didn't use a pre-encoder bias, a different activation function, etc. I explained this a bit more in the "Attach SAEs" section of the [demo](https://colab.research.google.com/github/ckkissane/TransformerLens/blob/hooked-sae-transformer/demos/HookedSAETransformerDemo.ipynb#scrollTo=t3q9DzPZY6lX)

transformer_lens/HookedSAE.py
ckkissane: Thanks, agreed. I modified so it hooks the SAE output in every case (error term vs no error)

transformer_lens/HookedSAE.py
ckkissane: Wouldn't this stop the gradient flow back to to the feature activations, which is important for attribution patching? I think this section is confusing and would appreciate feedback on ways to handle it better. Currently there are two things going on in this block:
1. Recompute the "true" error term. If we just use error = input - x_reconstruct at the end, then interventions that modifiy x_reconstruct will also effect the error term: error = input - x_recons_modified -> output = x_recons_modified + error = x_recons_modified + (input - x_recons_modified) = input, which means the intervention will have no affect downstream.
2. Detach the error term, for attribution patching on feature activations, as described in A.3 of [Marks et al.](https://arxiv.org/pdf/2403.19647.pdf) (I didn't fully follow all their gradient tricks in detail, so might be missing something)

transformer_lens/HookedSAE.py
jbloomAus: FWIW, I'd love to make my codebase also an inference codebase and build on this. Having this in T-lens seems good but I think will lead to lots of changes to adapt to new SAE methodologies which will create lots of work in T-lens which already has a lot going on. I think have this PR here accepted is probably good but in the long run, there should be other SAE analysis code bases.

ArthurConmy (changes requested):
Sorry to be a pain, but can you also add tests that this all works?

You can use `test_hooks.py` ([link](https://github.com/neelnanda-io/TransformerLens/blob/main/tests/unit/test_hooks.py)) and remove the conditional and permanent tests. But the rest of those tests should basically have an equivalent for HookedSAEs.

The HookedSAE you use in the test can be random for now. Later we should add tests that the SAE works well at reconstructing but we can skip that for now.

transformer_lens/HookedSAETransformer.py
```diff
@@ -0,0 +1,271 @@
...
+        ARgs:
+            *model_args: Positional arguments for the model forward pass
+            act_names: (List[str]) The act_names for the SAEs to turn on for this forward pass
+            fwd_hooks: (List[Tuple[Union[str, Callable], Callable]]) List of forward hooks to apply
+            bwd_hooks: (List[Tuple[Union[str, Callable], Callable]]) List of backward hooks to apply
+            reset_hooks_end: (bool) Whether to reset the hooks at the end of the forward pass (default: True)
+            clear_contexts: (bool) Whether to clear the contexts at the end of the forward pass (default: False)
+            **model_kwargs: Keyword arguments for the model forward pass
+        """
+        self.turn_saes_off()
```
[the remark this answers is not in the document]
ckkissane: IMO the main reason to use run_with_saes is if you just want to run with the specific SAEs you pass in for one forward pass, then have a clean slate. I can't imagine people using it in addition to other SAEs they already have turned on, unless by mistake. What do you think?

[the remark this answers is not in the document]
jbloomAus: that makes sense to me.

ArthurConmy (approved):
Looks great! LGTM after these tiny fixes are made:

1) Could you also edit the README, removing the October survey and adding a line to the effect that the library now supports running SAEs, but we recommend SAELens (built on top of TransformerLens) for training SAEs. Also add the ipynb demo to the list of demos in the README.
2) Edit the demo so that it installs the TransformerLens repo. We'll merge and publish the library soon, so that seems fine.

transformer_lens/HookedSAETransformer.py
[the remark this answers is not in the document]
jbloomAus: I like the idea of a flag, and think it should be by default off. I think that the most important thing here is that the API is intuitive and adding the flag makes this more obvious. FWIW, I share Connor's intuition about how it will be used but can imagine wanting to run with *marginal saes* in some cases (adding / removing SAEs on the margin and not having a reset). This seems more likely in a world with great SAEs (way better than we have now).

ArthurConmy:
> Making it so `model.run_with_saes` resets the SAE states to the same state as before the function was called would be a big improvement

@nix-apollo I suggested this here https://github.com/neelnanda-io/TransformerLens/pull/536#discussion_r1569800199 but deferred to Connor and Joseph who have used the library.

Persistent state seems OK to me? TransformerLens already has add_hook and reset_hooks which are really similar to attach_sae and turn_off_all_saes

nix-apollo:
To me `HookedSAETransformer.attach_sae` seems more analogous to `add_hook`. I like having this version of statefulness! Having an object that represents "GPT2 with this set of SAEs attached" seems useful to me. But being able to represent "GPT2 with these 12 SAEs attached, except only 4 of them are turned on" doesn't seem useful.

But I generally dislike persistent state, and avoid it when possible. This includes avoiding using `add_hook` and `reset_hook` whenever possible in normal TL. Maybe my opinion is too extreme here -- again, do feel free to ignore it!

The flag is probably a fine middle ground. Just imo adds a bunch of complexity for features that are more dangerous than useful.

ArthurConmy (changes requested):
Going to take back my LGTM since Nix brings up some good points I want to resolve.

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self.hook_sae_input = HookPoint()
+        self.hook_sae_acts_pre = HookPoint()
+        self.hook_sae_acts_post = HookPoint()
+        self.hook_sae_recons = HookPoint()
+        self.hook_sae_error = HookPoint()
+        self.hook_sae_output = HookPoint()
+
+        self.to(self.cfg.device)
+
```
neelnanda-io, line 53: You need a self.setup(), otherwise sae.run_with_cache and sae.run_with_hooks don't work. It would also be good to have a test to verify that you can do that, since clearly this is a gap in the tests!

transformer_lens/HookedSAE.py
neelnanda-io: I think it's worth saying this in the code! I would just reword a bit. Maybe:

"Note that HookedSAETransformer is fairly modular, and doesn't make strong assumptions about the architecture of the SAEs that get attached. We provide HookedSAE as a useful default class, but if you want to eg experiment with other SAE architectures, you can just copy the HookedSAE code, edit it, and add instances of the new SAE class to a HookedSAETransformer". Even better if you can enumerate what assumptions it does make.

neelnanda-io:
Thanks for the feedback Nix! I think that there's cases where I want a stateful workflow, eg I have a single residual stream SAE that I want to attach to the model, and then just do normal mech interp to. Or I have a fixed suite of SAEs, and want to include all of them (plus error terms) and do circuit analysis. I also think there's cases where I want an ephemeral stateless workflow, eg if I'm analysing feature splitting and have 8 different SAEs on the same layer's residual stream of different widths, and want to easily swap them out for each other. So it seems clearly good to support both. IMO the best solution to avoid footguns is to have `run_with_saes` return the SAEs to their original state after it runs, and to support both `attach_saes/turn_all_saes_off` and `run_with_saes/run_with_saes_and_cache/run_with_saes_and_hooks`

ArthurConmy:
@neelnanda-io I moved discussion to the Open Source slack about this issue as the interface is better.

TLDR support both permanently attached SAEs with add_sae and transiently attached SAEs with run_with_SAEs matching transformer lens symtax, do not use on/off at all

README.md
```diff
@@ -24,7 +24,7 @@ TransformerLens lets you load in 50+ different open source language models, and
 activations of the model to you. You can cache any internal activation in the model, and add in
 functions to edit, remove or replace these activations as the model runs.
 
-~~ [OCTOBER SURVEY HERE](https://forms.gle/bw7U3PfioacDtFmT8) ~~
+The library also now supports mechanistic interpretability with SAEs (sparse autoencoders)! With [HookedSAETransformer](https://colab.research.google.com/github/ckkissane/TransformerLens/blob/hooked-sae-transformer/demos/HookedSAETransformerDemo.ipynb), you can splice in SAEs during inference and cache + intervene on SAE activations. We recommend [SAELens](https://github.com/jbloomAus/SAELens) (built on top of TransformerLens) for training SAEs.
```
ArthurConmy, line 27: Please now use `github/neelnanda-io/` as we will merge this in.

README.md
```diff
@@ -51,6 +51,7 @@ logits, activations = model.run_with_cache("Hello World")
 * [Introduction to the Library and Mech
   Interp](https://arena-ch1-transformers.streamlit.app/[1.2]_Intro_to_Mech_Interp)
 * [Demo of Main TransformerLens Features](https://neelnanda.io/transformer-lens-demo)
+* [Demo of HookedSAETransformer Features](https://colab.research.google.com/github/ckkissane/TransformerLens/blob/hooked-sae-transformer/demos/HookedSAETransformerDemo.ipynb)
```
ArthurConmy, line 54: Again, use `/github/neelnanda-io/`

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,115 @@
...
+                + self.b_dec
+            ).reshape(input.shape)
+        )
+
+        if self.cfg.use_error_term:
+            with torch.no_grad():
+                # Recompute everything without hooks to get true error term
+                # Otherwise, the output with error term will always equal input, even for causal interventions that affect x_reconstruct
+                # This is in a no_grad context to detach the error, so we can compute SAE feature gradients (eg for attribution patching). See A.3 in https://arxiv.org/pdf/2403.19647.pdf for more detail
+                sae_acts_pre_clean = (
```
ArthurConmy, line 98:
```suggestion
                # NOTE: we can't just use `sae_error = input - x_reconstruct.detach()` or something simpler, since this would mean intervening on features would mean ablating features still results in perfect reconstruction.
                sae_acts_pre_clean = (
```

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,115 @@
...
+            assert self.cfg.hook_name.endswith(
+                "_z"
+            ), f"You passed in an input shape {input.shape} does not match SAE input size {self.cfg.d_in} for hook_name {self.cfg.hook_name}. This is only supported for attn output (hook_z) SAEs."
+            x = einops.rearrange(input, "... n_heads d_head -> ... (n_heads d_head)")
+        assert (
+            x.shape[-1] == self.cfg.d_in
+        ), f"Input shape {x.shape} does not match SAE input size {self.cfg.d_in}"
+
+        x_cent = x - self.b_dec
+        sae_acts_pre = self.hook_sae_acts_pre(
```
ArthurConmy, line 81:
```suggestion
        # WARNING: if editing this block of code, also edit the error computation inside `if self.cfg.use_error_term`
        sae_acts_pre = self.hook_sae_acts_pre(
```

transformer_lens/HookedSAE.py
```diff
@@ -0,0 +1,115 @@
...
+            einops.einsum(x_cent, self.W_enc, "... d_in, d_in d_sae -> ... d_sae")
+            + self.b_enc  # [..., d_sae]
+        )
+        sae_acts_post = self.hook_sae_acts_post(F.relu(sae_acts_pre))  # [..., d_sae]
+        x_reconstruct = self.hook_sae_recons(
+            (
+                einops.einsum(sae_acts_post, self.W_dec, "... d_sae, d_sae d_in -> ... d_in")
+                + self.b_dec
+            ).reshape(input.shape)
+        )
```
ArthurConmy, line 91:
```suggestion
        )
        # END WARNING
```

ArthurConmy (approved):
LGTM with minor comments :rocket:

ArthurConmy (commented):
Oh, also please fix indentation, the linter is not happy about that.

transformer_lens/HookedSAETransformer.py
```diff
@@ -0,0 +1,284 @@
...
+            reset_saes_end (bool): If True, removes all SAEs added by this context manager when the context manager exits, returning previously attached SAEs to their original state.
+
+        Example:
+            ```python
+            hooked_saes = [HookedSAE(...), HookedSAE(...)]
+            with model.saes(saes=hooked_saes):
+                spliced_loss = model(text, return_type="loss")
+            ```
+        """
+        act_names_to_reset = []
```
ArthurConmy, line 272:
```suggestion
    @contextmanager
    def saes(
        self,
        saes: Union[HookedSAE, List[HookedSAE]] = [],
        reset_saes_end: bool = True,
    ):
        """
        A context manager for adding temporary SAEs to the model.
        Attaches specified SAEs temporarily, and restores the previous state at exit.
        
        Args:
            saes (Union[HookedSAE, List[HookedSAE]]): The SAEs to attach temporarily.
            reset_saes_end (bool): If True, restores the SAEs' original state after exiting context.
        
        Example:
        ```python
        hooked_saes = [HookedSAE(...), HookedSAE(...)]
        with model.saes(saes=hooked_saes):
            spliced_loss = model(text, return_type="loss")
        ```
        """
        act_names_to_reset = []
```

ArthurConmy: @ckkissane this is where the error is -- maybe either tab/space spacing is wrong, or it hates the \n here?

ckkissane: For some reason the ```python block was causing issues, so I changed it to .. code-block:: python

PR #537, opened 2024-04-08
Add Xavier and Kaiming Initializations

Chanlaw:
# Description
Update initializations for `HookedTransformer` to support xavier/kaiming uniform/normal. MuP weight inits are supported but not 1/d_head attention or Adam learning rate rescaling (will be added soon). Also I haven't touched `HookedTransformerEncoder`, which seems to require a larger refactor. 

Note that default init is still "gpt2", so we don't break existing functionality. @jbloomAus should we just update the init to Kaiming normal, which I think is the best of the non-MuP initializations? 

The important stuff are in the `utils.py`, `test_utils.py`, and `HookedTransformer.py` files, the rest is mainly lot of small unimportant so the various linters aren't mad.

## Type of change

# Checklist:


jbloomAus:
Happy to update the default initialization strategy. I think ~95% of T-lens users are using pre-trained models anyway. As long as the docs are updated to reflect the default init, I think that's fine.

jbloomAus:
Accepting now and can change default in a follow up.

Chanlaw:
Sounds good, thanks for the quick response!

Yeah, most people are just using it to poke at pretrained 100m-10b models.

My main concern with changing the default is that it might screw over the various algorithmic task tutorials that involving training from scratch (esp the grokking ones, where weight norm at initialization sometimes matter). Happy to include the change when I'm done adding in MuP though.

PR #486, opened 2024-01-21
Loading of huggingface 4-bit quantized Llama

coolvision:
# Description

Adding support for loading HF Llama models, 4-bit quantized with bitsandbytes.
This allows to load Llama2-7b into 5-6G of GPU RAM.
Demo or running in in colab on T4 GPU:
https://github.com/coolvision/TransformerLens/blob/llama_4bit_v2/demos/LLaMA2_GPU_quantized.ipynb

The quantization is triggered by setting `load_in_4bit=True` in huggingface `from_pretrained` method.
Then, quantized model should be passed to TransformerLens `from_pretrained`.

```
LLAMA_2_7B_CHAT_PATH = "meta-llama/Llama-2-7b-chat-hf"
inference_dtype = torch.float32

hf_model = AutoModelForCausalLM.from_pretrained(LLAMA_2_7B_CHAT_PATH,
                                             torch_dtype=inference_dtype,
                                             device_map = "cuda:0",
                                             load_in_4bit=True)

tokenizer = AutoTokenizer.from_pretrained(LLAMA_2_7B_CHAT_PATH)

model = HookedTransformer.from_pretrained(LLAMA_2_7B_CHAT_PATH,
                                             hf_model=hf_model,
                                             dtype=inference_dtype,
                                             fold_ln=False,
                                             fold_value_biases=False,
                                             center_writing_weights=False,
                                             center_unembed=False,
                                             tokenizer=tokenizer)
```

This feature requires `torch >= 2.1.1`, `bitsandbytes` & `accelerate` packages:
```
poetry source add --priority=supplemental torch https://download.pytorch.org/whl/cu118
poetry add torch==2.1.1+cu118 --source torch
pip install bitsandbytes
pip install accelerate
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #547, opened 2024-04-19
Llama 4bit v2 typing

hannamw:
# Description

This change fixes typing problems that caused mypy issues for the 4bit-quanitized Llama PR, as discussed in #486. Concretely, it:
 - changes a use of `cfg.d_mlp` to `self.cfg.d_mlp`, as only the latter's type is known to mypy
 - changes the definition of `W_Q` and `W_O` in `AbstractAttention`, making them `bnb.Params4bit` if `cfg.load_in_4bit` is `True`, and `nn.Parameter` otherwise; in both cases, they're empty. It also removes the `W_Q` and `W_O` re-definitions in `Attention`.

The latter change is because `W_Q` and `W_O` were previously defined as `nn.Parameter` in `AbstractAttention` and then redefined (in `Attention`) as `bnb.Params4bit` if `cfg.load_in_4bit` was `True`. So, methods of `AbstractAttention` expected `nn.Parameter` instead of `bnb.Params4bit`. This change solves that problem, although one could also consider making `W_Q` and `W_O` into abstract attributes as with `W_K` and `W_V` .

Another option would be to just `assert`, where necessary, that relevant weights are `bnb.Params4bit` instead of `nn.Parameter`s, though I'm not sure this would solve the root issue.

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/components.py
+++ b/transformer_lens/components.py
@@ -1022,10 +1022,8 @@ class Attention(AbstractAttention):
         if cfg.load_in_4bit:
             # 4-bit quantization convention
             nq = int((cfg.d_model * cfg.d_model) / 2)
-            self.W_Q = Params4bit(torch.empty(nq, 1, dtype=torch.uint8), requires_grad=False)
             self.W_K = Params4bit(torch.empty(nq, 1, dtype=torch.uint8), requires_grad=False)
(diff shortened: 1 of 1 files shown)


bryce13950:
Thanks! This should be good to go.

Issue #574, opened 2024-05-06
[Bug Report] Bug in `get_caching_hooks` when `pos_slice=None`


hannamw:
**Describe the bug**
If you call `get_caching_hooks` without specifying `pos_slice` (or if you set `pos_slice=None`), and then run the model with these hooks, an `assert` checking that `pos_slice is not None` will be triggered. Looking at how the `pos_slice` is handled in `run_with_cache`, I believe that the correct solution here is to add in a couple lines that say 
```
if not isinstance(pos_slice, Slice):
    if isinstance(
        pos_slice, int
    ):  # slicing with an int collapses the dimension so this stops the pos dimension from collapsing
        pos_slice = [pos_slice]
    pos_slice = Slice(pos_slice)
```
It might also just make sense to move those lines into `get_caching_hooks`. Maybe this is user error on my end—I've been using `get_caching_hooks` and `with model.hooks()` to cache and intervene during the same forward pass, but let me know if I shouldn't be using `get_caching_hooks` for this at all.

**Code example**
```
from transformer_lens import HookedTransformer

model = HookedTransformer.from_pretrained('gpt2')
s = "This is a test sentence."

# This works
_, cache = model.run_with_cache(s)

# This doesn't
cache, hooks, _ = model.get_caching_hooks()
with model.hooks(hooks):
    model(s)
```

This code will break at line 545 (`assert pos_slice is not None  # keep mypy happy`) in `hook_points.py`.

**System Info**
Describe the characteristic of your environment:
 * Describe how `transformer_lens` was installed (pip, docker, source, ...): **pip (1.17.0)**
 * What OS are you using? (Linux, MacOS, Windows): **Linux**
 * Python version (We suppourt 3.7 -3.10 currently): **Python 3.10**

**Additional context**

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

janraasch:
Hi @hannamw,

I ran into the same issue when trying to run the "Get activations"-section 

```python
...
    with model.hooks(fwd_hooks=caching_hooks):
        _ = model(prompt)
    return cache[name]
...
```

of https://colab.research.google.com/drive/1x5EmvAdVmmHvOOd1jxvODj233wbvbl1U#scrollTo=2h6X8zdojKHN which is a notebook referenced in »Activation Addition: Steering Language Models Without Optimization« (see https://arxiv.org/pdf/2308.10248).

[...]

PR #578, opened 2024-05-08
Fix Pos Slice Issue

hannamw:
# Description
There is an issue with `get_caching_hooks` where if `pos_slice` is set to `None` (the default behavior), using these hooks causes cause an assertion error (for details see #574). This is because the code actually expected `pos_slice` to be `Slice` object, even though it is typed as an `Optional[Slice]` . This change makes it so that the `pos_slice` object is converted to a `Slice` if it is not one already; it also updates the typing to indicate that `pos_slice` is a `Slice` or `SliceInput`, as in `run_with_cache`.

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -12,7 +12,7 @@ from typing import Callable, Dict, List, Literal, Optional, Sequence, Tuple, Uni
 import torch.nn as nn
 import torch.utils.hooks as hooks
 
-from transformer_lens.utils import Slice
+from transformer_lens.utils import Slice, SliceInput
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


bryce13950:
Very good! I am very happy to see this change, but I do have a request. Can you setup your bit of logic here on the `Slice` class as a class method named `unwrap` that will take in the type you have defined in `get_caching_hooks`, and then always return a `Slice`? This logic is done in quite a few places, and I would like to start centralizing these blocks of logic so that we can test these things in isolation, and so we can start having these small repeated bits of logic in one place. 

As a bonus, `run_with_cache` is also doing the exact same thing as `get_caching_hooks`, so if you could update that function to be typed properly, and use the new unwrap function, that would be wonderful.

hannamw:
Sure, is this the sort of thing you were thinking of?

bryce13950:
That is perfect.

PR #579, opened 2024-05-08
Add Mistral 7B v0.2 Instruct

fakerybakery:
Prevents error when loading Mistral 7B v0.2 Instruct

### DIFF
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -941,7 +941,7 @@ def convert_hf_model_config(model_name: str, **kwargs):
             "act_fn": "silu",
             "normalization_type": "RMS",
             "positional_embedding_type": "rotary",
-            "window_size": 4096,
+            "window_size": hf_config.sliding_window,  # This will be 4096 on v0.1, None on later models as none was used
             "attn_types": ["local"] * 32,
             "eps": 1e-05,
             "n_key_value_heads": 8,
--- a/transformer_lens/loading_from_pretrained.py
+++ b/transformer_lens/loading_from_pretrained.py
@@ -558,7 +558,7 @@ MODEL_ALIASES = {
     ],
     "mistralai/Mistral-7B-v0.1": ["mistral-7b"],
     "mistralai/Mistral-7B-Instruct-v0.1": ["mistral-7b-instruct"],
-    "mistralai/Mistral-7B-Instruct-v0.2": ["mistral-7b-instruct"],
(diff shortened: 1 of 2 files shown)


transformer_lens/loading_from_pretrained.py
```diff
@@ -557,6 +558,7 @@
     ],
     "mistralai/Mistral-7B-v0.1": ["mistral-7b"],
     "mistralai/Mistral-7B-Instruct-v0.1": ["mistral-7b-instruct"],
+    "mistralai/Mistral-7B-Instruct-v0.2": ["mistral-7b-instruct"],
```
ArthurConmy, line 561: Wait, unless I miss something this means that this CHANGES which model `"mistral-7b-instruct"` points to. This seems unintended behavior -- can the alias be `"mistral-7b-instruct-v0.2"` instead?

ArthurConmy (changes requested):

transformer_lens/loading_from_pretrained.py
fakerybakery: Ah yes, sorry for the oversight. I changed the alias to `mistral-7b-instruct-v0.2` :)

ArthurConmy (approved):
Thanks! We can merge this once you do the following two things:

Add this model here: https://github.com/TransformerLensOrg/TransformerLens/blob/main/docs/source/content/getting_started.md

(CC @bryce13950 is there a way to automate populating that list of gated models? Can you make an issue tomorrow for that?)

Also can you confirm that this model didn't change any of the underlying hparams of Mistral (e.g. sequence length, rotary dim, rms eps)?

fakerybakery:
Sorry, there's one change on 0.2 (it doesn't use the sliding attention window and has a longer context length instead), I'll update the PR shortly

fakerybakery:
@ArthurConmy I updated the docs and the config

bryce13950:
@ArthurConmy there really isn't a way to get that list to repopulate. Once the agreements are accepted, the models behave exactly the same way as if they were any other model. The only way to conceivably do something like that is to run a REST request, and pull the meta information on the model to see if it is gated. That's quite a bit of work for what comes down to a pretty small block in the docs.
I am at the start of revamping the model submission process. One of the pieces of that is to setup a PR template for submitting new models. This situation will be noted in that template to make sure whoever is submitted is informing us of the requirement, and also to make sure they edit this doc.

ArthurConmy:
Thanks Bryce! No need to do a REST thing, thanks for clarifying this.

PR #596, opened 2024-05-15
Update Gemma to reflect upstream HF changes

### LINKED ISSUE #594: [Bug Report] Updates to Gemma
Since support for Gemma was merged [March 14th](https://github.com/TransformerLensOrg/TransformerLens/pull/511), there have been a number of changes to the upstream HF [model file](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/modeling_gemma.py) such that we no longer have good agreement across logits and cache. A large chunk of these changes exist [here](https://github.com/huggingface/transformers/commit/8e2fc52ea3c80d858cc7e7e5f835d4ce795c900d) but I think there are also more recent relevant changes in other commits. I will be opening a corresponding PR to address these issues tomorrow, in the meantime I would not trust Gemma outputs and activations.

### Checklist

cmathw:
# Description
Fixes #594.

Since Gemma was merged in March 14th, there have been a number of upstream HuggingFace changes that mean our activations/logits no longer matched this implementation well. This PR implements the changes described [here](https://github.com/huggingface/transformers/commit/8e2fc52ea3c80d858cc7e7e5f835d4ce795c900d), changes include:
- Changing the activation from standard GeLU to the tanh-approximated GeLU.
- Keeping RMSNorm weights in float32
- Setting the dtype of the embedding scale to match that in the HookedTransformer config.

A demo showing this agreement can be found [here](https://colab.research.google.com/drive/1MkzrjZ29MJY3hh_3-rifBJ2rpukzyWTv?usp=sharing) and summarised below. Note: There is currently a bug in various HF/transformer models where the default attention implementation used is not causally masking patterns, I opened an issue [here](https://github.com/huggingface/transformers/issues/30813) but it is already being addressed [here](https://github.com/huggingface/transformers/pull/30652). For this reason it is important when comparing activations/logits to specify `attn_implementation="eager"` when using HF/transformers' `from_pretrained` method.

Tolerances for both 2b and 7b models when running in float32 can be found below:
|       | 2b | 7b  |
|------------|-----|-------------|
| Logits   | 5e-04  | 5e-03    |
| Resid Pre Cache | 5e-05  | 5e-04    |
| Next Token Loss | 1e-06  | 5e-05   |

Tolerances when running float16 are much worse but I think this is commonly the case when adding models to TL with internal upcasting/downcasting, if this is not the case though and these tolerances should also be similarly close I can investigate this further.

## Type of change

# Checklist:


jbloomAus:
Thanks Chris! looks good.

PR #597, opened 2024-05-15
allow user to force trust_remote_code=true via from_pretrained kwargs

Butanium:
# Description

A simple code change that allows you to force TL to use `trust_remote_code=True` in its configuration. This is particularly useful when you need to use an unreferenced model with a supported architecture that requires it (e.g., m-a-p/CT-LLM-Base).

## Type of change


bryce13950:
Thanks!

Issue #540, opened 2024-04-15
[Bug Report] Test coverage missing on add_hook in hook_points


bryce13950:
Originally noted by @VasilGeorgiev39 while refactoring add_hook functions.

Thanks for merging! I also just checked it and seems good to me too. Actually it seemed to me that initially I made a mistake in add_hook where in the 'prepend' case I was always adding the hook to the forward hooks but you have fixed this, thanks. I am actually surprised that no tests caught this?

_Originally posted by @VasilGeorgiev39 in https://github.com/neelnanda-io/TransformerLens/issues/505#issuecomment-2053758445_
            

**Describe the bug**
The coverage needs to be improved, so that any further changes to this area can be tested properly in order to avoid bugs being introduced into the main branch.

anthonyduong9:
I can work on this today.

bryce13950:
@anthonyduong9 I just did something similar to this on a different part of the same file https://github.com/TransformerLensOrg/TransformerLens/pull/613/files. It was a lot easier to test the bit of code I tested by breaking it down into simpler functions. When you are doing this, if you are struggling to find an easy way to test things in isolation, breaking it out a bit might be something worth considering.

PR #617, opened 2024-05-29
Add tests for hook point add hook

anthonyduong9:
# Description

Adds tests for `HookPoint.add_hook()`. This makes it safer to change the method. @VasilGeorgiev39 wrote a bug while refactoring the method, and it wasn't caught by tests.

Fixes # https://github.com/TransformerLensOrg/TransformerLens/issues/540

## Type of change

### Screenshots

# Checklist:


bryce13950:
Thank you very much for adding this! I really appreciate it. If you are interested in helping a bit more, I would be very happy to help you find something to do. If you want to improve test coverage then there is a lot to be done there ehttps://transformerlensorg.github.io/TransformerLens/_static/coverage/. The report at the time of writing this is for 1.19, and 2.0 will be put up tomorrow, which will affect it a bit. Regardless, there is a lot of work that needs to be done there. If you want to do something a bit more interesting, then there is always the potential to add support for new models, and a whole bunch of additional things that are in the pipeline. If you are interested, and you need direction, reach out to me on slack @Bryce Meyer.

PR #627, opened 2024-06-06
Model config tests

curt-tigges:
# Description

Added a test to verify that all model configs currently listed in loading_from_pretrained are not None.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


bryce13950:
The failure is due to these tests now triggering http requests on gated models. GitHub does allow for two buckets of variables, and we could put the hf token into the variables section to make this work, but that would involve also allowing access to secrets, which is not possible. It seems like a lot of people are requesting this sort of functionality though, so maybe a solution will come up later https://github.com/orgs/community/discussions/44322. The other solution would be to manually drop on the config for these models, which is very feasible.

jlarson4:
This solution is now redundant due to #1231 & changes introduced by TransformerLens 3.0's `TransformerBridge`. The verification & testing of models is handled in the `model_registry` tool, and with the increase in models supported doing this test for every model would be difficult if not impossible.

Issue #612, opened 2024-05-27
[Proposal] Merge utils and utilities


bryce13950:
### Proposal 

Right now there are two modules in the code that are organized to basically do the same thing. The `utils.py` file should be broken up and organized into individual pieces within `utlities`. The pieces themselves can be a single function per file, or groups of functions that have a logical reason to go together. The `utils.py` will then need to be updated to import the newly reorganized functions in order to maintain compatibility until at least the next major version of TransformerLens. 

Once the code organization is done, the unit tests need to be reorganized a bit too. Any existing unit and integration tests should also be broken up, and moved into a subdirectory with a single test per file. e.g. if there is a util file named `tokenization.py`, the test should then be `test_tokenization.py` in a utils directory. This structure will make it really easy for us to immediately identify a file where a bug has been introduced, as well as make it easier to identify where tests are lacking.

#### Optional

Once the tests have been reorganized, if additional tests can be added to any functions we find lacking in testing, that would be super cool, but not required.

### Motivation

Let's get this all into a single module, so that the code itself can be better organized, and easier to understand.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

starship006:
will try and start tackling this sometime soon :)

PR #628, opened 2024-06-07
Refactor the utilities file into utilities folder

starship006:
# Description

Addresses #612 

As of this comment being written, this only currently just refactors utils - no extra testing has been added.

## Type of change

# Checklist:


starship006:
Okay, currently slightly confused about something. There is weird behavior around `USE_DEFAULT_VALUE = None` in the utilities. In some parts of the code this is being treated as an optional boolean, and other parts its an optional string.

Right now, we are facing 
`transformer_lens/utilities/exploratory_utils.py:23: error: Cannot determine type of "USE_DEFAULT_VALUE"  [has-type]`

But, whenever I add the type `USE_DEFAULT_VALUE: Optional[bool] = None` to it, it causes typing checks downstream to fail, such as:

```
transformer_lens/HookedTransformer.py:293: error: Argument 3 to "get_attention_mask" has incompatible type "bool | None"; expected "bool"  [arg-type]
transformer_lens/HookedTransformer.py:366: error: Incompatible default for argument "padding_side" (default has type "bool | None", argument has type "Literal['left', 'right'] | None")
```

@bryce13950 , what is the best way to proceed? is there something obvious here I'm missing?

bryce13950:
Sorry for not getting back to you earlier. Honestly, I think the whole `USE_DEFAULT_VALUE` is not adding value to the code, and it is mostly overcomplicating the readability. `if prepend_bos is None:` is a lot more readable than `if prepend_bos is USE_DEFAULT_VALUE:`. If you come across something like `= USE_DEFAULT_VALUE`, then I would just replace it with `USE_DEFAULT_VALUE = None`. I would like to remove it and replace it with None for all type hinting, but that is a bit beyond the scope of this task. We can pretty safely do it for booleans though.

bryce13950:
Thanks for getting this going! It was a huge help. There is an issue with how the utils were previously configured that are now causing issues in python 3.12, so I needed to get the wrapped up right away before continuing on that.

Issue #570, opened 2024-05-04
[Bug Report] Mixtral generates nonsense


joelburget:
**Describe the bug**
![Screenshot 2024-05-04 at 4 41 10 AM](https://github.com/neelnanda-io/TransformerLens/assets/310981/69c34618-015f-4cd9-9ed6-4e0b295982e9)

I followed the instructions in `docs/source/content/special_cases.md` as well as I could tell (ran the model in both full precision and with `HookedTransformer.from_pretrained_no_processing`), yet my model generations were nonsensical.

**Code example**
```
from transformer_lens import HookedTransformer
from transformers import AutoModelForCausalLM, AutoTokenizer

model = HookedTransformer.from_pretrained_no_processing(
    "mistralai/Mixtral-8x7B-v0.1",
    n_devices=4,
)

# test model actually works
for i in range(5):
    display(
        model.generate(
            "Once upon a time",
            verbose=False,
            max_new_tokens=50,
        )
    )
```

**System Info**
Describe the characteristic of your environment:
 * Describe how `transformer_lens` was installed: `pip install sae-lens`
 * What OS are you using? Ubuntu 22.04 (`runpod/pytorch:2.0.1-py3.10-cuda11.8.0-devel-ubuntu22.04`)
 * Python version: 3.10

**Additional context**
Running on a 4x A100 SXM system on Runpod.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

jbloomAus:
Generate is not a good way to check a model is running properly. Can you run the following and share the results?

```python
a_large_chunk_of_text = "Generate is not a good way to check a model is running properly. Can you run the following and share the results?"
loss = model(a_large_chunk_of_text, return_type="loss")
print(loss.item())
```

joelburget:
Looks like the loss is ~6.82

![Screenshot 2024-05-15 at 7 19 37 AM](https://github.com/TransformerLensOrg/TransformerLens/assets/310981/7a030147-a5ce-49c5-9892-d8235f440669)

jbloomAus:
This could be related to this issue: https://github.com/TransformerLensOrg/TransformerLens/issues/591 . If I were you, I'd `pip install opt_einsum`, then see if you get the same issue. opt_einsum should be automatically used by pytorch if it's installed and the other issue was solved by using it. Let me know if you still get garbage afterward.

[...]

PR #641, opened 2024-06-18
Match Huggingface MLP implementation exactly.

joelburget:
# Description

#570 has shown that there are small differences in outputs between the Huggingface implementation of models, which uses a fused add-multiply (this presumably matches the way the models were trained as well) and TransformerLens, which doesn't. It seems likely that with bigger models (e.g. Mixtral), these errors accumulate enough to be significant. This is my attempt to begin to fix the problem.

Note that we take a slight readability hit, but I believe this is worth it for correctness.

For now I've only applied this to the implemenation in `mlp.py`. There are several other places the same fix should probably be applied.

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/tests/integration/test_match_huggingface.py
+++ b/tests/integration/test_match_huggingface.py
@@ -1,0 +1,17 @@
+import torch
+from transformers import AutoModelForCausalLM
+
+from transformer_lens import HookedTransformer
+
+
+class TestMatchHuggingFace:
+    def test_test_mlp_outputs_match(self):
+        tl_model = HookedTransformer.from_pretrained_no_processing("gpt2", device="cpu")
+        hf_model = AutoModelForCausalLM.from_pretrained("gpt2", device_map="cpu")
+        test_tensor = torch.randn((1, 1, tl_model.cfg.d_model))
+
+        for layer_n in range(len(tl_model.blocks)):
+            tl_out = tl_model.blocks[layer_n].mlp(test_tensor)
+            hf_out = hf_model.transformer.h[layer_n].mlp(test_tensor)
+
+            assert torch.sum(tl_out == hf_out) == tl_model.cfg.d_model
--- a/tests/integration/test_match_huggingface.py
+++ b/tests/integration/test_match_huggingface.py
@@ -1,17 +1,26 @@
 import torch
+import pytest
 from transformers import AutoModelForCausalLM
 
 from transformer_lens import HookedTransformer
 
 
 class TestMatchHuggingFace:
-    def test_test_mlp_outputs_match(self):
-        tl_model = HookedTransformer.from_pretrained_no_processing("gpt2", device="cpu")
-        hf_model = AutoModelForCausalLM.from_pretrained("gpt2", device_map="cpu")
-        test_tensor = torch.randn((1, 1, tl_model.cfg.d_model))
+    # fixtures
+    @pytest.fixture(scope="class", params=["gpt2"])
+    def model_name(self, request):
+        return request.param
+
+    # tests
+    def test_compare_huggingface_logits_match_local_implementation(self, model_name):
+        tl_model = HookedTransformer.from_pretrained_no_processing(
+            model_name, device="cpu"
+        )
+        hf_model = AutoModelForCausalLM.from_pretrained(model_name, device_map="cpu")
+        test_tensor = torch.randn((2, 2, tl_model.cfg.d_model))
 
         for layer_n in range(len(tl_model.blocks)):
             tl_out = tl_model.blocks[layer_n].mlp(test_tensor)
             hf_out = hf_model.transformer.h[layer_n].mlp(test_tensor)
 
-            assert torch.sum(tl_out == hf_out) == tl_model.cfg.d_model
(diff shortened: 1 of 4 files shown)


transformer_lens/components/mlp.py
```diff
@@ -62,20 +75,13 @@ def __init__(self, cfg: Union[Dict, HookedTransformerConfig]):
...
-        )  # [batch, pos, d_mlp]
+        # This is equivalent to (roughly) W_in @ x + b_in. It's important to
+        # use a fused addmm to ensure it matches the Huggingface implementation
+        # exactly.
+        pre_act = addmm(self.b_in, self.W_in, x)  # [batch, pos, d_mlp]
         if self.cfg.act_fn is not None and not self.cfg.act_fn.endswith("_ln"):
             post_act = self.hook_post(self.act_fn(pre_act))  # [batch, pos, d_mlp]
         else:
             mid_act = self.hook_mid(self.act_fn(pre_act))  # [batch, pos, d_mlp]
             post_act = self.hook_post(self.ln(mid_act))
```
bryce13950, line 86: Could you move this into it's own function, so that we can easy apply a unit test to the block?

bryce13950:
Very nice. A couple things... I am not a huge fan of the test name. It's not super clear what exactly it is testing at a glance. A slightly more explicit name will make that clear, maybe something like `test_compare_huggingface_logits_match_local_implementation`. After debugging this mixtral issue, I imagine this is going to become a heavily used test to make sure our implementation matches. 
Also, do you have time to put together a unit test for the MLP? It is put together in a much better way now, and it is a lot easier to unit test after your change.

tests/integration/test_match_huggingface.py
```diff
@@ -0,0 +1,17 @@
...
+from transformers import AutoModelForCausalLM
+
+from transformer_lens import HookedTransformer
+
+
+class TestMatchHuggingFace:
+    def test_test_mlp_outputs_match(self):
+        tl_model = HookedTransformer.from_pretrained_no_processing("gpt2", device="cpu")
+        hf_model = AutoModelForCausalLM.from_pretrained("gpt2", device_map="cpu")
+        test_tensor = torch.randn((1, 1, tl_model.cfg.d_model))
```
bryce13950, line 11: Can you turn the model name into a parametrized fixture? That will make it really easy for us to expand this for multiple models without rewriting the test.

transformer_lens/components/mlp.py
joelburget: What exactly would be its own function? Something like this?

```
def pre_act(x: Float[torch.Tensor, "batch pos d_model"]) -> Float[torch.Tensor, "batch pos d_mlp"]:
  return addmm(self.b_in, self.W_in, x)
```

bryce13950: I was referring to the whole block that applies the hook points. I was imagining that we could have a function that applies the hook points, and then returns `post_act`. That way our entire forward pass turns into nothing more than a call to those two functions, and we can then more easily create unit tests for those pieces in isolation.

transformer_lens/components/mlp.py
```diff
@@ -16,6 +16,19 @@
 from transformer_lens.utils import gelu_fast, gelu_new, solu
 
 
+def addmm(bias, weight, x):
```
bryce13950, line 19: Could you actually move this into the class itself, in order to avoid potential name conflicts in the future?

joelburget: The reason I left it here is that I'm 85-90% sure that in the change I alluded to above ("there are several other places the same fix should probably be applied"), this will become a shared utility. From a quick glance I'd guess around half of the uses of `einsum` in the codebase should have this same fix applied. So we'd want to revert adding it to the class in the next change.

bryce13950: That makes sense. If that is the case, honestly I would probably add a new utilities file where we can have these equations. It's a bit hidden away here for other people, so putting in a shared place to begin with would probably be a good idea.

joelburget: `utils.py` seems like the current dumping ground for this kind of thing?

transformer_lens/components/mlp.py
joelburget: TBH I'm having a little bit of trouble interpreting this idea. I like that this PR is currently essentially changing two lines, while this change sounds a little more involved, so maybe it would make sense to do separately?

transformer_lens/components/mlp.py
bryce13950: `utils.py` is currently being refactored into separate files inside of `utilities`. I believe there is a pending PR to review for this, so putting it into a separate file here will make that guys work load a bit less.

transformer_lens/components/mlp.py
bryce13950: That's fine

bryce13950:
I am going to mess around with this now. I think it will probably be ready to merge shortly. For future reference, there is a make command (`make format`) that takes care of all formatting, which will then pass the CI

Issue #615, opened 2024-05-28
[Bug Report] The output from HookedTransformer is not identical compared to Huggingface model for Lllama 3 model.


iamsimha:
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
The generations from huggingface model (LlamaForCausalLM) and HookedTransformer are different

**Code example**
``` python
# relevant imports 
import torch
from transformer_lens import HookedTransformer
from transformers import AutoTokenizer, LlamaForCausalLM


# Generate sample using greedy sampling

def generate_tokens(model, tokenizer, prompt, max_tokens):
    tokenised_prompt = tokenizer(tokenizer.apply_chat_template([{"role": "user", "content": prompt}], tokenize=False),
                                 padding=False,truncation=False, return_tensors="pt").input_ids
    generated_response = ""
    prompt_length = tokenised_prompt.shape[1]
    all_tokens = torch.zeros(1, prompt_length + max_tokens, dtype=torch.long)
    all_tokens[:, 0:prompt_length] = tokenised_prompt
    with torch.inference_mode():
        for i in range(max_tokens):
            result = model(all_tokens[:, 0:prompt_length + i])
            if isinstance(model, LlamaForCausalLM):
                logits = result.logits
            else:
                logits = result
            next_tokens = logits[:, -1, :].argmax(dim=-1) # greedy sampling (temperature=0)
            all_tokens[:, prompt_length + i] = next_tokens
            generated_response += tokenizer.decode(next_tokens, skip_special_token=True)
    return generated_response


MODEL_PATH = 'meta-llama/Meta-Llama-3-8B-Instruct'

# Load huggingface model and tokenizer using LlamaForCausalLM and shard it on 8 gpus

hf_model = LlamaForCausalLM.from_pretrained(
        MODEL_PATH
)
hf_tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)

# Load HookedTransformer model 
hooked_model = HookedTransformer.from_pretrained_no_processing(
    MODEL_PATH,
    device="cuda",
    dtype=torch.float16,
    default_padding_side='left',
    n_devices=1,
)

prompt = """what is the population of Delhi"""

generate_tokens(hooked_model, hf_tokenizer, prompt, 30)

>> '<|start_header_id|>assistant<|end_header_id|>\n\nAs of 2021, the estimated population of Delhi is approximately 31.2 million (31,181,000) people'

[...]

PR #650, opened 2024-06-29
Added support for Gemma-2

neelnanda-io:
Added support for Gemma-2 models.

- Support for attention and output logit soft-capping (ie attn_scores -> 50 * tanh(attn_scores / 50), basically squashing it into the interval [-50, 50])
  - Gotcha: Output logits with softcap are *not* invariant to adding a constant.
- Support for before and after LN
- Minor refactors, eg now MLP activation functions are defined as a dict in utils.py rather than a stack of if statements in both mlp.py and gated_mlp.py

Key differences between Gemma-1 and Gemma-2:
* Gemma-2 has an RMS Norm at the start *and* end of each layer. You cannot fold the RMS Norm weights of the ones at the end.
* Attn_scores and final logits are softcapped (with constants 50 and 30 respectively)
* Every other attn layer (1, 3, 5, etc) uses local, sliding window attention, with window 4096 (the max ctx is 8192, so I don't know why they bother tbh)
* They use grouped query attention, with group size 2 (ie every adjacent pair of heads use the same value and key)

## Type of change

- [X] This change requires a documentation update

# Checklist:

- **Note**: I have not manually updated the model_properties_table - is this automatic or do I need to?
- Given that the smallest model is 9B, tests don't really seem practical. I have manually verified that the logits agree with HuggingFace for the 9B on the string "<BOS>The capital of France is", at max deviation 8.2016e-05 for float32, though notably worse for low precision: 0.5 for bfloat16, 0.0664 for float16. To me this seems annoying but acceptable. I have not tested that things work well on really long strings or generations.


neelnanda-io:
Looks like abstract_attention.py assumed that n_heads * d_head == d_model, which is normally true but not true for Gemma-2 27B, I fixed that in the latest commit (this should do nothing for any model where that relation is true)

neelnanda-io:
Hmm, there's at least a 0.2 difference in logits for the 27B in float32, which is concerning... And quite surprising tbh, I don't see any architectural difference between the 9B and 27B, which suggests this would only be cascading errors? I did get the HF logits on CPU and TL logits across 2 GPUs though, which might cause some additional divergence?

ArthurConmy:
> Hmm, there's at least a 0.2 difference in logits for the 27B in float32, which is concerning... And quite surprising tbh, I don't see any architectural difference between the 9B and 27B, which suggests this would only be cascading errors? I did get the HF logits on CPU and TL logits across 2 GPUs though, which might cause some additional divergence?

I think you should move on and put a loud warning when loading 27B model. We have lots of evidence that TransformerLens' slight differences from HuggingFace have cascading errors, e.g. [here](https://github.com/TransformerLensOrg/TransformerLens/issues/615) and [here](https://github.com/TransformerLensOrg/TransformerLens/issues/570). @bryce13950 has mentioned trying to improve the numerical errors, so I doubt you made the error in implementation. The only question to me is whether the 27B model should even be merged.

neelnanda-io:
Fair. I think 27B should be merged, but printing a warning sounds good. I was mostly surprised at such a big jump between 9B and 27B of the error, when even cascading errors shouldn't explain that IMO - it's like 42 vs 46 layers, just a fair bit wider and with about twice as many neurons per MLP layer

bryce13950:
I am going to try and fold in some accuracy improvements, specifically in MLPs, into this when I put it up. I wouldn’t worry about adding warnings or anything for the being. I have a list of models to try once that is done. This is second on the list now.

bryce13950:
We crossed paths a little bit with this. I have done quite a bit with MLPs recently, and we did a few things very similar. I have been thinking about how to put the two changes together, and I think I am going to wrap up my branch first, given that it affects accuracy with existing models. I am basically redoing the entire set of components, and I would bet that what I am doing is going to increase accuracy here.

transformer_lens/components/transformer_block.py
```diff
@@ -157,21 +168,33 @@ def forward(
...
             resid_mid = self.hook_resid_mid(resid_pre + attn_out)  # [batch, pos, d_model]
             mlp_in = (
                 resid_mid if not self.cfg.use_hook_mlp_in else self.hook_mlp_in(resid_mid.clone())
             )
             normalized_resid_mid = self.ln2(mlp_in)
-            mlp_out = self.hook_mlp_out(self.mlp(normalized_resid_mid))  # [batch, pos, d_model]
+            mlp_out = self.mlp(normalized_resid_mid)  # [batch, pos, d_model]
+            if self.cfg.use_normalization_before_and_after:
+                mlp_out = self.ln2_post(mlp_out)
+            mlp_out = self.hook_mlp_out(mlp_out)
```
bryce13950, line 186: Can we break this out into a function? Not a fan of repeated logic, as it is often where bugs can happen either through things like initial typos, or someone changing something down the line without realizing the logic exists twice. Maybe a function named `apply_mlp` that we can call here, and then move this block into that.

bryce13950:
Just one comment from me on the code so far. I am going to wrap up my work, and come back to this afterwards to test it, and load up the code locally to play around a bit.

transformer_lens/components/transformer_block.py
neelnanda-io: The repeated logic being that this is done after attn _and_ after mlp? I don't see a good way to combine it in abstract logic, since they use different LNs (I could make a 'apply_normalisation_after' function that takes as input the output and whether it's attn or mlp, and chooses the right LN accordingly, but that's kinda ugly and hard to read IMO.

We could pull it out of the forward pass into a separate fn, but I don't think this would do much to minimise risk if there's a separate apply_mlp and apply_attn function.

bryce13950: I am referring to this line, and line 194. They are currently identical outside of the input variable.

neelnanda-io: Oh, gotcha. Yeah that seems reasonable to me

JThh:
Hey @neelnanda-io thanks for making this pr to support Gemma2 series model. I am aware that this pr has not been ready, but I use what it has and am training a SAE on the `resid_post` position. The run is being recorded [here](https://wandb.ai/jiatongg/sae_semantic_entropy/runs/e88i5gcc?nw=nwuserjiatongg). It might also be good auxiliary reference for further code adjustments to this pr.

neelnanda-io:
Cool! Let me know if you run into any issues

bryce13950:
MLP outputs are now perfect on these models

bryce13950:
image

PR #654, opened 2024-07-04
Fix Out bias not being summed in attention component when using 4 bit precision

FlyingPumba:
# Description

The attention component was not using `b_O` for models in 4 bit precision. It looked like it was, but the statement had no effect since it was in a new line, without using `\`, and not surrounded by parenthesis. I found this out by looking at the automatic inspections generated by Pycharm.

**This was caused by the black formatter.** I had to add extra parenthesis to avoid formatting warnings. I.e., just moving `+ self.b_O` to the line above will get automatically reverted when running black.

I haven't added tests for this change (AFAIK there are not tests for 4 bit precision).

image

## Type of change

# Checklist:


bryce13950:
Will have to make a note to check for similar situations in the future.

PR #662, opened 2024-07-08
Add support for Qwen2 models

g-w1:
# Description

I added support for Qwen2 models. All this entailed was fixing the Qwen2 architecture loading code to use grouped query attention, since that is what Qwen2 expects anyways (Qwen1.5 just used a special case of grouped query attention where it was equivalent to regular attention, so this does not break Qwen1.5).

## Type of change

# Checklist:


bryce13950:
Great! I should be able to review this, and get it into a release early next week

bryce13950:
Hey! Sorry for not getting to this earlier. I got pulled away to wrap up a couple things. Looking at it now. Will let you know if anything odd pops up!

bryce13950:
I am going to go ahead and merge this. The implementation seems to be a bit inaccurate, but I don't think that has anything to do with what has been don here, since the same inaccuracies are in some Qwen-1X models. Just a word of caution if anyone sees this and decides to start using it.

PR #666, opened 2024-07-11
Fix attention result projection

callummcdougall:
Current result projection for attention is incorrect. Type annotations would suggest that `result` isn't being summed over `head_index`, but in fact it is. I've edited the function so that it's no longer being summed over `head_index`.

Note, this bug caused the ARENA material to fail for the first transformers chapter, I've tested it and it now works. Worth flagging that I think the tests should be improved for this reason; **I'd strongly advocate for having the ARENA introductory notebook as one of the checks in "Notebook Checks".**


bryce13950:
Thanks for fixing this. Someone else found an issue with the type annotations on the attention components very recently. I was surprised that it existed, but it does seem that there is some sort of issue with type annotations not testing properly across the board. Finding a way to get the CI to catch that is a very high priority right now for many reasons. Hopefully that problem will be solved relatively quickly. 

I am happy to get any number of ARENA notebooks into the CI checks. Would you be willing to make sure it is maintained? i.e. update the related cells when you make changes? 
What is probably more useful, and much easier to do rapidly would be to add a specific test for cells that we might not currently have coverage for. If you want to send me some info on specific cells that need to have coverage, I can probably add that tomorrow.

neelnanda-io:
Proposal: A cheap way to avoid this specific kind of error being introduced (namely, changing the shapes of activations) is to save a big dict of the expected shapes of activations, and test that run_with_cache produces those, including for all the niche hooks like hook_attn_result or hook_mlp_input that aren't as widely used and may not be tested properly.

It would also be good to explicitly add a test for hook_attn_result (may as well make it explicit, even though it's covered by the ARENA tutorials) that eg it has the right shape and adds up to attn_out

PR #607, opened 2024-05-23
 (v3) Draft PR: add Pyright static typing to hook_points.py #590 

starship006:
# Description

This is a copy of PR https://github.com/TransformerLensOrg/TransformerLens/pull/590 and #464 , but just remade so that @bryce13950 can close it.

-----

This draft PR adds Pyright static typing to hook_points.py. See [Alan's message on the slack](https://opensourcemechanistic.slack.com/archives/C04T79RAW8Z/p1699673111412959) for motivation/background. Specifically, this PR:
- modifies the HookPoint and HookedRootModule to support a baseline amount of pyright typing
- adds a [tool.pyright] configuration to pyproject.toml to configure the typing for this project

It's not entirely complete, but is in a state to where it would benefit from some review/assistance. Before being PR ready, we should:
- Address some TODO's I've marked on lines of code that I'm ucnertain about
- Clean up remaining type warnings on hook.py 
- If possible, adjust the pyproject.toml file so the type warnings quit applying to all files

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] My changes generate no new warnings
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #673, opened 2024-07-12
fix: fixing broken backward hooks change

### LINKED ISSUE #672: [Bug Report] Backward hooks are broken as of v2.0.0
**Describe the bug**
It looks like as part of a change to add pyright type checking in #607 ([line 119](https://github.com/TransformerLensOrg/TransformerLens/pull/607/files#diff-85fafabc38ed61e857e7db5df426fe1696f7a1dbe3a9c532ca8dff4718d0193dR119)), backwards hooks were changed from being added with `register_full_backward_hook()` to `register_backward_hook()`.

I don't know if this was intentional, but these are not the same function, and [register_backward_hook is deprecated and should not be used](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_backward_hook). Since this change, a hook I had added no longer works.

**Code example**
The code below will run fine on TLens < 2.0.0
```python
model = HookedTransformer.from_pretrained("gpt2")

def bwd_hook(output_grads, hook):
    return (output_grads, )

with model.hooks(bwd_hooks=[("blocks.0.hook_resid_post", bwd_hook)]):
    out = model("hi")
    out.sum().backward()
```

On TLens >= 2.0.0, the following error is raised:

```
RuntimeError: hook '<function bwd_hook at 0x7fc89a77b370>' has returned an incorrect number of values (got 1, but expected 2)
```

This error can be seen in https://colab.research.google.com/drive/1JG6SPwKL3If4p_pw1b1nquAHVF3uNaEi?usp=sharing

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

chanind:
# Description

This PR fixes a bug in v2.x.x of TransformerLens where backwards hooks were changed to use the deprecated and broken `register_backward_hook()` method instead of the correct `register_full_backward_hook()`. Before v2.x.x the correct method was used. The breaking change happened in #607, but it's not clear why this change was made.

This PR also adds a test which will not allow this regression to happen again.

Fixes #672

## Type of change

# Checklist:

### DIFF
--- a/tests/integration/test_hooks.py
+++ b/tests/integration/test_hooks.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import pytest
 import torch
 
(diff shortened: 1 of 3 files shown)


bryce13950:
Thanks for getting this in! The PR in question was done over a long period of time, and I think the wrong line was grabbed during a merge conflict. I will get this into PIP within the next hour.

PR #692, opened 2024-08-02
add a demo for Patchscopes and Generation with Patching

### LINKED ISSUE #680: [Proposal] Demo and Tutorial on Patchscopes and "Patching + Generation"
### Proposal 

UPDATE: Demo and Tutorial on Patchscopes and "Patching + Generation"

DEPRECATED: Replication of the original causal tracing from the ROME paper.

### Motivation

I found that the original causal tracing method hasn't been supported here, and I think it has some advantages over the current activation patching method. For example, corruption with Gaussian noise might preserve more semantic information from the original sentence than corruption by changing words.

### Pitch

To replicate the original causal tracing method from the ROME paper ([https://arxiv.org/abs/2202.05262](https://arxiv.org/abs/2202.05262))

### Alternatives

I also consider replicating the Patchscope here, which is also mentioned in issue #500. Since Patchscope can be considered as a more general framework for this kind of patching/intervention-based methods, implementing it here can also make causal tracing available. I'd like to open another issue for the replication of Patchscope later.

### Additional context

I've implemented a version locally, and would like to put some examples here, comparing results from my implementation and from the original implementation.

### Checklist

HenryCai11:
# Description

A demo for "Patchscopes and Generation with Patching" (also see issue #680).

I put both Patchscopes and generation with patching in the same notebook and implement the generation part using "Patchscopes' language". I demonstrate three applications mentioned in the original paper, i.e. LogitLens, Entity Description and Zero-shot Feature Extraction.

Fixes #680

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -133,6 +133,7 @@ jobs:
           - "Main_Demo"
           # - "No_Position_Experiment"
           - "Othello_GPT"
+          - "Patchscopes_Generation_Demo"
           # - "T5"
     steps:
       - uses: actions/checkout@v3
--- a/demos/Patchscopes_Generation_Demo.ipynb
+++ b/demos/Patchscopes_Generation_Demo.ipynb
@@ -215,6 +215,7 @@
     }
    ],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
     "# I'm using an M2 macbook air, so I use CPU for better support\n",
     "model = HookedTransformer.from_pretrained(\"gpt2-small\", device=\"cpu\")\n",
     "model.eval()"
@@ -401,6 +402,7 @@
     }
    ],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
     "clean_logits[:, 5], patched_logits[:, 5]"
    ]
   },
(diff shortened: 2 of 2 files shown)


bryce13950:
Hey @HenryCai11, thanks for putting this together. We are in the middle of slowly adding test coverage to all demos in TransformerLens. Any new demos should be immediately covered unless there is a reason they cannot be. e.g. Llama requires quite a bit of manual configuration that we can't really do in the CI. Can you try adding test coverage for this demo? To do that, you need to modify this block https://github.com/TransformerLensOrg/TransformerLens/blob/dd537ba713a943bb5d26b0c84f16174abd6bde5d/.github/workflows/checks.yml#L122. If you put your demo name in that list, the CI will start trying to cover the demo. 

You probably need to do a couple things to cover the notebook, but any notebook that is not commented out in that block can be referenced to see how to manage it. Some blocks will likely need to be skipped, i.e. setup/anything that outputs env data/anything that loads anything where bit rates are in the output. If you have any questions on the process, feel free to DM me on Slack. I have gotten all of those notebooks to pass, so I am very knowledgeable on how to get past any issues that may come up.

HenryCai11:
Oh @bryce13950 thank you for helping me handle this! I just got some time to take a look. Good to see it's been resolved!

PR #699, opened 2024-08-10
Improve attention masking

UFO-101:
# Description

**Allow `attention_mask` to override the default mask in `HookedTransformer.forward()`.**
Currently the `attention_mask` argument is a footgun because it does nothing if `start_at_layer` is not `None`. One solution could be to just assert that `start_at_layer` is `None` when `attention_mask` is not `None`. But it is useful to be able to control the attention masking directly, so instead this just makes it do the thing you would expect.

**Add `attention_mask` argument to `loss_fn()` and `lm_cross_entropy_loss()` and adjust the cross entropy calculation to ignore masked (padding) tokens.**
Currently the loss calculation will be wrong if there are any padding tokens (or other masked tokens). This fixes that.

## Type of change

# Checklist:

### DIFF
--- a/tests/integration/test_attention_mask.py
+++ b/tests/integration/test_attention_mask.py
@@ -45,3 +45,35 @@ def test_attention_mask():
     ]
 
     model.run_with_hooks(input, fwd_hooks=fwd_hooks)
+
+
+def test_masked_tokens():
+    """Test that masking tokens works as expected."""
+    MODEL = "solu-1l"
+    prompts = [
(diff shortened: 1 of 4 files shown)


bryce13950:
Hey @UFO-101! Thanks for putting this together. I need to sit down and play around with this given that it is a change at one of the most important parts of the code base. At a glance it looks good, but I want to play around with the branch before I merge this. If anything comes up I will let you know.

PR #725, opened 2024-09-19
Fix the bug that tokenize_and_concatenate function not working for small dataset

xy-z-code:
# Description

As stated in the docstring of the function ("Note: There is a bug when inputting very small datasets (eg, <1 batch per process) where it just outputs nothing. I'm not super sure why"), `tokenize_and_concatenate` currently cannot handle small datasets. This is because the `tokenize_function` inside `tokenize_and_concatenate` isn't producing any tokens when dealing with small datasets, i.e., the tokenization process doesn't produce any output, leading to the absence of the 'tokens' column.

Issue with small datasets:
- When the dataset is small, `full_text` may not be long enough to produce a significant number of tokens.
- `num_batches` can become zero if `num_tokens` is less than `seq_len`.
- When `num_batches` is zero, slicing and rearranging operations result in empty arrays.

Fixes # (issue)
- Check if `num_tokens` is greater than or equal to `seq_len` before proceeding with rearrangement.
- If not, set `num_batches` to one and handle the tokens differently to ensure they're still returned.

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/transformer_lens/utils.py
+++ b/transformer_lens/utils.py
@@ -312,8 +312,6 @@ def tokenize_and_concatenate(
 
     Returns:
         Dataset: Returns the tokenized dataset, as a dataset of tensors, with a single column called "tokens"
-
-    Note: There is a bug when inputting very small datasets (eg, <1 batch per process) where it just outputs nothing. I'm not super sure why
     """
     dataset = keep_single_column(dataset, column_name)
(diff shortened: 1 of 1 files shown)


xy-z-code:
The unsuccessful checks seem to be caused by a connection issue with Huggingface.

PR #728, opened 2024-09-25
`utils.test_prompt` compares multiple prompts

callummcdougall:
New version of `test_prompt` in utils, which works the same way as it did before when `answer` is a string, but when it's a list of strings then it's interpreted as a list of different answers which we compare to each other.

Note - `test_prompt` is meant to support `answer: str` consisting of multiple tokens (and it does multiple forward passes, with the tokens of `answer` being attached to the prompt at each stage). This obviously isn't compatible with passing `answers: list[str]`, so I've chosen to always crop the answers to their first token if we're passing them through as a string (in other words I assume that we only ever want to use a single multi-token answer or multiple single-token answers, but not both).

Here's what passing `answer = [" John", " Mary"]` for IOI looks like:

image

### DIFF
--- a/transformer_lens/utils.py
+++ b/transformer_lens/utils.py
@@ -817,6 +817,10 @@ def test_prompt(
                 print(
                     f"Top {i}th token. Logit: {logits[0, index-1, sorted_token_positions[0, i]].item():5.2f} Prob: {sorted_token_probs[0, i].item():6.2%} Token: |{model.to_string(sorted_token_positions[0, i])}|"
                 )
+
+    # If n_answers = 1 then unwrap answer ranks, so printed output matches original version of function
+    if not using_multiple_answers:
+        answer_ranks = [r[0] for r in answer_ranks]
     rprint(f"[b]Ranks of the answer tokens:[/b] {answer_ranks}")
 
 
--- a/transformer_lens/utils.py
+++ b/transformer_lens/utils.py
@@ -820,8 +820,10 @@ def test_prompt(
 
     # If n_answers = 1 then unwrap answer ranks, so printed output matches original version of function
     if not using_multiple_answers:
-        answer_ranks = [r[0] for r in answer_ranks]
-    rprint(f"[b]Ranks of the answer tokens:[/b] {answer_ranks}")
+        single_answer_ranks = [r[0] for r in answer_ranks]
+        rprint(f"[b]Ranks of the answer tokens:[/b] {single_answer_ranks}")
+    else:
+        rprint(f"[b]Ranks of the answer tokens:[/b] {answer_ranks}")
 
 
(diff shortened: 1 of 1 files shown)


callummcdougall:
@bryce13950 sorry, what do I do with this? Have tried to apply formatting but a bit stuck (maybe part of the problem is that the formatting tools require WSL if you're on windows? I don't seem to be able to install that on my laptop)

callummcdougall:
Noticing that some of the failing tests are because the specific function `test_prompt` is being tested so I'll fix those (and I'll also add tests for this new way of using the function), but not sure what to do about all the formatting tests failing

bryce13950:
@callummcdougall I'll take care of it by the end of the night!

bryce13950:
I'll see if I can give you further instructions, but honestly I am on Mac and Ubuntu, so if there is a specific issue with Windows I am not sure if I really good advice on how to tackle that. I have been considering installing Windows on my ubuntu machine, and maybe figuring this out will give me some motivation to do so. I'll let you know what I end up doing to get the CI to pass, and hopefully it's just a matter of knowing what commands to run. If it ends up being a Windows specific issue, maybe we can look into that further.

callummcdougall:
okay thanks! tbc I think it's only a windows-specific issue in the sense that the commands you need to run for formatting can't be installed on windows, you need WSL. I'm sure that would be fine for most laptops, but I've just run into a bug when trying to get WSL that I've not been able to fix

callummcdougall:
although maybe it would be useful if the process was easier for windows, i.e. formatting is applied in a way that doesn't require you to install wsl?

bryce13950:
I do agree that we should make it easier to run the tooling on Windows if possible. I can definitely play around on that end at some point.

For our purposes here, the problem is not actually the format, but the fact that the printed result is now wrapped in a list. The docstring test is what was actually picking that up, along with one of the demos where a cell output has changed. I fixed it locally for the docstring. I can fix it on the demo as well.

callummcdougall:
ah, I see - sorry for not figuring that out, in hindsight I should have spotted that (wasn't really in my hypothesis space cause I wasn't expecting error checks on printed output). I'll push a fix to that now

callummcdougall:
oh sweet it's worked! okay that's a lesson for me to not jump to conclusions about it being a formatting problem haha sorry

bryce13950:
No worries! Glad I could help. There are a lot of little tools on this project that keep track of a whole pile of different changes. This is ready to merge yes?

callummcdougall:
Yeah this is good to merge I think! Will get around to adding a test function to it for multiple answers but probably won't have time this week, and would be ideal to merge now if possible so the sae material functions (I'll add it to my TODOs for end of next week)

also the existence of a test for the single-answer case captures most of the value of these kinds of tests so don't think it's urgent, hope that's okay!

bryce13950:
Perfect! Thanks in advance for adding the new test

PR #742, opened 2024-10-03
Updated broken Slack link

neelnanda-io:
Argh, Slack invites break after 400 people use them, I can't see a way around this


callummcdougall:
> Argh, Slack invites break after 400 people use them

lol I did not know that but that explains so much

neelnanda-io:
Apparently if you email feedback@slack.com and ask they may give you a link with a higher cap, at least? But idk, this seems really dumb

bryce13950:
I'll ask

PR #743, opened 2024-10-04
`from_pretrained` has correct return type (i.e. `HookedSAETransformer.from_pretrained` returns `HookedSAETransformer`)

callummcdougall:
# Description

This changes the type signature of `from_pretrained`, so its return type is the generic type of the base class. For example, now when we use `HookedSAETransformer.from_pretrained`, the expected return type is `HookedSAETransformer` not `HookedTransformer`.

I changed this because in my [SAE intro material](https://colab.research.google.com/drive/1ePkM8oBHIEZ2kcqAiA3waeAmz8RSdHmq#scrollTo=47O2XjSCRm-9) I was having to explicitly type-annotate as `sae: HookedSAETransformer = HookedSAETransformer.from_pretrained(...)`, and so just changing the type signature seemed like the best option.

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -10,7 +10,18 @@ a deeper understanding of the internal workings of transformers like GPT-2.
 """
 import logging
 import os
-from typing import Dict, List, NamedTuple, Optional, Tuple, Union, cast, overload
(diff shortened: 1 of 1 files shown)


bryce13950:
Makes sense!

PR #751, opened 2024-10-09
Add support for `Mistral-Nemo-Base-2407` model

ryanhoangt:
# Description

This PR is to add support for a recent Mistral's [Mistral-Nemo-Base-2407](https://huggingface.co/mistralai/Mistral-Nemo-Base-2407).

## Type of change

### Screenshots

After the PR, the existing `mistral-7b` and the new `mistral-nemo-base-2407` work fine.

| `mistral-7b` | `mistral-nemo-base-2407` |
| ------ | ----- |
| Screenshot 2024-10-09 at 16 43 46 | Screenshot 2024-10-09 at 16 43 54 |

# Checklist:


bryce13950:
Thanks for adding this! I am wrapping up a big notebook that will outline all model compatibility with Colab. I only have to test the really big models at this point, and I will probably have that demo in the source before I merge this. This demo is going to add a new requirement to model submission, which will be figuring out what Colab environments models can be used in. If you can spin up Colab, and see if this model works on the free tier, that would be very helpful for getting this merged. If it does not work, if you could then check it on the A100 config, and finally on the high memory config via CPI if you cannot get it to load via CUDA. If you do not have access to a paid Colab account let me know, and I will do the check if need be!

bryce13950:
The notebook I mentioned previously is now in dev. If you can take a look at it, and organize your new model accordingly, that would be very much appreciated! You are the first person to submit a model with this new requirement, so if you have any questions, or if you need some help with running your model on paid Colab tier, let me know.

ryanhoangt:
I have put the model under `paid_gpu_models` list. Same with `mistral-7b`, this one can only be loaded via A100 GPU on Colab I think (even L4 doesn't have enough to load).

bryce13950:
In what context are you referring to CPI? For beartype, this seems to be an issue that a lot of people are running into these days. I normally set up the project with `poetry install` without the dev container, and that seems to work well. I will have to review the setup documents to see why so many people are running into this issue. For the time being, try running it locally with `poetry install`, and let me know if you can get the tests to pass.

ryanhoangt:
> In what context are you referring to CPI?

In the instruction above: "If it does not work, if you could then check it on the A100 config, and finally on the high memory config via CPI if you cannot get it to load via CUDA..."

Re: the beartype errors, I just switch my python version from 3.8 to 3.11 and that seems to resolve the errors

bryce13950:
Ahh! Typo. It's supposed to say CPU

bryce13950:
For 3.8, we honestly don't even officially support that version anymore. There may still be a handful of references to it in the docs, but it was removed from the CI a couple months ago based on feedback from the community. Everything should still hypothetically work in 3.8, but I would not be surprised if issues like these gradually began popping up. I will have to go through the docs to make sure references to it are removed.

ryanhoangt:
Yeah sounds good!

PR #760, opened 2024-10-18
Add py.typed for type hints

UFO-101:
# Description

I think that all libraries are supposed to include a py.typed file for types to be accessible to external users. See: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/.

However, I haven't researched this thoroughly, so someone should probably look into this a bit more.

## Type of change

# Checklist:


UFO-101:
Bump @bryce13950

bryce13950:
Sorry for the delay on getting this in, and thanks for bumping me. I updated it to the most recent version of TransformerLens, and that seems to have caused some issues. I am thinking of rolling this into 3.0. The issues that came up, I believe will already be fixed in dev-3.x. If you are fine with that, I will update the merge point to dev-3.x and work this into 3.0

PR #761, opened 2024-10-18
Add configurations for Llama 3.1 models(Llama-3.1-8B and Llama-3.1-70B)

vatsalrathod16:
# Description

This PR introduces new configuration dictionaries for the Llama 3.1 models, specifically the 8B and 70B parameter versions.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

Issue #691, opened 2024-07-31
[Proposal] Add Lllama 3.1 support


ssuukk:
### Proposal 

Add Llama 3.1 support. Currently trying to load it fails with:

`ValueError: meta-llama/Meta-Llama-3.1-8B-Instruct not found. Valid official model names (excl aliases): `

### Motivation

Llamma 3.1 is new bestest model, they say it's smarter than Rey Palpatine.

- [X] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

mylesgoose:
It works alredy with llama 3.1 you just have to put it into a folder with the name of the previous llama like so. MODEL_ID = "meta-llama/Meta-Llama-3-70B-Instruct"
NEW_MODEL_ID = "mylesgoose/Meta-Llama-3.1-70B-abliterated"
MODEL_TYPE = "meta-llama/Meta-Llama-3-70B-Instruct"

# Load model and tokenizer on CPU first to avoid memory issues
model = HookedTransformer.from_pretrained_no_processing(
    MODEL_TYPE,
    local_files_only=True,
    dtype=torch.bfloat16,
    default_padding_side='left',
    device="cpu"
)
tokenizer = AutoTokenizer.from_pretrained(MODEL_TYPE)
tokenizer.padding_side = 'left'
tokenizer.pad_token = tokenizer.eos_token or you can add the new model directly where you find the other models in that class. for exampel when you run the coee and it gives you an error. click on the url for the erro message and edit the page to incldue the details for the new model. so basically just copy and paste the4 cotnents of Meta-Llama-3.1-70B into the 3 folder and run from local files as above. https://huggingface.co/blog/mlabonne/abliteration

mylesgoose:
[loading_from_pretrained.zip](https://github.com/user-attachments/files/16535742/loading_from_pretrained.zip)

ssuukk:
Adding it to loading_from_pretrained.py fails with:

```
  File "E:\Projekty\_AI\abliteration\main.py", line 66, in <module>
    model = HookedTransformer.from_pretrained_no_processing(
  File "C:\Users\ssuuk\anaconda3\envs\rope2\lib\site-packages\transformer_lens\HookedTransformer.py", line 1331, in from_pretrained_no_processing
    return cls.from_pretrained(
[...]

PR #764, opened 2024-10-25
Add support for NTK-by-Part Rotary Embedding & set correct rotary base for Llama-3.1 series

Hzfinfdu:
# Description

Thanks to https://github.com/TransformerLensOrg/TransformerLens/issues/691 and https://github.com/TransformerLensOrg/TransformerLens/pull/761 for proposing and supporting Llama-3.1 models in TL.
However two modifications made on rotary embedding is not implemented here. 
- Llama 3.1 models use a rope_theta (which is rotary_base in TL) of 500,000 for [8B](https://huggingface.co/meta-llama/Llama-3.1-8B/blob/main/config.json#L28) and [70B](https://huggingface.co/meta-llama/Llama-3.1-70B/blob/main/config.json#L28) models.
- Llama 3.1 models use NTK-by-Parts Rotary Embedding introduced in Section 3.2 in https://arxiv.org/pdf/2309.00071, where they:
    > introduce two extra parameters α, β. All hidden dimensions d where r(d) < α are those where we linearly
interpolate by a scale s (exactly like PI, avoiding any extrapolation), and the d where r(d) > β are
those where we do not interpolate at all.

This PR implements both changes and significantly decreases max logit diff from >1 to about 0.2 for Llama-3.1-8B/

No dependency is required for this change.

## Type of change

### Screenshots

| Before | After |
| ------ | ----- |
| image| image |

# Checklist:


bryce13950:
I am curious... Do you know off the top of your head if this should apply for 3.2 as well?

Hzfinfdu:
Yes. It can be found [here](https://huggingface.co/meta-llama/Llama-3.2-3B/blob/main/config.json#L22). Most ROPE-related settings are the same except the `NTK_by_parts_factor` is 32 for 3.2 and 8 for 3.1.

Issue #747, opened 2024-10-06
[Bug Report] hook_normalized is inconsistent between RMSNorm and LayerNorm


neelnanda-io:
In [layer_norm.py](https://github.com/TransformerLensOrg/TransformerLens/blob/main/transformer_lens/components/layer_norm.py#L56) `hook_normalized` is after the gain and bias weights are used, in [rms_norm.py](https://github.com/TransformerLensOrg/TransformerLens/blob/main/transformer_lens/components/rms_norm.py#L44C9-L45C26) it's before. This is inconsistent and highly confusing IMO, and should be fixed

RMS
```
x = self.hook_normalized(x / scale).to(self.cfg.dtype)  # [batch, pos, length]
return x * self.w
```
LN
```
return self.hook_normalized(x * self.w + self.b).to(self.cfg.dtype)
```

This is an irritating situation, since I think it's super bad to be inconsistent as eg code won't transfer from an RMS Norm model to an LN model, and there'll be silent errors. However, making it consistent would be (technically) a breaking change, though I'd guess it's not widely used behaviour?

I personally think hook_normalized should be changed to be before the gain and bias weights in LN, since that's what normalized intuitively means. This is what it originally meant, I then changed it about two years ago in the early days of the library, I think because that was then "the thing that is input into the next layer". But now we have attn_input and mlp_input hooks so who cares. 

Note that this is not an issue if you fold the LN weights, it's equivalent

cc @ArthurConmy @bryce13950

bryce13950:
3.0 is coming sooner rather than later. We can definitely work this into that release.

jlarson4:
Now applied before gain/bias in both (PR #770); behavior is consistent as of the 3.0 release

PR #770, opened 2024-11-01
Restore consistency of hook_normalized between LayerNorm and RMSNorm

degenfabian:
# Description

This PR intends to fix issue #747. hook_normalized is applied after the gain and bias weights are used in layer_norm.py, whereas in rms_norm.py it's before. This inconsistency was fixed by moving hook_normalized before the gain and bias weights in layer_norm.py. According to @neelnanda-io this will be a breaking change, which after @bryce13950 could be worked into release 3.0. I figured since there was already so much guidance as to how to go about the change in the issue itself and since the actual change was so small, I would just go ahead and do it. I would greatly appreciate some guidance about whether or not I should add tests for this, since I am new to mech Interp and this library and wouldn't know how to go about that.

Thanks for maintaining this library!

Fixes # (issue) #747 

## Type of change

# Checklist:


bryce13950:
Thanks for putting this together! This is going to sit open for a while until we start testing some changes on 3.0. When we are ready to do that, I will get it merged right away!

PR #775, opened 2024-11-08
Set prepend_bos to false by default for Bloom model family

### LINKED ISSUE #774: [Proposal] prepend_bos should by default be set to false for the Bloom model family
### Proposal 

When loading the bloom models, prepend_bos should be set to false by default instead of true.

### Motivation

By default TransformerLens sets prepend_bos to true. However, for the bloom model family setting prepend_bos to true yields completely insensible output which is highly different from the HuggingFace implementation. Turning prepend_bos to false solves this issue (partially, use_kv_cache also needs to be set to false, I will create an issue about this in the future). One should be able to run a model in TransformerLens without knowing all the special options and parameters that need to be set, so in case of the bloom model family prepend_bos should be set to false by default.

I tested this on Google Colab (TL installed with pip) and my Mac (macOS Sequoia 15.0.1) with python 3.10.15 and they both yield the same results

I attached a screenshot of an
Bildschirmfoto 2024-11-08 um 04 16 28
Bildschirmfoto 2024-11-08 um 04 15 58
 example output which shows the difference between setting prepend_bos to true and setting it to false.

degenfabian:
# Description

This PR addresses Issue #774. Bloom models output insensible output when the prepend_bos flag is set to True, so for usability this flag should be set to False by default for the Bloom model family. I'm happy about ideas to implement this more cleanly, because with the current implementation the flag can practically not be set to True anymore, unless it is overwritten when specifically calling a function like generate. I think that's an unpretty solution but since I can't see a use case in which it would make sense to set the flag to True I think it's not terrible, but again very happy for suggestions.

Fixes #774 

## Type of change

### Screenshots
For Screenshots of the difference in the generation of the models please see Issue #774.

# Checklist:


degenfabian:
Interestingly, adding this if clause makes the acceptance test fail although it fixes the outputs being totally off. I will investigate this further tomorrow.

degenfabian:
Interestingly, I just ran the hugging face implementation on Google Colab to see if I would get a different value than on my Mac, and indeed I did. The loss value I got from Colab now matches what we get from TransformerLens with prepend_bos set to false close enough to be within the error margin, so hooray I guess. I'm slightly confused by this, does anyone have an idea as to why I would get a different loss from the hugging face implementation on my Mac then on Google Colab?

degenfabian:
I just tested this with gpt2-small, and here the output with prepend_bos set to True is not completely unusable but still deviates from Huggingface. If we just want to match HuggingFace as closely as possible, I could imagine adding this in loading_from_pretrained.py and setting the prepend_bos model accordingly for each model we load from Huggingface. I would be happy to do that and rename this PR and issue, what do you think @bryce13950 ?

bryce13950:
Let's talk about this a bit to get up to speed, and figure out if we can find where the divergence is.

tests/acceptance/test_hooked_transformer.py
```diff
@@ -83,6 +83,8 @@ def test_model(name, expected_loss):
     # Runs the model on short text and checks if the loss is as expected
     model = HookedTransformer.from_pretrained(name)
     loss = model(text, return_type="loss")
+    print(4e-5)
+    print(loss.item() - expected_loss)
```
bryce13950, line 87: Make sure to remove log points when you submit your PR!

bryce13950:
This looks good besides the log points! Once those are removed, we can merge this

tests/acceptance/test_hooked_transformer.py
degenfabian: So sorry, I forgot about those. I just removed them!

PR #782, opened 2024-11-17
Remove einsum in apply_causal_mask in abstract_attention.py

degenfabian:
# Description

There are known accuracy issues which are to some extent caused by the usage of einsum across large parts of the codebase. This PR removes one of these usages in the apply_causal_mask function in abstract_attention.py. There is no specific issue associated with this PR, although removing a lot of the einsum usages could help in removing implementation inaccuracies addressed in many issues. I have not added specific test cases for this change, but I ran all the models (except those that required authorization and the paid_cpu models) in the Colab_Compatibility notebook and have not run into any compatibility issues introduced by this change. Additionally, the changed part is executed over 230 times across the unit and acceptance tests.

## Type of change

# Checklist:


bryce13950:
Same as #783 and #781. I did verify that the model is working on the current public release of TransformerLens, so it is definitely a regression of this change.

PR #783, opened 2024-11-17
Remove einsum in forward pass in AbstractAttention

degenfabian:
# Description

There are known accuracy issues which are to some extent caused by the usage of einsum across large parts of the codebase. This PR removes one of these usages in the create_alibi_bias function. There is no specific issue associated with this PR, although removing a lot of the einsum usages could help in removing implementation inaccuracies addressed in many issues. I have not added any tests yet but the specific else clause that was modified is called 2 times within the test suite and all tests still pass. I am happy to write additional tests!

## Type of change

# Checklist:


degenfabian:
For documentation purposes:
This error occurred when calling the generate function on the BERT model. I was intrigued by the error occurring for all of these three PRs (#781, #782, #783) and somehow not on the latest release even though the generate function of the Bert model for some of the aforementioned PRs didn't even trigger the code I changed. Hence I tried to reproduce the error on the main and the dev branch and it turns out that it exists on both. I think this didn't happen when @bryce13950 tested this on the current public release because the error is highly irregular, sometimes it appears two to three times back to back but sometimes I have to run the same code 10+ times in order for the error to appear. To get to the bottom of this, I started tracking the error function through function and it turns out that the embedding matrix of the BERT model is just all zeros (every time, also when the error is not thrown) and when the generate function does end up throwing an error the NaN values stem from the positional embedding matrix which contains NaN values (only in the case when the error is triggered). These NaN values  then propagate through the different calculations and end up triggering the initial error in the sample_logits function.

The problem here is that in the Colab_Compatibility notebook that was used to test the change this PR introduced, the BERT model was initialized as a HookedTransformer. We get more information as to why this error occurs when initializing BERT as a HookedTransformer when looking into the weight loading of Bert:

```
embeddings = bert.bert.embeddings
    state_dict = {
        "embed.embed.W_E": embeddings.word_embeddings.weight,
        "embed.pos_embed.W_pos": embeddings.position_embeddings.weight,
        "embed.token_type_embed.W_token_type": embeddings.token_type_embeddings.weight,
        "embed.ln.w": embeddings.LayerNorm.weight,
        "embed.ln.b": embeddings.LayerNorm.bias,
    }
```

 The problem is that "embed.embed.W_E" and "embed.pos_embed.W_pos" is only used by the HookedEncoder class and therefore when loading BERT as an HookedTransformer, the embedding matrix and positional embedding matrix are not assigned values (because HookedTransformer uses the keys "embed.W_E" and "pos_embed.W_pos"). I guess that somehow torch.empty ends up producing NaN values (but not every time, which is why the error occurred so irregularly) which is how the positional embedding matrix got it in its initialization function. These NaN values then propagated through the code. 

This is more of a usage problem in the Colab_Compatibility notebook, since the HookedEncoder class was specifically created for BERT-style models and hence BERT shouldn't be loaded as a HookedTransformer. To prevent this from happening in the future, in PR #794 I fixed the incorrect usage of BERT in the Colab_Compatibility notebook (amongst other little bug fixes) and additionally in PR #795 an error is thrown when trying to load in BERT as a HookedTransformer, similarly as is already done for the t5 family of models.

PR #790, opened 2024-11-23
Remove einsum in complex_attn_linear

degenfabian:
# Description

There are known accuracy issues which are to some extent caused by the usage of einsum across large parts of the codebase. This PR removes one of these usages in the complex_attn_linear function in utilities/attention.py. There is no specific issue associated with this PR, although removing a lot of the einsum usages could help in removing implementation inaccuracies addressed in many issues. I ran all the models (except those that required authorization and the paid_cpu models) in the Colab_Compatibility notebook and have not run into any compatibility issues introduced by this change. Additionally, the changed part is executed 7 times across the unit and acceptance tests and I have added a little additional test case myself.

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/components/test_attention.py
+++ b/tests/unit/components/test_attention.py
@@ -1,3 +1,4 @@
+import einops
 import pytest
 import torch
 import torch.nn as nn
@@ -5,6 +6,7 @@ from transformers.utils import is_bitsandbytes_available
 
 from transformer_lens.components import Attention
 from transformer_lens.HookedTransformerConfig import HookedTransformerConfig
+from transformer_lens.utilities.attention import complex_attn_linear
 
 if is_bitsandbytes_available():
     from bitsandbytes.nn.modules import Params4bit
@@ -98,3 +100,31 @@ def test_attention_config_dict():
     assert attn.cfg.load_in_4bit == False
     assert attn.cfg.dtype == torch.float32
     assert attn.cfg.act_fn == "relu"
+
+
+def test_remove_einsum_from_complex_attn_linear():
+    batch = 64
+    pos = 128
+    head_index = 8
+    d_model = 512
+    d_head = 64
+    input = torch.randn(batch, pos, head_index, d_model)
+    w = torch.randn(head_index, d_model, d_head)
+    b = torch.randn(head_index, d_head)
+    result_new = complex_attn_linear(input, w, b)
+
+    # Check if new implementation without einsum produces correct shape
+    assert result_new.shape == (batch, pos, head_index, d_head)
+
+    # Old implementation used einsum
+    result_old = (
+        einops.einsum(
+            input,
+            w,
+            "batch pos head_index d_model, head_index d_model d_head -> batch pos head_index d_head",
+        )
+        + b
+    )
+
+    # Check if the results are the same
+    assert torch.allclose(result_new, result_old, atol=1e-4)
--- a/transformer_lens/utilities/attention.py
+++ b/transformer_lens/utilities/attention.py
@@ -2,6 +2,7 @@
 
 Utilities for attention components.
 """
+
 import einops
 import torch
 import torch.nn.functional as F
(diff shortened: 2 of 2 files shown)


bryce13950:
I think we may need to do some more thorough testing on this one. I ran some test generations from this branch, and I am getting some responses that seem to have degraded from the last major release. Check out the following examples...

bigscience/bloom-1b7
```
Hello my name is Node Jira!!!'}
        >>> from translate.
```
This model was already flagged as previously having issues, but the issue does not seem to have been as noticeable. The following generation is from a recent release of transformer lens
```
Hello my name is Mike and I am from OUR #Cousin
```

google/gemma-2b
```
Hello my name is Roee expermesting equation: } Solve the
```
This model previously did not have any really obvious inaccuracies, but with this change it seems to have new issues.

The good news is the the LLaMA family seems to generally be the same. There are only two models in that family, which seem to have inaccuracies, and I didn't notice anything visibly worse with this change. We should take a little bit more time to play with this one before it gets merged though, so that we can ensure these other models have not actually degraded, and figure out why they have if they really have degraded.

degenfabian:
Hi Bryce, thanks for your feedback! I took a look at this, and for the Bloom model I think the problem is that on this branch the change from PR #806 was not active yet when you were testing, which made the prepend_bos token be set to True which causes terrible outputs for the Bloom model. I ran some tests on logits and text output when the prepend_bos token is set to False again. To ensure that complex_attn_linear is executed, which is where I removed the einsum usage in this PR, I did:

```
model.set_use_attn_in(True)
```

Here are the results:
```
=== Logits Comparison ===
TransformerLens:
Mean: tensor(10.2282, device='mps:0', grad_fn=<MeanBackward0>)
Std: tensor(3.8330, device='mps:0', grad_fn=<StdBackward0>)
HuggingFace:
Mean: tensor(10.2282, device='mps:0', grad_fn=<MeanBackward0>)
Std: tensor(3.8330, device='mps:0', grad_fn=<StdBackward0>)
Max diff: tensor(2.2888e-05, device='mps:0', grad_fn=<MaxBackward1>)

TransformerLens output:
 Hello my name is Mike and I have this website called for the first time to tell my story and hopefully to help a few
HuggingFace output:
 Hello my name is Mike and I have this website called for the first time to tell my story and hopefully to help a few

TransformerLens output:
 Hello my name is Mike, I’m from Australia. I’m also the founder of the Project: Spending the Day Out project.
HuggingFace output:
 Hello my name is Mike, I’m from Australia. I’m also the founder of the Project: Spending the Day Out project.

TransformerLens output:
 Hello my name is Mike. I am looking for a good guy to chat with. I am new to the area and would
HuggingFace output:
 Hello my name is Mike. I am looking for a good guy to chat with. I am new to the area and would
```

For google/gemma-2b to ensure that complex_attn_linear is executed I do:

```
model.set_use_split_qkv_input(True)
```

Here are the results:

```
=== Logits Comparison ===
TransformerLens:
Mean: tensor(-26.8753, device='mps:0', grad_fn=<MeanBackward0>)
Std: tensor(4.7710, device='mps:0', grad_fn=<StdBackward0>)
HuggingFace:
Mean: tensor(-26.8753, device='mps:0', grad_fn=<MeanBackward0>)
Std: tensor(4.7710, device='mps:0', grad_fn=<StdBackward0>)
Max diff: tensor(0.0001, device='mps:0', grad_fn=<MaxBackward1>)

TransformerLens output:
 Hello my name is Mike and I am a 20 year old student from the UK. I am currently studying a degree
HuggingFace output:
 Hello my name is Mike and I am a 20 year old student from the UK. I am currently studying a degree
```

When I do
```
model.set_use_attn_in(True)
```
as I did for the Bloom model, the program crashes. This is because when adding the head_index dimension which is needed when using attn_in, TransformerLens does not account for the GroupedQueryAttention Gemma uses which has different number of heads for query and key/values. I will look for a nice way to fix this in another PR. Please let me know if you need anything else!

PR #795, opened 2024-11-24
Raise exception when BERT is loaded with HookedTransformer instead of…

degenfabian:
# Description
When the BERT model is loaded the incorrect way via HookedTransformer instead of HookedEncoder, currently no exception is raised although this is being done for the T5 model family with HookedEncoderDecoder. With the code change in this PR, an error is raised in a similar fashion when the BERT model is loaded with the HookedTransformer class. There is no specific issue addressed by this PR

# Checklist:


bryce13950:
Perfect! We are going to end up rolling this into the next major release. The T5 models did not load at all in HookedTransformer, so we weren't going to disrupt anyone with a loud warning. Someone could theoretically still load this model into HookedTransformer, which means that adding the warning in 2.x may mess with someones use.

PR #806, opened 2024-12-04
Set default_prepend_bos to False in Bloom model configuration

### LINKED ISSUE #774: [Proposal] prepend_bos should by default be set to false for the Bloom model family

degenfabian:
In the PR #775 I have built the necessary basis so that the default_prepend_bos parameter can have different default values for different models and not only the current global default value "True". I specifically implemented this change for the Bloom model family, because performance was significantly enhanced when the prepend_bos flag was set to False. In the PR that addresses this however, I forgot to actually add this to the configuration, so currently the prepend_bos flag still defaults to True also for Bloom models. This PR intends to fix this.

Fixes #774 

## Type of change

# Checklist:


bryce13950:
Make sure to update your fork with the latest changes when you start on a new branch. There were a bunch of commits in here that were merged into the main branch previously in a squash. It's not a big deal, but it makes it a bit more difficult to see what was changed in your branch at first.

degenfabian:
So sorry, I wanted to reuse this branch from the old PR and restored it but messed things up with the commits. Won't happen again, thanks for merging it so quickly!

PR #814, opened 2024-12-14
Fix that padding_side always defaults to "right" when no value is explicitly passed

### LINKED ISSUE #801: [Bug Report] Padding side inconsistency with Huggingface Transformers
**Describe the bug**
The HookedTransformer tokenizer has the padding side set to "right" for Gemma 2 2b. However, the huggingface autotokenizer has the padding side set to "left." I'm not sure why these are inconsistent.

**Code example**
```python
from transformers import AutoTokenizer

model = HookedTransformer.from_pretrained('google/gemma-2-2b')
tokenizer = AutoTokenizer.from_pretrained('google/gemma-2-2b')

print(model.tokenizer.padding_side)
print(tokenizer.padding_side)
```

Output:
```
right
left
```

**System Info**
Linux system: installed using pip in a Python 3.10.12 virtualenv. Package versions are:
* transformer_lens: 2.9.0
* transformers: 4.46.1

### Checklist

degenfabian:
# Description

Currently, when the user does not pass in an explicit value for default_padding_side, the tokenizer that is created/loaded for a model is always assigned the default value of "right", even when we load in a model from Huggingface whose tokenizer might already have a default_padding_side value assigned to it. There is an example of that in issue #801.  Since this default value could differ from the global default value "right" in TransformerLens, this PR creates the following resolution order:
1. If the user explicitly sets a value, use that value
2. If the tokenizer already has a default_padding_side value, use that one
3. Global default ("right")

This works similarly to the way the defaulting of prepend_bos was changed in PR #775 

Fixes #801 

For the example used in Issue #801:

```
from transformers import AutoTokenizer

model = HookedTransformer.from_pretrained("google/gemma-2-2b")
tokenizer = AutoTokenizer.from_pretrained("google/gemma-2-2b")

print(model.tokenizer.padding_side)
print(tokenizer.padding_side)

```

The output is now:
```
left
left
```

instead of:
```
right
left
```

I also fixed a little typo where an error message says that padding_side is set incorrectly although the assertion statement checks for the value of default_prepend_bos.

## Type of change

# Checklist:


bryce13950:
This is going to be put into 3.0. The default behavior is being changed, and even though it is currently not correct, people may not be setting the value for this explicitly when they want it to be right, since it already is right. Changing that in a minor release could cause a lot of subtle issues, and frustrations. We do have a conflict with the Othello demo now. If you run it again with what is currently in dev, that should resolve it.

PR #818, opened 2024-12-19
Add LLaVA support and make HookedTransformer.generate() able to get embedding as inputs 

zazamrykh:
Pipeline for work with LLaVA is next: 
1) Transform image into sequence of image embeddings
2) Project them into LLM space
3) Project tokens (text) sequence into the same space
4) Merge image features and tokens
So, here we have sequence of embeddings. And pass it to HookedTransformer.generate() is only available opportunity. Thus, generate function must support embeddings as input for LLaVA and other VLM models support.

At this pull request LLaVA support was provided. HookedTransformer.generate() function now supports embeddings input as well. Previous functionality was tested and it seems that nothing was broken. Also little demonstration of how LLaVA can be used with TransformerLens is provided. 

- [x] I have added tests that prove my fix is effective or that my feature works - _I have not lauch tests through corresponding test .py file but I run the same function on .ipynb and it passes_


bryce13950:
I see that you are trying to merge to 3.x. Is there a specific reason for doing that? It's not a big deal, but we could probably get this into 2.x if you wanted. There's quite a bit that needs to happen before 3.x goes up, so maybe makes sense to get LLaVA in a release before then.

transformer_lens/HookedTransformer.py
```diff
@@ -2032,7 +2032,7 @@ def process_weights_(
     @torch.inference_mode()
     def generate(
         self,
-        input: Union[str, Float[torch.Tensor, "batch pos"]] = "",
```
bryce13950, line 2035: Is there anyway around adding a dimension to the input here?

zazamrykh:
> I see that you are trying to merge to 3.x. Is there a specific reason for doing that? It's not a big deal, but we could probably get this into 2.x if you wanted. There's quite a bit that needs to happen before 3.x goes up, so maybe makes sense to get LLaVA in a release before then.

Hm, I think that I don't clearly understand how branches are organized here. I have python 3.11 so I thought that it is good idea for making pull request to dev3.x branch. Is it right that 2.x and 3.x are simply versions of project not python? As I understand, dev branch is corresponding to 2.x of project and in dev-3.x you are developing more long-term features. So, you want me to make pull request to dev branch instead of dev-3.x? 

And I think that I've made bad test function in test_hooked_transformer.py. I've write it but it appears not to work in that file. So I've run that function in .ipynb file and tested all cases of inputs for generate function for check if I've broke something. But I've decided to push that function too. Now I see that in generate function presents strange thing maybe bug. In list of argument specified that input is Float tensor but in description of function is saying that input is Int tensor. So in that test I pass Int tensor and jaxtyping invokes error, if I understand log correct.  

And Code Checks is just test for reformat? But I don't understand why it fails. In logits_utils.py and loading_from_pretrained.py I only add one string of code which is looking good and it fails here too. Am I suppose to do something with that?

bryce13950:
> > I see that you are trying to merge to 3.x. Is there a specific reason for doing that? It's not a big deal, but we could probably get this into 2.x if you wanted. There's quite a bit that needs to happen before 3.x goes up, so maybe makes sense to get LLaVA in a release before then.
> 
> Hm, I think that I don't clearly understand how branches are organized here. I have python 3.11 so I thought that it is good idea for making pull request to dev3.x branch. Is it right that 2.x and 3.x are simply versions of project not python? As I understand, dev branch is corresponding to 2.x of project and in dev-3.x you are developing more long-term features. So, you want me to make pull request to dev branch instead of dev-3.x?
> 
> And I think that I've made bad test function in test_hooked_transformer.py. I've write it but it appears not to work in that file. So I've run that function in .ipynb file and tested all cases of inputs for generate function for check if I've broke something. But I've decided to push that function too. Now I see that in generate function presents strange thing maybe bug. In list of argument specified that input is Float tensor but in description of function is saying that input is Int tensor. So in that test I pass Int tensor and jaxtyping invokes error, if I understand log correct.
> 
> And Code Checks is just test for reformat? But I don't understand why it fails. In logits_utils.py and loading_from_pretrained.py I only add one string of code which is looking good and it fails here too. Am I suppose to do something with that?

dev-3.x is for TransformerLens 3.x, which is currently under development. dev is the current stable development branch i.e. where releases currently come from. Once 3.0 is done, current `dev` will become `dev-2.x`, similar to `dev-1.x` in case we ever need to patch anything on the last 2.x release, and `dev-3.x` will be renamed to `dev` as it will become the current stable version. Putting this into 3.0 really isn't that big of a deal, it's just going to take it longer to get into a pip release as opposed to basing what you are doing on 2.x. You did make a change to a utility, which would have to be moved to back port it, but it would not be very difficult to do that.

For the format, there is a handy command you can run from the project root `make format`, which will get all of that fixed up for you automatically.

zazamrykh:
I've made branch from dev and now will make PR to dev branch

PR #829, opened 2025-01-06
Extend Bert support

degenfabian:
# Description

The current implementation of BERT only implements MaskedLanguageModelling and has a certain number of other limitations, like only being able to take tokens as input or not doing any kind of preprocessing like layer norm folding. This PR intends to enhance the BERT support of TransformerLens by addressing these limitations. So far I have added NextSentencePrediction and achieved that the logits of TransformerLens and HuggingFace match.

I will address the other issues mentioned above soon.

There is no specific issue attached to this PR.

# Checklist:


transformer_lens/HookedEncoder.py
```diff
@@ -83,40 +90,160 @@ def __init__(self, cfg, tokenizer=None, move_to_device=True, **kwargs):
...
-
-        one_zero_attention_mask: Optional[torch.Tensor]: A binary mask which indicates which tokens should be attended to (1) and which should be ignored (0). Primarily used for padding variable-length sentences in a batch. For instance, in a batch with sentences of differing lengths, shorter sentences are padded with 0s on the right. If not provided, the model assumes all tokens should be attended to.
+    ) -> Optional[
+        Union[
+            Float[torch.Tensor, "batch pos d_vocab"],
+            Float[torch.Tensor, "batch 2"],
+            str,
+            List[str],
+        ]
+    ]:
```
bryce13950, line 150: Is there anything that can be done to simplify this return type? This function could return 5 different states, and the worst part of that is that two of the potential states are the exact same type of data, but in different shapes. I understand that they will be used in completely different contexts, but it's for that reason that I am kind of wondering why we don't do a bit more with this, and create two distinct modules for the two different use cases. All of the shared functionality could be put into components that are used in both, and it may simply be easier to communicate the usage to end users if they are distinct. The real question is, is there any use case where someone will be using both use cases with the same instance of the module? If there is a valid use case for when that may happen, then leaving it as is is probably fine. However, if it's probably not going to happen, I think I would prefer to find a way to make these more simple and single purpose.

degenfabian: Thanks for your feedback! For the current implementation I based it a little on what HuggingFace does, but also added all these return types so it is more like what users usually expect from TransformerLens.
From HuggingFace, you can either use BertForMaskedLM or BertForNextSentencePrediction, which are both different classes each returning only the output relevant for the task. There is however another architecture, BertForPreTraining, which returns both the logits for MLM and the logits for NSP. I could implement it this way if you'd like, but since the other TransformerLens architectures usually just returns one logits tensor, I implemented it this way to keep it a little more uniform. 

I could imagine something like specifying the task you want to perform when you load the model, and that way we would use a different function based on what the user specified.

I could also use the normal forward function for MLM, since from what I've read it appears to be the more relevant use case, and then do an extra function for NSP, so that way each function would have less return states.

Since I haven't done interpretability research on BERT myself, I'm not sure how relevant having both use cases in the same module is, this was just a design decision based on the BertForPreTraining architecture from HuggingFace, but I'm really happy to adjust this! Would you be happy with either of the options above? Let me know if I should try to do something completely different, but off the top of my head I don't have an idea how we can make it more modular except for those two options.

bryce13950: tl;dr I wanted to share quite a bit of some of the reasoning behind developing TransformerLens API direction with you, so I got pretty into the weeds with quite a few things here. The answer to your actual question is at the bottom. The big explanation is for transparency, and to understand where is the code going over the next year or so.


Your reasons for doing it the way you did it are completely valid. In most cases, you want to keep new features in a similar standard as existing API in something like a library. However, TransformerLens is currently in the process of transitioning away from this style API for a couple of reasons, so let's break it down a bit...

Hugging face has a pretty interesting code base. If you look through the code for transformers quite a bit, you will find that every single model essentially has its own codebase. This has a lot of advantages, namely being that everything exists in isolation, so that you never have to worry about interfering with other implementations if you want to change something. Being able to do that is the ideal way to handle something like this, but able to do that is something that comes with millions of dollars of funding.

TransformerLens in contrast is in essence fitting a square peg into a round hole. There is so much shared code among every implementation, that it becomes almost debilitating to figure out how to add something without affecting existing features. This has resulted in a lot of functionality basically doing everything it possibly can at once, and that has lead to a lot of issues, which are in the process of being fixed.

Ideally, we would have a situation somewhere between current TransformerLens and transformers where we would still have a lot of shared components, and functions, but whenever something is significantly different, we would want to have more implementations of what is now shared functionality to make sure the parts that make them different are as isolated as possible. The advantage of that is that it becomes a lot simpler to write effective tests. 

When you have a function, every single bit of internal logic, and every single variance in the function declaration would be cause for writing a unit test in an ideal situation. e.g. if you had 4 possible return types, and let's say 2 layers of nested logic blocks, you would want to test all of those possible combinations with a unit test. That could mean somewhere between 8 and 64 unit tests in an ideal situation to ensure a function of this complexity is adequately tested in an ideal world.

All that being said, the example you followed to write your code is not following the ideal situation I am describing. That functionality basiclally needs to be rewritten to make it easier to maintain, and less bug prone. The problem is that you cannot change how a function works when going from something like 2.12 to 2.13. Everyone is using this code base relying on the assumption that it works in a certain way. You can change it when going from something like 2.12 to 3.0, because people expect that they need to consult documentation to see what has changed, so you can use that opportunity to make sure everyone knows what has changed. Even so, the functionality we are talking about is so fundamentally important that we can't really change it without being a little bit tactful on how we manage it. This will probably involve writing new functions that do all of the work, then fiting the old functions ontop of the new ones, and adding big warnings that point people in the direction of the new functions for quite some time. 


Now for the question of what do we do for the time being until we get to the point of redoing some of the less than ideal functionality, well we can basically do whatever we want. When you are adding new features, there really isn't a constraint on how you do it. As a code base evolves, you may come up with a completely new, and much better way to do something, and you don't really want to be constrained by trying to fit what you are doing into how old features worked. It's also a great opportunity to experiment with how you wish you could have done something that is already in production, so that you can get a new style into the hands of your end users to see what they think about doing it slightly different. That way you can then go back around to the original old implemetnation later, and redo it with feedback from the community in mind. 

The suggesting you had of keeping the forward function as is, and the adding an additional function is a pretty good option, and it probably aligns with where the overall code is going. This leaves a lot of the heavy lifting in a single function while also allowing our other functions to remain focused on the more direct task. Reworking it in this direction will be completely satisfactory for getting this feature out into the world.


As for the note about not sure how to make it more modular, here's something to think about. What if we took the HookedEncoderDecoder, left it almost entirely bare bones as to what it was doing. Then, let's say we created a seperate classes whose whole purpose was NSP. This class could then take a HookedEncoderDecoder as a dependency, which would allow us to mock out the entire functionality of the HookedEncoderDecoder for the purpose of testing. We can then completely control what we expect should go into our HookedEncoderDecoder and what should be returned by it through unit test expectations, along with allowing us to write much more simple and focused tests for our HookedEncoderDecoder. This sort of setup allows you to really isolate all of the functionality of both the HookedEncoderDecoder and NSP to ensure that they are both behaving exactly as expected in isolation. Then, ideally when you combine them, they should just work. That's the theory anyways.

degenfabian: Hey Bryce! Thank you so much for that message! It was very insightful and I learned a lot reading it, so thank you for taking the time to write this! This all makes perfect sense. Your suggestions were really helpful as well, I now created a separate class for NextSentencePrediction and the return types are a lot more clean because of it. 

However, since the input types are more strict than those from the HookedEncoder for the forward function, I am currently ignoring the types for the NextSentencePrediction class because the type checking fails otherwise. Should I just sacrifice some of the clarity for having less input types we get because of the NextSentencePrediction class in order for the type checking to pass, or could we do something else? I would love some guidance about this!
Also, the implementation of run_with_cache is the same in both classes, but since they have different return types run_with_cache is now also implemented in NextSentencePrediction. Do you have an idea how to make this a little bit more clean or is this just a sacrifice we have to make for clearer types?

bryce13950: OK, so there are a couple changes to be made with it that will resolve the last question here. Do you have time to talk tomorrow, the 24th? I want to talk a bit about how to go about adding tests for this, and then we can discuss a couple revisions that will allow you to full control the typing on it.

degenfabian: That sounds awesome, I'm free tomorrow!

bryce13950:
This looking better, but we do need to add some new tests for this. If you want my help on how to do that, let me know, and we can talk about how we should test these changes.

PR #838, opened 2025-01-19
Enables return of activation cache variables during generation

japhba:
This pull request changes the generate method in `transformer_lens/HookedTransformer.py`, enabling to optionally also return the activation cache in addition to the generated tokens via a return_cache flag.

This is implemented by concatenation to the context dimension of the existing cache as new tokens are being generated.


bryce13950:
Tests are failing, and there hasn't been any movement on it in 4 months. If you are still interested in doing this, please open a new PR

PR #845, opened 2025-01-28
Manually create repr for partial hooks

### LINKED ISSUE #631: [Proposal] Remove the overhead caused by full_hook.__name__ = (hook.__repr__())?
### Proposal 

Remove the full_hook.\__name\__ = (hook.\__repr\__()) line in HookPoint.add_hook, if this is not really necessary.

### Motivation

I am developing a forward hook recently, and use functools.partial. When I run the code I find it is very slow. After a long time of dig up, I realized that because I have a Dict[device, tensor] store some tensors on several device (for not need to copy tensor between devices when running hooks) as a parameter in functools.partial, when running hook.\__repr\__() it will query the \__repr\__ of Dict and then query the tensors inside Dict, which causes a long overhead when hook.\__repr\__() run a lot of times. I spent a long time and found this problem and maybe it is better to just remove it if not necessary? I am not sure here, maybe I just shouldn't write code in this way, what do you think?

### Pitch

Remove full_hook.\__name\__ = (hook.\__repr\__()) if not necessary.

### Alternatives

Avoid using hook.\__repr\__().

### Checklist

- I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo

danbraunai:
# Description

`HookPoint.add_hook()` contained the line:
```
full_hook.__name__ = (
    hook.__repr__()
)  # annotate the `full_hook` with the string representation of the `hook` function
```
This PR instead proposes
```
if isinstance(hook, partial):
    full_hook.__name__ = f"partial({hook.func.__repr__()},...)"
else:
    full_hook.__name__ = hook.__repr__()
```

This isn't an issue when the hook is a regular function, but the original code can be extremely expensive when hook is a partial function that contains arguments which may be large, since calling `partial.__repr__` will call `__repr__` on all of its arguments. A very natural and common usage of hooks is to create a partial function which takes a cache object as an argument. Some of the demos in the repo do this, e.g. `Exploratory_Analysis_Demo.ipynb` has:
```
hook_fn = partial(patch_residual_component, pos=position, clean_cache=cache)
        patched_logits = model.run_with_hooks(
            corrupted_tokens,
            fwd_hooks=[(utils.get_act_name("resid_pre", layer), hook_fn)],
            return_type="logits",
        )
```
where `cache` is a gpt2-small `model.run_with_cache()`.

I'll note that I think it's very unlikely that anyone would rely on a specific name of a pytorch-native hook which this affects. This would only break if a user did some kind of string search for one of the specific arguments of their partial function inside the raw `nn.Module._forward_hooks`.

I have not added tests as the use-case for accessing this name attribute when using a partial function is hyper-specific. But can do if I'm missing a reasonable use-case.

This change makes my code 3 times faster (I don't have a public version of it yet, but it does some model training and uses run_with_hooks twice in the forward pass).

Fixes #631 

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/hook_points.py
+++ b/transformer_lens/hook_points.py
@@ -108,9 +108,13 @@ class HookPoint(nn.Module):
                 module_output = module_output[0]
             return hook(module_output, hook=self)
 
-        full_hook.__name__ = (
-            hook.__repr__()
-        )  # annotate the `full_hook` with the string representation of the `hook` function
+        # annotate the `full_hook` with the string representation of the `hook` function
+        if isinstance(hook, partial):
+            # partial.__repr__() can be extremely slow if arguments contain large objects, which
+            # is common when caching tensors.
+            full_hook.__name__ = f"partial({hook.func.__repr__()},...)"
+        else:
+            full_hook.__name__ = hook.__repr__()
 
         if dir == "fwd":
             pt_handle = self.register_forward_hook(full_hook)


ArthurConmy:
LGTM but, thoughts on this? For this CL, or a future CL. 

Idea: show arg and kwarg names, separated by `*` if both are present.

```python
# Annotate the `full_hook` with the string representation of the `hook` function:
if isinstance(hook, partial):
    # Get function name safely
    func_name = getattr(hook.func, '__name__', hook.func.__class__.__name__)
    
    # Always show args as keyword arguments by getting their names from the signature
    sig = inspect.signature(hook.func)
    param_names = list(sig.parameters.keys())
    
    # Format args as keyword assignments using parameter names
    args_kwstr = ', '.join(f'{param_names[i]}=...' for i in range(len(hook.args)))
    
    # Format normal kwargs
    kwargs_str = ', '.join(f'{k}=...' for k in hook.keywords.keys())
    
    # Combine with * separator if both present
    params_str = args_kwstr
    if args_kwstr and kwargs_str:
        params_str += f', *, {kwargs_str}'
    elif kwargs_str:
        params_str = kwargs_str
    full_hook.__name__ = f"partial({func_name}({params_str}))"
else:
    full_hook.__name__ = hook.__repr__()
```

ArthurConmy (approved):

danbraunai:
Yeah good idea but I think getting the args might be overkill here: If the user knows it's a partial and gets the function name that is wrapped by the partial, they will know what the other argument names are just be the definition of the function. I don't think that can vary.

I'm also just unsure if anyone has ever accessed the `__name__` of the underlying pytorch hook, but maybe I'm missing a use case there.

PR #847, opened 2025-01-30
feat: streaming response for HookedTransformer.generate

hijohnnylin:
# Description

This adds a `stream_output` flag (default `False`) to HookedTransformer.generate which streams the tokens as they're generated, instead of waiting for the whole result to be generated first.

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/HookedTransformer.py
+++ b/transformer_lens/HookedTransformer.py
@@ -11,7 +11,6 @@ a deeper understanding of the internal workings of transformers like GPT-2.
 
 import logging
 import os
-from collections.abc import Generator
 from typing import (
     Dict,
     List,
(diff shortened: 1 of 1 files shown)


hijohnnylin:
still a draft as I have not written docs, tests, etc.

hijohnnylin:
the current implementation doesn't work because `yield` forces the method to become a Generator even if that code path is not used. for now just copying the method.

bryce13950:
Very nice! Looking forward to messing around with this. I will keep an eye out for a test, and then we can get it into a build right away.

bryce13950:
Is this still relevant, or are we passing on this in favor of 908?

hijohnnylin:
This can be closed. THanks!

PR #861, opened 2025-02-11
Fix new weight conversion for BERT model

degenfabian:
# Description

The new weight conversion for BERT, forgets to load in the weights and the bias for the unembedding matrix. This leads to inaccurate outputs as can be seen below. This PR fixes that by simply adding the missing conversions to the weight conversion for BERT.

This PR does not address any specific issue.

These are the output differences before the change:
```
Analyzing final_logits similarity:
Equality: False
Maximum difference: 30.604841
Mean difference: 5.686721
```

And this is after the change:
```
Analyzing final_logits similarity:
Equality: True
Maximum difference: 4.2915344e-05
Mean difference: 5.0457793e-06
```

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/weight_conversion/bert.py
+++ b/transformer_lens/weight_conversion/bert.py
@@ -7,34 +7,70 @@ from transformer_lens.weight_conversion.conversion_utils.conversion_steps import
 
 class BertWeightConversion(ArchitectureConversion):
     def __init__(self, cfg: HookedTransformerConfig) -> None:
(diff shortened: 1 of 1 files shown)


bryce13950:
Thanks for finding this! When I brought in the most recent release of 2.x, I found the issue you found, and I also needed to add the new keys, which overlaps this fix. The rest of your PRs should get merged tomorrow or Friday though.

PR #984, opened 2025-07-27
Create bridge for every module in Mixtral

degenfabian:
# Description

This PR maps a Bridge to every module of Mixtral thus creating hooks for every activation that happens during the forward pass in transformers. In order to make this possible, this PR changed the current code such that nested BlockBridges are now supported.

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/transformer_lens/model_bridge/component_setup.py
+++ b/transformer_lens/model_bridge/component_setup.py
@@ -11,6 +11,7 @@ from transformer_lens.model_bridge.architecture_adapter import ArchitectureAdapt
 from transformer_lens.model_bridge.generalized_components.base import (
     GeneralizedComponent,
 )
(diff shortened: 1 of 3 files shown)


transformer_lens/model_bridge/component_setup.py
```diff
@@ -74,6 +75,12 @@ def setup_submodules(
         original_model: The original model to get components from
     """
     for module_name, submodule in component.submodules.items():
+        if isinstance(submodule, BlockBridge):
```
bryce13950, line 78: There is a variable in the base component called `is_list_item` that should be used to detect things like this. Right now this will only work for a single type of list, but there could be a lot more in the future

degenfabian: Done!

PR #1044, opened 2025-09-06
Add support for layer norm and bias folding

degenfabian:
# Description

Fixes # (issue)

## Type of change

### Screenshots

# Checklist:


transformer_lens/config/TransformerLensConfig.py
bryce13950, line 1: Is there a reason why these config items were added for this?

transformer_lens/model_bridge/bridge.py
```diff
@@ -334,6 +345,373 @@ def set_compatibility_mode(component: Any) -> None:
...
+        def is_rms_norm(component: Any) -> bool:
+            if "rms" in str(type(component)).lower():
+                return True
+            return False
+
+        return_values = apply_fn_to_all_components(self, is_rms_norm)
+
+        if True in return_values.values():
+            return True
+        return False
```
bryce13950, line 367: I honestly think this should be configured at the architecture adapter level. The way it is being determined is not guaranteed to be true all the time. Someone could have a normalization layer that is theoretically an RMS norm, but simply has a different name, and in that case this fails. Additionally, the apply_fn_to_all_components can be pretty expensive. If we do something with it at the start, that's fine but as much as we can avoid using it, the better. We may need to look at all the current architectures and configure them manually.

degenfabian: That sounds good, I'm on it!

Issue #1062, opened 2025-09-17
[Bug Report] torch 2.8.0 makes model generation go haywire


hijohnnylin:
**Describe the bug**
- model.generate works fine on torch 2.7.1 but in 2.8.0 (on macos at least), it goes haywire
- workaround is to force version <=2.8.0 on macos

- macos 15.6.1
- python 3.11.8

I'll dig into this more later (eg test if it affects linux too) but just flagging this for now so it doesn't get lost.

**System Info**

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/TransformerLens/issues) in the repo (**required**)

gaabrielfranco:
I had the same problem. I was going crazy trying to find it. Forcing torch back to 2.7.1 also worked for me.

hijohnnylin:
@gaabrielfranco were you on cuda or mps or cpu?

gaabrielfranco:
@hijohnnylin I was using mps.

jlarson4:
I added a warning for any users that might attempt to run TransformerLens with Torch 2.8.0 on MPS which will now encourage them to either update to 2.9.0 or rollback their version to 2.7.1, whichever works best for them.

Issue #1008, opened 2025-08-16
[Bug Report] Qwen gives nonsensical output when used on 'mps'


A26-Projects:
**Describe the bug**
When using Qwen to generate some text, it gives me the output in other languages (mainly Chinese gibberish, but occasionally some other languages as well.

**This happens only when I try to use 'mps' on my MacBook, but I have no such problem when I ssh into an GPU on the cloud and use 'cuda'.**

**Code example**
 I'm using a method to generate content by taking the logit with the maximum probability (greedy sampling), decoding it to a string and appending it to the output... 

```python
from tqdm import tqdm
from transformer_lens.hook_points import HookPoint
from transformer_lens import HookedTransformer
import torch

DEVICE = torch.device('mps')

def get_model(model_name):
    model = HookedTransformer.from_pretrained_no_processing(model_name, device = DEVICE, dtype=torch.float16, default_padding_side='left', output_hidden_states=True)
    model.eval()
    model.to(DEVICE)
    return model

model = get_model("Qwen/Qwen1.5-4B-Chat")
# model = get_model("Qwen/Qwen1.5-1.8B-Chat")

def tokenize_prompt(model: HookedTransformer, prompt_str: str) -> str:
    
    prompt_message = [
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": prompt_str}
    ]

    prompt_chat_tokenized = model.tokenizer.apply_chat_template(
        prompt_message, tokenize=True, add_generation_prompt=True)
    prompt_chat_str = model.tokenizer.apply_chat_template(
        prompt_message, tokenize=False, add_generation_prompt=True)
        
    return prompt_chat_tokenized, prompt_chat_str

def generate_output(model: HookedTransformer, prompt_chat_str: str, max_new_tokens: int) -> tuple[str, dict, int]:
    output_str = prompt_chat_str
    for i in tqdm(range(max_new_tokens)):
        # Get the logits and cache for the current prompt
        logits, cache = model.run_with_cache(output_str)

        # Get the predicted next token (using argmax for temperature 0)
        next_token = logits[0, -1].argmax()

        # Convert the next token to a string
        next_token_str = model.to_string(next_token)

        # Append the new token to the prompt for the next iteration
        output_str += next_token_str
        
        if next_token.item() == model.tokenizer.eos_token_id:
            break
    
    return output_str

def normal_generation(model, prompt, max_tokens):
[...]

PR #1068, opened 2025-09-27
Fix attn mps

BrownianNotion:
# Description
Hi I'm new here, feedback is welcome/let me know if I've missed anything!

Due to a bug in PyTorch 2.8.0 F.linear for mps https://github.com/pytorch/pytorch/issues/161640, the lines below
```
out = F.linear(
                    z.reshape(z.shape[0], z.shape[1], self.cfg.d_head * self.cfg.n_heads),
                    w,
                    self.b_O,
                )
```
from https://github.com/TransformerLensOrg/TransformerLens/blob/main/transformer_lens/components/abstract_attention.py#L302-L306 
produce incorrect attention outputs on mps. Cpu works fine. I'm on Mac Sequoia 15.6.1. 

I've added a unit test which reproduces the issue in this commit f903629506cfbf692001c0bafa4c1c846c6987b5.

To fix this, I have replaced `F.linear` with `einops.einsum` which is also more consistent with the rest of the class.

Related issues: #1008 #1062 

## Type of change

# Checklist:


tests/unit/components/test_attention.py
```diff
@@ -128,3 +128,45 @@ def test_remove_einsum_from_complex_attn_linear():
 
     # Check if the results are the same
     assert torch.allclose(result_new, result_old, atol=1e-4)
+
+
+@pytest.mark.skipif(
+    not torch.backends.mps.is_available() and torch.__version__ != "2.8.0",
+    reason="Issue with F.linear issue exclusive to mps and PyTorch 2.8"
```
jlarson4: Add a newline character after `PyTorch 2.8`, so that the `https` and the `2.8` do not crash when the strings are concatenated.

PR #1111, opened 2025-11-07
HuBERT support rollout

david-wei-01001:
# Description

This add support for the HuBERT model, extending Transformer Lens to speech/audio models. Codes has been thoroughly tested, test codes can be found in demos/HuBERT_test folder. No additional dependencies needed

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/hubert_hook_test.py
+++ b/tests/unit/hubert_hook_test.py
@@ -17,10 +17,7 @@ def make_sine(sr=SAMPLE_RATE, duration=DURATION_S, freq=440.0, amp=0.1):
 
 
 device = "cuda" if torch.cuda.is_available() else "cpu"
-audio_model = HookedAudioEncoder.from_pretrained(
-    "facebook/hubert-base-ls960",
-    device=device
-)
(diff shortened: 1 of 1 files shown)


david-wei-01001:
The Notebook check is not necessarily failed. The reason why the list of supported models change is because we suppports new models now! Namely HuBERT variabts.

The reason why number of modules changes is because we add new modules to support HuBERT

david-wei-01001:
I fixed import format to the best of I can.

transformer_lens/components/bert_pooler.py
```diff
@@ -34,4 +34,5 @@ def forward(
         first_token_tensor = resid[:, 0]
         pooled_output = torch.matmul(first_token_tensor, self.W) + self.b
         pooled_output = self.hook_pooler_out(self.activation(pooled_output))
+        # pooled_output = self.hook_pooler_out(pooled_output)
```
jlarson4: Please clean up any commented out lines of code

PR #1219, opened 2026-03-28
Fix tests broken by a local GPU

brendanlong:
# Description

Two of the tests fail if you have a local GPU.

`test_apertus` failed because it creates CPU tensors and then compares them to tensors created on the default device. I fixed this by updating the model config to use the CPU as well.

`test_hooked_encoder` has a CUDA-only test (`test_cuda`) which expects an undefined `mlm_tokens` fixture. Presumably it passes in CI because CUDA isn't available. I updated this to use `tokens` instead.

Before:

```text
> poetry run pytest tests/unit/pretrained_weight_conversions/test_apertus.py tests/acceptance/test_hooked_encoder.py 
========================================================= test session starts =========================================================
platform linux -- Python 3.12.9, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/brendanlong/workspace/TransformerLens
configfile: pyproject.toml
plugins: jaxtyping-0.2.19, doctestplus-1.3.0, nbval-0.10.0, cov-5.0.0, anyio-4.5.2, typeguard-4.4.0
collected 29 items                                                                                                                    

tests/unit/pretrained_weight_conversions/test_apertus.py ........F...                                                           [ 41%]
tests/acceptance/test_hooked_encoder.py ................E                                                                       [100%]

=============================================================== ERRORS ================================================================
_____________________________________________________ ERROR at setup of test_cuda _____________________________________________________
file /home/brendanlong/workspace/TransformerLens/tests/acceptance/test_hooked_encoder.py, line 224
  @pytest.mark.skipif(not torch.cuda.is_available(), reason="Requires a CUDA device")
  def test_cuda(mlm_tokens):
E       fixture 'mlm_tokens' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, huggingface_bert, monkeypatch, no_cover, our_bert, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, tokenizer, tokens
>       use 'pytest --fixtures [testpath]' for help on them.

/home/brendanlong/workspace/TransformerLens/tests/acceptance/test_hooked_encoder.py:224
============================================================== FAILURES ===============================================================
___________________________________ TestApertusWeightConversion.test_xielu_params_fallback_defaults ___________________________________

[...]

PR #1230, opened 2026-04-03
Fix type of HookedTransformerConfig.device

brendanlong:
# Description

This is typed as `Optional[str]` but sometimes returns `torch.device`.
Updated the code to just return the `str` instead of wrapping with a
device.

I'm not confident that every function which takes a device will
always be passed a string, so I didn't change functions like
warn_if_mps.

Found while working on #1219.

## Type of change

# Checklist:


transformer_lens/utils.py
brendanlong: I'm actually unsure if we should change the type, or if we should change this function to just return the strings "cuda" or "cpu".

jlarson4: I am of the same mind, I lean towards updating the function to return strings rather than the `torch.device` object.

After looking through the code this morning it appears a majority of the code base is correctly expecting strings, but there are a handful of locations that rely on the `torch.device` implementation. I need to determine if this is a bug or an intentional choice by past maintainers that went undocumented. I'll let you know when I have a plan for the direction.

jlarson4: Ok, after some additional research and review of the code, we should correct this to always be string

brendanlong: @jlarson4 I updated this to change the return value to a string instead of changing the return type.

PR #1231, opened 2026-04-03
Fix HookedTransformerConfig rotary_base types

brendanlong:
# Description

rotary_base is frequently set to floats in the code but was typed as
an int, causing beartype errors if the configs get loaded in a test:
https://github.com/TransformerLensOrg/TransformerLens/blob/9c5a2a81674d5bcefa641c816b66e9827ccdf637/transformer_lens/loading_from_pretrained.py#L1984

HF confgs' allegedly always have rope_theta as a float:
https://github.com/huggingface/transformers/blob/c38b2fb78eaedd4261a0e446f7976345cd1c7f1b/src/transformers/modeling_rope_utils.py#L645

But sometimes they're actually ints, and beartype doesn't consider int
to be a subtype of float:
https://github.com/beartype/beartype/issues/66

This updates the type to Union[float, int] to be accurate while keeping
beartype happy.

See:

## Type of change

# Checklist:


brendanlong:
I'm tempted to add a TypedDict for the configs so this kind of mismatch would be caught by mypy instead of being a runtime error during tests, and I also noticed that there's a bunch of missing type annotations on function return types in loading_from_pretrained.py but wanted to avoid making a massive PR and wasn't sure if those improvements would actually be helpful.

brendanlong:
I could change the type to just `float` but we'd have to sprinkle float() casts everywhere since ints are valid in HF configs despite the float type, but they're invalid in beartype. We can't centralize this in HookedTransformerConfig since we'd need to do the cast before pass it into that class.

PR #1267, opened 2026-04-23
Fix generate() when tokenizer is unset and add regression tests

### LINKED ISSUE #483: [Bug Report] HookedTranformer.generate() with model.tokenizer unset gives pad_token_id error
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
When calling `HookedTranformer.generate()` without initializing the model with a tokenizer and having the default `use_past_kv_cache=True`, I am getting an error 'NoneType' object has no attribute 'pad_token_id'. I believe this is because, when `use_past_kv_cache=True`, the library needs to determine the pad tokens from an input, so it assumes `mode.tokenizer` is also set so it can the pad token id. However, in my use case, I am not using a tokenizer for my model. Maybe there can be a way to manually supply the pad_token_id in this case? 

**Code example**
```
from transformer_lens import HookedTransformer, HookedTransformerConfig
import torch as t

    n_layers=1,
    n_heads=1,
    d_model=4,
    d_vocab=4,
    n_ctx=4,
    d_head=2,
    act_fn='relu'
)

input = t.tensor([[0, 1, 2, 3], [0, 1, 2, 3]], dtype=t.long)
model.generate(input, eos_token_id=0)
```

**System Info**
Describe the characteristic of your environment:
Linux
transformer_lens version 1.12.0

**Additional context**

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

DityaChawla:
## Summary

Fixes tokenizer-free `HookedTransformer.generate()` paths where a tokenizer is not required.

Fixes #483.

## Changes

- removed the unconditional tokenizer assertion in `generate()`
- updated `get_attention_mask` to accept `tokenizer=None`
- added regression tests for tokenizer-free generation

## Validation

- verified tokenizer-free generation works with `stop_at_eos=False`
- verified tokenizer-free generation works with explicit `eos_token_id`
- added regression coverage in `tests/unit/test_generate_no_tokenizer.py`

### DIFF
--- a/tests/unit/test_generate_no_tokenizer.py
+++ b/tests/unit/test_generate_no_tokenizer.py
@@ -100,7 +100,9 @@ def test_generate_without_tokenizer_stop_at_eos_requires_eos_id():
 
     tokens = torch.zeros((1, 5), dtype=torch.long)
     try:
-        model.generate(tokens, max_new_tokens=3, stop_at_eos=True, return_type="tokens", verbose=False)
+        model.generate(
+            tokens, max_new_tokens=3, stop_at_eos=True, return_type="tokens", verbose=False
+        )
         raise AssertionError("Should have raised AssertionError")
     except AssertionError as e:
         assert "eos_token_id" in str(e), f"Unexpected error message: {e}"
--- a/tests/unit/test_generate_no_tokenizer.py
+++ b/tests/unit/test_generate_no_tokenizer.py
@@ -1,0 +1,118 @@
+"""Tests for HookedTransformer.generate() when no tokenizer is set.
+
+Regression test for https://github.com/TransformerLensOrg/TransformerLens/issues/483
+"""
+
+import torch
+
+from transformer_lens import HookedTransformer
+from transformer_lens.config import HookedTransformerConfig
+
+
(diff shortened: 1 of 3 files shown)


jlarson4:
Hi @DityaChawla! Thanks for taking on this change. A couple things before we can get this merged:

- [x] This is based on an older version of TransformerLens, could you please pull down the latest version of `dev` and apply this change there? (`get_attention_mask` now lives at `transformer_lens/utilities/tokenize_utils.py` around line 229)
- [x] Currently the function body does `tokens.ne(tokenizer.pad_token_id)` which crashes if `tokenizer is None`. It already has a `if tokenizer is None: return attention_mask` early return at line 249-250 (all-ones mask) — but that's gated on an implicit check that only fires when tokenizer is truthy.
- [x] Add a test with `use_past_kv_cache=True` explicit and another with `use_past_kv_cache=False` to prove both paths work
- [x] `generate()` calls several utilities that may assume a tokenizer. Worth checking for `self.tokenizer`. in the `generate` method and verifying each is guarded.

PR #1273, opened 2026-04-29
Improved Tokenize & Concatenate

### LINKED ISSUE #1133: [Bug Report] tokenize_and_concatenate doesn't tokenize correctly.
**Describe the bug**

tokenize_and_concatenate slices strings into 20 chunks by character before tokenizing. This can cut a token in two, leading to token pairs that would never normally occur.

**Code example**
I don't have an example, I found this while debugging a larger project.

In a debugger in that method after tokenizing and dropping padding tokens I observed the following:

```
> chunks[2][-10:], chunks[3][:10]
('t on the M', 'ilitary Ne')
> tokenizer.decode([4460])
' Mil'
> tokenizer.decode([337])
' M'
> tokenizer.decode([346])
'il'
> np.where((tokens[:-1] == 337) & (tokens[1:] == 346))[0]
array([79848])
> tokens[79848:79848+2]
array([337, 346]) #  SHOULD NEVER OCCUR

```

I think it's obvious from code inspection what the problem is.

### Checklist

jlarson4:
# Description

PR #1201 updated our chunking system to be whitespace bound, which narrowed the set of cases that #1133 could appear in, but did not completely eliminate the issue. This additional change should further close this gap in coverage.

Fixes #1133 

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -483,6 +483,61 @@ class TestTokenizeAndConcatenate:
                 f"This indicates a word was split across chunk boundaries."
             )
 
+    def test_no_split_tokens_in_no_whitespace_text(self):
+        """No-whitespace multi-doc input — the prior whitespace-lookahead fix
+        fell through to character cuts here. streaming=True keeps all docs in
+        one tokenize_function call so EOS markers actually exist to split on.
+        """
+        from datasets import Dataset
+        from transformers import AutoTokenizer
+
+        tokenizer = AutoTokenizer.from_pretrained("gpt2")
+
+        docs = ["a" * 200 + "MilitaryVehicleEngine" * 100] * 10
+        dataset = Dataset.from_dict({"text": docs})
+
+        result = utils.tokenize_and_concatenate(
+            dataset,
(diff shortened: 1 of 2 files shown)


jlarson4:
After doing more review of @BorisTheBrave's comments on #1133, and reviewing the historical design choices of `tokenize_and_concatenate`, I have made the decision to implement per-doc tokenizing. 

This is optimal for fast tokenizers and slightly slower (but still correct) for slow ones. If a slow-tokenizer user reports a real perf regression, we'll revisit. When `tokenize_and_concatenate` was originally written in 2022, fast tokenizers were in the minority, but that paradigm has shifted in the intervening years.

This change also has a small effect on SentencePiece tokenizers (T5, Mistral). Previously, a BOS token was emitted at every chunk boundary for LLaMA models, and an extra EOS token was emitted at every chunk boundary for T5 models. Any variance perceived by users of those models is due to the new per-doc approach eliminating those unnecessary tokens. In benchmark testing, this actually resulted in a 20-30% performance improvement for SentencePiece tokenizers, due to the removal of the spurious tokens.

Issue #1264, opened 2026-04-22
[Proposal] MPS CI Runner Support


jlarson4:
### Proposal 

Implement MPS CI runners for handling some testing on `main` PRs and releases. Originally proposed in #598, now possible.

### Motivation

A large portion of the community uses Mac laptops, which means providing solid support for MPS could be valuable to a lot of researchers. Having automated tests that confirm these features work at each release would be valuable.

### Pitch

Implement a set of tests for CI that run specifically on PRs merging into `main` (significant bug fixes) and on pushes to main that run a focused set of tests that fit within GitHubs 1GB limited Apple Silicon runners.

- Don't run Notebook tests, these can be ran in Colab for MPS users.
- Don't run any acceptance or integration tests that require large model loads that might exceed our memory limit

### Alternatives

A clear and concise description of any alternative solutions or features you've considered, if any.

### Additional context

Sources for Apple Silicon Support

- [Introducing the new, Apple silicon powered M1 macOS larger runner for GitHub Actions](https://github.blog/news-insights/product-news/introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/)
- [GitHub Actions M1 MPS Out-of-Memory issues on macos-15](https://github.com/orgs/community/discussions/155306)
- [GitHub Actions: Apple silicon (M1) macOS runners are now available in public beta!](https://github.blog/changelog/2023-10-02-github-actions-apple-silicon-m1-macos-runners-are-now-available-in-public-beta/)
- [GitHub Actions now offers free Apple Silicon macOS runners for public repositories](https://discourse.julialang.org/t/github-actions-now-offers-free-apple-silicon-macos-runners-for-public-repositories/109641)
- [Accelerated PyTorch training on Mac - Metal - Apple Developer](https://developer.apple.com/metal/pytorch/)

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

huseyincavusbi:
Hi @jlarson4 , I would love to work on this! Can I take it?

jlarson4:
@huseyincavusbi Absolutely! I'll mark you down as assigned, let me know if you have any questions

huseyincavusbi:
awesome @jlarson4 !

PR #1278, opened 2026-05-02
MPS CI Support

huseyincavusbi:
Hi @jlarson4,

This PR implements **MPS (Metal Performance Shaders) CI Runner Support** as proposed in #1264.

The goal is to provide automated testing for the Apple Silicon research community while working within the limits of GitHub's  Mac runners.

### Key Changes:

*   **New Test Suite:** Added `tests/mps/test_mps_basic.py` with 11 smoke tests covering device detection, core tensor ops on Metal, and `HookedTransformer` forward passes/caching with small models (`TinyStories-1M`).
*   **CI Automation:** Introduced the `mps-checks` job in `.github/workflows/checks.yml`. It uses `macos-latest` and runs only on PRs/pushes to main.
*   **Memory Management:**
    *   Updated `tests/conftest.py` to proactively clear the MPS cache after every test using `torch.mps.empty_cache()`.
    *   Configured the CI to ignore memory-intensive modules (e.g., `model_bridge`) to ensure stability.
*   **Opt-in Mechanism:** Respects `TRANSFORMERLENS_ALLOW_MPS=1` to ensure safe defaults for Mac users.

---

### Type of change

### Checklist:


huseyincavusbi:
Hi @jlarson4, I've updated the PR to address the automated feedback:

- API Stability: Reverted `get_device()` to return `torch.device` objects.
- Type Checks: Updated type hints across model classes to resolve `mypy` failures.
- CI Trigger: Strictly restricted `mps-checks` to the main branch

transformer_lens/utilities/devices.py
```diff
@@ -53,36 +53,36 @@ def _torch_mps_has_known_broken_bug() -> bool:
...
     ``TRANSFORMERLENS_ALLOW_MPS=1`` is set **and** the installed PyTorch
     version meets or exceeds ``_MPS_MIN_SAFE_TORCH_VERSION``.
 
     Returns:
-        torch.device: The best available device (cuda, mps, or cpu)
+        str: The best available device name (cuda, mps, or cpu)
     """
     if torch.cuda.is_available():
-        return torch.device("cuda")
+        return "cuda"
```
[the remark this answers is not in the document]
jlarson4: Roll back any changes related to `get_device`, there were some git history issues that caused this to crop up here for you. 

I discussed this with another contributor and setup #1230 to standardize the type returned by `get_device`. There was a typing inconsistency, `get_device() -> torch.device` only returned a `torch.device` object in the default case, and returned a string in all other instances, creating a situation where the typing was silently wrong. We standardized everything to `str`, but for some reason that change never made it to `dev`, I have integrated it into this branch and it should no longer be a problem.

.github/workflows/checks.yml
```diff
@@ -103,6 +103,60 @@ jobs:
...
+      - name: MPS Availability Check
+        run: |
+          uv run python -c "
+          import torch
+          print(f'PyTorch: {torch.__version__}')
+          print(f'MPS available: {torch.backends.mps.is_available()}')
+          print(f'MPS built: {torch.backends.mps.is_built()}')
+          assert torch.backends.mps.is_available(), 'MPS not available on this runner!'
+          "
+      - name: MPS Unit Tests
```
jlarson4, line 142: Make sure you're passing `$${{ secrets.HF_TOKEN }}` to  any tests that might load models so we can avoid HuggingFace rate limits where possible

.github/workflows/checks.yml
```diff
@@ -103,6 +103,60 @@ jobs:
...
+        run: >
+          uv run pytest tests/unit -v
+          --ignore=tests/unit/model_bridge/
```
jlarson4: We are ignoring a lot of tests here, some should be able to run within our memory limit. Try ignoring only these tests and let me know if you found something while setting this up that necessitates skipping more tests.
```
- name: MPS Unit Tests
  run: >
    uv run pytest tests/unit -v
    --ignore=tests/unit/model_bridge/test_optimizer_compatibility.py
    --ignore=tests/unit/model_bridge/test_gpt_oss_moe.py
    --ignore=tests/unit/model_bridge/test_component_inspection.py
    --ignore=tests/unit/model_bridge/test_key_analysis.py
    --ignore=tests/unit/model_bridge/test_benchmark_gated_hooks_fire.py
    --ignore=tests/unit/model_bridge/test_weight_processing_adapter_paths.py
    --ignore=tests/unit/model_bridge/test_bridge_generate_kv_cache.py
    --ignore=tests/unit/model_bridge/test_bridge_vs_hooked_transformer_patching.py
    --ignore=tests/unit/model_bridge/compatibility/
```

.github/workflows/checks.yml
```diff
@@ -103,6 +103,60 @@ jobs:
...
+      - name: MPS Integration Tests
+        run: >
+          uv run pytest tests/integration -v
```
jlarson4, line 158: We should still be able to run `grouped_query_attention` and `create_hooked_encoder` that are ignored here.

pyproject.toml
```diff
@@ -90,7 +90,10 @@
             "ignore:distutils Version classes are deprecated:DeprecationWarning",
             "ignore:pkg_resources is deprecated as an API:DeprecationWarning",
         ]
-        markers=["slow: marks tests as slow (deselect with '-m \"not slow\"')"]
+        markers=[
+            "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+            "no_mps: marks test as incompatible with MPS device (deselect with '-m \"not no_mps\"')",
```
jlarson4: If we aren't using this `no_mps` tag, we should drop it

huseyincavusbi:
Hi @jlarson4, here are the latest changes:
- Passed ${{ secrets.HF_TOKEN }} to all MPS job steps.
- Switched to targeted ignores for the modules you listed. I also added 3 more skips based on CI failures: generation tests (test_bridge_generation.py, test_bridge_integration.py) hit NaNs on MPS. For GQA, create_hooked_encoder runs and passes, but I kept test_grouped_query_attention ignored. It fails because Metal float arithmetic differs slightly from CPU, breaking strict torch.equal(). All other model_bridge tests now run and pass.
- Standardized to str and cleaned up stale imports.
- Dropped the unused no_mps marker.

Thanks for the review!

jlarson4:
Great work @huseyincavusbi! This looks good to me. One followup thought on the GQA test(s) that are still not working as expected:

GQA precision tests are a test-strictness issue. The diffs are last-bit fp32 noise (33226.9414 vs 33226.9453). `torch.equal` requires bit-exact equality. It should be safe to update the tests to `torch.allclose(..., atol=1e-4, rtol=1e-4)` in `test_grouped_query_attention.py`. Then re-include the test in MPS CI - should pass on CPU and MPS both.

huseyincavusbi:
Good point @jlarson4 ! I was too focused on getting them to pass without failure, but this is definitely a better approach.

On a related note, I noticed some notebook tests also fail intermittently due to similar floating point noise across different runs, even without code changes. For example, [this run](https://github.com/TransformerLensOrg/TransformerLens/actions/runs/25442166174/job/74636139719).
Would you like me to make those more flexible as well, or do you prefer keeping them strict?

Also, I'm seeing Node.js 20 deprecation warnings. I can update the CI workflows to use the latest versions. 
What do you think about these?

jlarson4:
@huseyincavusbi If you want to tackle those in a follow up PR, feel free!

The Notebook Checks that fail intermittently have given me pause. I want to keep some level of numerical comparison, but I haven't searched very deeply for a possible solution. Let me know if you find anything, I am going to merge this work as done! Thanks for taking this on, it looks great!

huseyincavusbi:
Thanks for the merge, @jlarson4! I really enjoyed working on this.
I'll dig into the notebook noise issue for a follow-up PR. I suspect giving them a little flexibility (similar to the allclose approach we used for GQA precision) would be the right balance to stabilize them while still keeping meaningful numerical checks. I'll also update the Node.js versions to clear deprecation warnings.
I'll be back soon with a PR to fix these!

Issue #448, opened 2023-11-14
[Bug Report] Fix `n_params` counts


ArthurConmy:
**Describe the bug**

The `n_params` counts calculated [here](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/HookedTransformerConfig.py#L242) are wrong. For example, LLAMA uses SwiGLU so the 2x factor in the [linked code ](https://github.com/neelnanda-io/TransformerLens/blob/f5a7d455546a88cfdfb26e781d5bd6447e8243eb/transformer_lens/HookedTransformerConfig.py#L242) is wrong.

**Code example**
```python
# Load in Llama-7B

llama.cfg.n_params # 5033164800 ...
```

**System Info**
N/A

**Additional context**
N/A

### Checklist

- [ x ] I have checked that there is no similar [issue](https://github.com/neelnanda-io/TransformerLens/issues) in the repo (**required**)

ArthurConmy:
Unclear what the solution should be. 

There are plausibly three different parameter counts that are helpful
1. Parameters in training 
2. Parameters ignoring embeddings
3. Parameters used now (e.g folding layer norm deletes some parameters)

neelnanda-io:
IMO this should be just total parameters for simplicity and alignment with
the Pythia suite. Who cares about LayerNorm

PR #1284, opened 2026-05-06
Add n_params_total property for total parameter count

DivijChawla:
## Summary

Related to #448.

Issue background: `cfg.n_params` counts only "hidden weight" parameters (attention projections + MLP weights, excluding embeddings, biases, and layer norms) — matching the [scaling laws paper](https://arxiv.org/pdf/2001.08361.pdf) convention. Users computing memory budgets or comparing with HuggingFace's `num_parameters()` were getting confused by the gap. In the original issue thread, the maintainer suggested a total-parameters surface for "simplicity and alignment with the Pythia suite."

This PR adds the missing total-count surface **without breaking the existing API**.

### `transformer_lens/HookedTransformer.py` — new `n_params_total` property

\`\`\`python
@property
def n_params_total(self) -> int:
    return sum(p.numel() for p in self.parameters())
\`\`\`

### `transformer_lens/config/HookedTransformerConfig.py` — docstring clarification

Makes `cfg.n_params`'s exclusions unambiguous and points users to the new property.

### `tests/unit/test_n_params_total.py` — 4 tests

- Equivalence with `sum(p.numel() ...)`
- Strictly larger than `cfg.n_params` (proves embeddings are included)
- Works on both attn-only and full-MLP models
- Returns a Python `int`

## Why this approach

- **Additive** — `cfg.n_params` semantics unchanged; nothing depending on the scaling-laws convention breaks
- **Lives on the model, not the config** — the config can't know actual parameter count without instantiation, so this is the natural home
- **Matches user mental model** — `model.n_params_total` reads exactly like what people expect

## Test plan

- [x] `pytest tests/unit/test_n_params_total.py` — 4 passed
- [x] `black --check` passes
- [x] No existing tests touched

## Out of scope

This PR does not modify the existing `cfg.n_params` formula. If reviewers want that semantics changed (per the original issue discussion), happy to do it as a follow-up.

### DIFF
--- a/tests/unit/test_n_params_total.py
+++ b/tests/unit/test_n_params_total.py
@@ -9,9 +9,9 @@ from transformer_lens import HookedTransformer
 from transformer_lens.config import HookedTransformerConfig
 
 
-def _make_small_model(attn_only: bool = True, gated_mlp: bool = False) -> HookedTransformer:
+def _make_small_model(attn_only: bool = True) -> HookedTransformer:
     """Build a tiny HookedTransformer for fast unit tests."""
-    cfg = HookedTransformerConfig(
+    kwargs = dict(
         n_layers=2,
         d_model=32,
         d_head=8,
@@ -19,10 +19,11 @@ def _make_small_model(attn_only: bool = True, gated_mlp: bool = False) -> Hooked
         n_ctx=16,
         d_vocab=50,
         attn_only=attn_only,
-        d_mlp=64 if not attn_only else None,
-        act_fn="gelu" if not attn_only else None,
-        gated_mlp=gated_mlp,
     )
+    if not attn_only:
+        kwargs["d_mlp"] = 64
+        kwargs["act_fn"] = "gelu"
+    cfg = HookedTransformerConfig(**kwargs)
     return HookedTransformer(cfg)
 
 
--- a/tests/unit/test_n_params_total.py
+++ b/tests/unit/test_n_params_total.py
@@ -5,8 +5,6 @@ where users asked for a total parameter count (including embeddings/biases),
 since cfg.n_params only counts the "hidden weight" subset.
 """
 
-import torch
-
 from transformer_lens import HookedTransformer
 from transformer_lens.config import HookedTransformerConfig
 
--- a/tests/unit/test_n_params_total.py
+++ b/tests/unit/test_n_params_total.py
@@ -1,0 +1,57 @@
+"""Tests for HookedTransformer.n_params_total property.
+
+Related to https://github.com/TransformerLensOrg/TransformerLens/issues/448
+where users asked for a total parameter count (including embeddings/biases),
+since cfg.n_params only counts the "hidden weight" subset.
+"""
+
+import torch
+
+from transformer_lens import HookedTransformer
+from transformer_lens.config import HookedTransformerConfig
+
+
+def _make_small_model(attn_only: bool = True, gated_mlp: bool = False) -> HookedTransformer:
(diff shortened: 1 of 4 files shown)


jlarson4 (commented):
There is a bug with adding individual comments at the moment, so I will resort to making a bulk comment:

Two things I'd like you to update for this PR:

1. We should add a test on a real world model for this issue. GPT2 is cached for CI, we could do this pretty simply with

```
def test_n_params_total_matches_huggingface():
    tl = HookedTransformer.from_pretrained("gpt2", device="cpu")
    hf = AutoModelForCausalLM.from_pretrained("gpt2")
    assert tl.n_params_total == hf.num_parameters()
```

2. `test_n_params_total_matches_sum_of_parameters()` is always true
`model.n_params_total` is doing the same math that `expected` is doing on line 33, meaning this test will always be `true`. It would be more beneficial to compute `expected` by hand for this small test fixture, and compare to `n_params_total`.

jlarson4:
There is one additional request I'd like to make, if you're willing to tackle it:

Can you mirror this change to the `TransformerBridge` system?

Thank you for your contributions @DivijChawla, please feel free to tag me when you've had an opportunity to make these small test changes and I will get this added to the next release

DivijChawla:
@jlarson4 thanks for the thorough review — pushed the changes:

**1. Tautological test fixed.** Replaced `test_n_params_total_matches_sum_of_parameters` with two hand-computed fixtures (`test_n_params_total_attn_only_matches_hand_computed` and `test_n_params_total_with_mlp_matches_hand_computed`). Each test's docstring shows the per-component breakdown (embed + pos_embed + per-block attn/ln/mlp + ln_final + unembed) so the expected constant is verifiable by inspection.

**2. TransformerBridge mirrored.** Added the same `n_params_total` property to `TransformerBridge` in `transformer_lens/model_bridge/bridge.py`.

**3. Real-model test — needs a quick design call.** Tried your suggested check verbatim:

```python
tl = HookedTransformer.from_pretrained("gpt2", device="cpu")
hf = AutoModelForCausalLM.from_pretrained("gpt2")
assert tl.n_params_total == hf.num_parameters()  # FAILS
```

It fails with `163,049,041 != 124,439,808`. The delta is `38,609,233`, of which `d_vocab * d_model = 38,597,376` comes from TL storing `W_E` and `W_U` as separate Parameters while HF GPT-2 ties `lm_head` to `wte`. The remaining ~12K is from layer-norm folding during `from_pretrained`. I tried `from_pretrained_no_processing` and Pythia-70m as alternatives — both still differ from HF (Pythia by ~37K) for similar reasons.

For now I added `test_n_params_total_real_model_gpt2` which loads GPT-2, asserts the property matches `sum(p.numel() for p in tl.parameters())` on the loaded model (proving the property reflects the actual stored params, not just the small fixture), and bounds the count in a sane band. The test docstring documents the HF discrepancy explicitly.

If you'd like a strict HF-parity check, options would be:
- a `from_pretrained` flag that copies HF's tying behavior into the TL model, or
- a helper like `n_params_total_huggingface_equivalent` that subtracts the untied lm_head delta when `cfg.tie_word_embeddings` was True in the source HF config.

Both feel like a separate PR. Happy to do whichever you prefer.

jlarson4:
I think this is fine for a sanity test for now! Thank you for tackling those changes @DivijChawla. The reality is that there are differences in params between TL and HT, and without doing lots of bending over backwards, we aren't going to get them to match exactly. At least this is closer. In the original issue, Neel explicitly said not to bother with LayerNorm, which is one of the elements here that is off.

PR #1290, opened 2026-05-08
fix: Bump CI actions and stabilize flaky notebook checks

huseyincavusbi:
## Description

Two changes:

**1. Node.js deprecation warnings in CI**  
Bumped all GitHub Actions to latest stable versions to clear Node.js 20 deprecation warnings (`checkout@v3-v4`, `cache@v3-v4`, `upload-artifact@v4-v7`, `download-artifact@v4-v8`, `setup-uv@v6-v7`). Also added `workflow_dispatch` to checks.yml for manual runs.

**2. Flaky notebook checks**  
Some notebooks (e.g. Santa_Coder) fail intermittently due to tiny floating-point noise across CI runs (~3.9515 vs ~3.9514). Following the same principle as the GQA `allclose` fix, I added a regex sanitizer to `doc_sanitize.cfg` that truncates floats to 3 decimal places before comparison. This preserves meaningful numerical checks while absorbing last-bit precision noise.

## Type of change

# Checklist:

### DIFF
--- a/demos/doc_sanitize.cfg
+++ b/demos/doc_sanitize.cfg
@@ -13,3 +13,7 @@ replace: HEX-CODE
 [regex4]
 regex: ,\s*device='[^']*'
 replace:
+
+[regex5]
+regex: (\d+\.\d{3})\d+
+replace: \1
--- a/demos/doc_sanitize.cfg
+++ b/demos/doc_sanitize.cfg
@@ -18,6 +18,18 @@ replace:
 regex: ([1-9]\d*\.\d{3})\d+
 replace: \1
 
-[regex6]
-regex: (0\.\d{4})\d+
+[regex6a]
+regex: (0\.000[1-9]\d{2})\d+
+replace: \1
+
+[regex6b]
+regex: (0\.00[1-9]\d{2})\d+
+replace: \1
+
+[regex6c]
+regex: (0\.0[1-9]\d{2})\d+
+replace: \1
+
+[regex6d]
+regex: (0\.[1-9]\d{2})\d+
 replace: \1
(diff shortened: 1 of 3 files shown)


demos/doc_sanitize.cfg
```diff
@@ -12,4 +12,8 @@ replace: HEX-CODE
 
 [regex4]
 regex: ,\s*device='[^']*'
-replace:
\ No newline at end of file
+replace:
+
+[regex5]
+regex: (\d+\.\d{3})\d+
```
jlarson4: The biggest flaw with this: This limits all demo notebook decimal outputs to 3 places. In instances with numbers like `0.001234`, it will resolve to `0.001`, which is a significant loss of precision at that scale. Is there a way to localize this, or improve the precision of the regex so that it is more selective in what it targets?

huseyincavusbi:
Thanks for pointing this out @jlarson4.  I changed it to be scale aware..
What do you think about this now: 
- Numbers ≥ 1 (e.g. 3.9515): truncated to 3 decimals → 3.951
- Numbers < 1 (e.g. 0.001234): truncated to 4 decimals → 0.0012

jlarson4:
@huseyincavusbi Would it be possible have numbers < 1 match to a certain amount of significant figures, say 3 decimals? Examples:

0.001234 to 0.00123
0.01234 to 0.0123

huseyincavusbi:
hey @jlarson4 . Updated the regex to use 3 significant figures for small numbers instead of fixed decimals. So 0.001234 now becomes 0.00123 instead of 0.001, keeps the precision where it actually matters

jlarson4:
This looks great! I am going to merge, thank you for tackling this issue.

PR #1296, opened 2026-05-09
Fix HuggingFace API rate limiting in CI (#1291)

### LINKED ISSUE #1291: [Proposal] CI HuggingFace Call Reduction
### Proposal 

Reduce HuggingFace calls done during the test suite in CI (Compatibility checks for 3.10, 3.11, 3.12 and Full Code Coverage Checks)

### Motivation

Currently if more than one PR, branch update, or release are running at one time, we hit a HuggingFace rate limit error 429, which causes CI to fail erroneously.

### Pitch

There are multiple ways we could go about this:
1. Identifying situations where HF calls are made that could be promoted to test fixtures (one call reused for many tests), or removed altogether in favor of creating small synthetic models that can be tested for the same details without losing strength of coverage
2. Determining if it is possible to have GitHub queue these CI checks, to ensure more than one of them is not running at once across multiple action workflows. We can use things like `cancel-in-progress` or concurrency checks to reduce the frequency of runs or restrict them from running at the same time. The flaw in limiting concurrency is an elongated feedback loop for contributors, we'd like to avoid this if possible.

### Additional context

[GitHub concurrency docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)

### Checklist

ak91456:
# Description

  Fixes #1291 — CI jobs were hitting HTTP 429 (Too Many Requests) from HuggingFace Hub when multiple PRs or pushes triggered simultaneous workflow runs.

  Even when model files are cached locally, `huggingface_hub` makes lightweight "resolve" API calls by default to check freshness. With 3 matrix Python
  versions + coverage + notebook jobs all running in parallel across multiple PRs, these calls exceeded HF's rate limit.

  **Changes:**
  - **Concurrency group** (`checks.yml`): Cancels the stale in-progress run on the same PR branch when a new push arrives. Push-to-main, tags, and
  `workflow_call` (release) events are exempt and never cancelled.
  - **Retry with backoff** (`hf_utils.py`): `download_file_from_hf` now retries up to 3 times (10s / 20s / 30s waits) on HTTP 429, matching the pattern
  already used in `hf_scraper.py`.

  Fixes #1291

  ## Type of change

  # Checklist:


transformer_lens/utilities/hf_utils.py
```diff
@@ -75,13 +76,27 @@ def download_file_from_hf(
...
+                repo_id=repo_name,
+                filename=file_name,
+                subfolder=subfolder,
+                cache_dir=cache_dir,
+                **select_compatible_kwargs(kwargs, hf_hub_download),
+            )
+            break
+        except Exception as exc:
+            if "429" in str(exc) and attempt < max_retries:
+                wait = 10 * (attempt + 1)
```
jlarson4, line 92: If many workers hit a 429 at the same time, all will retry at the same time, creating a repeated loop of failures. Can we add some random variance to the wait duration?

Additionally, we might want to increase the base duration of `wait` in this situation. HF's API timeout window is 5 minutes, if we have a large volume of requests at the same time we are going to want to spread out the calls on each retry using longer waits + variance to increase our chances of success.

jlarson4:
This is a solid start to fixing #1291. We will still need to do additional work on bundling and creating fixtures for model requests to reduce the request load further, but this will definitely help reduce the number of false failures caused by timeout. Thank you for contributing!

Issue #1302, opened 2026-05-12
[Proposal] Additional Architecture Adapter tests


jlarson4:
### Proposal 

Add config tests for remaining architecture adapters

### Motivation

Only about a third of our Architecture Adapters have focused tests in `tests/unit/model_bridge/supported_architectures/`. Use the existing available test suites as examples for how the new tests should be written.

### Pitch

Create a small test suite for each of the remaining architecture adapters that do not have one at present. Feel free to tackle as few or as many at once as you'd like, just leave a comment stating which architecture(s) you're working on to reduce duplication of effort.

Architectures that need tests
- [x] llama
- [x] mistral (@chandrudp29  **completed**)
- [x] mixtral (@RecreationalMath **completed**)
- [x] gemma1 (@jlarson4 **completed**)
- [x] gemma2 (@jlarson4 **completed**)
- [x] gpt2 (@sunny1401 **completed**)
- [x] qwen (@has9800 assigned)
- [x] qwen2 (@Rishik00 **completed**)
- [x] qwen3 (@sunny1401 **completed**)
- [x] t5 (@RecreationalMath assigned)
- [x] gpt_oss (@RecreationalMath **completed**)
- [x] bloom
- [x] opt (@willytop8 **completed**)
- [x] phi
- [x] phi3
- [x] neo (@chandrudp29  **completed**)
- [x] neox  (@chandrudp29  **completed**)
- [x] gptj (@along-l **completed**)
- [x] bert
- [x] apertus
- [x] hubert
- [x] olmo
- [x] olmo2 (@RecreationalMath **completed**)
- [x] olmo3 (@RecreationalMath assigned)
- [x] olmoe (@RecreationalMath **completed)**
- [x] openelm (@chandrudp29  **completed**)
- [x] granite
- [x] granite_moe
- [x] granite_moe_hybrid
- [x] qwen3_next
- [x] stablelm (@RecreationalMath  **completed**)
- [x] llava_next  (@sunny1401 **completed**)
- [x] llava_onevision (@chandrudp29  **completed**)
- [x] deepseek_v3
- [x] deepseek_v2 (@SanjidMzi **completed**)
- [x] mamba
- [x] mamba2

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

sunny1401:
I would like to take a stab at this if no one else is working. Can start with gpt2?

[...]

PR #1305, opened 2026-05-15
Add OPT architecture adapter tests

willytop8:
Adds tests for the OPT architecture adapter covering config defaults, post-norm behavior, weight conversions, component mappings, the OPT-350m projection path, and factory registration.

Tested with:
- uv run pytest tests/unit/model_bridge/supported_architectures/test_opt_adapter.py

Contributes to #1302


jlarson4:
Looks great, thanks @willytop8

PR #1329, opened 2026-05-26
Add Mixtral architecture adapter tests

RecreationalMath:
# Description

Adds a unit test suite for `MixtralArchitectureAdapter` under `tests/unit/model_bridge/supported_architectures/`, following the existing adapter-test pattern (modelled on the qwen3_moe and qwen2 suites). It needs no model downloads or real checkpoints, it uses tiny programmatic `TransformerBridgeConfig` objects, plus small synthetic tensors and a fake attention module for the behavioural tests, so it runs on CPU in seconds.

The suite (49 tests) covers:
- Adapter config defaults (RMSNorm, rotary, gated MoE MLP, `final_rms=False`).
- Weight conversions: QKVO weights plus Q/K/V biases, with GQA-aware head counts and the no-`n_key_value_heads` fallback.
- Numerical round-trips: the rearrange conversions are actually run on synthetic HF-shaped weight and bias tensors, asserting the split-head output shapes and lossless reversion.
- Component-mapping structure, bridge types, and HF module paths, including the `block_sparse_moe` MoE bridge with its `gate` router and the absence of Q/K-norm.
- GQA forward hook shapes: a fake attention module wired into the bridge confirms Q surfaces `n_heads` while K/V surface `n_key_value_heads`.
- `setup_component_testing` rotary-embedding wiring, eager-attention forcing, and robustness on a minimal HF model.

Contributes to #1302 (Mixtral checkbox).

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_mixtral_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_mixtral_adapter.py
@@ -1,0 +1,589 @@
+"""Unit tests for the MixtralArchitectureAdapter (download-free: tiny programmatic configs
+plus small synthetic tensors and a fake attention module, no real checkpoints).
+
+Covered:
+- Adapter config defaults (RMSNorm, rotary, gated MoE MLP).
+- Weight conversions: QKVO weights plus Q/K/V biases, with GQA-aware head counts.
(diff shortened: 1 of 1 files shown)


jlarson4:
Excellent work on this @RecreationalMath! Merging

PR #1330, opened 2026-05-26
added olmo adapter test

Rishik00:
## Description
Adds unit coverage for the OLMo architecture adapter.

## Changes
- Tests OLMo config defaults, component mapping, and HF module paths.
- Verifies Q/K/V/O weight conversions use the right head counts for GQA.
- Adds attention coverage for OLMo attention forward execution and Q/K/V/Z hook shapes.
- Tests prepare_model() clamp patching and setup_component_testing() rotary/eager-attention wiring.
- Checks factory registration for OlmoForCausalLM.

To run the test: uv run pytest tests\unit\model_bridge\supported_architectures\test_olmo_adapter.py 
All tests have passed. 

Relevant to Issue #1302 @jlarson4

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_olmo_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_olmo_adapter.py
@@ -44,7 +44,7 @@ from transformer_lens.model_bridge.supported_architectures.olmo import (
     OlmoArchitectureAdapter,
 )
 
-# Review for the whole file: would appreciate some comments and docstrings for each function/class highlighting what they do
+
 def _make_cfg(
     n_heads: int = 4,
     n_key_value_heads: int | None = 2,
@@ -54,6 +54,7 @@ def _make_cfg(
     d_vocab: int = 100,
     n_ctx: int = 64,
 ) -> TransformerBridgeConfig:
+    """Build a minimal bridge config for OLMo adapter unit tests."""
     return TransformerBridgeConfig(
         d_model=d_model,
         d_head=d_model // n_heads,
@@ -69,11 +70,13 @@ def _make_cfg(
 
 @pytest.fixture
 def cfg() -> TransformerBridgeConfig:
+    """Provide a small default config shared across OLMo adapter tests."""
     return _make_cfg()
 
 
 @pytest.fixture
 def adapter(cfg: TransformerBridgeConfig) -> OlmoArchitectureAdapter:
+    """Instantiate the adapter under test from the shared config fixture."""
     return OlmoArchitectureAdapter(cfg)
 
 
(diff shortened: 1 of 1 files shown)


tests/unit/model_bridge/supported_architectures/test_olmo_adapter.py
```diff
@@ -0,0 +1,450 @@
...
+        blocks = self._mapping(adapter)["blocks"]
+        mlp = blocks.submodules["mlp"]
+        assert isinstance(mlp, GatedMLPBridge)
+
+        assert mlp.name == "mlp"
+        assert mlp.submodules["gate"].name == "gate_proj"
+        assert mlp.submodules["in"].name == "up_proj"
+        assert mlp.submodules["out"].name == "down_proj"
+
+a
```
jlarson4: Character artifact that needs cleaning up

Rishik00: Noticed it as well, pushing again. Apologies!

jlarson4:
Just one small artifact that slipped into your PR. Also, don't forget to run `make check-format` & resolve any issues to make sure your code passes CI!

tests/unit/model_bridge/supported_architectures/test_olmo_adapter.py
jlarson4: Excellent! While you're pushing the format fix, please clean up the five `# Review:` comments left inline in the test file (line ~50 "would appreciate comments and docstrings", line ~177 "what is native autograd", line ~184 "what does this mean", and the two "same questions as above" repeats). They look like notes for the reviewer rather than code comments. I am happy to answer them in a PR comment, but they shouldn't ship.

To answer the underlying questions while we're here:

- `use_native_layernorm_autograd=True` tells `NormalizationBridge` to route gradients through PyTorch's native LayerNorm autograd rather than a custom path. It is needed for numerical stability and to keep gradients matching HF's reference behavior.
- On the "more docstrings" request: agreed, the Mixtral test PR (#1329) is a good template. It includes class-level docstrings explaining what each block of tests locks down, like `TestMixtralWeightConversionRoundTrips`'s "Run the rearrange conversions on synthetic HF-shaped tensors. The pattern/axis assertions only check metadata. These confirm the conversions actually reshape realistic weight and bias tensors and revert losslessly." Worth adopting that style here.

One small addition worth doing while you're touching the file:

- Add a round-trip test for the rearrange conversions (apply `convert` then `revert`, assert exact equality on a synthetic tensor). Currently you have shape assertions, but not the round-trip, which is the test that will catch a broken rearrange. The [Mixtral PR](https://github.com/TransformerLensOrg/TransformerLens/pull/1329/changes#diff-aef19edfdaef618a31241cc509b3345639eb95728d47aec166af2957ad795fb5R276) also has an example of this

Rishik00: I understand and fully agree on that, usually I'd add those comments for me to further dig into the codebase or let an agent go off and help me understand them. I remove them before commits but forgot to do so this time! Will get them removed in the next commit for the artifact! 

Thank you for taking the time to answer them regardless, appreciate it. Will get the round trip test added as well.

Rishik00:
The CI passes cleanly. Will get the round tests added in a few hours.

jlarson4:
@Rishik00 Another contributor, @TensorCruncher, accurately pointed out that architectures like Olmo do not at present require a roundtrip test. I am going to close and merge the PR as is, thank you for your work on this.

PR #1314, opened 2026-05-19
Add GPT-J architecture adapter tests

along-l:
# Description

Add unit tests for `GptjArchitectureAdapter`, contributing to the architecture adapter test coverage effort in #1302.

Coverage includes:
- Config attribute validation (6 required architecture flags)
- Component mapping structure (top-level + block-level + attn/mlp submodules)
- Weight conversion keys and rearrange patterns
- Architecture guards (no `pos_embed`, no top-level `rotary_emb`, no `ln2`, no MLP gate, exactly QKVO conversion keys)
- Factory registration (`GPTJForCausalLM` → `GptjArchitectureAdapter`)

49 tests, all passing locally via `uv run pytest`.

Related to #1302

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #1333, opened 2026-05-27
Add OLMoE architecture adapter tests

RecreationalMath:
# Description

Adds a unit test suite for `OlmoeArchitectureAdapter` under `tests/unit/model_bridge/supported_architectures/`, following the existing adapter-test pattern (modeled on the qwen2 / qwen3_moe / mixtral suites). It needs no model downloads or real checkpoints: it uses tiny programmatic `TransformerBridgeConfig` objects, plus small synthetic tensors and a fake attention module for the behavioral tests, so it runs on CPU in seconds.

The suite (50 tests) covers:
- Adapter config defaults (RMSNorm, rotary, gated MoE MLP, `final_rms=False`, eager attention).
- Weight conversions: the four QKVO weights (OLMoE has no projection biases), with GQA-aware head counts and the no-`n_key_value_heads` fallback.
- Numerical round-trips: the rearrange conversions are actually run on synthetic HF-shaped weight tensors, asserting the split-head output shapes and lossless reversion.
- Component-mapping structure, bridge types, and HF module paths, including the Q/K-norm RMSNorm submodules and the MoE bridge with its `gate` router.
- GQA forward hook shapes: a fake attention module (carrying OLMoE's pre-reshape Q/K norms) wired into the bridge confirms Q surfaces `n_heads` while K/V surface `n_key_value_heads`.
- `setup_component_testing` rotary-embedding wiring, eager-attention forcing, and robustness on a minimal HF model.
- `prepare_model` in-place-clamp patching: a no-op when `clip_qkv` is unset, when set it wraps attention forward so `clip_qkv` is disabled during the call and restored afterwards, and it tolerates a model without layers.

Contributes to #1302 (OLMoE checkbox). 

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_olmoe_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_olmoe_adapter.py
@@ -189,7 +189,7 @@ class TestOlmoeAdapterConfig:
         assert adapter.cfg.positional_embedding_type == "rotary"
 
     def test_final_rms_is_false(self, adapter: OlmoeArchitectureAdapter) -> None:
-        """OLMoE does not apply the final-RMS fold that Qwen-style adapters use."""
+        """OLMoE does not apply a final RMS fold."""
         assert adapter.cfg.final_rms is False
 
     def test_gated_mlp_is_true(self, adapter: OlmoeArchitectureAdapter) -> None:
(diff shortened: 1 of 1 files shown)


RecreationalMath:
The failing check is the Othello_GPT notebook at Cell 4, the `import transformer_lens` cell, which is the same cell flagged [previously](https://github.com/TransformerLensOrg/TransformerLens/pull/1314#issuecomment-4488788465). The mechanism here is a bit different from the rate-limit flake noted there: `nbval` fails it with `Unexpected output fields from running code: {'stderr'}`. The cell's saved output is empty, but in the CI environment the import emits a warning on stderr, so nbval sees an extra output field and fails the cell.

Running the exact CI command locally (`pytest --nbval-sanitize-with demos/doc_sanitize.cfg demos/Othello_GPT.ipynb`) passes all 13 cells with the same package versions CI used, so the stderr only appears in the CI runner environment. This PR only adds a new file under `tests/unit/`, so it can't affect that cell.

jlarson4:
Thanks for taking this on @RecreationalMath! Your tests look great, I will merge them shortly.

The Othello GPT notebook failure is a known issue, I reran the CI test and your code passed on a second run. Thanks for documenting the source of the problem, I'll see if we can't get that cleaned up.

Issue #1322, opened 2026-05-22
[Bug Report] [macOS-arm64] Cached eager attention NaNs in transformers v5 — blocks bridge KV-cache generation


jlarson4:
## Summary

`TransformerBridge.generate()` produces all-NaN logits at the cached step on macOS-arm64 + transformers v5 + PyTorch 2.7.1, even though the bridge already forces eager attention, the cache contents are finite, and the model is on CPU. The same forward computation without `past_key_values` returns correct logits. The bug appears to live in HuggingFace transformers' cached eager-attention path (and/or PyTorch's CPU matmul on Apple Silicon), not in transformer_lens.

Two tests are skipped on macOS-arm64 to unblock the MPS CI job; they remain active on Linux:

- `tests/unit/model_bridge/test_bridge_generate_no_tokenizer.py::test_generate_without_tokenizer_stop_at_eos_false_kv_cache`
- `tests/unit/model_bridge/test_bridge_generate_no_tokenizer.py::test_generate_return_type_str_without_tokenizer_errors`

## Reproduction (CI only - have not tried reproduction locally on M1 Mac)

CI matrix where it fires: `macos-latest` (GitHub Actions managed Apple Silicon) + Python 3.11.9 + PyTorch 2.7.1 + transformers 5.8.1.

```python
import torch
from transformer_lens.model_bridge import TransformerBridge

bridge = TransformerBridge.boot_transformers("distilgpt2", device="cpu")
bridge.tokenizer = None
tokens = torch.tensor([[15496, 11, 314, 1101, 257]], dtype=torch.long)

# NaNs on macOS-arm64 CI; finite locally and on Linux CI:
bridge.generate(
    tokens,
    max_new_tokens=3,
    stop_at_eos=False,
    use_past_kv_cache=True,
    return_type="tokens",
    verbose=False,
)
```

Direct HF reproduction (no bridge involved):

```python
o0 = bridge.original_model(tokens, use_cache=True)
next_id = o0.logits[:, -1, :].argmax(-1, keepdim=True)

# NaN on macOS-arm64 CI, finite elsewhere:
o1 = bridge.original_model(
    next_id,
    past_key_values=o0.past_key_values,
    use_cache=True,
    attention_mask=torch.ones((1, 6), dtype=torch.long),
    position_ids=torch.tensor([[5]], dtype=torch.long),
)

# Always finite (same effective input):
bridge.original_model(torch.cat([tokens, next_id], dim=1))
```

[...]

Issue #1336, opened 2026-05-27
[Bug Report] run_with_cache(device=...) permanently moves the model and leaves cfg.device stale


RecreationalMath:
**Describe the bug**
On a single-device model, `TransformerBridge.run_with_cache(input, device=...)` moves the underlying model (and the input tensors) to that device and never moves it back. The `device=` argument is meant to choose where cached activations are stored, not to relocate the model. Both the legacy [`get_caching_hooks`](https://github.com/TransformerLensOrg/TransformerLens/blob/v3.2.1/transformer_lens/hook_points.py#L799) (which documents `device` as "the device to store on") and [`ActivationCache.to`](https://github.com/TransformerLensOrg/TransformerLens/blob/v3.2.1/transformer_lens/ActivationCache.py#L192-L211) (whose `move_model` is deprecated) confirm this. After the call, model lives on `device` while `cfg.device` still reports the original device, and any subsequent forward/generate fails.

**Code example**
```python
import torch
from transformer_lens.model_bridge import TransformerBridge

m = TransformerBridge.boot_transformers("distilgpt2", device="mps")  # or "cuda"
toks = m.to_tokens("hello world")
_, cache = m.run_with_cache(toks, device="cpu")        # intent: offload the cache to CPU

print(next(m.original_model.parameters()).device)      # cpu  <- the MODEL was moved
print(m.cfg.device)                                     # mps  <- now stale / inconsistent
m.generate(m.to_tokens("again"), max_new_tokens=3)      # RuntimeError: Placeholder storage has not been allocated on MPS device!
```
[...]

PR #1337, opened 2026-05-27
Add `return_cache` option to `TransformerBridge.generate`

### LINKED ISSUE #697: How to get the hook while generating new tokens?
### Question

My prompt is "Which instrument does Henry Halstead mainly play? Please answer an instrument name. Answer: ", which is a question to LLM.
I want to get the cached hidden states of the _**LLM responsed tokens**_ while LLM is generating the response. How to do it?

On one hand, the code` logits, cache = model.run_with_cache(prompt, return_cache_object=True) ` only cache the hidden states of the prompt, because it dosen't run the generate function.

On the other hand, the code `output = model.generate(prompt, do_sample = False, max_new_tokens = 20)` only get the generated tokens or sentences, I can't get the Activation_cache of the generated Answer.

So how can I obtain the output answer and the Acativation_cache corresponding to the answer at the same time **_during one reasoning process_**?

RecreationalMath:
# Description
Adds an opt-in `return_cache` flag to `TransformerBridge.generate()`. When `return_cache=True`, `generate` returns `(output, cache)` where `cache` is a standard `ActivationCache` over the full prompt+generated sequence, identical to `run_with_cache(output)`. This resolves the gap in #697, where `run_with_cache` only covers the prompt and `generate` returns no activations. A `names_filter` argument lets callers scope the cache, and a `device` argument offloads the returned cache to another device (e.g. CPU); the cache over prompt+max_new_tokens can be large, so the docstring notes the memory cost.

Semantics are "recompute one clean forward over the generated sequence," so the cache is consistent with the rest of TransformerLens, includes attention patterns and all hook points, and avoids the cached-eager-attention path behind #1322. For a causal LM this is numerically identical to capturing during generation (verified), without the ragged per-step shapes. This PR covers single-sequence decoder-only text; encoder-decoder / SSM / multimodal / batched / inputs_embeds raise a clear error pointing to the `run_with_cache`-on-output workaround. Capturing during generation (for active-hook/steering scenarios) remains available via `with model.hooks(...)` around `generate` and can be added later as an explicit opt-in.

Fixes #697

## Type of change

# Checklist:

### DIFF
--- a/tests/integration/model_bridge/test_bridge_generate_return_cache.py
+++ b/tests/integration/model_bridge/test_bridge_generate_return_cache.py
@@ -1,0 +1,155 @@
+"""Tests for TransformerBridge.generate(return_cache=True).
+
+return_cache makes generate() also return an ActivationCache for the full prompt +
+generated sequence, identical to run_with_cache(output) (issue #697). v1 supports
+single-sequence, decoder-only text generation; other paths raise NotImplementedError.
+
+Uses distilgpt2 (CI-cached).
+"""
+
+import warnings
+
+import pytest
+import torch
+
+
+@pytest.fixture()
+def bridge(distilgpt2_bridge):
+    """Alias the session fixture for concise test signatures."""
+    return distilgpt2_bridge
+
+
+# Representative position-indexed hooks present in the non-compat distilgpt2 bridge cache.
+# All have the sequence axis at dim 1, so the cache length is shape[1].
+_PROBE_HOOKS = [
+    "hook_embed",
+    "blocks.0.hook_resid_pre",
+    "blocks.0.hook_resid_post",
+    "blocks.0.hook_mlp_out",
+    "blocks.0.attn.hook_q",
+    "blocks.0.attn.hook_z",
+    "blocks.5.hook_resid_post",
+    "ln_final.hook_normalized",
+]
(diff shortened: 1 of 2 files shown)


RecreationalMath:
Heads-up on a small follow-up, not a blocker for this PR.

The `device=` offload here moves the cache tensors directly (`cache.cache_dict = {k: v.to(device) ...}`) rather than passing `device=` into `run_with_cache`. That is deliberate as `run_with_cache(device=)` currently moves the whole model to the cache device and never restores it (filed as #1336). The direct move is correct and side-effect-free, but it offloads the cache after it is built, so it does not reduce peak memory.

Once #1336 is fixed, this can be simplified to a `run_with_cache(output_tokens, names_filter=names_filter, device=device)` passthrough, which offloads at capture time and therefore lowers peak memory for large caches.

I'd recommend keeping this as a self-contained version and switching to the passthrough in a small follow-up PR once #1336 lands. But if you prefer to hold this PR until #1336 is in and use the passthrough here directly, let me know @jlarson4

jlarson4:
I agree with your assessment here, and think it is fine to merge this as-is with the temporary solution. Typically, I'd ask you to add a note to #1336 that it should update this as a side-effect of the solution, but since you're the one handling that issue, I'll trust that you take care of it. 

Thank you for the thorough investigation of both this issue and the new one you discovered! Great work

PR #1339, opened 2026-05-28
Update broken Slack link again

APatelUIUC:
The Slack invite links in `README.md` and `.github/ISSUE_TEMPLATE/question.md` both point to expired invites (Slack invites cap at 400 uses). 
This is the same issue Neel noted here: https://github.com/TransformerLensOrg/TransformerLens/pull/742

https://join.slack.com/t/opensourcemechanistic/shared_invite/zt-3z9o0hxjl-MDX9pbATO2qESOazNDLpdQ


jlarson4:
Thanks for updating this link! It will be included in the next release

PR #1340, opened 2026-05-28
Fix FactoredMatrix indexing returning empty result for -1 index

Kymi808:
# Description

`FactoredMatrix.__getitem__` routes an integer index into the matrix (ldim/rdim) dimensions through `_convert_to_slice`, which turns an int `v` into `slice(v, v + 1)`. For `v == -1` this becomes `slice(-1, 0)` — an **empty** slice — so indexing the last row/column with a negative index silently returns a `(0, ...)` tensor instead of the requested element.

```python
import torch
from transformer_lens import FactoredMatrix

fm = FactoredMatrix(torch.rand(2, 4, 3), torch.rand(2, 3, 5))  # shape (2, 4, 5)

fm[0, -1].AB.squeeze().shape      # got (0, 5), expected (5,)
fm[0, 1, -1].AB.squeeze().shape   # got (0,),  expected ()
```

Only `-1` is affected; other negative indices (`-2`, `-3`, …) work because `v + 1` stays negative and produces a valid slice.

## Fix

In `_convert_to_slice`, use `None` as the slice stop when `v == -1`, so `slice(-1, None)` keeps the final element. Positive indices and other negatives are unchanged.

## Type of change

# Checklist:

- [x] I have added tests that prove my fix is effective (negative indices on each matrix dimension in `tests/unit/factored_matrix/test_get_item.py`; they fail on `main` and pass with this change)
- [x] New and existing unit tests pass locally with my changes (`pytest tests/unit/factored_matrix/` → 147 passed, 1 skipped)


jlarson4:
Excellent work on this @Kymi808! Thanks for your contribution, merging now

PR #1341, opened 2026-05-28
Add GPT-OSS architecture adapter tests

RecreationalMath:
# Description

Adds a unit test suite for `GPTOSSArchitectureAdapter` under `tests/unit/model_bridge/supported_architectures/`, following the existing adapter-test pattern (modeled on the merged mixtral and olmoe suites). It needs no model downloads or real checkpoints: it uses tiny programmatic `TransformerBridgeConfig` objects, plus small synthetic tensors and a fake attention module for the behavioral tests, so it runs on CPU in seconds.

The suite (44 tests) covers:
- Adapter config defaults (RMSNorm, rotary, gated MoE MLP, `eps_attr = "variance_epsilon"`).
- Weight conversions: the four QKVO weights (GPT-OSS has no projection biases), with GQA-aware head counts and the no-`n_key_value_heads` fallback.
- Numerical round-trips: the rearrange conversions are actually run on synthetic HF-shaped weight tensors, asserting the split-head output shapes and lossless reversion.
- Component-mapping structure, bridge types, and HF module paths, including the MoE bridge with no exposed submodules (the entire MoE block is wrapped opaquely).
- GQA forward hook shapes: a fake attention module wired into the bridge confirms Q surfaces `n_heads` while K/V surface `n_key_value_heads`.
- `setup_hook_compatibility` rotary-embedding wiring on a bridge model: a no-op when `bridge_model` is None or lacks `rotary_emb`, sets rotary on each block's attention bridge when present, skips blocks without attention, and `setup_no_processing_hooks` (the backward-compat alias) produces the same wiring.

Contributes to #1302 (GPT-OSS checkbox).

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_gpt_oss_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_gpt_oss_adapter.py
@@ -1,0 +1,548 @@
+"""Unit tests for the GPTOSSArchitectureAdapter (download-free, tiny programmatic configs
+plus small synthetic tensors and a fake attention module, no real checkpoints).
+
+Covered:
+- Adapter config defaults (RMSNorm with `variance_epsilon`, rotary, gated MoE MLP).
+- Weight conversions: QKVO weights (no biases) with GQA-aware head counts.
+- Numerical round-trips: the rearrange conversions actually reshape and revert losslessly.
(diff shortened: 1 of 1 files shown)


jlarson4:
Excellent test suite, thank you @RecreationalMath

Issue #1325, opened 2026-05-26
[Test Coverage] Parameterize Cohere `logit_scale` fold test with a non-trivial scale


jlarson4:
### Summary

The Cohere `logit_scale` fold test silently skips itself when `logit_scale == 1.0` because the fold becomes a no-op, but this means the test never exercises the fold logic it claims to cover.

### Affected test

- `tests/integration/model_bridge/test_cohere_adapter.py:217` has inline `pytest.skip("logit_scale=1.0 — fold is a no-op, skip this check")`

### What the test verifies (in intent)

That folding `logit_scale` into the unembedding correctly untiles the embed/unembed weights so they're no longer identical post-fold.

### Acceptance criteria

- [ ] The test parameterizes `logit_scale` with at least one non-1.0 value (and optionally `1.0` to keep the no-op case as a regression guard).
- [ ] The non-1.0 case exercises the fold logic and the assertion fires (no skip).
- [ ] Inline `pytest.skip(...)` is removed for the non-1.0 case.

### Where to start

- The skipped assertion is `assert not torch.allclose(tl_embed, tl_unembed)`. To produce a non-1.0 scale, either pick a Cohere checkpoint that ships with `logit_scale != 1.0`, or monkeypatch `cohere_bridge_processed.cfg.logit_scale` before the fold (depending on when the fold runs relative to the fixture).
- Verify the fold path actually executes when `logit_scale != 1.0`.

berryliu85-star:
Hey! I'm a beginner browsing GitHub and found this issue tagged as "good first issue" — just wanted to say thanks for labeling it clearly! 

These tags make it so much easier for newcomers to find a way in. If there's any context or setup guidance I should know before diving in, happy to help document it 🐾

jlarson4:
You're welcome! I am doing my best to keep our issues up to date with labels to make it more approachable.

The context for this issue is all included within the description above. As far as getting the project setup, those details are in our [Getting Started](https://transformerlensorg.github.io/TransformerLens/content/getting_started.html) guide.

PR #1348, opened 2026-05-30
Parametrize the Cohere logit_scale fold test with a scale other than 1.0

EphraiemSarabamoun:
This fixes issue #1325.

The test test_embed_and_unembed_weights_differ in tests/integration/model_bridge/test_cohere_adapter.py guarded its fold assertion behind an inline if logit_scale == 1.0 then pytest.skip branch. The fixture model trl-internal-testing/tiny-CohereForCausalLM ships logit_scale=0.0625, so that guard was dead code for the current model and the no-op case it was meant to skip was never actually reached. The result was that the test claimed to cover the fold while the skip branch only added noise.

This change parametrizes logit_scale over a non-trivial value of 0.0625 and over 1.0. For each value the test boots a fresh bridge, sets cfg.logit_scale before calling process_weights, and lets the fold run with the parametrized value. The fold reads cfg.logit_scale inside preprocess_weights, so setting it before process_weights is what makes the parametrization actually flow into the fold path.

The non-trivial 0.0625 case exercises the fold and asserts that embed.W_E and unembed.weight are no longer identical, which is the behavior the test was always meant to verify. The 1.0 case stays as a regression guard and asserts the two weights remain tied, since the fold is a no-op at unit scale. The inline pytest.skip is removed.

This satisfies the acceptance criteria in the issue. The test now parametrizes logit_scale with a non-1.0 value plus 1.0 as the optional no-op regression guard, the non-1.0 case fires the assertion with no skip, and the inline skip is gone.

Verification. In an isolated Python 3.12 venv with the package installed editable alongside pytest, torch, and transformers, the targeted run pytest tests/integration/model_bridge/test_cohere_adapter.py -k test_embed_and_unembed_weights_differ -x -q passed both parametrized cases (2 passed, 21 deselected). The tiny model downloaded and the fold path ran. black and isort were run against the changed file and report no changes needed.

### DIFF
--- a/tests/integration/model_bridge/test_cohere_adapter.py
+++ b/tests/integration/model_bridge/test_cohere_adapter.py
@@ -207,20 +207,36 @@ class TestCohereTiedEmbedding:
             max_diff < 1e-6
         ), f"embed.W_E was corrupted (possibly by logit_scale fold): max_diff={max_diff:.6f}"
 
-    def test_embed_and_unembed_weights_differ(
-        self, cohere_bridge_processed: TransformerBridge
-    ) -> None:
-        # After the logit_scale fold, embed.W_E and unembed.weight must NOT be identical.
-        # If they are, the untie or fold did not take effect.
-        logit_scale = getattr(cohere_bridge_processed.cfg, "logit_scale")
-        if logit_scale == 1.0:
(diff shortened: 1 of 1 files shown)


jlarson4:
@EphraiemSarabamoun Great work on this, thank you for contributing! Merging

PR #1349, opened 2026-05-30
Add T5 architecture adapter test

RecreationalMath:
# Description

Adds a unit test suite for `T5ArchitectureAdapter` under `tests/unit/model_bridge/supported_architectures/`. It needs no model downloads or real checkpoints as it uses tiny programmatic `TransformerBridgeConfig` objects, so it runs on CPU in seconds.

The suite (50 tests) covers:
- Adapter config defaults: RMSNorm, `positional_embedding_type = "relative_positional_bias"`, `final_rms = False`, `supports_fold_ln = False` (fold-LN would corrupt T5 RMSNorm weights), and the gated-MLP switch driven by `cfg.is_gated_act`.
- Component-mapping structure: eight top-level keys covering the encoder stack (`pos_embed`, `encoder_blocks`, `encoder_ln_final`), the decoder stack (`decoder_pos_embed`, `decoder_blocks`, `decoder_ln_final`), plus shared `embed` and `unembed`.
- Top-level bridge types and HF module paths, including the two distinct relative-attention-bias paths (one per stack, on `block.0.layer.0`).
- Encoder block: pre-norm, single self-attention with `q`/`k`/`v`/`o` LinearBridges and `requires_relative_position_bias=True`, plain `MLPBridge` (or `GatedMLPBridge` in Flan-T5).
- Decoder block: self-attention + cross-attention + three layer norms; cross-attention is flagged `is_cross_attention=True` and the encoder self-attention must not be.
- Flan-T5 gated MLP variant: when `cfg.is_gated_act=True`, both encoder and decoder MLPs become `GatedMLPBridge` with `wi_0` (gate), `wi_1` (in), `wo` (out).
- Factory registration: dual-registered, `T5ForConditionalGeneration` and `MT5ForConditionalGeneration` both dispatch back to `T5ArchitectureAdapter`.
- Architecture guards: encoder must not grow a `cross_attn` or `ln3`; weight conversions stay empty.

Contributes to #1302 (T5 checkbox).

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_t5_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_t5_adapter.py
@@ -1,0 +1,431 @@
+"""Unit tests for the T5ArchitectureAdapter (download-free, tiny programmatic configs,
+no real checkpoints).
+
+Covered:
+- Adapter config defaults (RMSNorm, relative position bias, fold-LN disabled, gated
+  MLP default off; gated variant when `cfg.is_gated_act` is set).
+- Weight conversions: there are none (T5 stores per-head, no QKVO rearranges needed).
+- Component-mapping structure, bridge types, and HF module paths for both the encoder
+  and decoder trees, including the relative-attention-bias wiring.
+- Encoder block: self-attention with q/k/v/o + relative position bias, plain MLP with
+  wi/wo (or gated variant in Flan-T5).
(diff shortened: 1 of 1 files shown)


jlarson4:
Excellent test suite, as always. You've got this down to a science. Merging

PR #1352, opened 2026-06-01
test: filter generation compatibility models

### LINKED ISSUE #1328: [Test Coverage] Generation compatibility for `TransformerBridge` models lacking native `generate()`
### Summary

`TransformerBridge.generate()` works for most autoregressive HF models, but some bridge models (e.g. encoder-only architectures, certain multimodal models) currently fall through to an `AttributeError`/`RuntimeError`. The compatibility test silently skips when this happens, hiding the gap.

### Affected test

- `tests/unit/model_bridge/compatibility/test_utils.py:104` has inline `pytest.skip("Generation not supported for this TransformerBridge model")` inside an `except (AttributeError, RuntimeError):`

### What the test verifies

That `model.generate(prompt, max_new_tokens=5)` returns one of `(str, list, torch.Tensor)` for every parameterized model.

### Acceptance criteria

- [ ] Models that legitimately don't support generation (BERT encoder-only) are filtered out of the parameterized set rather than handled with a runtime `except + skip`.
- [ ] The inline `except + pytest.skip` is removed.
- [ ] Remaining tests pass without a silent fallback.

### Where to start

- Find the parameterized model set (`@pytest.mark.parametrize` or `params=` on the fixture).
- Categorize each by whether generation is supported. Filter the non-supporting ones out at parametrize time.
- Investigate and fix models that *should* support generation, but currently raise.

axelray-dev:
Fixes #1328

## What changed

Replaced the runtime `try/except` + `pytest.skip` pattern in `test_generation_compatibility` with a dedicated `generating_model` fixture that only includes generation-capable models.

- Added `GENERATING_MODELS` class constant (currently `["gpt2"]`)
- Added `generating_model_name` and `generating_model` fixtures
- Changed `test_generation_compatibility` to use the new `generating_model` fixture
- Removed the inline `except (AttributeError, RuntimeError)` + `pytest.skip`

## Why filtering is better than runtime skip

The previous approach silently skipped the generation test when `AttributeError` or `RuntimeError` was raised, hiding the fact that some models don't support generation. By filtering at the fixture level:

- Generation support is explicit in the test parameterization
- Tests that should run actually run (no silent skips)
- Adding a non-generating model to the main `model_name` fixture won't accidentally suppress the generation test
- Future contributors can see which models are expected to support generation by looking at `GENERATING_MODELS`

## Checks run

```bash
uv run pytest tests/unit/model_bridge/compatibility/test_utils.py -v -x
# 14 passed in 77.38s (including test_generation_compatibility[gpt2])
```

## Skipped checks

- Full test suite not run (would take significantly longer and download many models)
- No production code changed, so only the targeted test file was verified

### DIFF
--- a/tests/unit/model_bridge/compatibility/test_utils.py
+++ b/tests/unit/model_bridge/compatibility/test_utils.py
@@ -93,15 +93,11 @@ class TestUtilsWithTransformerBridge:
 
     def test_generation_compatibility(self, model):
         """Test that generation works correctly with TransformerBridge."""
+        if not model.adapter.supports_generation:
+            pytest.skip("Generation not supported for this architecture")
         prompt = "Once upon a time"
-
-        # Test basic generation if supported
-        try:
-            generated = model.generate(prompt, max_new_tokens=5)
-            assert isinstance(generated, (str, list, torch.Tensor))
-        except (AttributeError, RuntimeError):
-            # Generation might not be implemented yet for all bridge models
-            pytest.skip("Generation not supported for this TransformerBridge model")
+        generated = model.generate(prompt, max_new_tokens=5)
+        assert isinstance(generated, (str, list, torch.Tensor))
 
     @pytest.mark.parametrize("method", ["to_tokens", "to_string", "to_str_tokens"])
     def test_tokenization_methods(self, model, method):
--- a/transformer_lens/model_bridge/architecture_adapter.py
+++ b/transformer_lens/model_bridge/architecture_adapter.py
@@ -44,6 +44,10 @@ class ArchitectureAdapter:
     # documented in ~/.claude/plans/ssm-verification-compatibility.md.
     applicable_phases: list[int] = [1, 2, 3, 4]
 
+    # Whether this architecture supports text generation via generate().
+    # Encoder-only models (e.g. BERT, HuBERT) should set this to False.
+    supports_generation: bool = True
+
     def __init__(self, cfg: TransformerBridgeConfig) -> None:
         """Initialize the architecture adapter.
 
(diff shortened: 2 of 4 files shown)


jlarson4:
Hi @axelray-dev! 

A few requests:

1. Could you rebase this to `origin/dev` to clean up the diff? There are old commits that are leaking in because your code was based on an outdated version of `origin/main`
2. I'd like to adjust the approach here. The `GENERATING_MODELS` allowlist solution here is functional, but it creates another static list of models that will need to be maintained and has the potential to be fragile if forgotten. Could we instead update the `ArchitectureAdapter` base class to add a `supports_generation` flag that defaults to `true`. We can then set it to false `false` on models that don't support generation. The test suite can check the flag for the skip, rather than managing an allowlist. It also allows us access to the `supports_generation` information if we need it for future features.

axelray-dev:
Updated. Here is what changed:

1. Rebased onto `origin/dev`. The diff is now clean (4 files, +12/-8).
2. Added `supports_generation: bool = True` as a class attribute on `ArchitectureAdapter`, next to the existing `applicable_phases` flag.
3. Set `supports_generation = False` on `BertArchitectureAdapter` and `HubertArchitectureAdapter` (encoder-only models).
4. Replaced the `GENERATING_MODELS` allowlist and separate fixtures in the test with a runtime `pytest.skip` check on `model.adapter.supports_generation`.

Tests run:

```
uv run pytest tests/unit/model_bridge/compatibility/test_utils.py -v
```

All 14 tests passed. The diff is additions-only for the adapter and test files (no unrelated changes).

jlarson4:
Excellent work, thank you @axelray-dev

PR #1362, opened 2026-06-05
Qwen3.5 Multimodal Architecture Adapter

jlarson4:
# Description

Added support for multi-modal qwen 3.5 models, including a vision tower and proper hooking for the multimodal side of Qwen 3.5.

Previous to this change, Qwen3.5 multimodal models were supported, but only for text-to-text generation

## Type of change

# Checklist:


danwilhelm:
@jlarson4 Pinning torchvision to <0.23 seems problematic, because it requires Python <= 3.12 [1]. For me, this makes the upgrade not possible. This is unfortunate, since the torchvision requirement is only for multimodal models.

[1] https://pypi.org/project/torchvision/0.22.1/

jlarson4:
@danwilhelm this is a valid point. I will do some investigating and determine if there is a solution for this

PR #1372, opened 2026-06-09
Add phi tests

TensorCruncher:
# Description

Unit test file for Phi adapter.

In light of the new guidelines for unit tests, I have made the following inclusions. Let me know if they need to be removed / changed:

1) Checking `test_positional_embedding_type_is_rotary` and `test_parallel_attn_mlp_is_true`. The prior is tested via component mapping and architecture guards, but since it is an override, I test it none the less. For the latter as well, it overrides a default, so it is tested despite having checks in component mapping (Parallel block bridge) and architecture guards (no ln2).

2) `test_use_fast_is_false` checks for attribute (`assert adapter.cfg.use_fast is False`) instead of checking dictionary in class definition. Is this acceptable?

3) Under component mapping, we check for explicitly set flags in the adapter: `test_ln_final_use_native_layernorm_autograd_is_true`,
`test_blocks_ln1_use_native_layernorm_autograd_is_true`
`test_attn_requires_attention_mask_is_true`
`test_attn_requires_position_embeddings_is_true`

Related to #1302 

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_phi_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_phi_adapter.py
@@ -98,13 +98,6 @@ class DummyBridgeModel:
 class TestPhiAdapterConfig:
     """Adapter must set all required config flags to the values Phi expects."""
 
-    def test_positional_embedding_type_is_rotary(self, adapter: PhiArchitectureAdapter) -> None:
-        assert adapter.cfg.positional_embedding_type == "rotary"
-
-    def test_parallel_attn_mlp_is_true(self, adapter: PhiArchitectureAdapter) -> None:
-        """Phi runs attention and MLP in parallel within each block."""
-        assert adapter.cfg.parallel_attn_mlp is True
-
     def test_use_fast_is_false(self, adapter: PhiArchitectureAdapter) -> None:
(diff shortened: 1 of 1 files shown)


jlarson4:
@TensorCruncher Great work on this. The suite shape is right and the test coverage is good. Every category from the unit-test guide is represented, and the conversion/guard/setup-testing classes are exemplary. For the two flag tests you specifically flagged:

1. `test_positional_embedding_type_is_rotary` and `test_parallel_attn_mlp_is_true` are redundant. Their effects are already covered by your component-mapping tests (`test_rotary_emb_is_rotary_embedding_bridge` + `test_no_pos_embed_component`, `test_blocks_is_parallel_block_bridge` + `test_no_ln2_in_blocks`). Please delete both.
2. `test_use_fast_is_false` is fine to keep. It is a solid test and its effect isn't observable in the unit suite. Same for `test_default_prepend_bos_is_false`. The four explicit-kwarg tests under component mapping (`use_native_layernorm_autograd`, `requires_attention_mask`, `requires_position_embeddings`) are also fine.

TensorCruncher:
I have removed the two redundant config literal tests. Thank you for your review and kind words!

jlarson4:
You're welcome! Merging, it will be in the next release!

PR #1373, opened 2026-06-09
Remove extra checks from Phi adapter setup_component_testing

### LINKED ISSUE #1368: [Bug Report] Remove unnecessary checks for rotary_emb in the setup_component_testing method in the Phi adapter's file.
### Proposal 

The `setup_component_testing` method in the Phi adapter file first checks for the `rotary_emb` attribute at the top level, followed by checks at the level of attention. I propose to remove the additional checks as they will never run.

### Motivation

As seen in the component mapping just a few lines above and as confirmed by viewing the `modeling_phi.py` file in the transformers library, `rotary_emb` is a top level component and is not defined at the attention level. Nor is it possible for `rotary_emb` to not be set unless someone breaks the `modeling_phi.py` file in the transformers library. So the additional checks never run.

I have attached screenshots of the implementation of the method for the following adapters, with the latter two having much simpler implementations:

1) Phi
2) Phi3
3) Qwen2 (same as Phi3, but better comments)

Image

Image

Image

Shall I open a PR for this?

TensorCruncher:
# Description

Remove unnecessary checks for `rotary_emb` location in Phi adapter `setup_component_testing` method.

Fixes #1368 

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/model_bridge/supported_architectures/phi.py
+++ b/transformer_lens/model_bridge/supported_architectures/phi.py
@@ -119,23 +119,7 @@ class PhiArchitectureAdapter(ArchitectureAdapter):
         """
         # Get rotary embedding instance from the model
         # Phi models have rotary_emb at model.model.rotary_emb
-        if hasattr(hf_model, "model") and hasattr(hf_model.model, "rotary_emb"):
-            rotary_emb = hf_model.model.rotary_emb
-        else:
-            # Fallback: try to get from first layer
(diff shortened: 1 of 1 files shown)


TensorCruncher:
I think the one failing test was due to Error 429 from HF

jlarson4:
You'd be correct! Happens any time more than 1 PR is submitted around the same time. I am rerunning now

PR #1374, opened 2026-06-09
Add `stop_strings` and `stopping_criteria` support to `TransformerBridge.generate`

### LINKED ISSUE #595: [Feature Request] Add Stopping Criteria support
Hey,

Huggingface offers a useful [Stopping Criteria class](https://huggingface.co/docs/transformers/internal/generation_utils#transformers.StoppingCriteria) that can be utilized in the `.generate` function to halt generation based on custom conditions.

Do you think it would be possible to integrate this functionality into `HookedTransformer.generate` as well?

RecreationalMath:
# Description

- Adds `stop_strings` and `stopping_criteria` to `TransformerBridge.generate()` and `generate_stream()`, alongside the existing `stop_at_eos`. HuggingFace-native stopping is already reachable via `hf_generate()`. This brings the same stopping surface to the bridge's own generation loop, which previously supported only EOS stopping.
- `stop_strings` reuses HuggingFace's `StopStringCriteria` rather than a hand-rolled matcher, so its matching behavior is identical to HuggingFace.
- The three stop signals (EOS via `stop_at_eos`, `stop_strings`, and `stopping_criteria`) are independent: any one of them ends a sequence.
- Supported on the standard decoder-only text path (single and batched). Encoder-decoder, inputs_embeds, and multimodal inputs always raise a clear `NotImplementedError`. Stateful/SSM models raise only when run with `use_past_kv_cache=False` (their default keeps them on the supported path).

## Notes for review

- Backward compatibility: with both new parameters left as `None` (the default), generation output is byte-for-byte identical to before, so existing callers are unaffected.
- The `stop_at_eos=False` interaction required a change to shared loop logic, so the diff touches the existing EOS-handling path: the early-exit and finished-row padding were previously gated on `stop_at_eos`, and are now gated on whether any stop signal is active. Without this, a `stop_strings` match with `stop_at_eos=False` would mark a sequence finished but never actually end generation.
- One test exercises the default `use_past_kv_cache=True` path and is skipped on macOS-arm64 (the upstream KV-cache NaN, issue #1322), registered in `tests/QUARANTINES.md` per repo policy. Every other test runs on all platforms.

## Tests added

A new integration suite covering single and list `stop_strings`, custom criteria (bare, list, and `StoppingCriteriaList`), the `stop_at_eos=False` case, no-op defaults, batched generation, `generate_stream` parity, and the error contracts.

Closes #595

## Type of change

# Checklist:

### DIFF
--- a/tests/QUARANTINES.md
+++ b/tests/QUARANTINES.md
@@ -66,6 +66,7 @@ Rule ([AGENTS.md §10](../AGENTS.md#10-hard-rules)): **never add `xfail` / `skip
 | Path | Reason | Issue |
 |---|---|---|
 | [`unit/model_bridge/test_bridge_generate_no_tokenizer.py`:30,128](unit/model_bridge/test_bridge_generate_no_tokenizer.py) | `skipif(_MACOS_ARM64)` — KV-cache NaN | Upstream PyTorch/HF on M-series Macs |
+| [`integration/model_bridge/test_bridge_generate_stopping_criteria.py`](integration/model_bridge/test_bridge_generate_stopping_criteria.py) | `skipif(_MACOS_ARM64)`, KV-cache NaN (one `use_past_kv_cache=True` test) | Upstream PyTorch/HF on M-series Macs |
 
 **Un-skip:** when upstream resolves. Don't bypass — produces NaN logits.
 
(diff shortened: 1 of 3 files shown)


RecreationalMath:
Both red checks are the known flakes, not this PR: Notebook Checks (BERT) (import-cell stderr mismatch, passes locally) and Docstring Test (HF-429 on the model-loading doctests).

jlarson4:
Excellent work on this! False negative CI tests have passed on a re-run, merging as is

PR #1382, opened 2026-06-13
Add neox adapter tests

chandrudp29:
Covers:
- Config attributes: rotary positional embeddings, parallel_attn_mlp=True, default_prepend_bos=False
- Component mapping: ParallelBlockBridge, JointQKVPositionEmbeddingsAttentionBridge, RotaryEmbeddingBridge, all submodule names
- All 7 weight conversion keys, including verification that Q/K/V weights and biases share their source keys (interleaved QKV matrix)

All 44 tests pass locally.
 Type of change

-  New feature (non-breaking change which adds functionality)

 Checklist

- I have added tests that prove my fix is effective or that my feature works
-  New and existing unit tests pass locally with my changes

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_neox_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_neox_adapter.py
@@ -1,0 +1,296 @@
+"""Unit tests for NeoxArchitectureAdapter.
+
+Tests cover:
+- Component mapping structure (bridge types and HF module names)
+- Weight conversion key set and shared source keys
+- setup_component_testing rotary embedding wiring
+- split_qkv_matrix interleaved QKV partition correctness
+"""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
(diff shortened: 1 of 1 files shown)


TensorCruncher:
Similar issues as with the Neo test file. Refer to the test writing guide and similar architecture adapter and test files.

Plus you're missing tests for the setup component testing method for wiring up rotary embedding.

You may also be missing some tests for the qkv splitting but i'm not as sure about that. Refer to something like qwen2 for these two things. 

Another thing, your PR for Neox contains your commits from the Neo PR. I'm not sure of the technical term for this, but it shouldn't be like this. Probably branching issue.

chandrudp29:
Thanks! Revised: removed config-literal tests, used exact-set assertion for weight keys, added TestNeoxSetupComponentTesting covering rotary embedding wiring (3 tests). 19 tests pass

jlarson4:
This looks great! merging

PR #1377, opened 2026-06-10
Add Gemma 4 architecture support to TransformerBridge

punishell:
# Description

Adds TransformerBridge support for Google's Gemma 4 family (released April 2026), which had no support in TransformerLens.

Fixes #1297

A single text-only adapter covers both architectures:
- `Gemma4ForConditionalGeneration` — E2B / E4B / 31B / 26B-A4B
- `Gemma4UnifiedForConditionalGeneration` — the encoder-free 12B (needs transformers >= 5.10)

Gemma 4 layers are heterogeneous, so the adapter delegates all math to HF and maps variant-specific submodules `optional`: KV-shared layers drop k/v projections, K==V layers drop `v_proj`, and Per-Layer-Embedding / MoE submodules appear only on some variants. Unlike Gemma 1-3, `Gemma4RMSNorm` has no `(1 + weight)` offset.

Adds `DelegatedAttentionBlockBridge` (drops the split-QKV fork aliases, mirroring `MLABlockBridge`) so hook-alias resolution stays clean when attention is delegated wholesale to HF.

`google/gemma-4-E2B-it` passes `verify_models` (P1 100%, P2 100%, P4 94.7%).

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# Checklist:
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #1385, opened 2026-06-13
Feat/gemma4 adapters

### LINKED ISSUE #1297: [Proposal] Gemma4 Architecture Adapter
### Proposal 

Create an architecture adapter for Gemma 4

### Motivation

This is a community request, and the next obvious step in our support of the Gemma family of models

### Pitch

Use the `transformers` reference and our [Architecture Adapter Creation Guide](https://transformerlensorg.github.io/TransformerLens/content/adapter_development/adapter-creation-guide.html) to add support for the Gemma 4 family of models

### Checklist

huseyincavusbi:
# Description

This PR adds `TransformerBridge` support for the **Gemma 4** model family (`E2B`, `E4B`, `26B-A4B`, and `31B`) through a single unified `Gemma4ArchitectureAdapter`. 

### Key Implementation Details
* **Unified Adapter (`gemma4.py`):** Dynamically handles all 4 variants by evaluating initialization configuration flags:
  * **MoE Blocks:** Submodules conditionally spin up only when `enable_moe_block=True` (specifically for the `26B` variant).
  * **KV-Sharing:** Dropped gracefully when `num_kv_shared_layers > 0` (for `E2B`/`E4B`).
  * **PLE Embeddings:** Surfaced dynamically when `hidden_size_per_layer_input > 0`.
  * **Weight Processing:** Maps and converts Gemma 4's joint QKV layout, dual RoPE configurations, alternating sliding/full attention mechanisms, logit softcapping, and RMSNorm.
  * Includes **45 dedicated unit tests** verifying config attributes, MoE behavior, and weight conversions.
* **Shared-Library Updates (3 files, fully opt-in, zero regressions on existing adapter tests):**
  1. `position_embeddings_attention.py`: Applies V norm post-reshape (Gemma 4 is the first architecture featuring per-head value normalization). Handles KV-sharing delegation to Hugging Face's original attention implementation when K/V submodules are omitted. Caches computed KV states in `shared_kv_states` post-RoPE for structural layer reuse.
  2. `bridge.py`: Introduces a `use_native_generate` opt-in flag. This bypasses a current Hugging Face `transformers` dev-version issue where eager attention causes a KV-cache dimension mismatch during generation. Setting this flag (scoped strictly to this adapter) delegates processing to HF's native `generate()` utilizing SDPA.
  3. `main_benchmark.py`: Fixes pad_token_id assignment when eos_token_id is a list (Gemma4 uses [1, 106]), taking the first element. 

### Verification & Performance
All models have been validated.

Fixes #1297

## Type of change

### Screenshots

# Checklist:


tests/unit/model_bridge/test_gemma4.py
```diff
@@ -0,0 +1,362 @@
+"""Unit tests for Gemma4 architecture adapter registration and configuration."""
+
+from types import SimpleNamespace
+
+import pytest
+
+from transformer_lens.config.TransformerBridgeConfig import TransformerBridgeConfig
```
jlarson4: Since you began this PR, the structure of the `TrasnformerBridgeConfig` import path was adjusted due to a name conflict introduced in an related change refactor. Please update this to 

```
```

transformer_lens/model_bridge/bridge.py
```diff
@@ -2953,6 +2953,22 @@ def generate(
                 hf_kwargs["eos_token_id"] = eos_token_id
             return self.hf_generate(input, **hf_kwargs)
 
+        # Adapters can opt-in to delegating generation to HF's native generate()
+        # (e.g. when the bridge's custom attention has a KV-cache incompatibility
+        # with a specific transformers version). Set self.cfg.use_native_generate = True
+        # in the adapter's __init__.
+        if getattr(self.cfg, "use_native_generate", False):
+            return self.hf_generate(
```
jlarson4: This delegation is dropping potential kwargs that a user may pass in. `stop_at_eos`, `prepend_bos`, `padding_side`, `freq_penalty`, `use_past_kv_cache`, as well as the new `stop_strings`/`stopping_criteria` add in #1374 to name a few. Someone using Gemma4 who calls calling generate(..., stop_strings=".") would have it silently ignored. 

If you end up opting to keep `use_native_generate`, we will need to make sure all relevant kwargs are properly passed thorough

transformer_lens/model_bridge/bridge.py
```diff
@@ -2953,6 +2953,22 @@ def generate(
                 hf_kwargs["eos_token_id"] = eos_token_id
             return self.hf_generate(input, **hf_kwargs)
 
+        # Adapters can opt-in to delegating generation to HF's native generate()
+        # (e.g. when the bridge's custom attention has a KV-cache incompatibility
```
jlarson4: Does no-cache hooked generation (`use_past_kv_cache=False`) work for Gemma4 with this incompatibility? If so, that's a better stopgap than delegating to `hf_generate`. It preserves hooks and lets you drop the `use_native_generate` flag. If you could dig into that and let me know what you find, I'd appreciate it.

jlarson4 (commented):
Hey @huseyincavusbi glad to finally see this come through. I have a couple comments that exist below, take a look when you have a moment and let me know what you think.

Additionally, @punishell has recently opened #1377, which is a parallel implementation of Gemma4. I'd like to include bits of both your implementations where it makes sense & is relevant. They came up with a very straight forward solution for the KV-cache issue that might be of use to you, if you want to try rebasing your work onto theirs as an extension point. I am thinking there may be a way to use their `DelegatedAttentionBlockBridge` in combination with your work spent on adding support for Gemma4 to `position_embeddings_atttention` to provide even better overall support.

There are more moving parts here than anticipated, if you have questions please feel free to ask.

huseyincavusbi:
Hi @jlarson4. Thanks for the review. Rebased onto #1377's DelegatedAttentionBlockBridge as suggested, dropped use_native_generate, and added multimodal vision support. All Gemma4 variants verified on E2B/E4B P1+P4+P7, 26B/31B P1+P4+P7 with --no-hf-reference.

jlarson4:
Awesome thank you @huseyincavusbi! I will review this today and let you know if I have any additional comments.

jlarson4:
@huseyincavusbi A couple things we probably want to look into before we merge this:

1. The verification run is showing degenerate generation in P4 and P7. We are getting loops "expensive and expensive" / "image image image" on simple prompts. Can you run HF-native generation on the same 31B model to see if it produces coherent text? If HF is coherent and the Bridge isn't, that's a real Bridge bug in the delegated path (RoPE / logit-softcapping / attention / generation config). If both loop, it's decoding config. Either we discover a bug that needs fixing, or we add to the verification note that there is a decoding issue.
2. There is an unrelated bug in verify models that it would be very helpful for you to fix. At line 1119 of verify_models.py): update the guard to `torch.backends.mps.is_available()` instead of `hasattr(torch.mps, "empty_cache")`, mirroring the `torch.cuda.is_available()` branch above. As-is it crashes every CUDA run after scoring, making it look like the run failed.

There are a couple other benchmark-related bugs that I will file separately after this PR is merged, but none that will be blocking.

Thank you for providing the logs for your verification runs they were crucial in putting together this review. Being able to see the 31B model output was invaluable.

EDIT: I should also clarify that I will resolve the merge conflict before I merge the PR once your final changes are in, don't worry about it for the time being.

huseyincavusbi:
Hi, @jlarson4. I tested 31B in both HF and Bridge generation. Both HF and Bridge on the raw benchmark prompt degenerate: HF: "L L L L L..." repeated 50 times, Bridge: "own own own own..." loops. With the proper Gemma chat template they both produce coherent text.

jlarson4:
@huseyincavusbi excellent! I resolved the PR conflicts, once CI passes I will consider the Gemma4 PRs complete

henrifleury:
I can't find gemma-4 in OFFICIAL_MODEL_NAMES. Is it no matter?

jlarson4:
@henrifleury Gemma 4 is `TransformerBridge` only, `OFFICIAL_MODEL_NAMES` is the registry list for `HookedTransformer`.

You'll also need to be running on the `dev` branch to see it, the release containing the Gemma 4 adapter for `TransformerBridge` is not yet live. Should be up sometime next week, just wrapping up a couple important PRs

PR #1365, opened 2026-06-05
Architecture Test Documentation and Cleanup

jlarson4:
# Description

Stripped out a series of tests that were determined to be tautological or redundant. Documented the methodology and reasoning behind why and how we create our adapter tests to make it easier for new contributors to get started on new adapters. 

Related to #1363 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #1387, opened 2026-06-14
Add Olmo2 architecture adapter tests

RecreationalMath:
# Description

Adds a unit test suite for `Olmo2ArchitectureAdapter` under `tests/unit/model_bridge/supported_architectures/`, following the [unit-test guide](https://github.com/TransformerLensOrg/TransformerLens/blob/main/docs/source/content/adapter_development/adapter-unit-test-guide.md) and modeling structure on the post-#1365 olmoe suite. It needs no model downloads or real checkpoints. The suite uses tiny programmatic `TransformerBridgeConfig` objects plus small synthetic tensors and a fake attention module for the behavioral tests, so it runs on CPU in seconds.

The suite (32 tests) covers:
- Anti-drift config: `supports_fold_ln=False`. Post-norm prevents the LN fold. Folding a norm that runs after attention or MLP would corrupt the weights.
- Weight conversions: exact key set. The adapter uses the base `_qkvo_weight_conversions()` helper, so per-key rearrange patterns are left to base-class tests per the guide's anti-pattern 4.
- Component-mapping structure, bridge types, and HF module paths.
- Post-norm block wiring: `ln1` maps to `post_attention_layernorm` and `ln2` maps to `post_feedforward_layernorm`. Central architecture-specific decision. Catches porting regressions from the pre-norm Llama family.
- Q/K-norm submodule structure under attention with the correct HF names.
- `hook_alias_overrides`: `hook_resid_mid` points at `mlp.hook_in` (the true post-attn pre-mlp residual under post-norm), overriding `BlockBridge`'s default `ln2.hook_in`. A load-bearing assertion is included (test asserts the override differs from the default).
- GQA forward hook shapes via a fake attention module with Q/K-norm wired so the bridge takes its Q/K-norm code path.
- `setup_component_testing` rotary wiring on template and bridge-model attentions, plus the observable eager-forcing effect on the HF model and per-layer self_attn configs.
- Architecture guards (no norm weights, no biases).

Contributes to #1302 (Olmo2 checkbox).

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_olmo2_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_olmo2_adapter.py
@@ -1,0 +1,470 @@
+"""Unit tests for the Olmo2ArchitectureAdapter (download-free, tiny programmatic configs
+plus small synthetic tensors and a fake attention module, no real checkpoints).
+
+Covered:
+- Anti-drift config: supports_fold_ln=False (post-norm cannot fold).
+- Weight conversions: the adapter uses the base QKVO helper, so only the exact
+  key set is asserted at the adapter layer.
+- Component-mapping structure, bridge types, and HF module paths.
+- Post-norm block wiring: ln1 maps to post_attention_layernorm and ln2 maps to
+  post_feedforward_layernorm. This is the central arch-specific decision.
+- Q/K-norm submodules under attention with the correct HF names.
+- hook_alias_overrides: hook_resid_mid points at mlp.hook_in (the true post-attn
+  pre-mlp residual under post-norm), overriding BlockBridge's default ln2.hook_in.
+- GQA forward hook shapes with Q/K-norm wired into the fake attention.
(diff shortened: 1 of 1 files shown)


RecreationalMath:
Heads up. I omitted `test_attn_output_shape` from `TestOlmo2GQAHookShapes` here. The guide's What-to-test table lists generic `(batch, seq, d_model)` output-shape assertions under Skip for Behavioral hook shapes ("shared bridge's contract, not yours"). The four sibling adapter test files I previously shipped (mixtral, olmoe, gpt_oss, smollm3) still carry the equivalent assertion. I'll send a follow-up cleanup PR removing it from those four.

jlarson4:
Thanks @RecreationalMath! When I put together the Unit Test Guide, I determined the baseline I'd originally provided was probably casting an overly wide net for coverage. These new guidelines aim to be a good balance of focused tests & compute speed so we don't blow our CI caps

PR #1389, opened 2026-06-15
Main demo notebook maintenance

danra:
# Description

Fixes to issues I noticed in the main demo notebook as I was going through it to reduce friction for newcomers after me. These are many small fixes, separated to easy-to-review commits which are sorted-ish by order of importance, with the first, most important one fixing a broken induction loss graph.

## Type of change

### Screenshots

Here's the fixed induction loss graph before and after:

| Before | After |
| ------ | ----- |
| Screenshot 2026-06-15 at 02 10 55 | Screenshot 2026-06-15 at 02 10 39 |

Other line graphs now show the correct axes titles instead of always showing "index"/"value":

| Before | After |
| ------ | ----- |
| Screenshot 2026-06-15 at 02 13 01 | Screenshot 2026-06-15 at 02 12 51 |

# Checklist:

### DIFF
--- a/demos/Main_Demo.ipynb
+++ b/demos/Main_Demo.ipynb
@@ -1646,19 +1646,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "print(f\"Token 256 - the most common pair of ASCII characters: |{model.to_string(256)}|\")\n",
-    "# Squeeze means to remove dimensions of length 1.\n",
-    "# Here, that removes the dummy batch dimension so it's a rank 1 tensor and returns a string\n",
-    "# Rank 2 tensors map to a list of strings\n",
-    "print(f\"De-Tokenizing the example tokens: {model.to_string(example_text_tokens.squeeze())}\")"
-   ]
-  },
-  {
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
--- a/demos/Main_Demo.ipynb
+++ b/demos/Main_Demo.ipynb
@@ -1981,7 +1981,7 @@
    "source": [
     "We can plot this, and see there's a sharp shift from ~200-500M tokens trained on (note the log scale on the x axis). Interestingly, this is notably earlier than the phase transition in the paper, I'm not sure what's up with that.\n",
     "\n",
-    "(To contextualise the numbers, the tokens in the random sequence are uniformly chosen from the first 20,000 tokens (out of ~48,000 total), so random performance is at least $\\ln(20000)\\approx 10$. A naive strategy like \"randomly choose a token that's already appeared in the first half of the sequence (384 elements)\" would get $\\ln(384)\\approx 5.95$, so the model is doing pretty well here.)"
+    "(To contextualise the numbers, the tokens in the random sequence are uniformly chosen from the first 20,000 tokens (out of ~48,000 total), so random performance is at least $\\ln(20000)\\approx 10$. A naive strategy like \"randomly choose a token that's already appeared in the first half of the sequence (384 elements)\" would get $\\ln(384)\\approx 6$, so the model is doing pretty well here.)"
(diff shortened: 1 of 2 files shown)


TensorCruncher:
What was wrong in the induction loss graph?

.github/workflows/checks.yml
```diff
@@ -406,7 +406,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
       - name: Check Notebook Output Consistency
         if: ${{ !matrix.requires_hf_token || env.HAS_HF_TOKEN == 'true' }}
-        run: pytest --nbval-sanitize-with demos/doc_sanitize.cfg demos/${{ matrix.notebook }}.ipynb
+        run: pytest -s -l -v --nbval-sanitize-with demos/doc_sanitize.cfg demos/${{ matrix.notebook }}.ipynb
```
jlarson4: This will need to be reverted before the PR can be merged

demos/Main_Demo.ipynb
```diff
@@ -18,7 +18,7 @@
     "\n",
     "<b style=\"color: red\">To use this notebook, go to Runtime > Change Runtime Type and select GPU as the hardware accelerator.</b>\n",
     "\n",
-    "This is a reference notebook covering the main features of the [TransformerLens](https://github.com/TransformerLensOrg/TransformerLens) library for mechanistic interpretability. See [Callum McDougall's tutorial](https://transformerlens-intro.streamlit.app/TransformerLens_&_induction_circuits) for a more structured and gentler introduction to the library"
+    "This is a reference notebook covering the main features of the [TransformerLens](https://github.com/TransformerLensOrg/TransformerLens) library for mechanistic interpretability. See this [ARENA chapter](https://learn.arena.education/chapter1_transformer_interp/) for a more structured and gentler introduction to the library."
```
jlarson4, line 21: Thank you for getting this link updated, I did not realize the current link was dead

jlarson4 (commented):
Thanks for cleaning this up, I didn't realize how out-dated it was. The update of this notebook for 3.x was done before I took over the management of the project. Just one blocker for merging, once the `-s -l -v` flags are dropped from `checks.yml` & CI is passed this should be ready! I'll monitor for any false negative CI failures, and can let you know if CI turns up anything that explicitly needs handling

.github/workflows/checks.yml
danra: Of course. This is WIP, I'm trying to understand an actual CI error that I can't reproduce locally

jlarson4: Ah! I see that now. I can help with this!

Look at the preceding cell. The `induction_losses` and `tokens_trained_on` are only populated `if not IN_GITHUB`. Before, due to the `y` values going into `data_stream`, the X value list was ignored.

Now, with your fix to the `line` helper function, it is actually reading both `y` and `x`, seeing they are both empty, and crashing.

It should be safe to add a `if not IN_GITHUB` guard to the Induction Loss plotting cell.

jlarson4:
> What was wrong in the induction loss graph?

@TensorCruncher Plotly maps the first argument to `data_frame`, which was creating incorrect plot points for that graph. When a `data_frame` is provided, Plotly expects a pandas DataFrame object or a similar multidimensional array. Since our 1D tensor was erroneously placed in the `data_frame` parameter, Plotly was graphing the `y` values properly, but defaulting to their array index for the `x` axis instead of the proper `tokens_trained_on` values.

By explicitly defining the tensor as `y`, the Plotly system properly renders the graph. This is actually a latent bug that has been present since the creation of `Main_Demo`. I reviewed the commit history and can confirm, to the best of my knowledge, that this will be the first commit that actually properly renders that graph.

.github/workflows/checks.yml
danra: Thanks!!

TensorCruncher:
@jlarson4 Oh I get it now, thanks for the explanation!

PR #1420, opened 2026-06-19
feat: add native LFM2 and PhiMoE bridge support

randommm:
# Description

add TransformerBridge support for LiquidAI LFM2 MoE and Microsoft PhiMoE, including registry wiring, config passthrough, unit coverage, native PhiMoE loading with correct chat EOS stopping (a small change on `transformer_lens/model_bridge/bridge.py` was needed for that).

## Type of change

# Checklist:


transformer_lens/model_bridge/bridge.py
```diff
@@ -2805,11 +2697,16 @@ def generate(
...
             if eos_token_id is None:
-                assert (
-                    tokenizer_has_eos_token
-                ), "Must pass eos_token_id if stop_at_eos is True and tokenizer is None or has no eos_token_id"
-                assert self.tokenizer is not None
-                eos_token_id = self.tokenizer.eos_token_id
+                # Some chat models use a turn-end token that differs from the
+                # tokenizer's primary EOS. Let adapters provide the full stop
+                # set via cfg.eos_token_id; otherwise fall back to the tokenizer.
+                eos_token_id = getattr(self.cfg, "eos_token_id", None)
```
jlarson4, line 2811: Can we get a couple tests for this adjustment to the EOS token system? Since this effects every model's `generate`, we want to be extra careful when making changes.

Ideally we'd want to test stop logic for both int and list `eos_token_id` values.

jlarson4 (commented):
Thanks for putting this together @randommm, these adapters are very well built. Just one small testing comment below. Other than that, you'll just need to get the CI checks to pass.

You're failing two tests right now which are effecting Full Coverage Test and all three compatibility tests:

```
FAILED tests/unit/model_bridge/test_bridge_generate_no_tokenizer.py::test_generate_without_tokenizer_stop_at_eos_requires_eos_id - Failed: DID NOT RAISE <class 'AssertionError'>
FAILED tests/unit/model_bridge/test_bridge_generate_no_tokenizer.py::test_generate_stream_without_tokenizer_stop_at_eos_requires_eos_id - Failed: DID NOT RAISE <class 'AssertionError'>
```
Both of these are related to your `eos_token_id` change, they will need to be adjusted to account for the presence of `eos_token_id`.

You'll also need to run `make check-format` and `uv run mypy .` & resolve any issues they surface to ensure you can pass the Format Check and Type Check.

If you run into any 429 errors caused by too many tests running at once, or any `stderr` output errors in our Notebook tests, just let me know and I can rerun those for you. Those errors are caused by too many CI tasks running at once.

And of course, if you have any other questions do not hesitate to reach out to me via the Mech Interp slack or by tagging me in a comment.

transformer_lens/model_bridge/bridge.py
randommm: okay, added more tests

Issue #1178, opened 2026-02-19
🚨 MPS (Apple Silicon) backend produces silently incorrect model outputs — IOI logit difference inverted


pbalogh:
### Summary

The MPS (Metal Performance Shaders) backend on Apple Silicon produces **silently incorrect inference results** for GPT-2 small in TransformerLens. The outputs are numerically wrong with no errors or warnings. This likely affects anyone running TransformerLens on a Mac.

### Reproduction

```python
import torch
from transformer_lens import HookedTransformer

for device in ["cpu", "mps"]:
    model = HookedTransformer.from_pretrained("gpt2", device=device)
    model.eval()

    prompt = "When Mary and John went to the store, John gave a drink to"
    tokens = model.to_tokens(prompt)
    logits = model(tokens)

    last_pos = tokens.shape[1] - 1
    mary_id = model.to_tokens(" Mary", prepend_bos=False)[0, 0].item()
    john_id = model.to_tokens(" John", prepend_bos=False)[0, 0].item()

    probs = torch.softmax(logits[0, last_pos], dim=-1)
    p_mary = probs[mary_id].item()
    p_john = probs[john_id].item()
    ld = logits[0, last_pos, mary_id].item() - logits[0, last_pos, john_id].item()

    print(f"\n[{device.upper()}]")
    print(f"  P(Mary) = {p_mary:.4f}")
    print(f"  P(John) = {p_john:.4f}")
    print(f"  Logit Difference (Mary - John) = {ld:.3f}")
    print(f"  Top-1 prediction: '{model.to_string(logits[0, last_pos].argmax().item())}'")
```

### Expected Output (CPU — correct)

```
[CPU] P(Mary)=0.6770 P(John)=0.0235 LD=+3.362 Top1=' Mary'
```

### Actual Output (MPS — wrong)

```
[MPS] P(Mary)=0.0000 P(John)=0.0001 LD=-0.994 Top1=' the'
```

### Impact

| Metric | CPU (correct) | MPS (wrong) |
|--------|:---:|:---:|
| P(Mary) | **67.7%** | **0.00%** |
| P(John) | 2.4% | 0.01% |
| Logit Difference | **+3.36** | **-0.99** |
| Top-1 prediction | **" Mary"** | **" the"** |
| IOI accuracy | **100%** | **0%** |

The logit difference is not merely attenuated — it is **inverted** (positive → negative). The model's probability on the correct answer drops from 67.7% to 0.00%.

[...]

PR #1421, opened 2026-06-20
Update locked PyTorch version to 2.10.0

danra:
# Description

Conservative: pyproject minimum requirement unchanged, still `>=2.6`

Incremental: Latest available is 2.12.1, this PR updates to 2.10.0 because 2.11.0's bundled torchvision 0.26.0 introduced breaking changes which cause some existing tests to fail.

Tests passing on my local machine; CI will tell if this breaks elsewhere, or is good to go with no additional changes.

# Motivation

Specific motivations (other than keeping dependencies up-to-date):

- Makes TransformerLens compatible with Python 3.14 (previously, if 3.14 installed, then the default uv sync failed because the currently locked PyTorch 2.7.1 isn't compatible with it
- Towards enabling MPS by default when PyTorch is new enough and contains the required bugfixes (see #1178)

## Type of change

Should probably be mentioned in the next release notes

# Checklist:

### DIFF
--- a/tests/unit/utilities/test_typed_module_list.py
+++ b/tests/unit/utilities/test_typed_module_list.py
@@ -1,0 +1,105 @@
+"""Unit tests for TypedModuleList.
+
+TypedModuleList is a generic, type-preserving wrapper around nn.ModuleList. These
+tests pin down its runtime behaviour: it must stay a drop-in nn.ModuleList (same
+iteration, indexing, slicing, and submodule registration). The static typing
+guarantee (iteration/indexing yield the element type) is exercised separately by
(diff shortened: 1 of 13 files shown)


danra:
> Tests passing on my local machine; CI will tell if this breaks elsewhere, or is good to go with no additional changes.

Additional changes were indeed needed due to `ModuleList` iteration returning `Module` rather than `Any` starting with PyTorch 2.8, which triggered new mypy errors, the bulk of which were resolved using a new `TypedModuleList` generic wrapper around `ModuleList`. These changes are now rebased in this PR ahead of the torch update.

PR #1422, opened 2026-06-20
tidy: PyTorch now always at 2.x, remove unnecessary code that thinks it can still be 1.x

danra:
## Type of change

Remove code to handle torch version 1.x which can longer have any effect: `pyproject.toml` requires torch>=2.6

# Checklist:


danra:
ugh, closing, reopening PR to dev instead of main

PR #1437, opened 2026-06-23
Fix error opening jupyter notebook

danra:
# Description

`uv run jupyter notebook demos/xxx.ipynb` currently fails with:
    
       ModuleNotFoundError: No module named 'jupyter_server.contents'
    
This was fixed in jupyer 6.5.6 (see https://github.com/jupyter/notebook/issues/7048)

The PR updates the locked jupyter notebook version from v6.5.4->v6.5.7, resolving the issue.
A couple of minor drive-by fixes are also applied.

## Type of change

# Checklist:


uv.lock
```diff
@@ -2194,18 +2203,20 @@ wheels = [
 
 [[package]]
 name = "jupyter-client"
-version = "8.6.3"
+version = "7.4.9"
```
jlarson4, line 2206: Do we know why the `jupyter-client` dependency was downgraded?

pyproject.toml
```diff
@@ -73,7 +72,13 @@
         "sphinxcontrib-napoleon>=0.7",
         "tabulate>=0.9.0",
     ]
-    jupyter=["ipywidgets>=8.1.1", "jupyterlab>=3.5.0"]
+    jupyter=[
+        "ipython>=8.0.0",
+        "ipywidgets>=8.1.1",
+        "jupyter>=1.0.0",
+        "jupyterlab>=3.5.0",
+        "notebook>=6.5.4",
```
jlarson4: Should we bump this to `6.5.6`? Your PR note says the fix is in 6.5.6, but the min version is below that currently.

danra: Sure. I was being conservative, keeping the previous minimum that the previous `jupyter` meta-package imposed

uv.lock
danra: I don't. I noticed that as well. Probably makes sense to lock that to the previously locked version (or see why we can't)

pyproject.toml
jlarson4: I poked around in the changelogs for the package, I don't see anything that should be overtly problematic, feel free to bump it. I'll monitor for issues just in case.

uv.lock
jlarson4: From reading the patch notes on Jupyter 6.5.5, this is a required downgrade. There are incompatibilities with version 6 of `notebook` and version 8 of `jupyter-client` that were not properly pinned in Jupyter<6.5.5

As long as it's not causing any issues, we should be safe to leave this as is. Our notebook CI is passing and pulling this down locally everything seems to work for me.

jlarson4: I am probably going to add upgrading to `notebook` version 7 as a separate issue, once I finish investigating to determine if it will cause issues in our system. This will allow us to run `jupyter-client` version 8 again.

pyproject.toml
danra: Cool. I'll bump it to 6.5.7 already then

PR #1439, opened 2026-06-24
feat: respect prepend_bos and add return_input_tokens flag

### LINKED ISSUE #1418: consequences of tokenizer divergence between TransformersLens and Huggingface Transformers
I don't know if this is intended and I couldn't find much about it, it seems that the tokenizers of TransformersLens add a BOS token by default when encoding while Transformers doesn't.

I know this can be disabled with `skip_special_tokens`, but my concern is whether this is the case when we can call `model.generate` with a prompt as text, adding an unexpected `bos` breaking the model expected first token distribution silently.

From what I understand here, it will call `to_tokens` which will call `tokenizer.encode` without `add_special_tokens=False`.

So if my understanding is correct, this is seems like a footgun. Could someone clarify this behavior?

Maybe a nice to have is some special argument on generate that have it return the tokens that it used to generate to help debugging and making sure nothing unexpected was added.

**Code example**
```python
import torch
from transformer_lens.model_bridge.bridge import TransformerBridge
from transformers import AutoModelForCausalLM, AutoTokenizer

prompt = "2 + 2?"

MODEL = "allenai/OLMoE-1B-7B-0125-Instruct"
    MODEL,
    dtype=torch.bfloat16,
    device_map="auto",
    trust_remote_code=False,
)
model.eval()
tl_tokenizer = model.tokenizer

hf_tokenizer = AutoTokenizer.from_pretrained(MODEL)

v1 = tl_tokenizer.apply_chat_template(
    [{"role": "user", "content": prompt}],
    add_generation_prompt=True,
    return_tensors="pt",
    tokenize=True,
)["input_ids"][0].tolist()
v2 = tl_tokenizer.encode(
    tl_tokenizer.apply_chat_template(
        [{"role": "user", "content": prompt}],
        add_generation_prompt=True,
        return_tensors="pt",
        tokenize=False,
    )
)
print("Token check for transformer lens is", v1 == v2)
print(v1[:3], v2[:4])

v1 = hf_tokenizer.apply_chat_template(
    [{"role": "user", "content": prompt}],
    add_generation_prompt=True,
    return_tensors="pt",
    tokenize=True,
)["input_ids"][0].tolist()
v2 = hf_tokenizer.encode(
    hf_tokenizer.apply_chat_template(
        [{"role": "user", "content": prompt}],
        add_generation_prompt=True,
        return_tensors="pt",
        tokenize=False,
    )
)
print("Token check for hf transformers is", v1 == v2)
```

Output:
```
Token check for transformer lens is False
[50279, 29, 93] [50279, 50279, 29, 93]
Token check for hf transformers is True
```

MdSadiqMd:
# Description

The Bridge tokenizer is loaded with `add_bos_token=True`, diverging from the default HF tokenizer behavior. This causes two silent footguns:

1. **`generate(prepend_bos=...)` was ignored** — the parameter existed but emitted a warning and fell back to `cfg.default_prepend_bos`, preventing users from passing `prepend_bos=False` to avoid double-BOS when using chat templates that embed `<|begin_of_text|>`.
2. **No way to inspect what tokens the model actually received** — when generation silently prepended an unexpected BOS, there was no way to detect it from the `generate()` return value.

This PR fixes both issues and documents the BOS contract so users know when and why to set `prepend_bos=False`.

**Changes:**

- **`generate()` now respects `prepend_bos`** — the parameter is forwarded to `to_tokens()` instead of being discarded with a warning. `generate(text, prepend_bos=False)` correctly strips tokenizer-auto-prepended BOS, which is the required pattern when the input is pre-formatted chat-template text that already contains a BOS token.
- **`return_input_tokens` flag added** — when `True`, `generate()` returns `(output, input_tokens)` where `input_tokens` is the token tensor that was actually fed to the model (after BOS handling). Compatible with `return_cache=True` (returns 3-tuple `(output, cache, input_tokens)`).
- **"Tokenization notes" docstring updated** — explicitly documents that `model.tokenizer` has `add_bos_token=True` and differs from the stock HF tokenizer, that direct `.encode()` prepends BOS, and shows the recommended `prepend_bos=False` pattern for chat-template text.

Fixes #1418 

## Type of change

### Screenshots

N/A — code changes only. Verified manually with gpt2:

| `prepend_bos` | Token length | Starts with BOS |
|---|---|---|
| `True` (default) | 2 tokens | Yes |
| `False` | 1 token | No |

# Checklist:

### DIFF
--- a/demos/Main_Demo.ipynb
+++ b/demos/Main_Demo.ipynb
@@ -1015,9 +1015,7 @@
     "Mathematically, centering is a linear map, normalizing is *not* a linear map, and scaling and translation are linear maps. \n",
     "* **Centering:** LayerNorm is applied every time a layer reads from the residual stream, so the mean of any residual stream vector can never matter - `center_writing_weights` set every weight matrix writing to the residual to have zero mean. \n",
     "* **Normalizing:** Normalizing is not a linear map, and cannot be factored out. The `hook_scale` hook point lets you access and control for this.\n",
-    "* **Scaling and Translation:** Scaling and translation are linear maps, and are always followed by another linear map. The composition of two linear maps is another linear map, so we can *fold* the scaling and translation weights into the weights of the subsequent layer, and simplify things without changing the underlying computation. \n",
-    "\n",
-    "[See the docs for more details](https://github.com/TransformerLensOrg/TransformerLens/blob/main/further_comments.md#what-is-layernorm-folding-fold_ln)"
+    "* **Scaling and Translation:** Scaling and translation are linear maps, and are always followed by another linear map. The composition of two linear maps is another linear map, so we can *fold* the scaling and translation weights into the weights of the subsequent layer, and simplify things without changing the underlying computation. \n"
    ]
   },
   {
--- a/tests/unit/model_bridge/test_bridge_generate_bos.py
+++ b/tests/unit/model_bridge/test_bridge_generate_bos.py
@@ -1,0 +1,277 @@
+"""Unit tests for Bridge.generate() BOS handling and return_input_tokens.
+
+Tests cover:
+- prepend_bos parameter being respected (not ignored)
+- return_input_tokens flag returning input tokens
+- return_input_tokens + return_cache combo
+- generate_stream respecting prepend_bos
+"""
+
+import pytest
+import torch
+
+from transformer_lens.model_bridge import TransformerBridge
+
+
+@pytest.fixture(scope="module")
+def gpt2_bridge():
+    """Load a small GPT-2 bridge for testing."""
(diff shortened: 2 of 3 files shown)


MdSadiqMd:
@jlarson4 PR is up, please review it

jlarson4:
Thank you for putting this together @MdSadiqMd! Just two notes:

- Can you add unit tests for the new feature `return_input_tokens` and for the change to `prepend_bos`?
- `generate_stream` also needs `prepend_bos` threading, could you touch that one as well?

MdSadiqMd:
Hey @jlarson4, thanks for the review! I've added the requested changes. Please have a look when you get a chance

jlarson4:
Thank you @MdSadiqMd! Great work, merging

PR #1441, opened 2026-06-24
Fix/verification

huseyincavusbi:
# Description

Fixes three verification bugs found during Gemma4 adapter testing:

- **Component benchmark false failures** — Models using `DelegatedAttentionBlockBridge` (Gemma4) reported 81-96 component failures per model, dragging P1 from 100% to 50%. These are benchmark infrastructure failures. The component comparison can't call delegated attention/rotary/PLE modules in isolation because they require model-specific kwargs the benchmark doesn't provide. Added skip logic in `component_outputs.py` that detects `DelegatedAttentionBlockBridge` (via missing `hook_q_input` in hook aliases) and skips untestable components (attn, rotary_emb, per_layer projections). The gold-standard `forward_pass_logits` test was already passing. This just stops the benchmark from reporting false negatives.

- **Phase 2 empty header** — `main_benchmark.py` printed a "PHASE 2:" header with zero tests on every run regardless of whether Phase 2 was selected. Moved the header inside the `should_run_phase(2)` guard.

- **Misleading "encoder-decoder" warning** — The HF model loading log printed "for encoder-decoder model" for multimodal architectures using `AutoModelForImageTextToText`. Removed the incorrect label.

**Verification (tiny-random/gemma-4-e):**
- Dev branch: 15/64 component failures
- Fix branch: 0/31 component failures (all untestable components skipped)
- Unit tests: 167/167 passed,

## Type of change

# Checklist:

### DIFF
--- a/transformer_lens/benchmarks/component_outputs.py
+++ b/transformer_lens/benchmarks/component_outputs.py
@@ -200,15 +200,13 @@ class ComponentBenchmarker:
     """Benchmarking utility for testing TransformerBridge components against HuggingFace."""
 
     def _is_delegated_block(self) -> bool:
-        """Return True if the blocks component uses DelegatedAttentionBlockBridge."""
+        """Return True if the blocks component has maintain_native_attention set."""
         blocks = (
             getattr(self.adapter, "component_mapping", {}).get("blocks")
             if self.adapter is not None
             else None
         )
-        return blocks is not None and (
-            "hook_q_input" not in getattr(blocks, "hook_aliases", {"hook_q_input": True})
-        )
+        return getattr(blocks, "maintain_native_attention", False)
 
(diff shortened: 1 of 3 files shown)


jlarson4:
Looks good! One small suggestion: `_is_delegated_block()` detects delegation by checking that `hook_q_input` is missing from the block's hook aliases, which is a bit indirect. The benchmark already skips attention via an explicit flag (`maintain_native_attention` / `requires_position_embeddings`), you could set `maintain_native_attention=True` on `DelegatedAttentionBlockBridge`, which would then reuse the existing skip for that case, rather than adding the hook-alias check?

huseyincavusbi:
Done @jlarson4 ! Thanks for suggestion

PR #1443, opened 2026-06-25
Fix: Prevent crash when trying to load SigLIP/CLIP encoders with transformers >= 5.6.0

Aaron-d-k:
# Description

Currently, trying to run any model using the SigLIP or CLIP encoder (Eg: google/gemma-3-4b-it) using a default installation of transformer-lens from PyPI crashes due to a change introduced in the transformers library in 5.6.0 (commits: https://github.com/huggingface/transformers/pull/44431). Both SiglipVisionModel and CLIPVisionModel had their structure slightly changed.
This issue arises because currently pyproject.toml in transformer-lens only specifies a minimum version.
There are three fixes I considered:
1) Adding <5.6.0 to the toml
2) Porting the entire library to 5.6.0
3) Adding a backward compatible method that checks for the version and corrects for it.

I have done 3. My change should not affect any users with transformers<5.6.0 or those who do not use the model, so is entirely backward compatible. It basically checks if the required attribute is missing when `set_original_component` is called and tries to monkey patch it in if it is necessary.

## Type of change

### Screenshots
Before (transformers==5.8.1, demos/Gemma3_Multimodal.ipynb):
Screenshot 2026-06-26 at 00-30-51 Gemma3_Multimodal - Jupyter Notebook
After fix:
Screenshot 2026-06-26 at 00-29-13 Gemma3_Multimodal - Jupyter Notebook

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works (I am unsure how I should add tests for this)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@
         "torch>=2.6",
         "tqdm>=4.64.1",
         "transformers-stream-generator>=0.0.5,<0.1",
-        "transformers>=5.6.0",
+        "transformers>=5.4.0",
         "typeguard>=4.2,<5",
         "typing-extensions",
         "wandb>=0.13.5",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@
         "torch>=2.6",
         "tqdm>=4.64.1",
         "transformers-stream-generator>=0.0.5,<0.1",
-        "transformers>=5.4.0",
+        "transformers>=5.6.0",
         "typeguard>=4.2,<5",
         "typing-extensions",
         "wandb>=0.13.5",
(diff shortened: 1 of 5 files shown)


jlarson4:
Hi @Aaron-d-k! Thank you for both finding and fixing this bug. This is an excellent solution to the problem. 

For testing, I'd recommend following the example set in `tests/unit/model_bridge/supported_architectures/test_gemma3_multimodal_adapter.py`. It build a stub using `SimpleNamespace` that doesn't download the model, but can validate shape.

jlarson4:
I put together the tests and pushed them to your branch so I can get this merged before the next release, thank you for putting it together.

PR #1434, opened 2026-06-22
feat: add NemotronH hybrid Mamba2-Transformer architecture adapter

mukund1985:
Implements TransformerBridge support for `NemotronHForCausalLM` (nvidia/Nemotron-H-8B-Base, Nemotron-H-47B-A13B).

## Architecture

NemotronH is a hybrid with ~8% attention layers and ~92% Mamba-2 SSM/MoE/MLP layers. Each block has a single pre-norm (`block.norm`) and a single `.mixer` attribute whose type varies by layer (`layers_block_type` config field). There is no `ln2` or model-level rotary module.

## Adapter design

**SSMBlockBridge** is used as the block container — it delegates the full forward to the HF block without enforcing the `ln2` that `BlockBridge` requires.

**SSM2MixerBridge(name="mixer")** wraps `.mixer` for all four layer types. Its `forward` is a pure passthrough (`original_component(*args, **kwargs)`), so it works correctly for attention, MLP, and MoE mixers as well as Mamba ones.

Mamba-specific submodules (`in_proj`, `conv1d`, `inner_norm`, `out_proj`) are declared `optional=True` so `component_setup` skips them gracefully on non-Mamba layers. Note: `GatedRMSNormBridge.__init__` does not accept `optional=` (unlike the `GeneralizedComponent` base), so the attribute is set directly post-construction.

**Stateful generation** uses `DynamicCache` (transformers >= 5.12), which carries both KV-cache entries and SSM conv/recurrent states in one object.

`applicable_phases = []` — `verify_models` is transformer-shaped; forward-pass correctness lives in integration tests.

## Changes

- `transformer_lens/model_bridge/supported_architectures/nemotron_h.py` — new adapter
- `transformer_lens/model_bridge/supported_architectures/__init__.py` — export
- `transformer_lens/factories/architecture_adapter_factory.py` — registration
- `transformer_lens/tools/model_registry/__init__.py` — HF_SUPPORTED_ARCHITECTURES + CANONICAL_AUTHORS_BY_ARCH
- `tests/unit/model_bridge/supported_architectures/test_nemotron_h_adapter.py` — 52 unit tests, all passing

Closes #1402

PR #1445, opened 2026-06-27
fix: wire `slow` pytest marker — exclude large-model tests from CI

mukund1985:
## Problem

`pyproject.toml` registers the `slow` marker with the hint `deselect with '-m "not slow"'`, but no `make` target or CI step actually passes that flag. Any test decorated with `pytestmark = pytest.mark.slow` (e.g. the NemotronH integration test that requires loading an 18 GB checkpoint) runs unconditionally in every CI job, including `coverage-report-test`.

This was first noticed in #1434 where `test_nemotron_h_adapter.py` kept failing Full Code Coverage even after the marker was added.

## Fix

Add `-m "not slow"` to the four pytest invocations that run in CI:

- `make unit-test`
- `make integration-test`
- `make acceptance-test`
- `make coverage-report-test`

Also add it to the MPS integration test step in `checks.yml`, which currently uses `--ignore` flags as a workaround for the same problem.

## Testing

All existing tests continue to pass. Tests tagged `@pytest.mark.slow` can still be run explicitly with:

```bash
pytest -m slow tests/
```

or by running them directly by path.

Closes the CI gap identified in #1434.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -169,7 +169,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
       - name: MPS Integration Tests
         run: >
-          uv run pytest tests/integration -v
+          uv run pytest tests/integration -v -m "not slow"
           --ignore=tests/integration/model_bridge/test_optimizer_compatibility.py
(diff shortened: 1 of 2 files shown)


mukund1985:
Closing — folding this fix into #1434 instead.

PR #1450, opened 2026-06-29
Construct causal masks on the fly

### LINKED ISSUE #479: [Proposal] Memory efficient causal mask implementation

abhinav-bellapu:
# Description

Construct HookedTransformer causal masks on the fly for the current attention-score shape instead of registering a full `(cfg.n_ctx, cfg.n_ctx)` mask buffer for every attention layer.

Fixes #479.

This revives the direction of #493, which was closed as abandoned rather than rejected. The change is intentionally narrow:

- keep `mask` registered as an empty bool buffer for state-dict/device compatibility
- build global/local causal masks inside `apply_causal_mask(...)` based on current query/key lengths and `past_kv_pos_offset`
- preserve strict loading of legacy state dicts by replacing old full-size saved masks with the new empty compatibility placeholder during load
- stop nanoGPT conversion from materializing full attention masks

I left the existing Qwen context-length caps unchanged. Those comments mention the old mask allocation problem, but lifting model-specific caps seems safer as a follow-up after model-specific validation.

Because this touches HookedTransformer attention internals, I am opening it as a draft and would appreciate maintainer feedback on whether this minimal per-attention implementation is preferred, or whether you would rather see a centralized model-level mask construction approach.

## Type of change

# Checklist

## Validation

Local checks run:

```bash
uv run --no-sync pytest tests/unit/components/test_attention.py tests/unit/components/test_abstract_attention.py tests/integration/test_grouped_query_attention.py -q
make format
make check-format
uv run mypy .
git diff --check
```

Results:

- targeted pytest: `13 passed, 4 skipped`
- `make format`: passed
- `make check-format`: passed
- `uv run mypy .`: passed, `Success: no issues found in 298 source files`
- `git diff --check`: passed

I did not run the full `make test-pr` locally to avoid pulling larger model-test artifacts onto my machine; this draft PR is intended to let CI exercise the heavier suite.

### DIFF
--- a/tests/unit/components/test_attention.py
+++ b/tests/unit/components/test_attention.py
@@ -121,6 +121,120 @@ def test_attention_config_dict():
     assert attn.cfg.act_fn == "relu"
 
 
+def test_attention_does_not_allocate_full_causal_mask():
+    cfg = HookedTransformerConfig(
+        n_layers=1,
+        d_model=4,
+        n_ctx=8192,
+        d_head=2,
+        n_heads=2,
+        act_fn="relu",
+    )
+
+    attn = Attention(cfg)
+
+    assert attn.mask.shape == (0, 0)
+    assert attn.state_dict()["mask"].numel() == 0
+
+
+def test_apply_causal_mask_global_matches_absolute_positions():
+    cfg = HookedTransformerConfig(
+        n_layers=1,
+        d_model=4,
+        n_ctx=16,
+        d_head=2,
+        n_heads=2,
+        act_fn="relu",
+    )
+    attn = Attention(cfg)
+    attn_scores = torch.zeros((1, 1, 3, 5))
+
+    masked_scores = attn.apply_causal_mask(attn_scores, past_kv_pos_offset=2)
+
+    expected_allowed = torch.tensor(
+        [
+            [True, True, True, False, False],
+            [True, True, True, True, False],
+            [True, True, True, True, True],
+        ]
+    )
+    assert torch.equal(torch.isfinite(masked_scores[0, 0]), expected_allowed)
+
+
+def test_apply_causal_mask_local_matches_window():
+    cfg = HookedTransformerConfig(
+        n_layers=1,
+        d_model=4,
+        n_ctx=16,
+        d_head=2,
+        n_heads=2,
+        act_fn="relu",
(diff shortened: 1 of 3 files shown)


jlarson4:
Great work on this @abhinav-bellapu! The per-attention approach was absolutely correct. The amount of compute we'd save making it a centralized model-level mask is not worth the effort it would take to make and maintain it.

You also did an excellent job of ensuring backwards compatibility. This also brings `HookedTransformer`'s causal mask system inline with `TransformerBridge`. Definitely nice to have for those who are still on `HookedTransofmer`. I am going to close and merge as is. If you'd like to tackle the Qwen updates as a follow up, feel free. If you'd prefer I post it as an issue for another contributor to tackle, I can do that as well, just let me know!

Merging as-is, thank you for putting this together!

PR #1453, opened 2026-06-29
Use Qwen config context lengths

abhinav-bellapu:
# Description

Follow-up to #1450.

Now that `HookedTransformer` builds causal masks on the fly, two Qwen config paths still had stale `n_ctx = 2048` caps whose comments said the cap existed because the full attention mask would be too large.

This PR removes those mask-era caps:

- `QWenLMHeadModel` now uses `hf_config.seq_length`
- `Qwen2ForCausalLM` now uses `hf_config.max_position_embeddings`

I intentionally left `Qwen3ForCausalLM` unchanged. Its `2048` value came from the later Qwen3 support PR and does not have the same stale attention-mask comment, so changing it should be a separate validation decision if desired.

One tradeoff worth calling out: this removes the old quadratic mask allocation problem, but long-context rotary models still precompute rotary sin/cos buffers proportional to `n_ctx`. That is much smaller than the old `(n_ctx, n_ctx)` mask allocation, but Qwen2 models can expose very large context lengths. Users can still pass `n_ctx=...` to choose a smaller context length if needed.

## Type of change

# Checklist

## Validation

Local checks run:

```bash
uv run --no-sync pytest tests/unit/test_loading_from_pretrained_utilities.py -q
uv run --no-sync python - <<'PY'
from transformer_lens.loading_from_pretrained import get_pretrained_model_config

for model in [
    "Qwen/Qwen-7B",
    "Qwen/Qwen1.5-0.5B",
    "Qwen/Qwen2-0.5B",
]:
    cfg = get_pretrained_model_config(model, trust_remote_code=True)
    print(f"{model}: architecture={cfg.original_architecture}, n_ctx={cfg.n_ctx}")
PY
make format
make check-format
uv run mypy .
git diff --check
```

Results:

- `tests/unit/test_loading_from_pretrained_utilities.py`: `11 passed`
- real config smoke check:
  - `Qwen/Qwen-7B`: `n_ctx=8192`
  - `Qwen/Qwen1.5-0.5B`: `n_ctx=32768`
  - `Qwen/Qwen2-0.5B`: `n_ctx=131072`
- `make format`: passed
- `make check-format`: passed
- `uv run mypy .`: passed, `Success: no issues found in 299 source files`
- `git diff --check`: passed

### DIFF
--- a/tests/unit/test_loading_from_pretrained_utilities.py
+++ b/tests/unit/test_loading_from_pretrained_utilities.py
@@ -1,3 +1,4 @@
+from types import SimpleNamespace
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


jlarson4:
Thanks for putting together this follow up! After investigating the tradeoff you mentioned, I have a couple additional details we probably want to address.

The description compares the new rotary buffers to "the old (n_ctx, n_ctx) mask," but for these models that full-size mask never existed: the 2048 cap held the mask to 2048², and #1450 dropped the stored mask entirely. The real baseline is the previous rotary buffer, which was also capped at 2048. `rotary_sin`/`rotary_cos` are built at full `n_ctx` per layer at construction (`abstract_attention.py:164-171`, fp32 by default), so this PR grows them ~67×:

| `n_ctx` | × 24 layers (Qwen2-0.5B) |
|---|---|
| 2048 (old) | 0.03 GB |
| 131072 (new) | **1.61 GB** |

That 1.6 GB lands on top of the ~2 GB of fp32 weights which is roughly +80% load memory for the 0.5B model, paid by every user at construction, even if they don't need a buffer that large.

We do have a solution that exists to help with this. `apply_rotary` extends these buffers on demand (`abstract_attention.py:747-750`), so they don't need full-`n_ctx` sizing up front. Bound the initial precompute (e.g. `min(n_ctx, 2048)`) and let the existing extension path grow it lazily. Then short-context users pay ~1 MB/layer and `n_ctx` still reflects the true context length. 

Two points you'll probably need to touch if you go this route: `_extend_rotary_embeddings` uses the wrong RoPE base for local-attention layers (`getattr(cfg, "rotary_base")` vs the `rotary_base_local` selection at lines 160-163). This issue is currently latent, but will become a problem once extension is the norm and the persistent rotary buffers will need the same `_load_from_state_dict` size shim #1450 added for mask.

Feel free to make that a separate PR first, or bundle it into this one, your call. Once this is complete, it should be safe to also remove the buffer from Qwen3.

abhinav-bellapu:
Opened #1455 as the separate follow-up for the rotary-buffer side of your comment. It keeps this Qwen config PR focused on using the HF context lengths, while #1455 handles the enabling rotary changes: bounding the initial cache to `min(n_ctx, 2048)`, using the local RoPE base during lazy extension, and adding the strict-load shim for legacy full-size rotary buffers.

abhinav-bellapu:
Update: #1455 is now ready for review and all checks are green. The intended merge order is #1455 first, then this PR, so the Qwen context-length change lands after the rotary cache memory concern is addressed.

jlarson4:
#1455 is merged, so I will now merge this one as well

PR #1455, opened 2026-06-30
Bound initial rotary cache size

abhinav-bellapu:
# Description

Follow-up/prep for #1453 based on maintainer feedback there.

`HookedTransformer` rotary attention currently precomputes persistent `rotary_sin` / `rotary_cos` buffers out to full `cfg.n_ctx` for every rotary attention layer during construction. For long-context configs, that means short-context users pay the full long-context rotary-cache memory cost up front.

This PR bounds the initial rotary precompute to `min(n_ctx, 2048)` and keeps the existing `apply_rotary` extension path responsible for growing the buffers on demand when a longer position is actually used. `cfg.n_ctx` still reflects the true supported context length.

It also fixes two compatibility details needed for making lazy extension the normal path:

- `_extend_rotary_embeddings` now uses the same local-vs-global RoPE base selection as construction, so local-attention layers use `rotary_base_local` when configured.
- `_load_from_state_dict` now replaces mismatched `rotary_sin` / `rotary_cos` buffers with the current buffers, mirroring the existing mask size shim, so older checkpoints with full-size rotary buffers can still load with `strict=True`.

## Type of change

# Checklist:

## Validation

Local checks run:

```bash
uv run --no-sync pytest tests/unit/components/test_attention.py -q
make format
make check-format
uv run mypy .
make unit-test
git diff --check
make test-pr
```

Results:

- `tests/unit/components/test_attention.py`: `13 passed, 4 skipped`
- `make format`: passed
- `make check-format`: passed
- `uv run mypy .`: passed, `Success: no issues found in 299 source files`
- `make unit-test`: passed, `3220 passed, 31 skipped, 42 deselected, 9 xfailed`
- `git diff --check`: passed
- `make test-pr` progressed through:
  - unit: passed, `3220 passed, 31 skipped, 42 deselected, 9 xfailed`
  - docstring: passed, `18 passed, 25 skipped`
  - acceptance: passed, `137 passed, 83 skipped, 11 deselected, 1 rerun`
  - integration: blocked locally before test execution by a macOS temporary-venv import hang in the notebook pytest plugin path (`nbval` / `jupyter_client` / `zmq`). CI should run this in a clean environment.

### DIFF
--- a/tests/unit/components/test_attention.py
+++ b/tests/unit/components/test_attention.py
@@ -137,6 +137,142 @@ def test_attention_does_not_allocate_full_causal_mask():
     assert attn.state_dict()["mask"].numel() == 0
 
 
+def test_rotary_embeddings_initial_cache_is_bounded():
+    cfg = HookedTransformerConfig(
+        n_layers=1,
+        d_model=8,
+        n_ctx=8192,
+        d_head=4,
+        n_heads=2,
+        act_fn="relu",
+        positional_embedding_type="rotary",
+    )
+
+    attn = Attention(cfg)
+    rotary_dim = cfg.rotary_dim
+    assert rotary_dim is not None
+
+    assert attn.cfg.n_ctx == 8192
+    assert attn.rotary_sin.shape == (2048, rotary_dim)
+    assert attn.rotary_cos.shape == (2048, rotary_dim)
+
+
+def test_rotary_embeddings_initial_cache_matches_short_context():
+    cfg = HookedTransformerConfig(
+        n_layers=1,
+        d_model=8,
+        n_ctx=128,
+        d_head=4,
+        n_heads=2,
+        act_fn="relu",
+        positional_embedding_type="rotary",
+    )
+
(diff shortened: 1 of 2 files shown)


jlarson4:
Great work! Approved and merging.

One additional note that's outside the scope of what I recommended originally:

`_extend_rotary_embeddings` recomputes the entire buffer on each growth, and the trigger extends to exactly `max_pos_needed`. For a long-prompt prefill that's one extension, but for token-by-token generation beyond 2048 (KV cache, `x_pos=1`, `max_pos_needed` +1/step), every step recomputes the whole growing buffer which means ~`O(n^2)` rotary recompute over a long generation. It's pre-existing behavior, but this PR makes lazy extension the normal path for exactly the long-context models that would hit it. A cheap mitigation: grow with headroom (example: `min(n_ctx, max(max_pos_needed, 2 * current_size))`) so extension fires `O(log n)` times instead of every step. 

Feel free to do another follow up PR if you're interested!

Issue #1280, opened 2026-05-05
[Proposal] Add support for `cpu`, `meta`, and `disk` to TransformerBridge `device_map`


jlarson4:
### Proposal 

Support `cpu`, `meta`, and `disk` for the `device_map` parameter of `boot_transformers`.

### Motivation

Please outline the motivation for the proposal.
Is your feature request related to a problem? e.g.,"I'm always frustrated when [...]".
If this is related to another GitHub issue, please link here too.

### Pitch

We want to allow `cpu` and other non-GPU devices to be used by torch for multi-device processing. At current, we error when any of the above types are requested in a multi-device setup.

### Additional context

The dtype-cast loop in transformers.py:550-552 is the blocker for `cpu` support:
```
for param in hf_model.parameters():
    if param.is_floating_point() and param.dtype != dtype:
        param.data = param.data.to(dtype=dtype)
```
For meta/offloaded params, .to(dtype) either no-ops or breaks accelerate's bookkeeping. Two small fixes should unlock CPU offload:

- [ ] Skip params on meta device (offload placeholders)
- [ ] Use accelerate's `align_module_device` context manager when iterating offloaded params

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

jlarson4:
See #1270 for details on the initial setup of `device_map` and #872 for the original discussion about this topic

snakefood3232:
I see the blocker is the dtype-cast loop in transformers.py, lines 550-552, which affects CPU support. I'd handle this by skipping parameters on the meta device and using Accelerate's `align_module_device` for offloaded parameters. I can have a PR ready in 3 days. I've built similar device-mapping solutions in distributed systems before.

jlarson4:
@snakefood3232 Excellent! I will assign this to you, and keep my eyes open for a PR. Thank you for your contribution!

jlarson4:
@snakefood3232 Just wanted to touch base and see if you've had an opportunity to work on this? I'd like to open it up to other contributors if not

hiro-nikaitou:
I have opened PR #1493 to add meta device target support to TransformerBridge device_map, building on the CPU support from #1459. This covers the meta portion of this issue — disk offload remains blocked pending the hook-routing fix described in #1459.

LightWork666:
[...]

PR #1459, opened 2026-06-30
Support CPU device_map in TransformerBridge

abhinav-bellapu:
# Description

Part of #1280.

This PR enables explicit CPU targets in `TransformerBridge` `device_map` dictionaries. Previously `resolve_device_map` rejected `"cpu"`, `"disk"`, and `"meta"` uniformly; this changes the boundary so CPU placement is passed through to HuggingFace while `disk` / `meta` remain rejected with a clearer error.

It also replaces the raw `hf_model.parameters()` dtype-cast loop with `cast_floating_params_to_dtype()`, which uses `accelerate.utils.align_module_device()` at the module level. That keeps dtype casting safe around Accelerate-managed offload placeholders and preserves float32 buffers.

This intentionally does **not** claim full disk/meta offload support yet. During local smoke testing, an explicit GPT-2 map with one block on `disk` loaded successfully but forward hit a `meta` tensor inside a Bridge-wrapped component, meaning disk/meta support needs a separate hook-routing fix.

## Type of change

# Checklist:

## Validation

- `make format`
- `make check-format`
- `uv run mypy .`
- `uv run pytest tests/acceptance/model_bridge/test_multi_gpu_bridge.py -q`
- Smoke: `TransformerBridge.boot_transformers("gpt2", device_map={"": "cpu"})` followed by a one-token forward pass

### DIFF
--- a/tests/acceptance/model_bridge/test_multi_gpu_bridge.py
+++ b/tests/acceptance/model_bridge/test_multi_gpu_bridge.py
@@ -206,6 +206,17 @@ class TestBootParamValidation:
         with pytest.raises(ValueError, match="mutually exclusive"):
             TransformerBridge.boot_transformers("gpt2", device="cpu", device_map="auto")
 
+    def test_explicit_cpu_device_map_boots_and_runs(self):
+        bridge = TransformerBridge.boot_transformers("gpt2", device_map={"": "cpu"})
+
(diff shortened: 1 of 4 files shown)


jlarson4:
This looks good, thank you for taking it on!

PR #1484, opened 2026-07-03
Add AST (Audio Spectrogram Transformer) Adapter

### LINKED ISSUE #1446: [Proposal] Add Audio Spectrogram Transformer (AST) Adapter (ASTForAudioClassification)
### Proposal 

Add a TransformerBridge adapter for the Audio Spectrogram Transformer family (ASTForAudioClassification)

### Motivation

TransformerLens currently supports audio through the HookedAudioEncoder (HuBERT/Wav2Vec2), and this proposal would add an adapter for the Vision-Transformer (ViT) based audio models. The AST is widely used and processes spectrograms, which are 2D images representing audio data, making it a prime candidate for Mechanistic Interpretability & TransformerLens, expanding the library's multi modal capabilities.

### Pitch

I plan to implement the AST adapter for the TransformerBridge architecture, including:
- mapping the HuggingFace AST configuration to HookedTransformerConfig
- writing the weight translation map for the dense ViT-style attention and MLP blocks
- implementing the patch embedding extraction layer (handling the 2D Conv2D projection)to ensure alignment between input spectrograms and the residual stream before the transformer blocks

I have a fair amount of experience with this particular model, and will gladly work on the implementation and testing for this adapter.

### Alternatives

I've considered doing this for ViT-base given its backbone is functionally identical to the AST (the AST input layers are altered, but the models' innards are the same), but decided to start with the AST since I'd be surprised if we don't have this for ViT-base already. Adapting the work of this proposal for ViT-base in the future should be simple.

### Additional context

HF for AST model: https://huggingface.co/docs/transformers/model_doc/audio-spectrogram-transformer

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/Transformerlens/issues) in the repo (**required**)

dylanberens:
# Description

Refactored the AST architecture adapter to align with the V3 'TransformerBridge' architecture and target 'dev'.

### Key Changes
* **V3 Bridge Architecture:** Implemented 'ASTArchitectureAdapter' using hierarchical 'component_mapping' ('BlockBridge', 'AttentionBridge', 'MLPBridge', 'NormalizationBridge', and 'UnembeddingBridge') & added ASTForAudioClassification to architecture adapter factory
* **Weight Processing:** Ported head-splitting tensor reshapes ('RearrangeTensorConversion') for Query/Key/Value/Output projections to handle dimensional conversion during loading
* **Dynamic Context Length:** Utilized 'prepare_model' hook to dynamically extract 'n_ctx' from positional embeddings length
* **Unit Test:** Updated unit tests under 'tests/unit/model_bridge/supported_architectures/test_ast_adapter.py' asserting exact bridge-vs-HF parity

Fixes # 1484

## Type of change

# Checklist:


jlarson4:
Thanks for putting this together @dylanberens! The AST domain work is great, the Q/K/V/O and bias reshapes, MLP transposes, the overlapping-patch Conv2d, the cls-distillation-patches token order, and the pooled dual-token + dual-LayerNorm head are all correct. 

We will need to address & adjust the approach: this is written against the legacy **HookedTransformer** `convert_weights` path, but TransformerBridge adapters are driven by **`component_mapping`** (wrap HF's live modules). As submitted it can't load or run, which is why your tests are failing in CI.

**Primary issues with the current code:**
1. Doesn't integrate with the bridge – `ASTAdapter` never sets `self.component_mapping` (it defines no `__init__`, so it inherits `None`), and it isn't registered in the factory. Booting AST raises `ValueError: Unsupported architecture`. `get_config_map`/`convert_weights` are legacy-HT classmethods no bridge path ever calls.
2. The test fails and never touches the bridge – It KeyErrors at `ast.py:125`, the HF keys are wrong on two axes: `layers.{i}` should be `encoder.layer.{i}`, and `q_proj/k_proj/v_proj/o_proj`/`mlp.fc1/fc2` should be `attention.attention.{query,key,value}` / `attention.output.dense` / `intermediate.dense` / `output.dense`. So the `max diff < 1e-4` claim isn't reproducible, and the test builds a raw `HookedTransformer`, not a bridge.

**To make this mergeable:**
- Rewrite as a `component_mapping` adapter: I'd recommend following our [Architecture Adapter Creation Guide](https://transformerlensorg.github.io/TransformerLens/content/adapter_development/adapter-creation-guide.html). The `hubert.py` adapter is our current working audio adapter, it would be a good reference point to work off.
- Delete `ASTEmbed` and both classmethods — the bridge wraps HF's own modules, so none of that math needs reimplementing.
- Fix `n_ctx = num_patches+2`; read patch/stride/`hidden_act` from config.
- Register `ASTForAudioClassification` in the factory + export it from `__init__`.
- Rewrite the test to boot a real `TransformerBridge` and assert bridge-vs-HF fp32 parity, under `tests/unit/...`.

Also, a couple repo items: make sure to drop `poetry.lock` from the PR. Starting from 3.0.0, the repo is uv-based. And, base this change to `dev`, we will be able to launch this as part of a 3.x release, it does not need to be gated behind version 4.

Happy to re-review once it's on the TransformerBridge pattern.

dylanberens:
thanks @jlarson4 for the feedback and roadmap! I've refactored the AST adapter to follow the V3 TransformerBridge pattern against dev

- removed custom embedding reimplementations and adopted hierarchical component_mapping
- ported q/k/v/o head splitting tensor conversions
- utilized prepare_model to resolve n_ctx from positional embeddings
- rebase to dev, added the ASTForAudioClassification to architecture adapter factory, added FP32 bridge parity unit tests (passing locally with exact 0.0 difference)

ready for re-review when you get a chance, thanks jonah!

jlarson4:
Excellent, thank you @dylanberens! I will make sure to let you know when I get around to reviewing

transformer_lens/model_bridge/supported_architectures/ast.py
```diff
@@ -0,0 +1,103 @@
...
+    MLPBridge,
+    NormalizationBridge,
+    UnembeddingBridge,
+)
+from transformer_lens.model_bridge.generalized_components.base import (
+    GeneralizedComponent,
+)
+
+
+class ASTArchitectureAdapter(ArchitectureAdapter):
```
jlarson4, line 26: This is the only adapter class in `supported_architectures/`) without a docstring, and the file also lacks a module docstring. AST has enough quirks that it would really benefit from one: the spectrogram input contract, the cls+distillation token layout, and especially the head structure.

It also needs the encoder-only class attribute:

```
class ASTArchitectureAdapter(ArchitectureAdapter):
    """Architecture adapter for AST (Audio Spectrogram Transformer) audio classifiers.

    Input is a [batch, time=1024, n_mels=128] spectrogram; positions 0/1 are the
    CLS and distillation tokens. HF pools (cls+dist)/2 after ln_final and applies
    an extra LayerNorm inside the classifier head, see the unembed mapping note.
    """

    supports_generation: bool = False
```

Without it, AST inherits `True` and every generation entry point will fail with unclear errors instead of the clean `NotImplementedError` gate at `bridge.py:2721`. I tested this and received `AssertionError: Must pass eos_token_id... from generate(spectrogram)`, and `IndexError` at `bridge.py:2631` with `stop_at_eos=False`.

transformer_lens/model_bridge/supported_architectures/ast.py
```diff
@@ -0,0 +1,103 @@
...
+            ),
+        }
+
+        # V3 bridge pattern: hierarchical mapping using bridge components
+        self.component_mapping = {
+            "embed": GeneralizedComponent(name="audio_spectrogram_transformer.embeddings"),
+            "ln_final": NormalizationBridge(
+                name="audio_spectrogram_transformer.layernorm", config=self.cfg
+            ),
+            "unembed": UnembeddingBridge(name="classifier.dense"),
```
jlarson4: HuggingFace's classification path after `ln_final` is: pool `(seq[:,0]+seq[:,1])/2` -> `classifier.layernorm` -> `classifier.dense`. Mapping unembed to `classifier.dense` alone means:

- `unembed.hook_in` is the pooled, post-head-LayerNorm `[batch, d_model]` tensor. There are two unhooked ops downstream of `ln_final.hook_out` `[batch, seq, d_model]`. On every text model, unembed input is the `ln_final` output. Users expecting that convention get silently different semantics here.
- The pooling step and `classifier.layernorm` have no hook points at all, neither is observable or intervenable.

The actual results are unaffected, so I'm fine shipping this mapping if documented in the class docstring, but consider mapping the head LN as a component (or the whole classifier) so the head is hookable. 

Either way, a couple related small bugs:

- `ASTConfig(num_labels=0)` makes `classifier.dense` an `nn.Identity`, and boot then dies with `AssertionError: Component classifier.dense has no weight attribute (bridge.py:667)`. Guard it in `prepare_model` (drop the unembed mapping, like HuBERT's conditional unembed injection) or raise a clear unsupported-config error.
- `cfg.d_vocab`/`d_vocab_out` stay at the -1 sentinel after boot even though `W_U` is a real `(d_model, num_labels)` head. Set `self.cfg.d_vocab_out` (and arguably `d_vocab`) from `num_labels` so downstream consumers don't read -1.

transformer_lens/model_bridge/supported_architectures/ast.py
```diff
@@ -0,0 +1,103 @@
...
+                        },
+                    ),
+                },
+            ),
+        }
+
+    def prepare_model(self, hf_model: Any) -> None:
+        # hook to access the live Huggingface model before boot
+        # calculate n_ctx dynamically from the instantiated position embeddings
+        self.cfg.n_ctx = (
```
jlarson4: The dynamic `n_ctx` is the right call, but the hardcoded `hf_model.audio_spectrogram_transformer`. access is one of the two things that breaks real boots: a bare `ASTModel` has no such attribute. 

Please make this and the component_mapping prefix-aware following `hubert.py`'s `_build_component_mapping(prefix=...)` pattern. It solves the bare-encoder case and a real boot blocker. This is also the natural place for the `num_labels` -> `d_vocab_out` propagation mentioned above.

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
```diff
@@ -0,0 +1,56 @@
+import torch
+from transformers import ASTConfig, ASTForAudioClassification
+
+from transformer_lens.model_bridge import TransformerBridge
+
+
+def test_ast_parity():
```
jlarson4: We could use a bit more coverage from the testing. We have a guide on what we typically expect from adapter tests [here](https://transformerlensorg.github.io/TransformerLens/content/adapter_development/adapter-unit-test-guide.html). Let me know if you have any questions!

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # explicitly define the architecture list for the local config
+    hf_config.architectures = ["ASTForAudioClassification"]
+
+    hf_model = ASTForAudioClassification(hf_config)
+    hf_model.eval()
+
+    # 2. boot the v3 transformerbridge natively
+    # bridge auotmatically detecty ASTForAudioClassification and use new ASTArchitectureAdapater
+    tl_model = TransformerBridge.boot_transformers("ast", hf_model=hf_model)
```
jlarson4: Because the adapter sets `is_audio_model=True`, boot reaches the processor block (sources/transformers.py:977-989) and calls `AutoFeatureExtractor.from_pretrained("ast")`. This attempts to find a HF repo literally named `ast` on every unit-test run. If anyone ever publishes a repo named `ast` with a `preprocessor_config.json`, this test silently loads third-party content. Use an unmistakably-local placeholder and/or monkeypatch the processor load.

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
```diff
@@ -0,0 +1,56 @@
...
+        # apply AST pooling logic to the bridges residual stream
+        cls_token = resid[:, 0, :]
+        dist_token = resid[:, 1, :]
+        pooled_out = (cls_token + dist_token) / 2.0
+
+        # apply the final classifier (which holds 2nd layernorm + dense)
+        tl_logits = hf_model.classifier(pooled_out)
+
+    diff = (hf_logits - tl_logits).abs().max().item()
+    print(f"Bridge vs HF Max Logit Diff: {diff:.6e}")
```
jlarson4: Drop the `print()`s (pytest reporting covers it) and the `if __name__ == "__main__": block. The TransformerLens standard are assertions with messages.

transformer_lens/tools/model_registry/__init__.py
```diff
@@ -47,6 +47,7 @@
 HF_SUPPORTED_ARCHITECTURES: set[str] = {
     "ApertusForCausalLM",
     "ArceeForCausalLM",
+    "ASTForAudioClassification",
```
jlarson4, line 51: There are a couple other locations you'll need to register this:

1. `AUDIO_ARCHITECTURES` in `transformer_lens/utilities/architectures.py` needs "ASTForAudioClassification". This is routes `verify_models` to the audio phase set `{1, 8}` instead of text phases (see review body) when running benchmarks to confirm individual model support. 
2. `ARCHITECTURE_DESCRIPTIONS` in `generate_report.py` this is a nice-to-have, not essential for PR merge

jlarson4 (commented):
Hey @dylanberens sorry it took me so long to get around to reviewing this. Full detailed review broken down below. The overall structure and content of your adapter are very good! There are just a couple small items that need addressing before I can merge. Once the comments below are resolved, this should be good to go!

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
dylanberens: resolved, dropped print in most recent commit

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
dylanberens: resolved, expanded test_ast_adapter.py with additional tests and adjusted for preferred structure

transformer_lens/model_bridge/supported_architectures/ast.py
dylanberens: resolved, prefix aware fix replaced hardcoded .audio_spectrogram_transformer. with variable

transformer_lens/model_bridge/supported_architectures/ast.py
dylanberens: resolved, most recent commit adds docstrings and encoder-only class attribute

transformer_lens/model_bridge/supported_architectures/ast.py
dylanberens: resolved, added docstring for the pooling/layernorm behavior explanation, added unembed guard in prepare_model and propogated num_labels to d_vocab/d_vocab_out for downstream users to get actual label count

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
dylanberens: resolve, used monkeypatch for processor loading, avoiding silently loading ast from huggingface

transformer_lens/tools/model_registry/__init__.py
dylanberens: resolved, registered in AUDIO_ARCHITECTURES. 

did not add to ARCHITECTURE_DESCRIPTIONS in generate_report.py in this commit

jlarson4:
Thanks @dylanberens for the update! The fixes look good, I verified them all locally. Three things left before merge:

1. The real load path still can't reach your fixes – `AUDIO_ARCHITECTURES` routes AST to bare `AutoModel`, so a real `boot_transformers("MIT/ast-...")` loads a headless `ASTModel` without the classifier, and the `audio_spectrogram_transformer.` prefix branch never fires. #1546 added an perfect example: mirror `VISION_CLASSIFICATION_ARCHITECTURES` -> `AutoModelForAudioClassification` in `utilities/architectures.py` and `sources/transformers.py`. Please also add a boot test,  `boot_transformers(<checkpoint>, load_weights=False)` is cheap and fails today, so it'll pin the fix.

2. `"ASTModel"` in `AUDIO_ARCHITECTURES` is currently inert – it isn't registered in the factory or registry, so bare-encoder checkpoints still fail at adapter selection while the docstring claims support. Either register it in the other sites or drop the entry and narrow the docstring.

3. The parity test belongs in the integration folder (`tests/integration/model_bridge/test_ast_tiny.py`). A new adapter needs one there regardless.

transformer_lens/tools/model_registry/__init__.py
dylanberens: most recent commit adds to ARCHITECTURE_DESCRIPTIONS in generate_report.py just to be thorough

dylanberens:
@jlarson4 most recent commit split parity to integration folder, add load_weights boot test, fix audio classification load path, and add to ARCHITECTURE_DESCRIPTIONS in generate_report.py just to be thorough

also for ASTModel i opted to drop the entry and narrow the docstring, thanks man

transformer_lens/utilities/architectures.py
```diff
@@ -69,7 +69,10 @@
     "HubertForCTC",
     "HubertModel",
     "HubertForSequenceClassification",
-    "ASTModel",
+}
+
+# Audio spectrogram models for classification
+AUDIO_CLASSIFICATION_ARCHITECTURES: set[str] = {
```
jlarson4: We've now taken `ASTForAudioClassification` out of `AUDIO_ARCHITECTURES`, meaning`classify_architecture`  now returns `"causal_lm"` again and `verify_models` will put `ASTForAudioClassification` models through text phases `{1,2,3,4}`, undoing last round's registration fix. 

On the boot path, your new branch is checked first so this was a bit tricky to catch. Only functionality that uses `classify_architecture` breaks. The vision precedent unions the two sets (`VISION_ARCHITECTURES = VISION_MODEL_ARCHITECTURES | VISION_CLASSIFICATION_ARCHITECTURES`), it should be safe to replicate that here.


You'll just need to move `AUDIO_CLASSIFICATION_ARCHITECTURES` above `AUDIO_ARCHITECTURES` for this to work.

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
```diff
@@ -73,37 +72,13 @@ def test_classification_prefix_rebinding(self, hf_config, tl_config):
...
-        diff = (hf_logits - tl_logits).abs().max().item()
-
-        # 5. assert parity
-        assert diff < 1e-4, "Parity failed: Bridge tensors do not match HuggingFace."
+    def test_ast_boot_transformers_load_weights_false(self):
+        # load_weights=False prevents downloading massive model binaries during unit tests
+        tl_model = TransformerBridge.boot_transformers(
+            "MIT/ast-finetuned-audioset-10-10-0.4593", load_weights=False
+        )
+        assert tl_model.cfg.architecture == "ASTForAudioClassification"
```
jlarson4: I ran these two assertions against an early commit where they should fail and both still pass, so this test doesn't fail on the bug it's guarding. Lets assert something only the fixed path produces:
```
assert type(tl_model.original_model).__name__ == "ASTForAudioClassification"
assert "unembed" in tl_model.adapter.component_mapping
```

transformer_lens/model_bridge/supported_architectures/ast.py
```diff
@@ -0,0 +1,141 @@
...
+    GeneralizedComponent,
+)
+
+
+class ASTArchitectureAdapter(ArchitectureAdapter):
+    """Architecture adapter for AST (Audio Spectrogram Transformer) audio classifiers.
+
+    Input is a [batch, time=1024, n_mels=128] spectrogram; positions 0/1 are the
+    CLS and distillation tokens. HF pools (cls+dist)/2 after ln_final and applies
+    and extra LayerNorm inside the classifier head, see the unembed mapping note.
```
jlarson4: "applies and extra LayerNorm" should be "an extra".

transformer_lens/model_bridge/supported_architectures/ast.py
```diff
@@ -0,0 +1,141 @@
...
+    def prepare_model(self, hf_model: Any) -> None:
+        """Detect classification head, rebind prefixes, guard unembed, and set n_ctx."""
+        # 1. handle classification model vs bare encoder
+        if hasattr(hf_model, "audio_spectrogram_transformer"):
+            self.component_mapping = self._build_component_mapping(
+                prefix="audio_spectrogram_transformer."
+            )
+            base_model = hf_model.audio_spectrogram_transformer
+
+            # guard unembed: only map if num_labls > 0 and dense head is real
```
jlarson4: `num_labls` should be `num_labels`

jlarson4 (commented):
Thanks @dylanberens! Great work on this, `boot_transformers` is now loading and returning logits as expected! A few new requests below. I appreciate your hard work on this, sorry for all the back and forth

tests/unit/model_bridge/supported_architectures/test_ast_adapter.py
dylanberens: resolved, replaced with these specific assertions and passed local pytest

transformer_lens/utilities/architectures.py
dylanberens: resolved, moved classification architectures set above and added union to audio_architectures to include that set

PR #1487, opened 2026-07-06
Added HRM-Text two-timescale recurrent adapter

### LINKED ISSUE #1471: [Proposal] Add HRM-Text two-timescale recurrent adapter (HrmTextForCausalLM)
### Proposal

Add a TransformerBridge adapter for `HrmTextForCausalLM` (Sapient Intelligence HRM-Text), a hierarchical reasoning model with nested two-timescale depth recurrence.

### Motivation

HRM is a third, distinct flavor of the depth-recurrence axis (alongside Huginn's variable-count loop and Ouro's halting loop): a **slow high-level stack** iterates over a **fast low-level stack**, with cross-stack additive state coupling. This gives interp researchers a clean substrate to separate slow "planning" from fast "computing" state, patch later refinement back into earlier steps, and watch attractor/fixed-point dynamics, none of which are expressible on a flat block list. There is direct mech-interp on the HRM architecture family (fixed-point / spurious-attractor / grokking analysis), though so far on the 27M puzzle HRM rather than HRM-Text specifically. HRM's claims are debated, which is itself a reason interp tooling on it is valuable.

Gap scan (2026-06-25): ~6 models, ~297K downloads.

### Scope note (medium-effort adapter)

**Native** — `model_type: hrm_text` is supported in `transformers >= 5.9.0` (model card: "ships native support for the hrm_text model class"). The repo currently pins `transformers` to 5.8.1, so picking this up needs a routine pin bump first. Verified recurrence fields: `H_cycles=2`, `L_cycles=3` (plus `L_bp_cycles=[0,3]`); per-stack depth is `num_hidden_layers` (=16, "16 layers per stack"). The adapter maps the slow/fast recurrence onto the reused physical weights.

### Pitch

Map the slow (H) and fast (L) stacks and their cross-stack coupling, exposing each as a hookable stream so the two timescales can be ablated and patched independently.

- Claude Code users can scaffold with `/add-model-support sapientinc/HRM-Text-1B`.
- Verify smallest-first: `sapientinc/HRM-Text-1B`, then `LLM-OS-Models/KoHRM-Text-1.4B`.

### Additional context

- HRM origin paper: https://arxiv.org/abs/2506.21734
- Interp on the HRM family (fixed-point / attractor analysis): https://arxiv.org/abs/2601.10679
- Checkpoint: https://huggingface.co/sapientinc/HRM-Text-1B

### Checklist

MdSadiqMd:
# Description

Add a TransformerBridge adapter for `HrmTextForCausalLM` (Sapient Intelligence HRM-Text), a hierarchical two-timescale recurrent model. The adapter exposes the fast (L) and slow (H) stacks as side-by-side hookable block lists (`L_blocks`, `H_blocks`) so the two timescales can be ablated and patched independently.

HRM-Text's nested recurrence (`H_cycles × (L_cycles + 1)` forward passes through the reused physical stacks) is a third flavor of depth-recurrence alongside Huginn's variable-count loop and Ouro's halting loop. The dual-stack architecture gives interp researchers a clean substrate to study fixed-point attractor dynamics, cross-timescale state coupling, and latent-space reasoning — none of which are expressible on a flat block list.

**Key adapter properties:**
- Two `BlockBridge` entries (`L_blocks` → `model.L_module.layers`, `H_blocks` → `model.H_module.layers`) with identical submodule structure (RMSNorm, PositionEmbeddingsAttentionBridge with sigmoid gate, GatedMLPBridge)
- Parameterless RMSNorm → `supports_fold_ln = False`
- Non-standard block names → `supports_center_writing_weights = False`
- `embedding_scale` folded into `embed.weight` in `preprocess_weights` for compat-mode parity
- Passthrough attrs: `H_cycles`, `L_cycles`, `L_bp_cycles`, `num_layers_per_stack`, `embedding_scale`, `prefix_lm`

**Required dependency bump:** `transformers >= 5.9.0` (HF introduced `HrmTextForCausalLM` in v5.9.0, currently at v5.12.1).

Closes #1471 

## Type of change

### Screenshots
N/A

# Checklist:


MdSadiqMd:
hey @jlarson4 PR is up, please review it

jlarson4:
@MdSadiqMd It'll be a few hours before I can do a full review of this, if you want to work on it in the meantime, it appears your CI run has a couple test failures. I'll make sure to ping you when I've done a more thorough review, thank you for putting this together!

Also, I resolved conflicts with the latest version of `dev`, I'd recommend pulling those changes down before making any changes.

tests/integration/model_bridge/test_hrm_text_adapter.py
```diff
@@ -0,0 +1,194 @@
...
+            output = bridge(tokens, use_cache=False)
+        assert output.shape == (1, 4, 128)
+        assert not torch.isnan(output).any()
+
+    def test_forward_matches_hf(self, bridge):
+        tokens = torch.randint(0, 128, (1, 4))
+        with torch.no_grad():
+            hf_logits = bridge.original_model(tokens, use_cache=False).logits
+            bridge_logits = bridge(tokens, use_cache=False)
+        assert hf_logits.shape == bridge_logits.shape
```
jlarson4, line 165: The bridge wraps the HF modules **in place**, so `original_model` is the same hook-wrapped module tree. No adapter bug can ever fail this assertion, because you're comparing the logits to themselves. Compare against an independently loaded HF model (PR 1488's `test_ouro_adapter.py`  in PR #1488 documents exactly this rationale and is a good template).

transformer_lens/model_bridge/generalized_components/position_embeddings_attention.py
```diff
@@ -466,6 +466,12 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
                 q_gate = self.hook_q_gate(q_gate)
             attn_output = attn_output * torch.sigmoid(q_gate)
 
+        # --- Gated attention (HRM-Text: separate gate_proj on hidden_states) ---
+        gate_comp = getattr(self, "gate", None)
+        if gate_comp is not None and gate_comp.original_component is not None and q_gate is None:
```
jlarson4, line 475: The gate branch added to `PositionEmbeddingsAttentionBridge.forward()` (position_embeddings_attention.py:469-473, comment names HRM-Text) is numerically correct, however it runs into a couple issues:
- it lives in shared code used by every RoPE adapter and no tests were added to guard it
- `getattr(self, "gate", None)` falls through `__getattr__` to the wrapped HF module's raw attributes – a future HF attention exposing `.gate` crashes on `.original_component`. Please scope the lookup to declared submodules (e.g. `self._modules.get("gate")`);
- the gate multiplication has no attention-level HookPoint (only the LinearBridge's `gate.hook_in/out` fire), and in split-qkv mode the gate reads the un-forked `hidden_states`, so `hook_q_input`-style patches don't flow into it.

transformer_lens/model_bridge/supported_architectures/hrm_text.py
```diff
@@ -0,0 +1,233 @@
...
+        ):
+            if hasattr(cfg, attr):
+                setattr(self.cfg, attr, getattr(cfg, attr))
+
+        n_kv_heads = (
+            self.cfg.n_key_value_heads
+            if hasattr(self.cfg, "n_key_value_heads") and self.cfg.n_key_value_heads is not None
+            else self.cfg.n_heads
+        )
+        self.weight_processing_conversions = self._build_weight_conversions(n_kv_heads)  # type: ignore[assignment]
```
jlarson4: Rather than type ignore here, lets widen `_build_weight_conversions`' return annotation to the declared attribute type (`dict[str, ParamProcessingConversion | str]`).

jlarson4 (commented):
Hey @MdSadiqMd! Great progress on this so far, you've done a great job handling all the tricky bits that come with this architecture. There is a bit of work yet to be done before we can merge this, however. I have added a few specific comments as well as some general points below.

### 1. Booted bridge crashes on default forward and `generate()` (KV-cache / `cycle_offset`)
Booting through the production path (`TransformerBridge.boot_transformers` on a saved tiny HrmText checkpoint) succeeds, but a plain `bridge(ids)` and `bridge.generate()` both crash: `RuntimeError: The size of tensor a (16) must match the size of tensor b (8) at non-singleton dimension 3`. HF's HrmText attention updates the KV cache at `self.layer_idx + cycle_offset` (modeling_hrm_text.py:215) so each of the `H_cycles×(L_cycles+1)` stack invocations gets its own cache slot; the bridge's reimplemented attention updates at `self._layer_idx` only (generalized_components/attention.py:370-379, ignores the `cycle_offset` kwarg), so repeated invocations of the same physical layer within one forward append K/V into one slot. An 8-token prompt yields 16 cached keys and a mask-shape mismatch.

When `use_cache=False` the bridge is exactly correct, which is why none of the PR's tests catch this: the integration tests bypass `boot()` entirely and hardcode `use_cache=False` on every call. You'll need to fix the cache slotting (respect `cycle_offset`).

### 2. Compat mode breaks parity
The `embedding_scale` fold double-applies `preprocess_weights` folds `embedding_scale` (×8 on the tiny config) into `embed.weight`, and compat mode writes the folded weight back into the **live** HF module,  but HF's `HrmTextModel.forward` still multiplies `inputs_embeds * self.embedding_scale` at runtime (modeling_hrm_text.py:467), so the scale applies twice. gemma1.py:51-56 documents this trap ("must NOT pre-scale weights" when HF applies the scale at runtime). The fix: drop the fold, using the Gemma example as a point of reference.

### 3. The transformers 5.13 adaptation needs a little more work 
Commit d2d2b4a9 ("pipeline fix") legitimately fixed the unit-tier MLA failure and the unit tier is now fully green, But the Full Code Coverage job fails 4 integration tests, all transformers-5.13 fallout.

- **glm_moe_dsa parity: max diff 0.0016867, bit-identical across runs**
  HF 5.13 switched GLM-MoE-DSA main-attention RoPE from split-half NeoX style to interleaved-pair rotation (`apply_rotary_pos_emb_interleave`). The bridge's `_apply_rotary_pos_emb_single` (glm_moe_dsa_attention.py:19-24, applied at :138/:140) still implements the 5.8.1 form.
- **glm4_moe: beartype tuple violation.** 5.13's `Glm4MoeTopkRouter.forward` now returns a 3-tuple (was a bare Tensor. The adapter wraps the gate in a Tensor-typed `LinearBridge`. The fix: use the tuple-aware `GeneralizedComponent(name="gate", optional=True)`. deepseek_v3.py:82 and glm_moe_dsa.py:75 are good examples of how this works.
- **granite_moe_hybrid + nemotron_h `layers_block_type`:** HF 5.13 canonically remaps legacy layer
  type names at config construction (`'mamba' maps to 'linear_attention'`, `'attention' maps to 'full_attention'`, configuration_utils.py:82-92). The adapters are pure passthroughs and nothing in TL branches on the strings (SSM detection is structural), so the right fix is updating the two tests' expected lists to the canonical names plus the stale legacy vocabulary mentioned in nemotron_h.py's docstring.

Latent 5.13 gaps in the same component, unexercised by tests:
With `attention_mask=None` the bridge's DSA index mask lacks HF's causal constraint (future-token leakage in standalone use), and a shared layer with `prev_topk_indices=None` calls a `None` indexer (TypeError) where HF raises a clean ValueError.

### 4. Missing a "verify_models run" for a HRM-Text model
Add a `sapientinc/HRM-Text-1B` entry to `tools/model_registry/data/supported_models.json` and run `verify_models --model sapientinc/HRM-Text-1B`. A verification run will catch a lot of the latent bugs that have not yet been addressed.

### 5. `L_blocks`/`H_blocks` silently miss every `blocks`-hardcoded bridge feature The adapter spec says the block container is "always named `blocks`", but this is an instance that proves this false. The T5 adapter's `encoder_blocks`/`decoder_blocks` are the only special-cased exceptions that are currently in place. The fix: implement adapter-level `setup_hook_compatibility` over L_blocks/H_blocks (the bridge already has an adapter branch for this) so q/k/v/z hooks get head-split shapes, and make `stop_at_layer` raise instead of silently no-op. 

Longer-term, I will need to make an issue for us to generalize the bridge's block-list discovery from the literal `blocks` attribute to all BlockBridge lists in `component_mapping`.

MdSadiqMd:
hey @jlarson4, thanks for the comments. Please review it once again

transformer_lens/model_bridge/architecture_adapter.py
```diff
@@ -683,45 +683,47 @@ def convert_hf_key_to_tl_key(self, hf_key: str) -> str:
         if self.component_mapping is None:
             return hf_key
         for tl_name, component in self.component_mapping.items():
-            if tl_name == "blocks":
+            if tl_name in ("blocks", "L_blocks", "H_blocks", "encoder_blocks", "decoder_blocks"):
```
jlarson4: Drop the encoder/decoder blocks from here and line 692, they are causing failures in the CI. We do not want those blocks being accepted by these statements, they were previously blocked on purpose. I am sorry if my PR comment was unclear on this subject.

MdSadiqMd: Hey @jlarson4 did it, not an issue
On a side note, thanks for maintaining this

MdSadiqMd:
hey @jlarson4, solved the pipeline errors. Please review it once again

jlarson4:
Great work on this @MdSadiqMd! Apologies for the delay, I have been off the grid for the last few days. 

I am going to hold on merging this for the moment, until I get caught up on all of other contributions for the last week or so. Because this includes a bundled `transformers` version increase, it could cause unexpected failures in those PRs. Once I'm caught up in the next couple days, I'll get this merged.

MdSadiqMd:
Hey @jlarson4, sounds good to me
I see merge conflicts. Let me know when you're done with the review, I'll resolve them

jlarson4:
@MdSadiqMd Other reviews are finally wrapped up! Take a look at the conflicts at your leisure, once those are resolved I will merge this

MdSadiqMd:
Hey @jlarson4 solved the merge conflicts and the pipeline runs clean, please merge when ready

PR #1488, opened 2026-07-06
Added Ouro architecture adapter and integration tests for OuroForCausalLM

### LINKED ISSUE #1470: [Proposal] Add Ouro LoopLM looped-depth adapter (OuroForCausalLM)
### Proposal

Add a TransformerBridge adapter for `OuroForCausalLM` (ByteDance Ouro / LoopLM), a Universal-Transformer-style looped model: one weight-tied layer stack iterated T times, plus a learned adaptive-exit halting gate.

### Motivation

Ouro adds two surfaces no supported model has: **weight-sharing across depth** — the same L-layer stack is applied iteratively (default T=4), refining the hidden state in latent space; and a **learned per-step exit/halting gate** (ACT-style adaptive computation). Together they let researchers ask questions TL currently cannot pose: how an answer evolves/revises across loop iterations (cross-step logit lens, answer-flip analysis), what features drive halting (probe/patch the exit gate), and whether latent iterations are causally faithful vs post-hoc. It complements Huginn (`RavenForCausalLM`) by pairing the depth axis with adaptive-compute halting.

Gap scan (2026-06-25): ~15 models, ~288K downloads.

### Scope note (medium-effort adapter)

Remote-code (`auto_map` to `modeling_ouro`), but the loop is a conventional block run in a `for`-loop — closer to a standard adapter than Huginn's variable-count recurrence. Remote-code loading is already supported (see [openelm.py](../../transformer_lens/model_bridge/supported_architectures/openelm.py)).

### Pitch

Map the single shared stack iterated T times (expose each iteration's residual stream as a hookable step) plus the adaptive-exit gate head running parallel to the LM head.

- Claude Code users can scaffold with `/add-model-support ByteDance/Ouro-1.4B`.
- Verify smallest-first: `ByteDance/Ouro-1.4B`, then `ByteDance/Ouro-2.6B-Thinking`.

### Additional context

- Paper (LoopLM): https://arxiv.org/abs/2510.25741
- Adjacent latent-reasoning interp (sibling depth-recurrent model): https://arxiv.org/abs/2507.02199
- Checkpoint: https://huggingface.co/ByteDance/Ouro-1.4B

### Checklist

Canonik:
# Description

Adds a TransformerBridge architecture adapter for `OuroForCausalLM` (ByteDance Ouro / LoopLM), a looped-depth "Universal Transformer" decoder: one weight-tied stack of 24 physical layers is applied `total_ut_steps=4` times per forward pass, with a learned early-exit gate scored alongside the LM head.

What"s included:

- "supported_architectures/ouro.py`, including a `prepare_loading` patch for two transformers-v5 incompatibilities in the Ouro remote code (written against 4.55): the removed `ROPE_INIT_FUNCTIONS["default"]` entry, and the missing `compute_default_rope_parameters` static method that the v5 `_init_weights` machinery expects on rotary classes. Both patches are scoped to the imported `modeling_ouro` module; the shared transformers dict is untouched.
- Four-site registration, an `"ouro"` model_type fallback in `sources/transformers.py`, and a `"ByteDance/Ouro-"` entry in `_BRIDGE_REMOTE_CODE_PREFIXES` in `verify_models.py` so registry verification auto-enables `trust_remote_code`.
- A `ByteDance/Ouro-1.4B` registry row (status 0, null scores).
- Unit adapter test (23 structural tests, runs in CI) and a gated integration test (`slow` + `skipif(CI)`, QUARANTINES.md row) asserting logit parity against an independently loaded HF eager model at 1e-3, delegation identities, and the per-step hook firing count.

Verified locally on real Ouro-1.4B weights: 11/11 integration tests pass. Loading requires `trust_remote_code=True`; no new package dependencies.

Note for reviewers: plain `AutoModelForCausalLM.from_pretrained` on Ouro is currently broken under transformers v5 regardless of TransformerLens (the upstream remote code reads `config.pad_token_id`, which v5 no longer materializes). The bridge load path already shims this generically; the reference fixture in the integration test mirrors the same shim.

Fixes #1470

## Type of change

# Checklist:


tests/QUARANTINES.md
```diff
@@ -46,6 +46,7 @@ Rule ([AGENTS.md §10](../AGENTS.md#10-hard-rules)): **never add `xfail` / `skip
 | [`unit/model_bridge/supported_architectures/test_gemma2_adapter.py`:49](unit/model_bridge/supported_architectures/test_gemma2_adapter.py) | "Network/disk fetch of tiny Gemma2" |
 | [`integration/model_bridge/test_bridge_integration.py`:801](integration/model_bridge/test_bridge_integration.py) | "Skip Gemma2 in CI to avoid timeout" |
 | [`acceptance/model_bridge/compatibility/test_hook_completeness.py`:156](acceptance/model_bridge/compatibility/test_hook_completeness.py) | "Gemma2 too large for CI" |
+| [`integration/model_bridge/test_ouro_adapter.py`](integration/model_bridge/test_ouro_adapter.py) module-level (+ `slow` marker) | "ByteDance/Ouro-1.4B: 2.8GB download + ~11GB RAM, too large for CI" |
```
jlarson4: Incredibly tiny nitpick, apologies, can you add the line number for the test to the link text similar to the other lines in this table?

jlarson4:
This looks pretty good @Canonik! Just a couple small notes:

1. Missing `supports_fold_ln = False` – compat mode corrupts the UT loop. Ouro applies `model.norm` (mapped as `ln_final`, a **weighted** RMSNorm) after EVERY UT pass, with the normed output feeding the next pass and the early-exit gate (modeling_ouro.py:592-608), it is not a final-only norm. The adapter leaves `supports_fold_ln` at the default True, so `enable_compatibility_mode()` / verify_models Phase 3 folds `ln_final.w` into W_U, resets the norm weight to ones **in the live HF module the loop reuses**, and corrupts passes 1..N-1.]. The fix is one line plus a comment: `self.supports_fold_ln = False` in `__init__`Related docstring note: `ln_final` hooks fire `total_ut_steps`× per forward and `run_with_cache` keeps only the last pass, worth a comment alongside the existing per-block note.

2. Run `verify_models` on a sample model, if possible – no models have been verified thus far. The error documented above would've been caught by a verification pass. If you can't run the verification suite on your machine, no worries, just let me know and I can run the verification run & push the results.

3. Surface `total_ut_steps` (and possibly `early_exit_threshold`) on `bridge.cfg` per the config-attr propagation rule. Then the integration test won't have to reach into `hf_eager.config` for it.

Canonik:
Thanks I’ll address these in a small follow-up commit.

For `verify_models`, I’ll try running it locally on the sample Ouro model and report back if I hit an OOM or any local machine limitation. I think I should be able to fit it cleanly, but I’ll confirm with the actual run.

Canonik:
- 'verify_models' ran clean (Ouro-1.4B, P1-P4 = 100/100/100/95.9).
- I applied the 'supports_fold_ln = False' fix with a unit test covering it.
- Reworked the integration tests as requested: loop config attrs now read off bridge.cfg, plus a propagation test.
- Fixed the QUARANTINES.md line anchor.

Should be ready to merge :)

jlarson4:
Looks awesome! Thank you @Canonik, merging now

Canonik:
Glad to hear 😊

PR #1491, opened 2026-07-07
Centralize soft-cap logic behind a shared helper and a named disabled…

### LINKED ISSUE #1489: [Proposal] Centralize soft-cap logic behind a shared helper and a named "disabled" sentinel
### Proposal

Consolidate the tanh soft-capping used for attention scores and output logits (Gemma-2 style) into a single shared helper such as `apply_softcap(x, cap)` in `transformer_lens/utilities/activation_functions.py`. Also replace the bare `-1.0` "disabled" default with a named constant such as `SOFTCAP_DISABLED`.

Today both the `cap * tanh(x / cap)` formula and its `cap > 0` enable-guard are re-implemented independently at every call site, and the disabled sentinel `-1.0` is a magic number whose meaning lives only in a docstring.

### Motivation

Soft-capping is currently implemented in several independent places. Each site separately (1) checks whether the cap is enabled with `cap > 0` and (2) applies `cap * tanh(x / cap)`:

- `transformer_lens/HookedTransformer.py` – final-logits soft-cap (the `output_logits_soft_cap > 0.0` block, ~L672).
- `transformer_lens/components/abstract_attention.py` – attention-score soft-cap (the `attn_scores_soft_cap > 0` block, ~L525).
- `transformer_lens/model_bridge/sources/native/model.py` – attention soft-cap (~L272) **and** output-logits soft-cap (~L471).

The disabled sentinel is likewise a bare literal in two config files:

- `transformer_lens/config/hooked_transformer_config.py` – `attn_scores_soft_cap: float = -1.0`, `output_logits_soft_cap: float = -1.0` (~L283–284).
- `transformer_lens/config/transformer_bridge_config.py` – same two fields default to `-1.0` (~L80–81).

Two problems follow:

1. **Magic sentinel.** `-1.0` means "disabled," but that contract is documented only in a docstring and enforced only by convention. A reader seeing `output_logits_soft_cap = -1.0` has no in-code signal that this is a sentinel rather than a real cap value.

2. **The duplicated guard is easy to get subtly wrong.** Because the correct enable-check is `cap > 0`  the natural-looking `if cap:` is a latent bug: `-1.0` is truthy, so a truthiness check would apply `-1.0 * tanh(x / -1.0) = tanh(x)`, which saturates every score above ~9 to `1.0` in fp32 and silently corrupts the argmax. Every hand-rolled copy of the guard is one careless refactor away from this. A shared helper makes the mistake structurally impossible.

### Pitch

Add a small, well-tested helper (leaf-level, already-imported module):

```python
# transformer_lens/utilities/activation_functions.py
from typing import Optional

import torch

SOFTCAP_DISABLED: float = -1.0


def softcap_enabled(cap: Optional[float]) -> bool:
    """A soft-cap is active only for a strictly positive value; -1.0 / 0 / None mean disabled."""
    return cap is not None and cap > 0


def apply_softcap(x: torch.Tensor, cap: Optional[float]) -> torch.Tensor:
    """Gemma-style ``cap * tanh(x / cap)`` when enabled, identity otherwise."""
    return cap * torch.tanh(x / cap) if softcap_enabled(cap) else x
```

Then collapse the call sites to one line each, e.g.:

```python
# before (HookedTransformer.py)
if self.cfg.output_logits_soft_cap > 0.0:
    logits = self.cfg.output_logits_soft_cap * F.tanh(logits / self.cfg.output_logits_soft_cap)

# after
logits = apply_softcap(logits, self.cfg.output_logits_soft_cap)
```

Sites to update (all on `dev`; `git grep "soft_cap" transformer_lens` finds them):

- `HookedTransformer.py` — output-logits soft-cap (~L672).
- `components/abstract_attention.py` — attention-score soft-cap (~L525).
- `model_bridge/sources/native/model.py` — attention (~L272) and output logits (~L471).
- `config/hooked_transformer_config.py` and `config/transformer_bridge_config.py` — point the `-1.0` defaults at `SOFTCAP_DISABLED`.
- The `center_unembed` compatibility checks in `HookedTransformer.py` (~L1423) and `model_bridge/bridge.py` (~L923) can read more clearly as `softcap_enabled(cfg.output_logits_soft_cap)`.

This is a **pure refactor**, behaviour is unchanged. Gemma-2 already exercises the enabled path, so a before/after logit-parity check on a Gemma-2 model (plus any non-capped model to confirm the disabled path stays identity) is sufficient verification.

### Additional context

- **Leave the HF-native attention path alone.** `model_bridge/generalized_components/position_embeddings_attention.py` (~L426) reads a raw HF attribute (`getattr(hf_attn, "attn_logit_softcapping", None)`, guarded by `if softcap is not None`). That is a *different* convention (`None` means disabled, sourced from the HF config object, not the TL config field).
- **Config-mapping sites are out of scope.** `sources/_bridge_builder.py` and `sources/transformers.py` translate HF's `final_logit_softcapping` / `attn_logit_softcapping` into the TL config fields; they assign the field, they don't apply the cap.
- **Import direction.** If the config defaults reference `SOFTCAP_DISABLED`, confirm `config` to `utilities.activation_functions` introduces no import cycle (the module is leaf-level, so it should be fine). If it does, keep the constant beside the config fields and import it into the helper instead.

### Checklist

delaidam:
Fixes #1489

Consolidates the tanh soft-capping (Gemma-2 style) used for attention scores
and output logits into `apply_softcap()` / `softcap_enabled()` in
`transformer_lens/utilities/activation_functions.py`, replacing five
independent `cap * tanh(x / cap)` implementations:

- HookedTransformer.py - output-logits soft-cap
- components/abstract_attention.py - attention-score soft-cap
- model_bridge/sources/native/model.py - both attention and output-logits soft-cap
- HookedTransformer.py + model_bridge/bridge.py - center_unembed compatibility checks

Also replaces the bare `-1.0` disabled sentinel with a named
`SOFTCAP_DISABLED` constant in both `HookedTransformerConfig` and
`TransformerBridgeConfig`.

Pure refactor, no behavior change. Left the HF-native attention path
(`generalized_components/position_embeddings_attention.py`) and the
config-mapping sites (`sources/_bridge_builder.py`, `sources/transformers.py`)
untouched, per the issue's scope notes.

## Verification
- `uv run mypy .` - clean, 307 files
- `make unit-test` - 3204 passed, 32 skipped, 9 xfailed, 0 failed
- `pytest` (docstrings) on all changed files - clean
- New unit tests in `tests/unit/utilities/test_activation_functions.py`,
  including a regression guard for the exact truthiness trap the issue
  describes (`-1.0` must stay identity, not saturate to 1.0)

### DIFF
--- a/tests/unit/utilities/test_activation_functions.py
+++ b/tests/unit/utilities/test_activation_functions.py
@@ -1,0 +1,70 @@
+"""Unit tests for the shared soft-cap helpers.
+
+Covers ``softcap_enabled`` and ``apply_softcap`` in
+``transformer_lens.utilities.activation_functions`` (see issue #1489).
+"""
+
+import torch
+
+from transformer_lens.utilities.activation_functions import (
+    SOFTCAP_DISABLED,
+    apply_softcap,
+    softcap_enabled,
+)
+
+
+class TestSoftcapEnabled:
+    """Test cases for softcap_enabled."""
+
+    def test_none_is_disabled(self):
+        assert softcap_enabled(None) is False
+
+    def test_zero_is_disabled(self):
+        assert softcap_enabled(0.0) is False
+
+    def test_disabled_sentinel_is_disabled(self):
+        assert softcap_enabled(SOFTCAP_DISABLED) is False
+
+    def test_negative_is_disabled(self):
+        assert softcap_enabled(-5.0) is False
+
+    def test_positive_is_enabled(self):
+        assert softcap_enabled(30.0) is True
+
+
+class TestApplySoftcap:
+    """Test cases for apply_softcap."""
+
+    def test_disabled_is_identity(self):
+        x = torch.tensor([-100.0, -1.0, 0.0, 1.0, 100.0])
+        result = apply_softcap(x, SOFTCAP_DISABLED)
+        assert torch.equal(result, x)
+
+    def test_none_is_identity(self):
+        x = torch.tensor([-100.0, 0.0, 100.0])
+        result = apply_softcap(x, None)
+        assert torch.equal(result, x)
+
+    def test_enabled_matches_capped_tanh_formula(self):
+        x = torch.tensor([-50.0, -10.0, 0.0, 10.0, 50.0])
+        cap = 30.0
+        result = apply_softcap(x, cap)
+        expected = cap * torch.tanh(x / cap)
+        assert torch.allclose(result, expected)
+
+    def test_enabled_output_stays_within_cap(self):
+        x = torch.tensor([-1000.0, 1000.0])
+        cap = 30.0
+        result = apply_softcap(x, cap)
+        assert torch.all(result.abs() <= cap)
+
+    def test_disabled_sentinel_does_not_saturate_large_values(self):
+        """Regression guard for the truthiness trap the issue describes:
+        a naive ``if cap:`` check treats -1.0 as enabled, which computes
+        ``-1.0 * tanh(x / -1.0) == tanh(x)`` and saturates every large score
+        to 1.0. The disabled path must instead be a pure identity.
+        """
+        x = torch.tensor([50.0, 100.0])
+        result = apply_softcap(x, SOFTCAP_DISABLED)
+        assert torch.equal(result, x)
+        assert not torch.allclose(result, torch.ones_like(x))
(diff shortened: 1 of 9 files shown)


jlarson4:
Great work on this @delaidam! I appreciate the thoroughness & additional testing you added. Merging now

PR #1497, opened 2026-07-09
Mistral: use fork-capable attention bridge (enable hook_attn_in / set_use_attn_in)

almogtavor:
Mistral loads through the plain `AttentionBridge`, which delegates q/k/v to HF and exposes no per-receiver fork point. So on a bridged Mistral, `set_use_attn_in(True)` raises and `blocks.{i}.hook_attn_in` (and the q/k/v input forks) never exist, making per-head attention-input interventions (activation/path patching that edits a receiver input) impossible.

Mistral is structurally identical to Qwen2 here (separate `q/k/v/o_proj`, RoPE, GQA, RMSNorm, gated MLP), which uses `PositionEmbeddingsAttentionBridge`.

## Fix
- Point the Mistral adapter at `PositionEmbeddingsAttentionBridge` (same bridge as Qwen2/Qwen3).
- Pass `config` to Mistral's `BlockBridge` so `hook_mlp_in` is wired too.

## Test
`tests/unit/model_bridge/test_mistral_attn_in_fork.py` boots `hf-internal-testing/tiny-random-MistralForCausalLM` and checks the bridge is `PositionEmbeddingsAttentionBridge`, `hook_attn_in` fires at `[batch, pos, n_heads, d_model]`, and zeroing a single head's forked input changes the logits.


jlarson4:
@almogtavor This looks good to me, from my review. Please just update or remove the stale test that is causing CI to fail. `test_mistral_adapter.py::test_attn_is_not_position_embeddings_subclass` currently asserts "Mistral uses plain AttentionBridge, not PositionEmbeddingsAttentionBridge." This is no longer true, and is tripping CI.

jlarson4:
@almogtavor I did find one additional item while running some `verify_models` runs against this change. It breaks quantized models.

Run `pytest tests/integration/model_bridge/test_bridge_integration.py -k quantized` and fix the regression (most likely solution will be porting the fp-dtype guard into `PositionEmbeddingsAttentionBridge` from `AttentionBridge.forward`)

almogtavor:
@jlarson4 Awesome, I'll fix it thanks !

jlarson4:
Hey @almogtavor! Just wanted to touch base and see if you'll have an opportunity to wrap this up within the next couple days. I am getting ready to do a release, and would love to include this if it will be ready. 

Thank you!

almogtavor:
@jlarson4 Of course, I'll wrap it up on the weekend :)

almogtavor:
@jlarson4 Both fixed :)

The stale assert is inverted now. For quantized, ported the non-fp param skip into `PositionEmbeddingsAttentionBridge`. Had to touch that test too: it captured dtype by stubbing the HF attention forward, and the reimplemented bridge doesn't call it, so it reads the dtype at `q_proj` now. Verified it still fails without the guard.

jlarson4:
Looks great on my end! Thanks for putting this together! Merging now @almogtavor

PR #1502, opened 2026-07-10
Fix 4-bit split QKV projection

### LINKED ISSUE #737: [Bug Report] Q cannot be reshaped correctly when model is loaded in 4bit
**Describe the bug**
Query_input's shape is [batch, pos, n_heads, d_model], and the purpose of the code where the error occurred is to reshape query_input to [batch, pos, n_heads, d_head].
I found that the shape of output of bnb.matmul_4bit is still [batch, pos, n_heads, d_model] so it cannot be reshaped to [batch, pos, n_heads, d_head].

The reason for this error may be the following code in abstract_attention.py：
```python
if self.cfg.load_in_4bit:
            nq = int((self.cfg.d_model * self.cfg.d_head * self.cfg.n_heads) / 2)
            self.W_Q = Params4bit(torch.empty(nq, 1, dtype=torch.uint8), requires_grad=False)
            self.W_O = Params4bit(torch.empty(nq, 1, dtype=torch.uint8), requires_grad=False)
        else:
            self.W_Q = nn.Parameter(
                torch.empty(
                    self.cfg.n_heads,
                    self.cfg.d_model,
                    self.cfg.d_head,
                    dtype=self.cfg.dtype,
                )
            )
            self.W_O = nn.Parameter(
                torch.empty(
                    self.cfg.n_heads,
                    self.cfg.d_head,
                    self.cfg.d_model,
                    dtype=self.cfg.dtype,
                )
            )
```
When model is loaded in 4bit, the shape of matrix W_Q is [(self.cfg.d_model * self.cfg.d_head * self.cfg.n_heads) / 2, 1] which leads to the unexpected shape of the output from the bnb.matmul_4bit function. 
When model is not loaded in 4bit, the shape of matrix W_Q is [n_heads, d_model, d_head] which does nor trigger the bug mentioned above.

**Code example**
Eorro message:
```python
File ~/python_library/TransformerLens/transformer_lens/components/abstract_attention.py:364, in AbstractAttention.calculate_qkv_matrices(self, query_input, key_input, value_input)
    339        if self.cfg.load_in_4bit:
    340            q = self.hook_q(
    341                # call bitsandbytes method to dequantize and multiply
    342                bnb.matmul_4bit(
    343                    query_input,
    344                    self.W_Q.t(),
    345                    bias=None,
    346                    quant_state=self.W_Q.quant_state,
-->347                ).reshape(
    348                    query_input.shape[0],
    349                    query_input.shape[1],
    350                    self.cfg.n_heads,
    351                    self.cfg.d_head,
    352                )
    353                + self.b_Q
RuntimeError: shape '[20, 22, 32, 128]' is invalid for input of size 57671680"
```

code:
```python
with torch.inference_mode():
            with model.hooks(fwd_hooks=fwd_hooks_corrupted):
                _ = model(corrupted)
```

**System Info**
Describe the characteristic of your environment:
 * git clone
 * Linux
 * python 3.11.4

### Checklist

- [√] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/TransformerLens/issues) in the repo (**required**)

abhinav-bellapu:
Fixes #737.

## Summary
- Keep the fused bitsandbytes matmul for ordinary `[batch, pos, d_model]` Q/K/V inputs.
- For split `[batch, pos, head, d_model]` inputs, dequantize the 4-bit weight once and reuse `complex_attn_linear` for the head-wise projection.
- Centralize `Params4bit` validation in the shared projection helper so Q, K, and V report the same clear error.
- Add CPU-safe fake-bitsandbytes regression coverage for both ordinary and split inputs.
- Refresh the affected quarantine line references.

## Why
The old 4-bit path reshaped split Q/K/V projections incorrectly, producing the shape error in #737. Projecting every input head through every output head before selecting the diagonal would fix the shape but introduce an `n_heads`-fold compute and memory cost. Dequantizing once and sharing the established head-wise projection path preserves the intended semantics without that transient.

## Validation
- `env UV_PYTHON=3.12 make format`
- `env UV_PYTHON=3.12 make check-format`
- `uv run --python 3.12 pytest tests/unit/components/test_attention.py -q` (15 passed, 4 skipped)
- `uv run --python 3.12 mypy .` (300 source files clean)
- `env UV_PYTHON=3.12 make unit-test` (3,027 passed, 33 skipped, 9 xfailed)

### DIFF
--- a/transformer_lens/components/abstract_attention.py
+++ b/transformer_lens/components/abstract_attention.py
@@ -446,59 +446,20 @@ class AbstractAttention(ABC, nn.Module):
         )
         if self.cfg.load_in_4bit:
             W_Q_4bit = cast(Params4bit, self.W_Q)
-            q = self.hook_q(
-                # call bitsandbytes method to dequantize and multiply
-                bnb.matmul_4bit(
-                    query_input,
-                    W_Q_4bit.t(),
-                    bias=None,
-                    quant_state=W_Q_4bit.quant_state,
-                ).reshape(
-                    query_input.shape[0],
-                    query_input.shape[1],
-                    self.cfg.n_heads,
-                    self.cfg.d_head,
-                )
-                + self.b_Q
-            )
+            q = self.hook_q(self._project_4bit_qkv(query_input, W_Q_4bit, self.b_Q))
         else:
             q = self.hook_q(attn_fn(query_input, self.W_Q, self.b_Q))
         if self.cfg.load_in_4bit:
             if not isinstance(self.W_K, Params4bit):
                 raise ValueError("W_K must be a Params4bit object if load_in_4bit is True")
-            k = self.hook_k(
-                # call bitsandbytes method to dequantize and multiply
-                bnb.matmul_4bit(
-                    key_input, self.W_K.t(), bias=None, quant_state=self.W_K.quant_state
-                ).reshape(
-                    key_input.shape[0],
-                    key_input.shape[1],
-                    self.cfg.n_heads,
-                    self.cfg.d_head,
-                )
-                + self.b_K
-            )
+            k = self.hook_k(self._project_4bit_qkv(key_input, self.W_K, self.b_K))
         else:
             k = self.hook_k(attn_fn(key_input, self.W_K, self.b_K))
 
         if self.cfg.load_in_4bit:
             if not isinstance(self.W_V, Params4bit):
                 raise ValueError("W_V must be a Params4bit object if load_in_4bit is True")
-            v = self.hook_v(
-                # call bitsandbytes method to dequantize and multiply
-                bnb.matmul_4bit(
-                    value_input,
-                    self.W_V.t(),
-                    bias=None,
-                    quant_state=self.W_V.quant_state,
-                ).reshape(
-                    value_input.shape[0],
-                    value_input.shape[1],
-                    self.cfg.n_heads,
-                    self.cfg.d_head,
-                )
-                + self.b_V
-            )
+            v = self.hook_v(self._project_4bit_qkv(value_input, self.W_V, self.b_V))
         else:
             v = self.hook_v(attn_fn(value_input, self.W_V, self.b_V))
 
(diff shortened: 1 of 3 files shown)


abhinav-bellapu:
@jlarson4 ready for review

transformer_lens/components/abstract_attention.py
```diff
@@ -510,6 +471,42 @@ def calculate_qkv_matrices(
 
         return q, k, v
 
+    def _project_4bit_qkv(
```
jlarson4, line 469: The 4D path use `n_heads` times the necessary compute and memory – please dequantize once and reuse `complex_attn_linear` instead.

`_project_4bit_qkv` projects every input head through all heads' weights and then keeps only the diagonal. On a 4-bit Llama-7B (32 heads) at `pos=2048`, that's a ~512 MiB fp16 (~1 GiB fp32) transient per projection, three times per layer, of which only the ~16 MiB diagonal survives. This is a real OOM risk on the memory-constrained setups people use 4bit models to avoid.

Since `matmul_4bit` dequantizes internally anyway, dequantizing the weight lets the split path share the already-tested einsum instead of maintaining a parallel implementation. Divergence between these two paths is how #737 happened in the first place.

transformer_lens/components/abstract_attention.py
```diff
@@ -446,59 +446,20 @@ def calculate_qkv_matrices(
         )
         if self.cfg.load_in_4bit:
             W_Q_4bit = cast(Params4bit, self.W_Q)
```
jlarson4: `W_Q` currently gets a static `cast` while `W_K`/`W_V` get the runtime `isinstance` + `ValueError`. A non-quantized `W_Q` under `load_in_4bit=True` dies with an opaque `AttributeError`. 

Let's move the `isinstance` guard from the `W_K`/`W_V` cases into the helper and drop the Q/K/V asymmetry. Checking `isinstance(weight, Params4bit)` inside `_project_4bit_qkv` gives all three weights the clear error and collapses the three near-identical call-site blocks.

transformer_lens/components/abstract_attention.py
jlarson4: Additionally, if you could possibly add a docstring comment to this function?

jlarson4 (commented):
Thanks for this fix! I verified it locally. The unit-tier, fake-bnb approach for CPU-safe coverage is a nice touch.

There are a couple organization changes outlined below that we should address before merging, but altogether the solution is sound.

Outside of the current edits in this PR, it would be appreciated if you could update `tests/QUARANTINES.md`.

The new test moves the two quarantined entries in `tests/unit/components/test_attention.py` (cited there as `:48` and `:83`) down by ~72 lines. Please refresh those line references.

jlarson4:
@abhinav-bellapu Update looks good, once the CI passes I will merge

abhinav-bellapu:
@jlarson4 CI looks good!

PR #1503, opened 2026-07-11
Add Apertus adapter tests and fix nested key translation

abhinav-bellapu:
Part of #1302.

## Summary

- Add download-free focused coverage for `ApertusArchitectureAdapter` config routing, registration, component paths, GQA weight conversions, key translation, rotary/eager test setup, and the XiELU meta-device compatibility patch.
- Make Hugging Face-to-TransformerLens key translation prefer nested component paths over their parent prefix.
- Verify Q/K/V/O reshapes numerically round-trip and use the correct query vs. KV head counts.

## Why

Apertus was the remaining unclaimed architecture in #1302 without a dedicated supported-architecture test suite.

The new key-translation cases also exposed a shared bug: `self_attn` or `mlp` matched before nested paths such as `self_attn.q_proj` and `mlp.up_proj`. This emitted keys such as `blocks.0.attn.q_proj.weight` instead of the canonical `blocks.0.attn.q.weight`, so the result was not the inverse of the existing TL-to-HF mapping and did not line up with adapter weight-conversion keys. Checking the more specific nested path first fixes that while preserving parent-component fallback behavior.

## Validation

- `uv run --locked --python 3.11 pytest tests/unit/model_bridge/supported_architectures/test_apertus_adapter.py -q` (`31 passed`)
- `pytest tests/unit/model_bridge -q` (`2404 passed, 6 skipped, 9 xfailed`)
- `pytest tests/unit -m "not slow" --reruns 2 --reruns-delay 5 -q` (`3260 passed, 33 skipped, 42 deselected, 9 xfailed`)
- `mypy .` (`Success: no issues found in 308 source files`)
- `pycln --check --all . --exclude __init__.py`
- `isort --check-only .`
- `black --check .`

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_apertus_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_apertus_adapter.py
@@ -1,0 +1,362 @@
+"""Download-free unit tests for :class:`ApertusArchitectureAdapter`.
+
+The suite covers HF config routing, adapter selection, component paths, GQA
+weight reshaping, key translation, rotary test wiring, and the temporary XiELU
+meta-device compatibility patch.
+"""
+
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+import torch
+import torch.nn as nn
+from transformers import ApertusConfig
+from transformers import activations as transformers_activations
+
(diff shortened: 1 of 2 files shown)


jlarson4:
Looks good! Thanks @abhinav-bellapu

PR #1507, opened 2026-07-11
Add Jacobian lens (J-lens): published-artifact loading, readout, native fitting, and interventions

### LINKED ISSUE #1505: [Proposal] Jacobian Lens (J-lens): load published lenses, fit natively, read out and intervene
### Proposal

Add first-class support for the **Jacobian lens (J-lens)** — the interpretabilitytechnique from Anthropic's July 2026 paper [*Verbalizable Representations Form aGlobal Workspace in Language Models*](https://transformer-circuits.pub/2026/workspace/index.html)— as a tool in `transformer_lens/tools/analysis/`, working with both`TransformerBridge` and `HookedTransformer`.

Concretely, one `JacobianLens` component that can:

1. **Load published lens artifacts** from the HF Hub ([`neuronpedia/jacobian-lens`](https://huggingface.co/neuronpedia/jacobian-lens) ships pre-fitted lenses for 38 open models, e.g. Gemma 2/3/4, Qwen 2.5/3/3.5/3.6, Llama 3.x, GPT-2, Pythia, OLMo 3, gpt-oss) in the official artifact format.
2. **Read out** per-layer, per-position vocabulary distributions: `lens(h_l) = softmax(W_U · norm(J_l · h_l))`, using the model's own final norm and unembedding (plus the config's logit softcap where the architecture has one, e.g. Gemma-2), with the logit lens available as the `J_l = I` baseline through the same code path.
3. **Fit lenses natively** on any hooked model with TransformerLens backward hooks, reproducing the official estimator ([`anthropics/jacobian-lens`](https://github.com/anthropics/jacobian-lens)); fit parity would be verified numerically against the official implementation on small configs as part of the PR. Plus `merge()` for parallelized fitting.
4. **Intervene** in J-lens coordinates: steering along a J-lens vector, ablation (projecting a vector out), and the paper's pseudoinverse coordinate swap (`V = [v_s, v_t]`, `c = V⁺h`, `h ← h + V(σ(c) − c)`).

### Motivation

The J-lens is, in the paper's framing, the principled correction of the logit lens: `J_l` is the corpus-averaged, causal linear map relating layer-`l` residual directions to their final-layer counterparts (the logit lens is the special case `J_l = I`; the tuned lens is trained on output-matching, which is correlational and tends to "skip ahead" past intermediate computation). It surfaces interpretable content at depths where the logit lens reads out noise, and its vectors support causal interventions (steering, ablation, concept swaps). Beyond the paper's workspace results, it is already being used for alignment auditing readouts.

Right now there is no maintained, general implementation anyone can build on:

- Anthropic's companion repo (`anthropics/jacobian-lens`, Apache-2.0) is explicitly frozen: "Not maintained and not accepting contributions." It also contains no intervention code (readout + fitting only), and at least one community offer to contribute there is stranded (anthropics/jacobian-lens#2).
- Neuronpedia serves J-lens readouts at [neuronpedia.org/jlens](https://www.neuronpedia.org/jlens) — and their inference server internally runs them **on top of `transformer_lens.HookedTransformer`** — but that lives inside their platform, not as a reusable library API.
- In the nine days since the code release, roughly forty standalone replications/ports (MLX, visualizers, single-model demos) have appeared, none of them a maintained, tested library integration.

TransformerLens already has the right substrate (residual-stream hooks, backward hooks, `ActivationCache`, HF Hub utilities) and an established home for exactly this kind of feature (`tools/analysis/`, cf. #1369). J-lens would be TL's first fitted-lens tool and the first to consume published lens artifacts from the Hub.

### Pitch

Proposed API (names/placement happily adjusted to maintainer preference):

```python
from transformer_lens.tools.analysis import JacobianLens

model = TransformerBridge.boot_transformers("google/gemma-2-2b", dtype=torch.bfloat16)

# 1) load a published lens (fp16 artifact, fp32 in memory; validated against model)
lens = JacobianLens.from_pretrained(
    "neuronpedia/jacobian-lens",
    filename="gemma-2-2b/jlens/Salesforce-wikitext/gemma-2-2b_jacobian_lens.pt",
    model=model,  # validates d_model / layer count; errors on processed weights
)

# 2) per-layer readout on a prompt (JacobianLensResult dataclass)
result = lens.readout(model, "The capital of France is", layers=[6, 13, 19], positions=[-1])
result.top_tokens(k=5)                  # J-lens top-k per (layer, position)
result_ll = lens.readout(model, ..., use_jacobian=False)  # logit-lens baseline

# 3) fit natively (backward hooks; parity with the official estimator)
lens = JacobianLens.fit(model, prompts, dim_batch=8, max_seq_len=128)
merged = JacobianLens.merge([lens_a, lens_b])   # parallelize across prompt slices

# 4) interventions (returned as hook fns usable with model.hooks / run_with_hooks)
band = range(10, 24)  # an intermediate-layer band
hooks = lens.steer(" Paris", layers=band, alpha=4.0)
hooks = lens.ablate(" Paris", layers=band)
hooks = lens.swap(" France", " China", layers=band)   # pseudoinverse coordinate swap
```

**Coordinate-basis compatibility plan** (the main correctness risk). Published lenses are fitted on **raw HF activations**: `J[l]` maps the output of block `l` (`blocks.{l}.hook_resid_post`) to the final block output, pre-`ln_final`, and the readout applies the model's own final norm + unembed. TL's classic weight processing (`fold_ln`, `center_writing_weights`, `center_unembed`) changes the residual basis, so:

- `TransformerBridge.boot_transformers(...)` (raw HF weights by default) and `HookedTransformer.from_pretrained_no_processing(...)` are the supported paths.
- `JacobianLens` checks the model's processing state on attach and **raises** (not warns) when lens artifacts meet a fold_ln/centered model, with an actionable message. Same treatment for `d_model`/layer-count mismatches.
- Artifacts saved by TL carry metadata (model name, TL version, processing flags, hook convention, corpus, n_prompts, dtype) while staying loadable by the official package (official keys preserved; extras additive). Loading official artifacts — which carry only `{J, n_prompts, source_layers, d_model}` — stays fully supported.
- Readout reuses the model's own `ln_final` + `unembed` modules, so RMSNorm vs LayerNorm and Gemma-2's `final_logit_softcapping` are handled by construction (cf. the logit-lens normalization fixes in #1077/#1180 and the processing discussion in #1272).

**Validation plan** (all numbers to be posted in the PR):

1. *Artifact parity*: loading the published gemma-2-2b lens reproduces reference readouts (official `jlens` package on HF transformers as the numerical oracle, with top-k overlap and rank correlation thresholds on fixed prompts, per layer; Neuronpedia's public lens API as a secondary qualitative sanity check — their serving stack resolves bf16 near-ties differently, so it isn't a bitwise reference).
2. *Late-layer agreement*: J-lens ≈ logit lens top-k in the final layers (`J_l → I`).
3. *Kurtosis profile*: excess kurtosis of readout logits ≈ 0 in the first ~third of depth, rising through the workspace band (paper §4 structure result).
4. *Fit convergence*: native TL fit at small n on wikitext trends toward the published readouts (official repo: "~100 prompts is usable"; published artifacts ship per-prompt convergence CSVs to compare against).
5. *Causal smoke test*: the France→China coordinate swap on "The capital of France is", following the paper's flexible-generalization protocol (swap clamped at every position across an intermediate-layer band), with observed success rates reported honestly — the paper itself reports 42/48 for country swaps at α=1 on Sonnet 4.5, and no open-weights numbers exist yet.

**Scope of the first PR** (deliberately small; a working vertical slice):

- `transformer_lens/tools/analysis/jacobian_lens.py` (load/save/fit/merge/readout/ interventions, result dataclass), exported like the existing analysis tools.
- Unit tests (synthetic tiny model, no HF downloads: estimator orientation on a linear toy model where `J` is known in closed form, save/load round-trip, merge weighting, basis-mismatch errors) + integration tests on CI-cached models (`gpt2`: fit + readout invariants; the published `gpt2-small` lens exists for parity) + `@pytest.mark.slow` gemma-2-2b parity tests.
- One demo notebook (`demos/Jacobian_Lens_Demo.ipynb`): J-lens vs logit lens on a two-hop prompt with an unspoken intermediate, and the France→China swap; gemma-2-2b (+ a Bridge-only modern model, see below).

**Follow-up work (PR2+, separate issue once this lands):** artifact registry/aliases so `JacobianLens.from_pretrained("gemma-2-2b")` just works; converters for community formats (fit-checkpoint files, other lens zoos); gradient-pursuit sparse decomposition (J-space coordinates, occupancy); a community fitting guide (the fit parallelizes cleanly with `merge`, so lens coverage of new models is a natural community contribution surface).

Model coverage in the first PR: **gemma-2-2b** as the correctness reference (supported by both HookedTransformer and TransformerBridge; published artifact; CI already caches the `-it` variant, which also has a published lens, for any cached-tier tests) and **Qwen/Qwen3.5-4B** as the Bridge-only modern-architecture demo (the model used in the official walkthrough, with a published n=1000 artifact). I've smoke-tested both: the Qwen3.5 text-only Bridge adapter loads the model, its `blocks.{i}.hook_out` dims match the published artifact (32 layers, d_model 2560), and backward hooks deliver gradients — so both readout and native fitting are feasible on it.

### Alternatives

- **Standalone package** (like the MLX ports): works, but fragments the ecosystem — TL users would still lack lens-aware hooks/tests, and the technique pairs naturally with TL's existing cache/hook idioms. The official repo being frozen makes the canonical mech-interp library the right long-term home.
- **Depend on the official `jlens` package**: it's frozen, pins `transformers>=5.5`, and its model adapter duplicates what TL already abstracts (and it would still leave interventions unimplemented).
- **Readout-only integration** (no fitting): smaller, but native fitting is what makes the feature self-sufficient for models without published artifacts, and it reuses TL's backward-hook machinery — the marginal complexity is contained.

### Additional context

- Paper: Gurnee et al., *Verbalizable Representations Form a Global Workspace in Language Models*, Transformer Circuits Thread, 2026. https://transformer-circuits.pub/2026/workspace/index.html
- Official reference implementation (Apache-2.0, frozen): https://github.com/anthropics/jacobian-lens — the fitting estimator and artifact format to match; attribution will be included in module docstrings.
- Published artifacts: https://huggingface.co/neuronpedia/jacobian-lens (MIT card; Gemma-derived lenses remain subject to the Gemma Terms of Use, so they are always downloaded at runtime, never vendored).
- Interactive reference: https://www.neuronpedia.org/jlens
- Estimator note for reviewers: the official fit is deterministic — for each output dimension, a one-hot cotangent is planted at every valid target position at once and backpropagated (causal masking turns the per-source-position gradient into the sum over targets `t' ≥ t` for free); sum over targets, mean over source positions (first 16 and the final position excluded), unweighted mean over prompts; cost 1 forward + `ceil(d_model/dim_batch)` backwards per prompt. This maps directly onto `model.hooks(bwd_hooks=...)`.

I have a working spike of the loader/readout path: loading the published gemma-2-2b lens through a raw-weights `TransformerBridge` and reading out via the model's own `ln_final`/`unembed` reproduces the official implementation's per-layer top-8 readouts (worst case 7/8 top-8 overlap across 75 layer×prompt cells on 3 fixed prompts, top-1 identical in 70/75; the residue is bf16 near-ties). If this direction looks right, I'll follow up with the PR against `dev` shortly. Happy to adjust API names, placement, scope, and test tiers to whatever the maintainers prefer.

### Checklist

danielxmed:
# Description

Implements the Jacobian lens (J-lens) as a TransformerBridge-only analysis component, following the design proposal in #1505 and the estimator from Gurnee et al., [*Verbalizable Representations Form a Global Workspace in Language Models*](https://transformer-circuits.pub/2026/workspace/index.html).

The J-lens characterizes an intermediate residual activation by its corpus-averaged first-order causal effect on the final residual stream. Each fitted source layer has one `d_model × d_model` transport matrix; readout then uses the model's own final norm, unembedding, and architecture-specific output-logit transform. The logit lens is available through the same path by disabling Jacobian transport.

Fixes #1505.

## What is included

- `JacobianLens.from_pretrained`, `load`, and `save`, compatible with the published `neuronpedia/jacobian-lens` artifact format. TransformerLens artifacts may add recursively validated provenance metadata while remaining loadable with `weights_only=True`.
- `readout`, returning retained top-k values and token ids by default, with full vocabulary logits available as an explicit opt-in. The final layer reuses the model's actual logits rather than recomputing them.
- Native deterministic fitting with Bridge backward hooks. The estimator sums causal target effects `t' >= t`, averages valid source positions per prompt, then averages prompts equally. `merge` combines disjoint prompt shards using positive `n_prompts` weights and requires matching provenance.
- `steering_hooks`, `ablation_hooks`, and pseudoinverse-coordinate `swap_hooks`, with shared position, shape, dtype, and device validation.
- Per-layer/per-device Jacobian caching, Hub retry integration, optional Hub revision pinning, and complete fit provenance including model name, resolved model revision, TransformerLens version, raw-weight contract, hook convention, corpus, prompt count, estimator settings, and dtype.
- An executable Gemma-2 demo in `demos/Jacobian_Lens_Demo.ipynb`.

## Supported model contract

JacobianLens requires a freshly booted, raw-weight, causal decoder-only `TransformerBridge` with a single rank-3 residual stream per block and a direct final-norm/unembedding path. `HookedTransformer`, compatibility mode, manually processed weights, encoder/seq2seq models, recurrent physical-block reuse, multi-stream block outputs, and unsupported final projections are rejected explicitly before fitting or readout.

Output-logit behavior is delegated to architecture adapters. This PR adds explicit contracts for Cohere/Cohere2 scaling, Granite division by `logits_scaling`, Falcon-H1 multiplication by `lm_head_multiplier`, Gemma-family softcaps, and fail-closed handling for ambiguous MPT `logit_scale` configurations.

## Review-driven changes

- Corrected and pinned the exact causal estimator with a cross-position toy test.
- Removed HookedTransformer support and the configurable public `target_layer` path.
- Centralized model validation across fitting, readout, lens vectors, and all intervention builders.
- Hardened metadata safety, Hub downloads, shard merging, top-k memory behavior, device caching, interventions, and swap conditioning.
- Added fail-closed topology, processed-weight, output-transform, model-name, and resolved-revision validation.
- Added real TransformerBridge steering/fitting coverage, a small regular-CI Gemma2 test, and the requested slow `google/gemma-2-2b-it` published-artifact parity test covering top-k and softcapping.

## Type of change

## Validation

All results below were produced from the final reviewed tree on local WSL/Ubuntu with the local RTX GPU where applicable:

- `make test-pr`: 3,521 unit, 18 docstring, 150 acceptance, and 856 integration tests passed.
- Slow model tests: the published Gemma-2 parity/top-k/softcap test passed, and both GPT-2 fitting/merge/convergence tests passed.
- Notebook: 8/8 cells passed under `nbval`.
- `make check-format`: 640 files clean.
- `uv run mypy .`: 317 source files clean.
- `uv build`: sdist and wheel built successfully.
- `git diff --check` and a secret-pattern scan passed.

# Checklist

- [x] I have added tests that prove the feature works
- [x] I have not rewritten tests relating to key interfaces in a way that affects backward compatibility

### DIFF
--- a/tests/integration/test_jacobian_lens.py
+++ b/tests/integration/test_jacobian_lens.py
@@ -1,14 +1,9 @@
-"""Integration tests for the Jacobian lens tool on gpt2 (cached).
+"""Real-model integration tests for the TransformerBridge Jacobian lens paths.
 
-These exercise the real-model paths: loading the published gpt2-small lens from
-the Hugging Face Hub (``neuronpedia/jacobian-lens``, ~13 MB), readout invariants
-(the final-layer row equals the model's own logits; late-layer readouts align
-with the model output), HookedTransformer/TransformerBridge agreement in the raw
-weight basis, an exact fit/merge consistency identity, and a causal steering
-smoke test.
-
-gpt2 is in the CI model cache, so these live in ``integration/`` per
-``tests/AGENTS.md``; the small fits are marked ``slow``.
+The cached GPT-2 model exercises published-artifact loading, top-k and full-logit
+readouts, a native fit, and causal steering. A tiny random Gemma2 bridge keeps
+RMSNorm and logit-softcap coverage in the regular CI suite, while a slow
+Gemma-2-2b-it test checks the published artifact on the real architecture.
 """
 
 import pytest
@@ -16,7 +11,11 @@ import torch
 
 PROMPT = "The Eiffel Tower is in the city of"
 LENS_REPO = "neuronpedia/jacobian-lens"
-LENS_FILE = "gpt2-small/jlens/Salesforce-wikitext/gpt2_jacobian_lens.pt"
+LENS_REVISION = "a4114d7752d11eb546e6cf372213d7e75526d3a1"
+GPT2_LENS_FILE = "gpt2-small/jlens/Salesforce-wikitext/gpt2_jacobian_lens.pt"
+GEMMA_MODEL = "google/gemma-2-2b-it"
+GEMMA_LENS_FILE = "gemma-2-2b-it/jlens/Salesforce-wikitext/gemma-2-2b-it_jacobian_lens.pt"
+FIT_CORPUS = "jacobian-lens-integration-prompts-v1"
 
 # Long enough to clear the default 16-position source skip.
 FIT_PROMPTS = [
@@ -30,47 +29,54 @@ FIT_PROMPTS = [
 
 
 @pytest.fixture(scope="module")
-def ht_gpt2():
-    from transformer_lens import HookedTransformer
-
-    return HookedTransformer.from_pretrained_no_processing("gpt2", device="cpu")
-
-
-@pytest.fixture(scope="module")
-def bridge_gpt2():
+def gpt2_bridge():
     from transformer_lens.model_bridge import TransformerBridge
 
-    return TransformerBridge.boot_transformers("gpt2", dtype=torch.float32, device="cpu")
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+    return TransformerBridge.boot_transformers("gpt2", dtype=torch.float32, device=device)
 
 
 @pytest.fixture(scope="module")
-def published_lens(ht_gpt2):
+def published_gpt2_lens(gpt2_bridge):
     from transformer_lens.tools.analysis import JacobianLens
 
-    return JacobianLens.from_pretrained(LENS_REPO, filename=LENS_FILE, model=ht_gpt2)
+    return JacobianLens.from_pretrained(
+        LENS_REPO,
+        filename=GPT2_LENS_FILE,
+        revision=LENS_REVISION,
+        model=gpt2_bridge,
+    )
 
 
-def test_published_lens_loads_and_validates(published_lens, ht_gpt2):
-    assert published_lens.d_model == 768
-    assert published_lens.source_layers == list(range(11))
-    assert published_lens.n_prompts > 0
-    assert published_lens.validate_model(ht_gpt2) is published_lens
+def test_published_lens_loads_and_validates(published_gpt2_lens, gpt2_bridge):
+    assert published_gpt2_lens.d_model == 768
+    assert published_gpt2_lens.source_layers == list(range(11))
+    assert published_gpt2_lens.n_prompts > 0
+    assert published_gpt2_lens.validate_model(gpt2_bridge) is published_gpt2_lens
 
 
-def test_final_layer_readout_equals_model_logits(published_lens, ht_gpt2):
-    result = published_lens.readout(ht_gpt2, PROMPT, layers=[11], positions=[-1])
-    torch.testing.assert_close(result.lens_logits[11], result.model_logits)
+def test_readout_defaults_to_topk_and_final_layer_matches_model(published_gpt2_lens, gpt2_bridge):
+    result = published_gpt2_lens.readout(gpt2_bridge, PROMPT, layers=[11], positions=[-1])
+
+    assert result.lens_logits is None
+    assert result.model_logits is None
+    assert result.lens_topk_values[11].shape == (1, 10)
+    assert result.lens_topk_indices[11].shape == (1, 10)
+    torch.testing.assert_close(result.lens_topk_values[11], result.model_topk_values)
+    assert torch.equal(result.lens_topk_indices[11], result.model_topk_indices)
 
 
-def test_late_layer_readout_aligns_with_model_output(published_lens, ht_gpt2):
-    """J_l -> I near the output: the rank of the model's top-1 token under the
-    J-lens readout must collapse from thousands early to near zero late.
-    (Top-k overlap is deliberately not asserted mid-stack — there the lens
-    surfaces verbalizable content, not output predictions.)
-
-    Measured on gpt2 for this prompt: rank 8577 at L0, 12 at L10, 0 at L11.
-    """
-    result = published_lens.readout(ht_gpt2, PROMPT, layers=[0, 10, 11], positions=[-1])
+def test_late_layer_readout_aligns_with_model_output(published_gpt2_lens, gpt2_bridge):
+    """The model's top-1 token moves from a poor early rank to first at output."""
+    result = published_gpt2_lens.readout(
+        gpt2_bridge,
+        PROMPT,
+        layers=[0, 10, 11],
+        positions=[-1],
+        return_full_logits=True,
+    )
+    assert result.lens_logits is not None
+    assert result.model_logits is not None
     model_top1 = result.model_logits[-1].argmax().item()
 
     def rank_of_model_top1(layer):
@@ -82,77 +88,63 @@ def test_late_layer_readout_aligns_with_model_output(published_lens, ht_gpt2):
     assert rank_of_model_top1(11) == 0
 
 
-def test_logit_lens_baseline_differs_mid_stack(published_lens, ht_gpt2):
-    """use_jacobian=False is the logit lens; mid-stack the two must diverge
-    (the whole point of the lens) while sharing the identical code path."""
-    lens_result = published_lens.readout(ht_gpt2, PROMPT, layers=[5], positions=[-1])
-    baseline = published_lens.readout(
-        ht_gpt2, PROMPT, layers=[5], positions=[-1], use_jacobian=False
+def test_logit_lens_baseline_differs_mid_stack(published_gpt2_lens, gpt2_bridge):
+    """The Jacobian and identity transports must diverge in the middle."""
+    lens_result = published_gpt2_lens.readout(
+        gpt2_bridge,
+        PROMPT,
+        layers=[5],
+        positions=[-1],
+        return_full_logits=True,
     )
+    baseline = published_gpt2_lens.readout(
+        gpt2_bridge,
+        PROMPT,
+        layers=[5],
+        positions=[-1],
+        use_jacobian=False,
+        return_full_logits=True,
+    )
+    assert lens_result.lens_logits is not None
+    assert baseline.lens_logits is not None
     assert not torch.allclose(lens_result.lens_logits[5], baseline.lens_logits[5])
 
 
-def test_bridge_and_ht_readouts_agree(published_lens, ht_gpt2, bridge_gpt2):
-    """Raw-basis Bridge and no-processing HT must produce near-identical
-    readouts (same weights, fp32); allow one near-tie swap in the top-8."""
-    ht_result = published_lens.readout(ht_gpt2, PROMPT, positions=[-1])
-    bridge_result = published_lens.readout(bridge_gpt2, PROMPT, positions=[-1])
-    assert ht_result.tokens.tolist() == bridge_result.tokens.tolist()
-    for layer in published_lens.source_layers:
-        ht_top8 = ht_result.lens_logits[layer][-1].topk(8).indices.tolist()
-        bridge_top8 = bridge_result.lens_logits[layer][-1].topk(8).indices.tolist()
-        assert len(set(ht_top8) & set(bridge_top8)) >= 7, f"layer {layer}"
-        assert ht_top8[0] == bridge_top8[0], f"layer {layer} top-1 mismatch"
-
-
-def test_steering_raises_target_logit(published_lens, ht_gpt2):
-    """Adding a token's J-lens direction mid-stack must raise that token's
-    output logit (directional causal smoke test)."""
+def test_bridge_steering_raises_target_logit(published_gpt2_lens, gpt2_bridge):
+    """A real Bridge run must honor hooks built from a published lens."""
     target = " Paris"
-    target_id = ht_gpt2.to_single_token(target)
-    tokens = ht_gpt2.to_tokens(PROMPT)
+    target_id = gpt2_bridge.to_single_token(target)
+    tokens = gpt2_bridge.to_tokens(PROMPT)
+    hooks = published_gpt2_lens.steering_hooks(gpt2_bridge, target, layers=[6, 7, 8], alpha=6.0)
+
     with torch.no_grad():
-        baseline = ht_gpt2(tokens)[0, -1, target_id].item()
-        hooks = published_lens.steering_hooks(ht_gpt2, target, layers=[6, 7, 8], alpha=6.0)
-        with ht_gpt2.hooks(fwd_hooks=hooks):
-            steered = ht_gpt2(tokens)[0, -1, target_id].item()
+        baseline = gpt2_bridge(tokens)[0, -1, target_id].item()
+        steered = gpt2_bridge.run_with_hooks(tokens, fwd_hooks=hooks)[0, -1, target_id].item()
+
     assert steered > baseline
 
 
-def test_processed_models_are_refused(published_lens):
-    from transformer_lens import HookedTransformer
-    from transformer_lens.tools.analysis import JacobianLens
-
-    processed = HookedTransformer.from_pretrained("gpt2", device="cpu")
-    with pytest.raises(ValueError, match="fold_ln"):
-        published_lens.validate_model(processed)
-    with pytest.raises(ValueError, match="fold_ln"):
-        JacobianLens.fit(processed, FIT_PROMPTS, show_progress=False)
-    # fold_ln=False still centers weights (normalization_type stays "LN"), so the
-    # centered-unembed signature is the marker that must catch it.
-    half_processed = HookedTransformer.from_pretrained("gpt2", fold_ln=False, device="cpu")
-    with pytest.raises(ValueError, match="mean-centered"):
-        published_lens.validate_model(half_processed)
-
-
 @pytest.mark.slow
-def test_fit_merge_consistency_and_finiteness(ht_gpt2):
-    """Exact identity: fitting two prompts jointly equals merging two
-    single-prompt fits (the estimator averages prompts unweighted and merge
-    weights by n_prompts)."""
+def test_bridge_fit_merge_consistency_and_finiteness(gpt2_bridge):
+    """Joint fitting equals merging per-prompt fits for a real Bridge."""
     from transformer_lens.tools.analysis import JacobianLens
 
-    kwargs = dict(
+    fit_kwargs = dict(
         source_layers=[0, 5, 10],
         dim_batch=96,
         max_seq_len=32,
         show_progress=False,
     )
-    joint = JacobianLens.fit(ht_gpt2, FIT_PROMPTS, **kwargs)
+    joint = JacobianLens.fit(gpt2_bridge, FIT_PROMPTS, corpus=FIT_CORPUS, **fit_kwargs)
     merged = JacobianLens.merge(
-        [JacobianLens.fit(ht_gpt2, [prompt], **kwargs) for prompt in FIT_PROMPTS]
+        [
+            JacobianLens.fit(gpt2_bridge, [prompt], corpus=FIT_CORPUS, **fit_kwargs)
+            for prompt in FIT_PROMPTS
+        ]
     )
+
     assert joint.n_prompts == merged.n_prompts == 2
+    assert joint.metadata["corpus"] == FIT_CORPUS
     for layer in joint.source_layers:
         assert torch.isfinite(joint.jacobians[layer]).all()
         torch.testing.assert_close(
@@ -161,16 +153,14 @@ def test_fit_merge_consistency_and_finiteness(ht_gpt2):
 
 
 @pytest.mark.slow
-def test_small_fit_converges_toward_published_lens(published_lens, ht_gpt2):
-    """A native 2-prompt fit must already point in the published lens's
-    direction (measured: cosine 0.51 / 0.66 / 0.91 at layers 0 / 5 / 10 vs the
-    n=277 published gpt2 lens) — the estimator converges fast, and any basis
-    or orientation bug would drive these toward zero."""
+def test_small_bridge_fit_converges_toward_published_lens(published_gpt2_lens, gpt2_bridge):
+    """A two-prompt Bridge fit already points toward the published estimator."""
     from transformer_lens.tools.analysis import JacobianLens
 
     fitted = JacobianLens.fit(
-        ht_gpt2,
+        gpt2_bridge,
         FIT_PROMPTS,
+        corpus=FIT_CORPUS,
         source_layers=[0, 5, 10],
         dim_batch=96,
         max_seq_len=32,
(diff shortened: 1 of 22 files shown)


transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
+"""Jacobian Lens (J-lens).
+
+The Jacobian lens characterizes an intermediate residual-stream activation by its
+first-order causal effect on the model's output, averaged over a corpus of contexts.
+For each layer :math:`\\ell` it fits a single :math:`d_{model} \\times d_{model}` matrix
+
+.. math::
+
+    J_\\ell = \\mathbb{E}_{\\text{prompt},\\,t,\\,t' \\geq t}
```
jlarson4: The header formula is a mean over (prompt, t, t' >= t) triples, but the implementation takes the sum over targets and averages only over sources and prompts (~56× difference at seq 128, skip 16). Anyone calibrating raw J magnitudes from this formula will be misled; the header is the only dissenter, so fix it here.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+            One vector per token, fp32, on the model's device.
+        """
+        token_ids = _to_token_ids(model, tokens)
+        matrix = self.jacobians.get(layer)
+        if matrix is None:
+            raise ValueError(
+                f"layer {layer} is not in this lens's source layers "
+                f"({self.source_layers[0]}..{self.source_layers[-1]})"
+            )
+        unembed_columns = model.W_U[:, token_ids].float()  # [d_model, n]
```
jlarson4, line 601: `lens_vectors` reads `model.W_U` with no `validate_model(model)`. It's the gateway for `steering_hooks`/`ablation_hooks`/`swap_hooks`, so all three build and run on a `fold_ln`+`center_unembed` HT (or compat-mode Bridge; the hook names still resolve) while computing vectors in the wrong residual basis. That's the silent-wrongness L34–37 promises to refuse loudly, and #1505's plan validated on attach, the per-call model API change dropped that invariant. One `validate_model(model)` here restores it for all three builders.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+        """
+        payload: Dict[str, Any] = {
+            "J": {layer: matrix.to(dtype) for layer, matrix in self.jacobians.items()},
+            "n_prompts": self.n_prompts,
+            "source_layers": self.source_layers,
+            "d_model": self.d_model,
+        }
+        if self.metadata:
+            payload["metadata"] = self.metadata
+        torch.save(payload, path)
```
jlarson4, line 225: `save()` stores arbitrary metadata, but `load()` (L213) uses `weights_only=True`. A numpy scalar in metadata saves fine and then fails this module's own `load()` with an opaque `UnpicklingError`, before the friendly no-'J'-key error is reached. Consider validating metadata values are weights-only-safe primitives at save time.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+        import os
+
+        if os.path.isfile(name_or_path):
+            lens = cls.load(name_or_path)
+        elif os.path.isdir(name_or_path):
+            lens = cls.load(os.path.join(name_or_path, filename))
+        else:
+            from huggingface_hub import hf_hub_download
+
+            local_path = hf_hub_download(name_or_path, filename, revision=revision)
```
jlarson4: Two robustness gaps: 
1. `hf_hub_download` here isn't wrapped by `call_hf_with_retry`, so lens downloads lack the 429 backoff every other Hub fetch in the repo gets.
2. `revision=None` tracks the mutable main of a third-party repo. Wrap the call, and consider documenting revision-pinning in the docstring.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+        """
+        if not lenses:
+            raise ValueError("cannot merge an empty sequence of lenses")
+        first = lenses[0]
+        for other in lenses[1:]:
+            if other.source_layers != first.source_layers or other.d_model != first.d_model:
+                raise ValueError(
+                    "all lenses being merged must share the same source_layers and d_model"
+                )
+        total = sum(lens.n_prompts for lens in lenses)
```
jlarson4, line 346: `load()` defaults a missing `n_prompts` to 0 (L221) and this weighting multiplies that shard by zero. Raise or warn when any merged lens has `n_prompts == 0`.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+        Raises:
+            ValueError: If the model fails :meth:`validate_model`, ``input`` is
+                batched, or a requested layer has no transport matrix.
+        """
+        self.validate_model(model)
+        tokens = model.to_tokens(input) if isinstance(input, str) else input
+        if tokens.ndim != 2 or tokens.shape[0] != 1:
+            raise ValueError(f"readout expects a single prompt; got shape {tuple(tokens.shape)}")
+        n_layers = model.cfg.n_layers
+        final_layer = n_layers - 1
```
jlarson4, line 509: `target_layer` is a documented public kwarg (L651, L686–688) but lands only in metadata (L762). The readout always uses the final-layer-identity convention and presents `lens_logits` as approximating the model's output. A `fit(target_layer=6)` lens reads out under silently mixed conventions. Either consult the stored `target_layer` (and skip/adjust the identity row) or raise on non-default-target lenses.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+            if out_of_range:
+                raise ValueError(
+                    f"positions {out_of_range} out of range for a {seq_len}-token prompt"
+                )
+
+        hook_names = {layer: _resid_post_hook_name(model, layer) for layer in layers}
+        wanted = set(hook_names.values())
+        logits, cache = model.run_with_cache(tokens, names_filter=lambda name: name in wanted)
+
+        lens_logits: Dict[int, torch.Tensor] = {}
```
jlarson4: At gemma-scale vocab this is ~1.7 GB of CPU tensors for a 64-token prompt (26 × [64, 256k] fp32). `top_tokens` needs only top-k. A `top_k` retention mode (storing values+indices) would cut it to KBs with the full-tensor path opt-in. 

Also, the `layer == final_layer` branch recomputes logits that `run_with_cache` already returned on L431. Reusing them saves a `pos * d_model * d_vocab` matmul per readout.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+        Args:
+            residual: Activations from the output of block ``layer``.
+            layer: Source layer index.
+        """
+        if layer not in self.jacobians:
+            raise ValueError(
+                f"layer {layer} is not in this lens's source layers "
+                f"({self.source_layers[0]}..{self.source_layers[-1]})"
+            )
+        matrix = self.jacobians[layer].to(residual.device)
```
jlarson4: Every call of `transport`/`lens_vectors` moves `self.jacobians[layer]` to the model device. ~530 MB over PCIe per full gemma-2-2b readout, re-paid by every hook build. A per-device cache (or an explicit `lens.to(device)`) makes it a one-time transfer.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+            direction = self.lens_vectors(model, token, layer)[0]
+            unit = direction / direction.norm()
+
+            def steer_fn(
+                activation: Float[torch.Tensor, "batch pos d_model"],
+                hook: Any,
+                unit: torch.Tensor = unit,
+            ) -> Float[torch.Tensor, "batch pos d_model"]:
+                scale = alpha * activation.float().norm(dim=-1).median()
+                delta = (scale * unit).to(activation.dtype)
```
jlarson4: Two hardening gaps, both also present in `ablate_fn` (L572/L580) and `swap_fn` (L628/L634): 
1. closure tensors are pinned to `W_U's` device (L523/L565/L619) and only ever cast to `activation.dtype`. On sharded models (`n_devices=2`, accelerate `device_map`) early-layer hooks raise cross-device `RuntimeErrors`, the one surface the device-robustness commit didn't cover. Move tensors to `activation.device` in the hook or at build time.
2. `activation[:, list(positions), :]` has no bounds check or negative-index normalization, unlike `readout` (L420–425). Out-of-range crashes mid-forward with a bare `IndexError`, and any explicit positions list breaks under KV-cache decode (`seq_len==1`). Factor out `readout`'s normalization so you can reuse it and since the three builders already share ~150 lines of near-identical scaffolding (per-layer loop, `lens_vectors` fetch, positions slice/writeback), extracting one shared wrapper gives this hardening a single home instead of three.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+                intermediate-layer band).
+            alpha: Swap strength.
+            positions: Positions to swap. Defaults to all positions.
+
+        Returns:
+            ``[(hook_name, fn), ...]`` for ``model.hooks(fwd_hooks=...)``.
+        """
+        hooks = []
+        for layer in layers:
+            basis = self.lens_vectors(model, [source_token, target_token], layer).T  # [d, 2]
```
jlarson4: Reproduced both failure legs: `source_token == target_token` yields a rank-1 basis whose `pinv` truncates the zero singular value. The swap becomes a silent no-op; near-parallel vectors amplify as ~1/sin(angle), silently destroying the forward pass for near-synonym swaps. Assert `source != target` and raise (or warn) past a cosine-similarity threshold before building the hook.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+                )
+                for layer in resolved_sources:
+                    jacobian_sum[layer] += per_prompt[layer]
+                n_done += 1
+        if n_done == 0:
+            raise ValueError(
+                "every prompt was too short to contribute valid positions; nothing was fitted"
+            )
+
+        full_metadata: Dict[str, Any] = {
```
jlarson4: #1505 committed artifacts to carry model name, TL version, processing flags, hook convention, corpus, n_prompts, and dtype; this dict has no TL version, no processing flags, no corpus identifier, and no dtype. The missing dtype matters most, `bf16` fits are silently degraded and nothing records it.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+                "model (raw weights) for Jacobian-lens analyses."
+            )
+        return
+    unembed_weight = getattr(model, "W_U", None)
+    if unembed_weight is not None:
+        # A 64-row slice is enough: centering zeroes every row mean exactly, while
+        # raw checkpoints sit orders of magnitude above the 0.01 threshold
+        # (measured: gpt2 3.4, gemma-2-2b 17.2; centered gpt2 ~1e-7).
+        rows = unembed_weight[:64].float()
+        if rows.mean(dim=-1).abs().max() < 0.01 * rows.abs().mean():
```
jlarson4: With this threshold, random-init `W_U` trips the guard in 20/20 seeds at `d_vocab=256k` (0/20 at GPT-2's 50k). Random row means scale as ~`1/sqrt(d_vocab)` and cross 0.01 right around Gemma/Llama-3 vocabs, so fitting on an untrained control model (a standard interp baseline) errors out claiming `center_unembed`. Your own measurement on L836 (centered ≈ 1e-7 vs trained ≥ 3.4) supports a much tighter threshold with ~10^4 margin both ways:
```
if rows.mean(dim=-1).abs().max() < 1e-3 * rows.abs().mean():
```

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+    target = captured[_resid_post_hook_name(model, target_layer)]
+    sources = [captured[_resid_post_hook_name(model, layer)] for layer in source_layers]
+    device = target.device
+    positions_index = torch.tensor(valid_positions, device=device)
+    batch_index = torch.arange(dim_batch, device=device)
+
+    jacobians = {
+        layer: torch.zeros(d_model, d_model, dtype=torch.float32) for layer in source_layers
+    }
+    cotangent = torch.zeros_like(target)
```
jlarson4, line 1275: `zeros_like(target)` makes the cotangent run in the model dtype. `fp32` begins only at the post-hoc `.float()` on L971. For bf16 models that's ~8-bit-mantissa accumulation with no test pinning it and no recorded fit dtype. At minimum record the dtype in metadata and recommend fp32 fits in the docstring, upcasting the captured activations would fix it outright.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -0,0 +1,973 @@
...
+) -> Float[torch.Tensor, "pos d_vocab"]:
+    """Apply the model's own final norm, unembedding, and logit soft cap.
+
+    The norm/unembed components contract on ``[batch, pos, d_model]``, so the
+    position rows are passed through with a singleton batch axis. The residual
+    is moved to the unembedding's device for sharded/multi-GPU models.
+    """
+    compute_dtype = model.W_U.dtype
+    batched = residual.to(device=model.W_U.device, dtype=compute_dtype).unsqueeze(0)
+    logits = model.unembed(model.ln_final(batched)).float().squeeze(0)
```
jlarson4: This line assumes every validated model owns an `ln_final` module, but the two ends of the chain disagree: `_require_unprocessed_weights` normalizes `normalization_type=None` to `""` (see L812), so None-norm models pass the `fold_ln` check, While `HookedTransformer.__init__` explicitly skips creating `ln_final` when `normalization_type is None` (see `HookedTransformer.py:247-249`). `nn.Module.__getattr__` then raises on the missing submodule. The failure ordering is what we're aiming to fix here: `fit()` never unembeds anything, so it validates and completes, the crash waits for the first readout.
```
model = HookedTransformer(HookedTransformerConfig(..., normalization_type=None))  # attn-only toy
lens = JacobianLens.fit(model, prompts)  # validator passes; full fit succeeds
lens.readout(model, prompt)  # AttributeError: 'HookedTransformer' object has no attribute 'ln_final'
```
The user pays the whole fitting cost, then gets a bare `AttributeError` from deep inside `_unembed` that names neither normalization nor the lens contract. 

Two possible solutions: 
1. Reject `normalization_type in (None, "")` in `_require_unprocessed_weights` with an actionable message. This option is loudest, though it also bans `fit`, which works fine on these models
2. Mirror the model's own forward semantics: `final = model.ln_final(batched) if getattr(model, "ln_final", None) is not None else batched`. Two lines that keep attn-only toy models fully supported and unembed them exactly the way `HookedTransformer` itself does.

tests/integration/test_jacobian_lens.py
```diff
@@ -0,0 +1,186 @@
...
+
+@pytest.fixture(scope="module")
+def bridge_gpt2():
+    from transformer_lens.model_bridge import TransformerBridge
+
+    return TransformerBridge.boot_transformers("gpt2", dtype=torch.float32, device="cpu")
+
+
+@pytest.fixture(scope="module")
+def published_lens(ht_gpt2):
```
jlarson4: This fixture is only consumed by the readout-agreement test. `fit()` and all three intervention builders never execute on a real Bridge in CI, and those are exactly the code paths the module-side comments flag. One small Bridge fit-or-steering test here would cover them.

tests/unit/tools/test_jacobian_lens.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class _ToyBlock(nn.Module):
+    def __init__(self, d_model: int):
+        super().__init__()
+        self.linear = nn.Linear(d_model, d_model, bias=False)
+        nn.init.normal_(self.linear.weight, std=0.2)
+        self.hook_resid_post = HookPoint()
+
+    def forward(self, resid):
```
jlarson4: With per-position `Linear` blocks there's no cross-position interaction, so only the `t' == t` term is nonzero. A deliberately wrong fit planting cotangents at ALL positions passes the closed-form test bit-for-bit. The integration tests are cosine-based or fit-vs-fit, so nothing pins the target set. One toy variant with attention-style mixing closes the hole.

jlarson4 (commented):
Hi @danielxmed! Thanks for putting this together, finally getting around to a full review for you. This is a very comprehensive review with a lot of comments, apologies in advance. For something like this I want to be extra thorough. Please let me know if you run into any questions as you work through them

I have a couple general suggestions that don't tie to any specific line:

1. gemma parity test: #1505 committed to slow `gemma-2-2b` parity tests and itself noted the CI-cached `-it` variant has a published lens. A `@pytest.mark.slow gemma-2-2b-it` parity test is feasible and would give the `softcap` path and a second architecture CI coverage.
2. I noted in the $1505 that we should drop support for `HookedTransformer` from this tool. Please do that, and then begin making your way through my comments. Some of them may be obsoleted by dropping that support, ping me if you are uncertain at any point.

danielxmed:
Thanks for the thorough review. I reworked the PR around a strict TransformerBridge-only contract and addressed all inline points.

- The estimator definition now matches the causal target sum, with a cross-position test pinning `t' >= t`.
- Validation now covers fitting, readout, lens vectors, and all interventions. HookedTransformer and configurable `target_layer` support were removed; incompatible processing, topology, attention, and output paths fail explicitly.
- Persistence, Hub retries and revision pinning, shard merging, device caching, top-k retention, final-logit reuse, intervention position/device handling, and swap conditioning are hardened.
- Fit provenance now records and validates `model_name` and the resolved `model_revision`, alongside the TransformerLens version, raw-weight contract, hook convention, corpus, prompt count, estimator settings, and dtype. Reduced-precision fits warn and recommend fp32.
- Coverage now includes real Bridge steering/fitting, causal position mixing, regular-CI Gemma2, and the requested slow `google/gemma-2-2b-it` published-artifact parity test covering top-k and softcapping.

Final validation is clean: `make test-pr` passed 3,521 unit, 18 docstring, 150 acceptance, and 856 integration tests; the Gemma slow test passed, both GPT-2 slow tests passed, and notebook validation passed 8/8. Mypy passed across 317 source files, formatting passed across 640 files, and both the sdist and wheel built successfully.

jlarson4:
Great work on updating this @danielxmed! You've covered everything I asked for an then some, thank you. Merging now, and going to include the baseline for this in the next release. Are you intending to move on to the next parts of your issue, or should I open it up to other contributors?

PR #1509, opened 2026-07-11
Keep input_tokens on the model device in generate()

### LINKED ISSUE #1508: [Bug Report] generate() device mismatch on CUDA machines: input_tokens left on original device after input is moved
**Describe the bug**
On a CUDA machine, `HookedTransformer.generate()` raises `RuntimeError: Expected all tensors to be on the same device ...` when the input is a token tensor on a different device than the model (e.g. CPU tokens passed to a model whose `cfg.device` is `"cuda"`).

Root cause: in `generate()`, `input_tokens` is bound to the original input tensor *before* `input = input.to(device)` moves the input to the model's device, so `input_tokens` keeps pointing at the original CPU tensor. Sampled tokens are produced on the model's device, so `torch.cat((input_tokens, ...))` mixes devices — both in the freq-penalty sampling path (`HookedTransformer.py` ~line 2187) and in the final output concatenation (~line 2222, hit with `do_sample=False` too).

This is easiest to hit with tokenizer-free models (algorithmic tasks): `HookedTransformerConfig` defaults `device` to `"cuda"` when available, and hand-built token tensors typically live on CPU. Concretely, 4 of the 6 existing unit tests in `tests/unit/test_generate_no_tokenizer.py` fail on any CUDA-visible machine at current `dev`, while passing on CPU-only CI (no cross-device mismatch is possible there).

**Code example**
```python
import torch
from transformer_lens.config import HookedTransformerConfig

    n_layers=1, d_model=16, d_head=4, n_heads=4, n_ctx=32, d_vocab=20, attn_only=True
)
model = HookedTransformer(cfg)  # cfg.device defaults to "cuda" when available
tokens = torch.zeros((1, 5), dtype=torch.long)  # cpu
model.generate(tokens, max_new_tokens=3, stop_at_eos=False, return_type="tokens", verbose=False)
```

```
File ".../transformer_lens/HookedTransformer.py", line 2187, in generate
    tokens=torch.cat(
RuntimeError: Expected all tensors to be on the same device, but got tensors is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA_cat)
```

Or simply run `pytest tests/unit/test_generate_no_tokenizer.py` on a GPU machine (4 failed, 2 passed at `dev` 8d01cbdb).

**System Info**
- transformer_lens installed from source (`dev` branch, commit 8d01cbdb)
- Linux (Ubuntu on WSL2), NVIDIA RTX 5070 Laptop GPU
- Python 3.12.13, torch 2.10.0+cu128

**Additional context**
The ordering was introduced in #999 (`input_tokens = input ...` now runs before `input = input.to(device)`). `generate_stream()` and `TransformerBridge.generate()` are not affected — both move their token tensor to the model device before use.

Happy to submit a fix: re-alias `input_tokens` after the move so it lives on the model's device, plus a CUDA-gated regression test for the explicit cross-device case.

### Checklist

danielxmed:
# Description

`HookedTransformer.generate()` crashes on CUDA machines when the input is a token tensor on a different device than the model (e.g. CPU tokens passed to a model whose `cfg.device` is `"cuda"`, the default when CUDA is available):

```
RuntimeError: Expected all tensors to be on the same device, but got tensors is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA_cat)
```

**Root cause:** `input_tokens` is bound to the original input tensor *before* `input = input.to(device)` moves the input to the model's device, so it keeps pointing at the original CPU tensor. Sampled tokens are produced on the model's device, so `torch.cat((input_tokens, ...))` mixes devices — both in the freq-penalty sampling path and in the final output concatenation (the latter also hits `do_sample=False`). The ordering came in with #999.

**Fix:** after `input = input.to(device)`, re-alias `input_tokens = input` (guarded by `input_tokens is not None`, which is exactly the str/tokens input case). Re-aliasing rather than calling `.to(device)` again avoids a second host-to-device copy of the same data.

This is easiest to hit with tokenizer-free models (algorithmic tasks), which is why 4 of the 6 existing tests in `tests/unit/test_generate_no_tokenizer.py` fail on any CUDA-visible machine at current `dev` while passing on CPU-only CI. With this fix they pass on both. `generate_stream()` and `TransformerBridge.generate()` are not affected (both already move their token tensor before use), so no Bridge mirroring is needed.

**Tests:** added `test_generate_cpu_tokens_cuda_model`, a CUDA-gated (`skipif`) regression test that explicitly passes CPU tokens to a CUDA model and covers both the sampling and greedy paths. Verified it fails without the fix and passes with it.

Ran locally on a CUDA machine (RTX 5070, torch 2.10.0+cu128, Python 3.12):
- `pytest tests/unit/test_generate_no_tokenizer.py` — 4 failed / 2 passed before, 7 passed after
- `pytest tests/unit -m "not slow"` — passes
- `mypy transformer_lens/HookedTransformer.py`, `black`/`isort`/`pycln` checks — clean

Fixes #1508

## Type of change

# Checklist:

### DIFF
--- a/tests/unit/test_generate_no_tokenizer.py
+++ b/tests/unit/test_generate_no_tokenizer.py
@@ -3,6 +3,7 @@
 Regression test for https://github.com/TransformerLensOrg/TransformerLens/issues/483
 """
 
+import pytest
 import torch
 
 from transformer_lens import HookedTransformer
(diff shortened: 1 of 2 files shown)


jlarson4:
Great work on this! Verified and approved

PR #1516, opened 2026-07-15
Add TransformerBridge adapter for lightweight decoder-only pretrainin…

Cacapice:
# Description

## Summary

Adds an interoperability adapter for live model modules loaded from
lightweight decoder-only pretraining checkpoints (RoPE, RMSNorm, SwiGLU,
optional MoE), enabling mechanistic interpretability experiments on
models trained from scratch.

Rather than converting checkpoints into a separate TransformerLens
implementation, the adapter wraps the existing module and delegates
execution to its native `forward`, preserving the source model's semantics
(including RoPE conventions) while exposing the standard TransformerBridge
interface.

## Motivation

Interpretability research on small from-scratch models (circuit discovery,
induction heads, toy superposition models, controlled probing) needs
precise control over architecture, data, and checkpoints. This PR adds
that interoperability path: train a small decoder-only transformer, wrap
it with `build_pretrain_bridge`, confirm the wrapping is transparent (the
bridge delegates to and matches the source model's own forward exactly,
rather than running a separately reimplemented TransformerLens forward),
then use TransformerLens's caching, hooks, and patching. The goal is
architecture interoperability, not a general-purpose training framework.

## What changed

- **`PretrainArchitectureAdapter`**
  (`transformer_lens/model_bridge/supported_architectures/pretrain.py`)
  maps token embeddings, decoder blocks (RoPE attention, RMSNorm, gated
  SwiGLU MLP), final norm, and tied/untied unembedding.
- **`DenseOrMoEFeedForwardBridge`** dispatches each block's MLP
  structurally (`router` + `experts` versus `gate` + `up` + `down`), so
  dense, MoE, and mixed models use the same component mapping.
- The selected delegate is excluded from normal module registration to
  prevent duplicate named hook points. Structural validation requires
  projection/router fields to be modules and `experts` to be a registered
  `ModuleList` or `ModuleDict`, producing clear construction-time errors
  for unsupported module shapes.
- **`build_pretrain_bridge(model, cfg)`**: the public entry point. Wraps
  the source model, normalizes its output contract, and ensures bridge
  mode changes propagate to the source model (see Scope boundaries).
  Direct `build_bridge_from_module` use still works for advanced cases.
- Registered `"SwiGLUMoEForCausalLM"` in `SUPPORTED_ARCHITECTURES` and
  `INTENTIONAL_EXCLUDES` (no HF Hub checkpoint backs it).
- Config mutation: like `nanogpt.py`, the adapter mutates the passed-in
  `cfg` in place rather than copying it.
- Unit tests split by concern: `test_pretrain_adapter.py` (mapping,
  dispatch, tied embeddings) and `test_pretrain_model_container.py`
  (container, kwarg filtering, output normalization, hooks, train/eval,
  dtype, persistence -- see Scope boundaries), sharing mocks from
  `_pretrain_mocks.py`. Integration tests use a
  real reference model (`tests/mocks/tiny_pretrain_model.py`) with genuine
  adjacent-pair RoPE/RMSNorm math for logit and hook-placement parity
  (see the integration test module docstring for the caveat on what
  logit parity does and doesn't prove).
- **`MoEBridge` defensive validation:** added validation for tuple
  outputs -- rejecting empty tuples and non-tensor first elements with a
  clear `TypeError` (instead of a lower-level `HookPoint` type-check
  error), and guarding `hook_router_scores` so it is only invoked for
  tensor-valued router scores. Covered directly by new `MoEBridge` tests
  (empty tuple, non-tensor first element, non-tensor metadata preserved
  without firing the router hook, tensor router scores still firing it),
  not just indirectly through the pretrain adapter's tests. This is a
  small compatibility and error-reporting improvement discovered while
  integrating the pretrain adapter and does not change the behavior of
  standard `(hidden_states, router_scores)` MoE outputs.

## Supported behavior

`build_pretrain_bridge()` wraps the supplied model in place. Callers that
need a raw source-model checkpoint should capture its `state_dict()`
before bridge construction.

Requires a specific module protocol (documented in the adapter file's
docstring) rather than any decoder-only architecture generally. Within
that protocol:

- preserves the source model's rotary convention for native-forward
  execution, since the adapter delegates rather than reimplementing it —
  the tradeoff is that this adapter does not expose per-head attention
  hooks (`hook_q`/`hook_k`/`hook_v`/`hook_pattern`);
- RMS norm + gated SwiGLU MLP, dense/MoE/mixed, dispatched structurally;
- tied or untied embeddings; `forward` with or without a `**kwargs`
  catch-all (only `output_attentions` is stripped — anything else,
  including a caller typo, passes through and raises naturally);
- output as a `"logits"` dict, tensor, tensor-first tuple, or
  `.logits`-exposing object, all validated, with clear errors on
  malformed shapes.

## Container wrapping

`PretrainModelContainer` resolves attribute-name collisions with
`TransformerBridge` and normalizes output to the `.logits` contract
(raising immediately on missing/malformed output). Full mechanics are in
the class docstring. Kept local rather than generalizing
`TransformerBridge`'s output handling, to keep this PR architecture-scoped
— a candidate for later centralization, not addressed here.

The dense/MoE delegate is excluded from the public module tree to avoid
duplicate hook points; forward execution, gradients, dtype/device
conversion, `state_dict`-based reconstruction, hook lifecycle, and (via a
small generated subclass) train/eval mode all still reach the wrapped
source model — see Scope boundaries for why train/eval needed a
workaround here.

## Hook coverage

Block-level hooks only (`resid_pre`/`resid_mid`/`resid_post`, MLP
`hook_in`/`hook_out`) — no per-head `hook_q`/`hook_k`/`hook_v`/
`hook_pattern`. TransformerLens's HF-oriented attention bridges use
rotate-half RoPE, which does not match this adapter's adjacent-pair
convention and would break numerical equivalence, so it uses a plain
`AttentionBridge` delegating to the source model's own attention instead.
Per-head hooks would need a new attention bridge built for that
convention.

## Scope boundaries

No training loop, optimizers/schedules, dataset/tokenizer infrastructure,
experiment tracking, distributed training, or checkpoint-shard
reconstruction — those stay with the source framework. This PR only adds
the TransformerBridge mapping and its tests.

Worth flagging for discussion: `build_pretrain_bridge` reassigns
`bridge.__class__` to propagate train/eval mode (see code comments).
Reassigning `__class__` at runtime is unconventional; a
`TransformerBridge.train()` fix upstream would be cleaner if maintainers
are open to that scope.

Persistence is tested by saving the source model's pre-bridge
`state_dict`, reconstructing the source model and bridge from
configuration, and verifying output and mode-propagation equivalence.
Whole-object pickling is not asserted because it depends on unrelated
`TransformerBridge` internals, including locally defined attention
hook-conversion classes. A standardized `TransformerBridge` checkpoint
interface could provide a stable, versioned save/load contract,
improving usability and scalability without requiring whole-object
serialization.

## Testing

```bash
python -m pytest tests/unit/model_bridge/supported_architectures/test_pretrain_adapter.py -v
python -m pytest tests/unit/model_bridge/supported_architectures/test_pretrain_model_container.py -v
python -m pytest tests/integration/model_bridge/test_pretrain_adapter.py -v
python -m pytest tests/unit/model_bridge/generalized_components/test_moe_bridge_tuple_output.py -v
python -m pytest tests/unit/tools/test_model_registry.py -k TestRegistrySyncedWithFactory -v
```

Locally: 77 tests passed (26 adapter unit, 34 container unit, 9
integration, 4 generalized-component unit tests (MoEBridge tuple-output
validation), and 4 registry-sync tests). Covers: component
mapping and MLP dispatch (dense/MoE/mixed, checking actual
`run_with_cache` keys, not just `isinstance`, plus basic
type validation when an attribute name matches the MoE/gated-MLP protocol
but the value doesn't); logit and residual-stream parity; that
integration fixtures supply a `cfg` truthfully describing the model they
build (`build_pretrain_bridge` itself performs no cfg-vs-model
validation, so a mismatch silently reports wrong numbers on `bridge.cfg`
-- see `TestIntegrationFixturesSupplyATruthfulConfig`); activation
caching and hook interventions; kwarg filtering and output-normalization
edge cases; hook registration, gradients, dtype/device, `state_dict`-based
reconstruction; train/eval mode propagation (a cached generated subclass,
rather than per-instance method closures); malformed-architecture errors;
`MoEBridge`'s own tuple-output validation and router-score hook gating
(empty tuple, non-tensor first element, non-tensor metadata preserved
without firing the hook, tensor router scores still firing it).

Exact logit parity mainly demonstrates that wrapping does not alter the
delegated forward; it does not independently establish that the source
RoPE implementation is correct. The residual-stream decomposition and
hook-intervention tests provide the stronger evidence on hook placement.

Numerical parity is tested at float32 on CPU; float64 is tested for
preservation through construction only, not full parity at that dtype.

## Type of change

# Checklist:

### DIFF
--- a/TransformerLens/tests/integration/model_bridge/test_pretrain_adapter.py
+++ b/TransformerLens/tests/integration/model_bridge/test_pretrain_adapter.py
@@ -1,0 +1,312 @@
+"""Integration tests for PretrainArchitectureAdapter -- numerical parity
+against a real (not structural-mock) reference model.
+
+Uses `tests/mocks/tiny_pretrain_model.py`'s `TinyPretrainModel`, which
+implements genuine adjacent-pair RoPE and RMSNorm math (unlike
+`_pretrain_mocks.py`'s unit-test mocks, which stub RoPE with constant
+cos/sin and only need to get attribute names and call-signatures right).
+
+Covers: source/bridge logit parity (dense, full-MoE, mixed dense/MoE,
+untied embeddings); that this file's fixtures supply a `cfg` truthfully
+describing the wrapped model's head/vocab dimensions (a fixture property,
+not something `build_pretrain_bridge` itself enforces -- see
+`TestIntegrationFixturesSupplyATruthfulConfig`); residual-stream
+decomposition via `run_with_cache`; a forward-hook intervention
+cross-checked against an independent way of expressing the same edit; and
+float64 construction-time preservation.
+
+Note on scope: since the adapter delegates the whole forward to the
+source model rather than reimplementing attention/RoPE, exact logit
+parity mainly demonstrates that wrapping and output normalization don't
+alter an unhooked forward, not that the RoPE convention itself is
+correct (the same implementation runs on both sides of the comparison).
+`TestResidualStreamDecomposition` and `TestHookIntervention` are the
+stronger evidence for correct intervention points.
+"""
+
+from __future__ import annotations
+
+import torch
+
+from tests.mocks.tiny_pretrain_model import TinyPretrainModel
+from transformer_lens.config import TransformerBridgeConfig
+from transformer_lens.model_bridge.supported_architectures.pretrain import (
+    ARCHITECTURE_NAME,
+    build_pretrain_bridge,
+)
+
+
+def _make_cfg(
+    *,
+    d_model: int,
+    n_layers: int,
+    n_heads: int = 2,
+    d_vocab: int = 64,
+) -> TransformerBridgeConfig:
+    """Build a bridge config matching this file's TinyPretrainModel
+    fixtures (n_heads=2, vocab_size=64 by default). See
+    `TestIntegrationFixturesSupplyATruthfulConfig` for why this matching
+    matters and what happens when it doesn't."""
+    return TransformerBridgeConfig(
+        d_model=d_model,
+        d_head=d_model // n_heads,
+        n_layers=n_layers,
+        n_ctx=128,
+        n_heads=n_heads,
+        d_vocab=d_vocab,
+        d_mlp=d_model * 2,
+        architecture=ARCHITECTURE_NAME,
+    )
+
+
+class TestSourceBridgeLogitParity:
+    """Bridge output must equal the source model's own `"logits"` output
+    exactly -- not within tolerance. The adapter never translates weights
+    into a second parameter layout, so there is no floating-point drift
+    to tolerate; any mismatch here is a real bug, not numerical noise."""
+
+    def test_dense_model_logit_parity_exact(self) -> None:
+        torch.manual_seed(0)
+        model = TinyPretrainModel(d_model=16, n_heads=2, n_layers=2, d_ff=32, vocab_size=64)
+        model.eval()
+        tokens = torch.randint(0, 64, (1, 5))
+        with torch.no_grad():
+            source_logits = model(tokens)["logits"]
+
+        bridge = build_pretrain_bridge(model, _make_cfg(d_model=16, n_layers=2))
+        with torch.no_grad():
+            bridge_logits = bridge(tokens)
+
+        torch.testing.assert_close(bridge_logits, source_logits, atol=0, rtol=0)
+
+    def test_full_moe_model_logit_parity_exact(self) -> None:
+        """Every block's MLP is MoE (moe_layer_indices covers all layers)."""
+        torch.manual_seed(0)
+        model = TinyPretrainModel(
+            d_model=16,
+            n_heads=2,
+            n_layers=2,
+            d_ff=32,
+            vocab_size=64,
+            n_experts=4,
+            top_k=2,
+            moe_layer_indices=frozenset({0, 1}),
+        )
+        model.eval()
+        tokens = torch.randint(0, 64, (1, 5))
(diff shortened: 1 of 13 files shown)


jlarson4:
@Cacapice Could you please fix the formatting of your PR comment? I will review once that is done, thank you!

Cacapice:
@jlarson4 Good to go! Thank you for your review. I look forward to any feedback or discussion on the design decisions. Katherine

jlarson4:
Hi @Cacapice! Great work on this, an exceptionally well created adapter. I have just one small note related to naming:

Can we rename the `SwiGLUMoEForCausalLM` key to `TransformerLensPretrain`, matching the convention established by the `TransformerLensNative` key & `NativeArchitectureAdapter`. We want to make sure any custom architectures don't have a potential to conflict with future possible architecture keys, even though it is unlikely `SwiGLUMoEForCausalLM` would be chosen.

Cacapice:
Due to a local Git repository issue while recovering my branch history, I closed this PR and recreated it with the same changes on a branch correctly based on dev. The replacement PR is #1519. Thank you for the review and apologies for the extra churn.

jlarson4:
No worries at all @Cacapice, I will take a look at the new PR shortly.

PR #1527, opened 2026-07-21
Fix silent hook_normalized/hook_scale no-op on NormalizationBridge native-autograd path

### LINKED ISSUE #1526: [Bug Report] hook_normalized / hook_scale interventions and backward hooks are silently ignored on NormalizationBridge's native-autograd path
**Describe the bug**

When a `NormalizationBridge` is constructed with `use_native_layernorm_autograd=True`, its forward path (`_hf_autograd_forward_with_hooks` in `transformer_lens/model_bridge/generalized_components/normalization.py`) computes the values for `hook_scale` and `hook_normalized` inside a `torch.no_grad()` block and **discards the hooks' return values**:

```python
with torch.no_grad():
    ...
    x_normalized = x_centered * inv_rms
_ = self.hook_scale(scale)
_ = self.hook_normalized(x_normalized)
...
result = self.original_component(x)
```

The actual output always comes from `self.original_component(x)`. Two consequences, both silent:

1. **Forward interventions are discarded.** A hook that edits `hook_normalized` or `hook_scale` fires normally but has zero effect on model output. Anything built on patching these points — including the standard "freeze LN scale" linearization used in attribution patching — returns unpatched results with no error or warning.
2. **Backward hooks never fire.** The hook tensors are created under `no_grad`, so gradient-based attribution through these hook points silently produces nothing.

Caching/observation is unaffected (`ActivationCache.apply_ln_to_stack` etc. work fine) — which makes the failure harder to notice: the hooks *look* live.

The flag is hardwired in 13 adapters under `supported_architectures/`: `bart`, `bert`, `gemma3`, `gemma3_multimodal`, `gpt_oss`, `hubert`, `neox` (Pythia), `olmo`, `opt`, `phi`, `stablelm`, `t5gemma2`, `xglm`. On the default python-norm path (all other architectures) interventions and backward hooks work correctly, so the same user code behaves differently depending on which model is loaded.

**Code example**

Minimal repro against `dev` (`a041da3a`), CPU-only, no model download. The control (default path) passes; the native-autograd path fails both checks:

```python
import torch
import torch.nn as nn
from transformer_lens.model_bridge.generalized_components.normalization import NormalizationBridge

class Cfg:
    eps = 1e-5
    layer_norm_folding = False

def run(native: bool):
    torch.manual_seed(0)
    ln = nn.LayerNorm(8)
    bridge = NormalizationBridge(name="ln", config=Cfg(), use_native_layernorm_autograd=native)
    bridge.set_original_component(ln)
    x = torch.randn(2, 3, 8, requires_grad=True)

    baseline = bridge(x)
    bridge.hook_normalized.add_hook(lambda t, hook=None: torch.zeros_like(t))
    patched = bridge(x)
    bridge.hook_normalized.remove_hooks()
    edit_propagates = not torch.allclose(baseline, patched)

    bwd_fired = {}
    bridge.hook_normalized.add_hook(lambda g, hook=None: bwd_fired.setdefault("y", True) and g, dir="bwd")
    bridge(x).sum().backward()
    bridge.hook_normalized.remove_hooks()
    print(f"native={native}: edit propagates={edit_propagates}, bwd hook fired={'y' in bwd_fired}")

run(native=False)  # control:  edit propagates=True,  bwd hook fired=True
run(native=True)   # bug:      edit propagates=False, bwd hook fired=False
```

Output:

```
native=False: edit propagates=True, bwd hook fired=True
native=True: edit propagates=False, bwd hook fired=False
```

**System Info**

* Installed from source (`dev` branch, `a041da3a`)
* Linux, Python 3.11, torch 2.10.0, transformers 5.13.0, CPU

**Additional context**

The native path was introduced in #1101 for GPT-OSS dtype correctness, and its docstring is explicit that it exists "to preserve exact gradients" — so the observation-only design is understandable. The problem is that it is silent and undocumented: existing adapter tests only assert the flag is set, and no test covers intervention semantics on this path.

Possible directions, in increasing order of ambition:

1. **Warn or raise** when a hook attached to `hook_scale`/`hook_normalized` returns a modified value (or when a backward hook is attached) on this path — cheap, honest, stops silent wrong results.
2. **Fall back to the python-norm path when hooks are attached** to these points — preserves intervention semantics at the cost of slightly different numerics only while hooks are present.
3. Rework the native path to consume hook outputs (e.g. reconstruct the output from the hooked `x_normalized` and the module's weight/bias) — full fidelity, but touches the dtype/gradient guarantees the path exists for.

I'm happy to take this: I'd start with (1) plus a regression test that fails on current `dev` (checked via a clean worktree), and would like maintainer input on whether (2)/(3) are worth pursuing.

Two related minor observations in the same file (can be a separate PR if preferred):
- `_last_input_before_norm` (line 80) is written on every forward but never read anywhere in the repo since #1129 — it pins one `[batch, seq, d_model]` activation (and, when `requires_grad`, its upstream graph) per norm module between forwards.
- The `layer_norm_folding` config flag consumed at line 83 is never set `True` anywhere in the repo, so that branch appears unreachable in practice.

*This bug was found and this report drafted with the assistance of Claude (Anthropic's AI assistant), working under my direction; I've reviewed the analysis and the repro output is from a real run on my machine.*

### Checklist

Hetens:
# Description

Fixes #1526.

On the native-autograd path (`use_native_layernorm_autograd=True`, hardwired in 13 adapters incl. Gemma 3, NeoX/Pythia, OLMo, Phi, GPT-OSS), `NormalizationBridge` fired `hook_scale` / `hook_normalized` on detached observation values and discarded their return values: forward edits silently no-oped and backward hooks never fired, while the default python-norm path honored both.

Per the dispatch design agreed in #1526:

- **Backward hooks attached**: falls back to the python-norm path so hook tensors stay in the autograd graph.
- **Forward hook edits**: detected by object identity (a hook returning non-`None` replaces the tensor, matching `HookPoint` semantics); the output is reconstructed from the hooked values and stays grad-connected to the input. When `hook_scale` is edited, `hook_normalized` is recomputed from the edited scale first, mirroring the python path's ordering.
- **Observation-only hooks** (`run_with_cache` caching hooks return `None`) and hook-free forwards: unchanged, bit-identical native HF output and exact autograd. Guarded by a dedicated regression test that fails on any warning.
- Both fallbacks emit a `UserWarning` (the "solution 1" stopgap), since fallback numerics differ from HF's at float-rounding scale.

Known limitation (documented in a comment): a hook that mutates the tensor in place and returns `None` is not detected as an edit. Return the tensor from the hook to edit.

Based on `main` per maintainer request in #1526 (patch release). `normalization.py` is currently identical on `main` and `dev`, so this merges cleanly upward. The `_last_input_before_norm` dead-code cleanup from #1526 will follow as a separate PR to `dev`.

## Type of change

# Checklist:

Notes on the checklist:
- *No new warnings*: no warnings on any previously-working configuration; the two `UserWarning`s introduced fire only in the situations that previously silently produced wrong results, which is the point of the fix.
- *Docs*: no docs page covers these hook points' per-path semantics today; happy to add one if wanted.

## Validation

- New test file red on unfixed `main` (verified by swapping in `origin/main`'s `normalization.py`): `5 failed, 4 passed`. The 5 failures are exactly the bug-demonstrating tests; the 4 passes are the behavior-preservation guards.
- On this branch: all 9 new tests pass; full `tests/unit -m "not slow"` suite green: 3035 passed, 0 failed, 32 skipped, 9 xfailed (20 min, CPU).
- `make format` / `make check-format` clean, `uv run mypy .` clean (no `# type: ignore` added).

*Found, diagnosed, and fixed with the assistance of Claude (Anthropic's AI assistant), working under my direction; I reviewed the changes and the validation runs are from my machine.*

### DIFF
--- a/transformer_lens/model_bridge/generalized_components/normalization.py
+++ b/transformer_lens/model_bridge/generalized_components/normalization.py
@@ -1,11 +1,27 @@
 """Normalization bridge component implementation."""
-from typing import Any, Dict, Optional, cast
+import contextlib
+import warnings
+from typing import Any, ContextManager, Dict, Optional, cast
 
 import torch
 
 from transformer_lens.hook_points import HookPoint
 from transformer_lens.model_bridge.generalized_components.base import (
     GeneralizedComponent,
+)
+
+# The native-autograd path returns HF's own output, so hook edits and backward hooks
+# can only be honored by switching to the python-norm computation, whose numerics
+# differ from HF's at float-rounding scale.
+NATIVE_PATH_BWD_FALLBACK_WARNING = (
+    "Backward hooks on hook_scale/hook_normalized require grad-connected hook tensors; "
+    "falling back from the native-autograd path to the python-norm path. Output numerics "
+    "may differ from the unhooked forward at float-rounding scale."
+)
+NATIVE_PATH_EDIT_FALLBACK_WARNING = (
+    "A forward hook edited hook_scale/hook_normalized on the native-autograd path; the "
+    "output is reconstructed from the hooked values instead of HF's native forward. "
+    "Output numerics may differ from the unhooked forward at float-rounding scale."
 )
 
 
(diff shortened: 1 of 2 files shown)


transformer_lens/model_bridge/generalized_components/normalization.py
```diff
@@ -83,48 +99,69 @@ def forward(self, hidden_states: torch.Tensor, **kwargs: Any) -> torch.Tensor:
...
+            ).sqrt()
+        )
+        hidden_states = self.hook_normalized(hidden_states / scale)
+        return self._apply_weight_and_bias(hidden_states, input_dtype)
+
+    def _apply_weight_and_bias(
+        self, hidden_states: torch.Tensor, input_dtype: torch.dtype
+    ) -> torch.Tensor:
+        """Apply weight/bias in float32 before casting back (matches HF precision)."""
+        hidden_states = hidden_states * self.weight
```
jlarson4: @Hetens great work on this! Just this one small note from me.

The Gemma family of models applies an offset of +1 to their weight. Without a check for that here, we will produce incorrect results. Should be an easy fix, just add `w = (1.0 + self.weight) if getattr(self.config, "rmsnorm_uses_offset", False) else self.weight`. We will want to also add a test that covers this specific case.

Hetens: Good catch, done!

jlarson4:
Looks great, thank you for getting to this so quickly @Hetens! Merging now

PR #1529, opened 2026-07-22
Add config tests for DeepSeek V2 architecture adapter

SanjidMzi:
Part of #1302.

This PR adds unit tests for `DeepSeekV2ArchitectureAdapter` covering the config flags it
sets, the component-mapping structure and HF module paths, the `optional` flags
that distinguish the V2-full and V2-Lite attention paths, and the (empty)
weight-conversion set.

Behavioural coverage (forward pass, hook firing) already lives in
`tests/integration/model_bridge/test_deepseek_v2_adapter.py`; these tests
complement it at the unit tier, following the pattern of the existing
`test_granite_adapter.py` / `test_granite_moe_hybrid_adapter.py`.

`make format`, `uv run mypy .`, and `make unit-test` all pass locally.


jlarson4:
Looks good, thanks @SanjidMzi

PR #1530, opened 2026-07-22
Add lfm2 integration test

TensorCruncher:
# Description

Adds integration test for the recently added Lfm2 adapter. Modelled on the integration test file for NemotronH.

Let me know if I'm missing any tests / categories.

Do we need to add "LiquidAI/LFM2.5-230M" to the list of cached models for CI? Or is it okay to let it download each time since it's a relatively small model?

Related to  #1510 

## Type of change

### DIFF
--- a/tests/integration/model_bridge/test_lfm2_adapter.py
+++ b/tests/integration/model_bridge/test_lfm2_adapter.py
@@ -18,6 +18,8 @@ from transformer_lens.model_bridge.generalized_components import (
     RMSNormalizationBridge,
 )
 
+pytestmark = pytest.mark.slow
+
 MODEL = "LiquidAI/LFM2.5-230M"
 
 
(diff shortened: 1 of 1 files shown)


jlarson4:
Great tests @TensorCruncher! Thanks for putting this together.

The real 230M checkpoint is fine to keep, but we should mark it `@pytest.mark.slow`. This will allow it to run locally when I test PRs and when contributors run the tests before the submit, without bogging down CI with a minute+ download of a new model.

To cover this in CI without the big download, we'd typically use a tiny `from_config` test. You'll find a good example of that in `test_nemotron_h_tiny.py`. It builds a small model with random weights via `AutoModelForCausalLM.from_config`, so there's no Hub download and it runs in CI in a second or two. Could you add a `test_lfm2_tiny.py` alongside this one? Keep it small (a couple of layers so it covers both a `conv` and an `attention` layer) and it can carry the wiring, hook-coverage, and bridge-vs-HF forward checks.

Let me know if you have any questions!

jlarson4:
The docs are pretty vague on this specific point, I am going to update them to clarify this pattern for future adapter creators. Thanks again for putting these together!

PR #1531, opened 2026-07-22
Add config tests for DeepSeek V3 architecture adapter

SanjidMzi:
Part of #1302.

This pr adds unit tests for `DeepSeekV3ArchitectureAdapter` covering the config flags it
sets, the component-mapping structure and HF module paths, the MLA attention
submodule mapping, and the (empty) weight-conversion set.

V3 differs from V2 at the unit level: V3 always compresses Q via
the two-stage LoRA path (that's why there is no direct `q_proj` fallback and every
attention projection is required), `q_a_layernorm` is an RMSNorm rather than a
plain component, and the router `gate` is bridged as an optional component.

Behavioural coverage (forward pass, hook firing) already lives in
`tests/integration/model_bridge/test_deepseek_adapter.py`; these tests
complement it at the unit tier, following the same pattern as the existing
`test_deepseek_v2_adapter.py`.

`make format`, `uv run mypy .`, and `make unit-test` all pass locally.

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_deepseek_v3_adapter.py
+++ b/tests/unit/model_bridge/supported_architectures/test_deepseek_v3_adapter.py
@@ -1,0 +1,272 @@
+"""Unit tests for DeepSeekV3ArchitectureAdapter.
+
+Tests cover:
+- Config flags set by the adapter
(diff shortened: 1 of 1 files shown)


jlarson4:
Looks great! Thank you for putting this together @SanjidMzi!

PR #1533, opened 2026-07-22
Add Starcoder2 architecture adapter

SanjidMzi:
Adds a TransformerBridge adapter for Starcoder2 (`Starcoder2ForCausalLM`), which had no support (StarCoder1 / `gpt_bigcode` is already supported).

Starcoder2 is a Llama-shaped decoder (sequential pre-norm, GQA, rotary) that differs from Llama in three ways it shares with GPTBigCode: LayerNorm+bias instead of RMSNorm, a non-gated GELU MLP (`c_fc`/`c_proj`), and biases on all projections.

- Adapter + four-place registration (factory, `__init__`, model registry, report).
- Verified numerically: bridge logits match HuggingFace on `tiny-random-Starcoder2ForCausalLM` (max abs diff ~1e-7, 100% argmax agreement).
- Unit tests (config, component mapping, GQA weight/bias conversions) + an integration test asserting the forward pass matches HuggingFace.

`make format`, `uv run mypy .`, and unit + integration + registry tests pass locally.


jlarson4:
Hey @SanjidMzi! Thanks for putting this together. A Starcoder2 adapter was added independently from this, but I have rebased your PR `dev` and preserved the unique components of your adapter that will benefit TransformerLens. It'll be merged shortly

PR #1543, opened 2026-07-25
test(integration): add oracle parity test for JacobianLens (#1539 Tier-1)

### LINKED ISSUE #1539: [Proposal] Jacobian Lens follow-up: validation deliverables + deferred PR2 scope
### Proposal

The Jacobian lens vertical slice from #1505 has landed on `dev` (#1507, plus registry short-name resolution in #1537). This tracks the #1505 commitments that did **not** land with the first PR, in two tiers so the correctness items can ship independently:

- [ ] **Tier 1** — the three unrealized validation-plan checks.
- [ ] **Tier 2** — the second demo model + the explicitly-deferred PR2 scope.

### Motivation

#1505 committed to a five-point validation plan ("all numbers to be posted in the PR") and named PR2+ follow-ups. #1507 shipped the code, tests, and a demo. This issue covers all follow-ups and any trailing items that didn't make the first PR

##### Tier 1  – Validation Deliverables

- [x] External-oracle parity – Slow test comparing readouts against the official `anthropics/jacobian-lens` package on fixed prompts, with per-layer top-k overlap + rank-correlation thresholds. Current parity tests are self-consistency + load/validate only. Lock in the #1505 spike number: worst-case 7/8 top-8 overlap across 75 layer×prompt cells (top-1 identical in 70/75) on gemma-2-2b, using an overlap threshold rather than bitwise equality for bf16 near-ties. @mukund1985 working on this (#1543)
- [x] Kurtosis profile – Check that excess kurtosis of readout logits is roughly 0 in the first third of depth and rises through the workspace band. @rbellec working on this 
- [x] Causal swap success-rate – `swap_hooks` works and the demo shows rank-of-answer qualitatively, but there's no quantified success rate across an intermediate-layer band. Report France→China swap success honestly on an open-weights model. @emerardd working on this PR #1545 

##### Tier 2 – Demo + Remaining PR2 scope

- [x] Second demo model – The demo uses only `gemma-2-2b`; add the promised Bridge-only modern-arch demo (`Qwen3.5-4B`, the official-walkthrough model). PR #1547 
- [x] Community-format converters – Import fit-checkpoint files / other lens zoos into the `JacobianLens` artifact format. @priyanka25aug working on this
- [x] Gradient-pursuit sparse decomposition – J-space coordinate / occupancy analysis over fitted lenses. @janmenjayap working on this.
- [x] Community fitting guide – Short doc on fitting + `merge`-parallelizing new-model lenses. Task address by #1544

### Additional context

- Original proposal: #1505 · Implementation: #1507 · Registry follow-up: #1537
- Reference implementation (Apache-2.0, frozen): https://github.com/anthropics/jacobian-lens — the Tier 1 parity oracle
- Paper: Gurnee et al., *Verbalizable Representations Form a Global Workspace in Language Models*, Transformer Circuits Thread, 2026 — https://transformer-circuits.pub/2026/workspace/index.html (§4 kurtosis; Methods for the swap protocol)
- Published artifacts: https://huggingface.co/neuronpedia/jacobian-lens

### Checklist

mukund1985:
Closes #1539 Tier-1 item 1.

## What

Adds `tests/integration/test_jacobian_lens_oracle_parity.py` — a `@pytest.mark.slow` integration test that compares `TransformerBridge` `JacobianLens.readout()` against the reference `anthropics/jacobian-lens` oracle across 75 layer×prompt cells.

## Why

#1539 Tier-1 items require a test asserting numerical parity between the TL implementation and the oracle. The oracle is pinned to commit `581d398` (Apache-2.0, frozen) so the threshold is reproducible independent of upstream drift.

## How it works

The oracle (`jlens.lens.JacobianLens`) and TL's `JacobianLens` both load the same published checkpoint from `neuronpedia/jacobian-lens`. The oracle is installed at test time via pip from the pinned commit:

```
git+https://github.com/anthropics/jacobian-lens.git@581d398613e5602a5af361e1c34d3a92ea82ba8e
```

Both lenses run on `google/gemma-2-2b-it`. The oracle wraps `bridge.original_model` and `bridge.tokenizer` via `jlens.hf.from_hf()` — no second model copy is loaded.

## Pass criteria (per #1539 spec, matching #1505 spike numbers)

75 cells = 5 prompts × 15 sampled layers (0–24 plus final)

- Worst-case top-8 token overlap **>= 7/8** in every cell
- Spearman rank-correlation on top-64 logit union **>= 0.95** in every cell

Final layer (identity transport) is verified against oracle's `model_logits` rather than lens output, since the oracle raises if passed a layer not in `source_layers`.

## Test structure

```
tests/integration/test_jacobian_lens_oracle_parity.py
  oracle_package        fixture — pip-installs pinned oracle, returns (OracleJL, from_hf)
  gemma_bridge          fixture — loads google/gemma-2-2b-it as TransformerBridge
  tl_lens               fixture — JacobianLens.from_pretrained(neuronpedia/jacobian-lens)
  oracle_lens_and_model fixture — oracle JacobianLens + HFLensModel (shared HF model)
  test_jacobian_lens_oracle_parity  @pytest.mark.slow — runs all 75 cells, reports all failures
```

Spearman is computed inline with numpy (no scipy dependency).

### DIFF
--- a/tests/integration/test_jacobian_lens_oracle_parity.py
+++ b/tests/integration/test_jacobian_lens_oracle_parity.py
@@ -10,11 +10,12 @@ Pass criteria (per the #1539 Tier-1 spec, matching the #1505 spike numbers):
 The oracle is pinned to commit 581d398613e5602a5af361e1c34d3a92ea82ba8e
 (Apache-2.0) so the threshold is reproducible independent of upstream drift.
 
+Install the oracle before running this test:
+    uv pip install git+https://github.com/anthropics/jacobian-lens.git@<ORACLE_COMMIT>
+
 Marked @pytest.mark.slow; requires ~4 GB VRAM and network access.
 """
 
-import subprocess
-import sys
 from typing import List, Set
 
 import numpy as np
@@ -89,16 +90,19 @@ def _spearman_top_union(tl_logits: torch.Tensor, oracle_logits: torch.Tensor, k:
 
 @pytest.fixture(scope="module")
 def oracle_package():
-    """Install the pinned anthropics/jacobian-lens and return its key symbols."""
-    subprocess.run(
-        [sys.executable, "-m", "pip", "install", "--quiet", ORACLE_PKG],
-        check=True,
+    """Return oracle symbols; skips if the pinned package is not installed.
+
+    Install out-of-band before running:
+        uv pip install {ORACLE_PKG}
+    Or let .github/workflows/oracle-parity.yml handle it in CI.
+    """
+    jlens_hf = pytest.importorskip(
+        "jlens.hf",
+        reason=(f"requires the pinned oracle: uv pip install {ORACLE_PKG}"),
     )
-    # Imports must happen after install so they resolve to the just-installed pkg.
-    from jlens.hf import from_hf
     from jlens.lens import JacobianLens as OracleJL
 
-    return OracleJL, from_hf
+    return OracleJL, jlens_hf.from_hf
 
 
 @pytest.fixture(scope="module")
--- a/tests/integration/test_jacobian_lens_oracle_parity.py
+++ b/tests/integration/test_jacobian_lens_oracle_parity.py
@@ -1,0 +1,221 @@
+"""External-oracle parity test for TL JacobianLens.
+
+Compares TransformerBridge-based JacobianLens readouts against the reference
+anthropics/jacobian-lens oracle on google/gemma-2-2b-it.
+
+Pass criteria (per the #1539 Tier-1 spec, matching the #1505 spike numbers):
+  - Worst-case top-8 token overlap >= 7/8 across all 75 layer x prompt cells
+  - Spearman rank-correlation >= 0.95 on the top-64 logit union per cell
+
+The oracle is pinned to commit 581d398613e5602a5af361e1c34d3a92ea82ba8e
+(Apache-2.0) so the threshold is reproducible independent of upstream drift.
+
+Marked @pytest.mark.slow; requires ~4 GB VRAM and network access.
+"""
+
+import subprocess
+import sys
+from typing import Dict, List, Set, Tuple
+
+import numpy as np
+import pytest
+import torch
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+ORACLE_COMMIT = "581d398613e5602a5af361e1c34d3a92ea82ba8e"
+ORACLE_PKG = f"git+https://github.com/anthropics/jacobian-lens.git@{ORACLE_COMMIT}"
+
+LENS_REPO = "neuronpedia/jacobian-lens"
+LENS_REVISION = "a4114d7752d11eb546e6cf372213d7e75526d3a1"
+GEMMA_MODEL = "google/gemma-2-2b-it"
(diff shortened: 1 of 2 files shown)


mukund1985:
@jlarson4 when you get a chance — this is the external-oracle parity test for #1539 Tier-1 item 1. Pinned the oracle to 581d398 so the thresholds are reproducible. Let me know if you want the tolerance numbers adjusted or the prompt set changed.

jlarson4:
Awesome, thanks @mukund1985! I will make a look at this first thing Monday.

tests/integration/test_jacobian_lens_oracle_parity.py
```diff
@@ -0,0 +1,221 @@
...
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def oracle_package():
+    """Install the pinned anthropics/jacobian-lens and return its key symbols."""
+    subprocess.run(
```
jlarson4: Our test path is `uv run pytest`, `uv` venvs don't include pip. `python -m pip` at that point is "No module named pip", so `check=True` will make this fixture error. The convention for an optional external dep is `pytest.importorskip`. Installing at runtime also mutates the developer's environment with no cleanup.

Suggested replacement:
```
@pytest.fixture(scope="module")
def oracle_package():
    jlens_hf = pytest.importorskip(
        "jlens.hf",
        reason=(
            "requires the pinned oracle: "
            "uv pip install {ORACLE_PKG}"
        ),
    )
    from jlens.lens import JacobianLens as OracleJL
    return OracleJL, jlens_hf.from_hf
```
We should install the oracle out-of-band, either as a CI step, or a dedicated test dependency group in `pyproject` so that `uv sync --group <name>` installs it. Keep `ORACLE_COMMIT` as the single source of truth for the pin (reference it in the CI install too).

jlarson4 (commented):
Hi @mukund1985! Thanks for putting this together, just one small note on the `jlens` package import in the test file. Let me know if you have any questions

tests/integration/test_jacobian_lens_oracle_parity.py
mukund1985: Good catch. Addressed in this push:

- `oracle_package` now uses `pytest.importorskip("jlens.hf", ...)` — skips gracefully if the oracle is absent, no pip install at test time.
- `import subprocess` and `import sys` removed.
- Added `.github/workflows/oracle-parity.yml` — reads `ORACLE_COMMIT` from the test file at CI time (single source of truth) and runs `uv pip install git+...@${ORACLE_COMMIT}` before the slow suite. Triggers on `workflow_dispatch` and on pushes that touch the test or workflow file, keeping oracle runs opt-in for PR checks.

mukund1985:
Pushed a small follow-up to address your note on the fixture import:

- Replaced the `subprocess.run` pip-install in `oracle_package` with `pytest.importorskip("jlens.hf", ...)` — the test now skips cleanly if the oracle isn't installed rather than failing on pip at runtime
- Added `.github/workflows/oracle-parity.yml` — a dedicated workflow that installs the oracle via `uv pip install` (reads `ORACLE_COMMIT` from the test file directly so there's a single source of truth) and then runs the marked test. Kept the slow test out of the standard PR gate.

Let me know if anything else needs tweaking.

jlarson4:
Great work @mukund1985, merging now

PR #1558, opened 2026-07-30
fix(bridge): return W_in/W_out/W_gate in TL orientation for nn.Linear-backed models

### LINKED ISSUE #1554: [Bug Report] Bridge W_in/W_out returned in torch orientation on nn.Linear-backed models
**Describe the bug**
On `nn.Linear`-backed (native) `TransformerBridge` models, `W_in`/`W_out`/`W_gate` come back transposed relative to TL convention, while `Conv1D`-backed models (GPT-2) return TL orientation. Neuron-level analysis migrated off deprecating `HookedTransformer` (Grokking/Othello/superposition demos read `.W_in`/`.W_out`) silently computes garbage.

**Code example**
```python

ht = HookedTransformer.from_pretrained("gpt2")  # Conv1D
br = TransformerBridge.boot("some-native-nn-linear-model")  # nn.Linear
print(ht.W_in.shape)  # [n_layers, d_model, d_mlp]
print(br.W_in.shape)  # [n_layers, d_mlp, d_model]  <-- transposed
```

**System Info**

`W_in`/`W_out`/`W_gate` should return TL orientation regardless of backing module. Add a reshape callback at `transformer_lens/model_bridge/transformer_bridge.py:971-986` (W_in `:971-974`, W_gate `:976-981`, W_out `:983-986`) that transposes `nn.Linear` weights and no-ops for `Conv1D`, mirroring `_reshape_qkv`/`_reshape_o` (`:937-949`). Mirror HT MLP orientation in `transformer_lens/components/mlps/`.

The better mirror is the attention path's layout awareness: `AttentionBridge._weight_layout_in_out` (`attention.py:546-558`) branches on Conv1D vs `nn.Linear` and feeds `_reshape_weight_to_3d` (`attention.py:560-594`); and `_stack_block_params` already accepts the needed `reshape_fn` hook (`transformer_bridge.py:887`, applied `:926-927`), so the fix is a small helper plus one line per property.

Acceptance:
- [ ] Native and GPT-2 bridges return matching TL orientation for `W_in`/`W_out`/`W_gate`
- [ ] Orientation matches equivalent `HookedTransformer`
- [ ] Strengthen guard to assert inner dims at `tests/acceptance/model_bridge/compatibility/test_legacy_hooked_transformer_coverage.py:169-179`; that block is wrapped in `try/except AttributeError: pass` (`:179-180`), which silently swallows failures — worth fixing in the same PR
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` clean

**Additional context**
Size: M. No dependencies.

### Checklist

MdSadiqMd:
# Description

On nn.Linear-backed (native) TransformerBridge models, `W_in`/`W_out`/`W_gate` were returned transposed relative to TL convention, while Conv1D-backed models (GPT-2) returned TL orientation. Neuron-level analysis migrated from HookedTransformer (Grokking/Othello/superposition demos read `.W_in`/`.W_out`) silently computed garbage.

**Root cause**: `TransformerBridge.W_in` called `_stack_block_params("mlp.W_in")` without a reshape callback, unlike attention weights which pass `_reshape_qkv`/`_reshape_o`. MLP weights passed through raw from `original_component.weight`.

**Fix**: Add `_weight_layout_in_out()` helper to MLPBridge (same pattern as AttentionBridge), replace property_aliases with actual `@property` methods that transpose nn.Linear weights to match TL convention.

| Model Type | Before | After |
|------------|--------|-------|
| Pythia (nn.Linear) | `[n_layers, d_mlp, d_model]` | `[n_layers, d_model, d_mlp]` |
| GPT-2 (Conv1D) | `[n_layers, d_model, d_mlp]` | `[n_layers, d_model, d_mlp]` |

Fixes #1554 

## Type of change

# Checklist:


MdSadiqMd:
Hey @jlarson4, the pipeline passed. Please review it when you're free

jlarson4:
Will do, I probably will not have time to review it today but will get reviewing all of these new PRs first thing tomorrow

MdSadiqMd:
Sure, thank you for maintaining this repo

jlarson4:
Of course! Happy to do it!

Full review coming shortly

transformer_lens/model_bridge/generalized_components/mlp.py
```diff
@@ -81,3 +80,55 @@ def forward(self, *args, **kwargs) -> Any:
...
+    def _weight_layout_in_out(self, proj: Any) -> Optional[bool]:
+        """Whether proj's wrapped module stores its weight as [in, out].
+
+        Conv1D (GPT-2 style) stores [in_features, out_features]; nn.Linear stores
+        [out_features, in_features]. Returns None when the wrapped module is
+        neither, so callers can fall back to a shape heuristic.
+        """
+        from transformers.pytorch_utils import Conv1D
+
+        component = getattr(proj, "original_component", None)
```
jlarson4, line 93: All three properties branch only on `layout is False` (:108, :120, :132), so `None` unintentionally means "already TL-oriented".The doctoring mentions a shape fallback, it would be good to implement that here.`attention.py:578-591` implements a similar pattern and may work as a reference.

tests/acceptance/model_bridge/compatibility/test_legacy_hooked_transformer_coverage.py
```diff
@@ -166,17 +166,18 @@ def test_weight_access_functionality(self, bridge_model):
...
-            assert w_out.shape[0] == bridge_model.cfg.n_layers
-
-            weight_tests.append("mlp_weights")
-
-        except AttributeError:
-            pass  # Weight access might not be implemented yet
+        w_in = bridge_model.W_in
+        w_out = bridge_model.W_out
+
+        assert w_in.shape[0] == bridge_model.cfg.n_layers
```
jlarson4, line 172: The acceptance test you strengthened only ever runs against `gpt2`, which is `Conv1D`, so it does not fail if the fix is reverted. Could you add a unit test pinning `W_in`/`W_gate`/`W_out` against an `nn.Linear`-backed MLP by checking that `resid @ W_in` reproduces what the wrapped projection computes?

jlarson4 (commented):
This looks great! Thanks for tackling it @MdSadiqMd. Just a couple small changes listed below

One side effect of the fix I noticed: but `get_params()` now disagrees with `bridge.W_in` on the same object. `get_params_util.py:128-129` and `:148` still return raw `.weight` (measured `(64,32)` vs `(32,64)`). It should be fixable by sourcing from the properties you just setup.

MdSadiqMd:
Hey @jlarson4 implemented the changes. Let me know if anything is required from my side

transformer_lens/model_bridge/generalized_components/mlp.py
```diff
@@ -81,3 +80,74 @@ def forward(self, *args, **kwargs) -> Any:
...
+    ) -> torch.Tensor:
+        """Normalize MLP weight to TL orientation.
+
+        Args:
+            weight: 2D weight tensor from the projection
+            layout: True if [in, out] (Conv1D), False if [out, in] (nn.Linear),
+                None falls back to shape heuristic.
+            pattern: "in" for W_in/W_gate [d_model, d_mlp], "out" for W_out [d_mlp, d_model]
+        """
+        if layout is None:
```
jlarson4, line 116: GIDD's `ScaledLinear` is a bare `nn.Module` and the size-ordering guess silently mis-orients whenever `d_mlp < d_model`. We assumed "typical" here, but untested. Could this key off the module's declared `in_features`/`out_features` instead? If not, can we at least document the blind spot?

transformer_lens/model_bridge/generalized_components/mlp.py
```diff
@@ -81,3 +80,74 @@ def forward(self, *args, **kwargs) -> Any:
...
+
+    @property
+    def W_in(self) -> torch.Tensor:
+        """MLP input weight in TL orientation [d_model, d_mlp]."""
+        in_module = getattr(self, "in", None)
+        if in_module is None:
+            raise AttributeError("No 'in' submodule on this MLP bridge")
+        weight = in_module.weight
+        layout = self._weight_layout_in_out(in_module)
+        return self._normalize_mlp_weight(weight, layout, pattern="in")
```
jlarson4: `bridge.get_params()["blocks.0.mlp.W_in"]` still returns `[d_mlp, d_model]` while this property now returns `[d_model, d_mlp]`. Could the params path reuse these accessors, the way it already reshapes KV weights in `get_params_util.py` on line 9Z?

jlarson4 (commented):
Two more small comments on this one. Should be good after these are resolved

PR #1574, opened 2026-07-30
support loading fit checkpoints in JacobianLens.load()

priyanka25aug:
## What this does

Extends `JacobianLens.load()` to accept fit checkpoints (files saved by the fitting pipeline with a `jacobian_sum` key) in addition to the existing artifact format (files with a `J` key).

The conversion lives in a new `_from_checkpoint_payload()` classmethod:
- Divides the running Jacobian sums by `n_prompts` to recover the per-prompt mean
- Harvests safe scalar provenance keys (`model_name`, `model_revision`, `corpus`) from the flat payload namespace into `metadata`
- Strips fit-reserved keys (`transformer_lens_fit`, `transformer_lens_version`, `model_system`, etc.) so they don't leak into the converted lens
- Drops tensor-valued metadata fields that can't survive `weights_only=True` reload, recording their names and shapes in `dropped_fields` for transparency
- Sets `converted_from: "jacobian_lens_checkpoint"` so `merge()` naturally refuses to mix converted and natively TL-fitted lenses (provenance keys differ)

Raises `ValueError` with a clear message if the file has neither key, or if `n_prompts <= 0`. Tuned-lens support deferred.

## Files changed

- `transformer_lens/tools/analysis/jacobian_lens.py` — new `_FIT_RESERVED_KEYS` and `_CHECKPOINT_FLAT_PROVENANCE` frozensets; updated `load()`; new `_from_checkpoint_payload()`
- `tests/unit/tools/test_jacobian_lens_import.py` — unit tests on synthetic fixtures covering artifact regression, checkpoint round-trip, dtype preservation, `n_prompts=0` guard, tensor field dropping, fit-key stripping, and merge provenance rejection; no model/oracle dependency
- `docs/source/content/jacobian_lens_fitting.md` — "Importing an existing lens" section added

## Type of change

## Checklist

## Testing

```bash
uv run pytest tests/unit/tools/test_jacobian_lens_import.py -v
```

### DIFF
--- a/docs/source/content/jacobian_lens_fitting.md
+++ b/docs/source/content/jacobian_lens_fitting.md
@@ -251,18 +251,25 @@ means automatically:
 lens = JacobianLens.load("path/to/checkpoint.pt")
 ```
 
-The expected top-level keys are:
+The reference implementation (`anthropics/jacobian-lens`) writes exactly six
+top-level keys. `load()` also accepts optional flat-provenance keys from alternative
+writers. The full set of recognised keys is:
 
 | Key | Type | Required | Description |
 |---|---|---|---|
 | `jacobian_sum` | `dict[int, Tensor[d, d]]` | yes | Running sum of per-prompt Jacobians (float32) |
-| `n_prompts` | `int` | yes | Number of prompts accumulated into the sums (must be > 0) |
-| `d_model` | `int` | yes | Residual stream width |
+| `n_done` | `int` | yes | Number of prompts accumulated into the sums (must be > 0) |
+| `next_idx` | `int` | no | Index of the next prompt to process (informational; not used by `load()`) |
 | `source_layers` | `list[int]` | no | Documented layer indices (informational only) |
-| `metadata` | `dict[str, ...]` | no | Nested provenance (scalars, lists, string-keyed dicts) |
+| `target_layer` | `int` | no | Target layer fitted against — harvested into `metadata` so `validate_model()` can check it |
+| `skip_first` | `int` | no | Leading positions excluded from the source average (informational; not used by `load()`) |
 | `model_name` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
 | `model_revision` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
 | `corpus` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
+| `metadata` | `dict[str, ...]` | no | Nested provenance from alternative writers (scalars, lists, string-keyed dicts) |
+
+`d_model` is **not** read from the payload; it is inferred from the shape of the
+first matrix in `jacobian_sum`.
 
 `load()` always sets `converted_from: "jacobian_lens_checkpoint"` in the resulting
 lens's metadata. This sentinel prevents `merge()` from silently combining a
@@ -286,6 +293,9 @@ Three things are guaranteed when loading a checkpoint:
 TransformerLens internals (`transformer_lens_fit`, `transformer_lens_version`,
 `model_system`, `hook_convention`, etc.) are not carried over, since they describe
 the *native* fitting pipeline and would be wrong for an imported lens.
+`target_layer` is a deliberate exception: it is preserved in the resulting metadata
+so `validate_model()` can detect and reject checkpoints fitted against a non-final
+target layer.
 
 **Tensor-valued fields are dropped.** `JacobianLens.save()` uses
 `weights_only=True` for reload safety, which restricts metadata to plain Python
--- a/docs/source/content/jacobian_lens_fitting.md
+++ b/docs/source/content/jacobian_lens_fitting.md
@@ -232,3 +232,82 @@ lens = JacobianLens.from_pretrained(
 To propose a short-name entry in TransformerLens, open a pull request that adds the
 published file to `transformer_lens/tools/analysis/jacobian_lens_registry.json` and
 include the fitting provenance and validation results.
+
+## Importing an existing lens
+
+`JacobianLens.load()` accepts two file schemas: the standard artifact format (four
+official keys — `J`, `n_prompts`, `source_layers`, `d_model`) written by
+`JacobianLens.save()` and compatible with the Anthropic reference package, and the
+fit-checkpoint format described below.
+
+### Fit checkpoint format
+
+A fit checkpoint holds the running Jacobian sums from an interrupted or staged fit,
+before the final per-prompt average is applied. `JacobianLens.load()` detects a
+checkpoint by the presence of a `jacobian_sum` key and reconstructs the per-layer
+means automatically:
+
+```python
+lens = JacobianLens.load("path/to/checkpoint.pt")
+```
+
+The expected top-level keys are:
+
+| Key | Type | Required | Description |
+|---|---|---|---|
+| `jacobian_sum` | `dict[int, Tensor[d, d]]` | yes | Running sum of per-prompt Jacobians (float32) |
+| `n_prompts` | `int` | yes | Number of prompts accumulated into the sums (must be > 0) |
+| `d_model` | `int` | yes | Residual stream width |
+| `source_layers` | `list[int]` | no | Documented layer indices (informational only) |
+| `metadata` | `dict[str, ...]` | no | Nested provenance (scalars, lists, string-keyed dicts) |
+| `model_name` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
+| `model_revision` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
+| `corpus` | `str` | no | Flat provenance shortcut (harvested into `metadata`) |
+
+`load()` always sets `converted_from: "jacobian_lens_checkpoint"` in the resulting
+lens's metadata. This sentinel prevents `merge()` from silently combining a
+converted checkpoint with a natively TL-fitted lens, since the two provenance
+dictionaries will differ. To merge shards that were all loaded from checkpoints, they
+must otherwise share identical provenance (same `model_name`, `corpus`, etc.):
+
+```python
+shards = [JacobianLens.load(str(p)) for p in checkpoint_paths]
+# All shards get converted_from="jacobian_lens_checkpoint";
+# merge() requires identical provenance apart from n_prompts.
+merged = JacobianLens.merge(shards)
+merged.save("merged_lens.pt")
+```
+
+### Metadata handling
+
+Three things are guaranteed when loading a checkpoint:
+
+**Fit-reserved keys are stripped.** Keys written by `JacobianLens.fit()` to record
+TransformerLens internals (`transformer_lens_fit`, `transformer_lens_version`,
+`model_system`, `hook_convention`, etc.) are not carried over, since they describe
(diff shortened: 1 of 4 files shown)


priyanka25aug:
Hi @jlarson4 — all format, type, docstring, and benchmark checks are passing. The remaining long-running jobs (compatibility across Python 3.10/3.11/3.12 and full coverage) are still in progress but looking clean so far.

Would you mind taking a look when you get a chance? Happy to make any changes based on your feedback.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -266,28 +287,121 @@ def save(self, path: str, *, dtype: torch.dtype = torch.float16) -> None:
...
+    @classmethod
+    def _from_checkpoint_payload(cls, path: str, payload: Dict[str, Any]) -> "JacobianLens":
+        """Reconstruct a JacobianLens from a fit-checkpoint payload.
+
+        Divides the running Jacobian sums by ``n_prompts``, strips fit-reserved
+        provenance keys, harvests safe scalar metadata, records dropped tensor
+        fields, and marks the result as converted so :meth:`merge` refuses to
+        mix it with natively TL-fitted lenses.
+        """
+        n_prompts = int(payload.get("n_prompts", 0))
```
jlarson4: The reference package's `write_checkpoint()` stores the count as `n_done`, so genuine checkpoints are rejected here with `n_prompts=0`. Could the loader read the key set that writer actually emits, as the pre-existing fixture does (`tests/unit/tools/test_jacobian_lens.py:507`)?

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -121,6 +121,27 @@ def _resolve_registry_entry(name_or_path: str) -> Optional[Tuple[str, str]]:
 _SWAP_WARN_COSINE = 0.99
 _SWAP_ERROR_COSINE = 0.999
 
+# Keys written by fit() that must not appear in converted-lens metadata so that
+# merge() can refuse to mix TL-fitted lenses with externally converted ones.
+_FIT_RESERVED_KEYS: frozenset = frozenset(
```
jlarson4, line 129: Real checkpoints can record a non-final `target_layer`, but this conversion discards the key, so the target-convention refusal in `validate_model` silently passes (`jacobian_lens.py:605`). Could the conversion preserve `target_layer`, or refuse checkpoints recording a non-final target?

tests/unit/tools/test_jacobian_lens_import.py
```diff
@@ -0,0 +1,325 @@
...
+    metadata=None,
+    flat_provenance=None,
+    n_prompts_override=None,
+):
+    """Write a minimal fit-checkpoint (jacobian_sum format) to *path*.
+
+    *n_prompts_override* lets callers inject a bad value without changing the
+    sums (useful for zero/negative n_prompts tests).
+    """
+    payload = {
```
jlarson4, line 44: These fixtures encode the loader's own invented schema, so the suite continues to pass while real reference checkpoints fail to load. Once the schema is fixed, could one fixture mirror the reference `write_checkpoint()` payload key-for-key so format drift causes a failure?

docs/source/content/jacobian_lens_fitting.md
```diff
@@ -232,3 +232,82 @@ lens = JacobianLens.from_pretrained(
...
+    print("fields not carried over:", lens.metadata["dropped_fields"])
+```
+
+**Dtype is not preserved.** `JacobianLens.__init__` casts all matrices to float32
+and `save()` defaults to float16, so the source checkpoint's storage dtype cannot
+be preserved end-to-end. Value precision is preserved within those two conversions.
+
+### Note on tuned-lens
+
+Tuned-lens checkpoints are not currently supported. The tuned-lens format has no
```
jlarson4: Tuned-lens translators are affine and carry the bias. It is the Jacobian artifact format that has no bias slot to receive it. Can we ensure this detail is included as part of this note?

jlarson4 (commented):
Hi @priyanka25aug! Thanks for taking this on. Checkpoint import is a genuinely useful capability, and the conversion math, artifact-path safety, and merge sentinel are all done right.

A couple comments below that we should address before merging

docs/source/content/jacobian_lens_fitting.md
priyanka25aug: Fixed in `08738f86`. The note now reads: "Tuned-lens translators are affine (weight + bias), but the Jacobian artifact format has no bias slot to receive the translation component; importing would therefore silently drop the bias and produce wrong-but-plausible readouts." Thanks for catching the direction — I had it backwards.

transformer_lens/tools/analysis/jacobian_lens.py
priyanka25aug: Fixed in `08738f86` — removed `target_layer` from `_FIT_RESERVED_KEYS` with an explanatory comment. It now survives checkpoint conversion into the lens metadata, so `validate_model()` will correctly raise for any checkpoint that recorded a non-final target layer rather than silently passing.

tests/unit/tools/test_jacobian_lens_import.py
priyanka25aug: Fixed in `08738f86` — added `test_load_checkpoint_mirrors_fit_payload_schema` which uses the exact payload layout `fit()` produces: `jacobian_sum`, `n_prompts`, `d_model`, flat provenance keys (`model_name`, `model_revision`, `corpus`), and nested metadata containing `transformer_lens_fit`, `target_layer`, and other fit-reserved fields. The test asserts that `target_layer` is preserved (not stripped), all fit-reserved keys are gone from the converted lens, and `converted_from` is set — so any schema drift in the real payload will cause a failure.

priyanka25aug:
@jlarson4 All four of your review comments have been addressed in `08738f86` and `7aec2e7b`:

1. **`target_layer` preservation** — removed from `_FIT_RESERVED_KEYS` so it survives checkpoint conversion and `validate_model()` correctly raises for non-final-target checkpoints.
2. 2. **Reference fixture** — added `test_load_checkpoint_mirrors_fit_payload_schema` using the exact payload layout `fit()` produces, including flat provenance keys and nested fit-reserved metadata, with assertions that verify `target_layer` is preserved and fit-reserved keys are stripped.
3. 3. **Tuned-lens note** — corrected the direction: tuned-lens translators are affine (weight + bias), the Jacobian artifact format has no bias slot to receive the translation component.
4. 4. **`n_done` key** — `_from_checkpoint_payload` now reads `payload.get("n_done", payload.get("n_prompts", 0))`, preferring `n_done` (the key the reference writer emits) with `n_prompts` as a fallback so genuine checkpoints aren't silently rejected.
All threads are resolved. Could you please re-review and merge when you're happy? Thank you!

transformer_lens/tools/analysis/jacobian_lens.py
priyanka25aug: Fixed in `7aec2e7b` — `_from_checkpoint_payload` now reads `payload.get("n_done", payload.get("n_prompts", 0))`, preferring `n_done` (the key the reference writer emits) with `n_prompts` as a fallback. Good catch — genuine checkpoints would have been silently rejected with `n_prompts=0`.

jlarson4:
@priyanka25aug There is a CI failure that needs it be addressed still it looks like. The failure is an indicator of the remaining bug. `n_done` was one of three keys the reference writer disagrees with us about. Real checkpoints also carry no `d_model` (derivable from any `jacobian_sum` matrix's shape) and put `target_layer` at the top level rather than inside a nested `metadata` dict. With only the count fixed, a genuine checkpoint still cannot load, it just fails one line later with an unhandled `KeyError`. Worth opening `write_checkpoint()` (`jlens/fitting.py:315-327` in `anthropics/jacobian-lens@581d398`) and working from its six keys directly rather than from my comments one at a time.

If the intent was a *TransformerLens* checkpoint format rather than the reference one, say so and we'll take the other fork: that would mean teaching `fit()` to write it (`checkpoint_path` / `checkpoint_every`, resumable), because nothing in either codebase emits the documented layout. The PR docstring describes the running-sum file "written during or after a fitting run", which reads as the reference format, so that is what I was assuming.

Suggested order:

1. Derive `d_model` from a `jacobian_sum` matrix's shape rather than requiring the key, and harvest top-level `target_layer` alongside the flat provenance keys.
2. Then make one fixture the verbatim six-key payload and assert both that it loads and that a checkpoint recording a non-final `target_layer` is refused by `validate_model()`.
3. `test_load_checkpoint_with_zero_n_prompts_raises` needs a decision rather than a patch: its `jacobian_sum` is empty, so it can never load however the schema lands. Give it a matrix and let it assert the round trip, and test the non-positive-count guard on its own synthetic file. Whichever error the empty-sums case raises, it should name the empty sums, rather than using `n_prompts=0` or a  bare `KeyError`.
4. Update the docs schema table to the real key set – as written it documents `n_prompts`/`d_model` as required, which sends readers off to hand-build files.
5. Small doc gap from `08738f86`: the "Metadata handling" section still says keys written by `fit()` "are not carried over ... (`transformer_lens_fit`, `transformer_lens_version`, `model_system`, `hook_convention`, etc.)". `target_layer` is now a deliberate exception and documented in the code comment, but not the doc.

priyanka25aug:
Hi @jlarson4 — all five points from your second review are addressed in commit ff6cef73. Here's a summary of what changed:

1. **`n_done` key support**: `_from_checkpoint_payload()` now reads `n_done` first (the real key written by `write_checkpoint()`), falling back to `n_prompts` for backward compatibility.
2. **`d_model` inferred from matrix shape**: Removed the `payload["d_model"]` lookup (which would KeyError on real checkpoints). `d_model` is now derived from `first_matrix.shape[0]` of the first entry in `jacobian_sum`, with an explicit error if `jacobian_sum` is empty.
3. **Top-level `target_layer` harvested**: Added a targeted harvest of the top-level `target_layer` key (where the reference implementation writes it) into metadata, so `validate_model()` can check the fitting target.
4. **Tests updated to use the verbatim 6-key reference schema**: Replaced the old test (which had a structurally wrong payload) with:
5.    - `test_load_checkpoint_with_zero_n_done_raises` — checks `n_done=0` triggers the `n_prompts=0` error
6.    - `test_load_checkpoint_with_empty_jacobian_sum_raises` — checks empty `jacobian_sum` triggers the new error
7.    - `test_load_checkpoint_mirrors_fit_payload_schema` — rewritten to use the exact 6-key payload `write_checkpoint()` produces
8.    - `test_load_checkpoint_harvests_flat_provenance_and_strips_fit_keys` — new test covering flat-provenance harvest and fit-key stripping
5. **Docs updated**: Schema table in `jacobian_lens_fitting.md` now reflects the real 6-key format (`n_done`, `next_idx`, `target_layer`, `skip_first`); added a note that `d_model` is inferred from the matrix shape; added a paragraph noting `target_layer` is a deliberate exception to fit-key stripping.
CI is all green — the only failure is the pre-existing `Activation_Patching_in_TL_Demo` notebook flake, unrelated to this PR. Ready for re-review when you get a chance!

jlarson4:
Thanks for covering all that @priyanka25aug, great work! Merging now

PR #1576, opened 2026-07-31
docs: migrate executable doctests to TransformerBridge

### LINKED ISSUE #1565: [Proposal] Rewrite executable doctests that construct HookedTransformer onto TransformerBridge
### Proposal
Base on `dev-4.x`. Rewrite the executable `>>>` doctests that build a `HookedTransformer` so they construct a `TransformerBridge` instead, in the modules that keep running doctests after `HookedTransformer` is removed: `transformer_lens/ActivationCache.py` (4 HT-constructing doctests), `transformer_lens/evals.py` (2 HT-constructing doctests, only 1 executable at `:326-329` — the other is already `# doctest: +SKIP` at `:554-559`), `transformer_lens/utilities/exploratory_utils.py` (1).

### Motivation
`HookedTransformer` is being deprecated in favour of `TransformerBridge`. `ActivationCache`, `evals`, and `exploratory_utils` are staying, but their doctests still do `HookedTransformer.from_pretrained(...)`. `pytest` runs these (`pyproject.toml` enables `--doctest-modules` over `transformer_lens/`), so they will hard-fail the moment `HookedTransformer` is removed. Migrating them now keeps the examples current and unblocks removal.

### Pitch
- Replace `HookedTransformer.from_pretrained("gpt2"...)` with `TransformerBridge.boot_transformers("gpt2")` (add `.enable_compatibility_mode()` where the example relies on HookedTransformer-exact hook names / numerics).
- Update expected doctest output to whatever the bridge prints, and any `# doctest: +SKIP` markers that are no longer needed. Note every executable example also asserts the literal output line `Loaded pretrained model ... into HookedTransformer`, so each rewrite must update that expected-output line too.
- Keep each example's teaching intent identical; only the model construction changes.

Acceptance:
- [ ] No doctest in `ActivationCache.py`, `evals.py`, `utilities/exploratory_utils.py` constructs `HookedTransformer`
- [ ] `make docstring-test` (or `pytest --doctest-modules transformer_lens/`) passes
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

Mirror: existing bridge examples in `docs/source/content/` and the `boot_transformers` / `enable_compatibility_mode` usage in `tests/unit/model_bridge/`.

### Checklist

Austin1serb:
# Description

Migrates executable doctests in `ActivationCache`, `evals`, and `exploratory_utils` from the deprecated `HookedTransformer` loader to `TransformerBridge`.

The examples now use official Hugging Face model IDs and enable compatibility mode where their existing numerics and legacy hook aliases require it. Obsolete loader-output assertions were removed, and cache-key examples were updated to reflect the bridge's canonical hooks alongside compatibility aliases.

The skipped MMLU example was also migrated so no doctest in the three surviving modules constructs `HookedTransformer`.

Fixes #1565

## Type of change

# Checklist

- [x] No additional explanatory source comments were required
- [x] The executable documentation was updated
- [x] The focused and full docstring suites pass locally
- [x] New and existing unit tests pass locally

## Testing

- `uv run pytest --doctest-modules transformer_lens/ActivationCache.py transformer_lens/evals.py transformer_lens/utilities/exploratory_utils.py -q` (`9 passed`, `1 skipped`)
- `make docstring-test` (`17 passed`, `24 skipped`)
- `make unit-test` (`5139 passed`, `61 skipped`, `44 deselected`, `10 xfailed`)
- `uv run mypy .`
- `make check-format`
- `git diff --check`

### DIFF
--- a/transformer_lens/ActivationCache.py
+++ b/transformer_lens/ActivationCache.py
@@ -71,9 +71,9 @@ class ActivationCache:
     the model predicting "road". This kind of analysis commonly falls under the category of "logit
     attribution" or "direct logit attribution" (DLA).
 
-    >>> from transformer_lens import HookedTransformer
-    >>> model = HookedTransformer.from_pretrained("tiny-stories-1M")
-    Loaded pretrained model tiny-stories-1M into HookedTransformer
+    >>> from transformer_lens.model_bridge import TransformerBridge
+    >>> model = TransformerBridge.boot_transformers("roneneldan/TinyStories-1M")
+    >>> model.enable_compatibility_mode()
 
     >>> _logits, cache = model.run_with_cache("Why did the chicken cross the")
     >>> residual_stream, labels = cache.decompose_resid(return_labels=True, mode="attn")
@@ -270,12 +270,12 @@ class ActivationCache:
 
         Examples:
 
-            >>> from transformer_lens import HookedTransformer
-            >>> model = HookedTransformer.from_pretrained("tiny-stories-1M")
-            Loaded pretrained model tiny-stories-1M into HookedTransformer
+            >>> from transformer_lens.model_bridge import TransformerBridge
+            >>> model = TransformerBridge.boot_transformers("roneneldan/TinyStories-1M")
+            >>> model.enable_compatibility_mode()
             >>> _logits, cache = model.run_with_cache("Some prompt")
             >>> list(cache.keys())[0:3]
-            ['hook_embed', 'hook_pos_embed', 'blocks.0.hook_resid_pre']
+            ['embed.hook_in', 'hook_embed', 'embed.hook_out']
 
         Returns:
             List of all keys.
@@ -306,16 +306,16 @@ class ActivationCache:
 
         Examples:
 
-            >>> from transformer_lens import HookedTransformer
-            >>> model = HookedTransformer.from_pretrained("tiny-stories-1M")
-            Loaded pretrained model tiny-stories-1M into HookedTransformer
+            >>> from transformer_lens.model_bridge import TransformerBridge
+            >>> model = TransformerBridge.boot_transformers("roneneldan/TinyStories-1M")
+            >>> model.enable_compatibility_mode()
             >>> _logits, cache = model.run_with_cache("Some prompt")
             >>> cache_interesting_names = []
             >>> for key in cache:
             ...     if not key.startswith("blocks.") or key.startswith("blocks.0"):
             ...         cache_interesting_names.append(key)
             >>> print(cache_interesting_names[0:3])
-            ['hook_embed', 'hook_pos_embed', 'blocks.0.hook_resid_pre']
+            ['embed.hook_in', 'hook_embed', 'embed.hook_out']
 
         Returns:
             Iterator over the cache.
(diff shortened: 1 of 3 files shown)


transformer_lens/ActivationCache.py
jlarson4: This block still says `ActivationCache` is designed for `HookedTransformer` and "will not work with other models" Can we update this comment and any surrounding examples?

transformer_lens/ActivationCache.py
```diff
@@ -270,12 +270,12 @@ def keys(self):
...
-            >>> from transformer_lens import HookedTransformer
-            >>> model = HookedTransformer.from_pretrained("tiny-stories-1M")
-            Loaded pretrained model tiny-stories-1M into HookedTransformer
+            >>> from transformer_lens.model_bridge import TransformerBridge
+            >>> model = TransformerBridge.boot_transformers("roneneldan/TinyStories-1M")
+            >>> model.enable_compatibility_mode()
             >>> _logits, cache = model.run_with_cache("Some prompt")
             >>> list(cache.keys())[0:3]
-            ['hook_embed', 'hook_pos_embed', 'blocks.0.hook_resid_pre']
+            ['embed.hook_in', 'hook_embed', 'embed.hook_out']
```
jlarson4: Two of these three keys are the same tensor, `hook_embed` is the `compat` alias for `embed.hook_out`. The slice shows duplication rather than cache structure, and `__iter__` now prints an identical list below (line 318). Can the slice span enough keys to still teach the layout?

jlarson4 (commented):
Thanks for putting this together @Austin1serb! Looks great! Just one couple small notes below

transformer_lens/ActivationCache.py
Austin1serb: Okay yeah. I revised the ActivationCache documentation to show it supports both HookedTransformer and TransformerBridge, and clarified that the remaining limitation is that the advanced helpers expect the TransformerLens weight-processing interface and a complete activation cache.

transformer_lens/ActivationCache.py
Austin1serb: Got it. 
I updated both examples so they show a better section of the cache instead of highlighting the alias

Canonik (changes requested):
Ran the doctests from this branch locally to double-check. `ActivationCache.py` and
`exploratory_utils.py` pass as-is (nice, the compat bridge reproduces the old float table
in `test_prompt` exactly). The `IOIDataset` one breaks on a CUDA machine though:

    RuntimeError: Expected all tensors to be on the same device, but got index is on cpu,
    different from other tensors on cuda:0

`boot_transformers("gpt2")` with no device arg picks cuda when available, but `ioi_eval`
feeds CPU tensors straight from its DataLoader, and the bridge forward passes `input_ids`
through to the HF model unchanged. The old HookedTransformer version wasn't affected
because `input_to_embed` moves tokens to the model device (HookedTransformer.py:412).
CI is green here only because the runners are CPU-only. Easiest fix:

    >>> model = TransformerBridge.boot_transformers("gpt2", device="cpu")

same pattern as the `gpt2_bridge_compat` fixture in tests/conftest.py. Verified locally
that with `device="cpu"` the doctest passes end to end on this branch.

Two small things while you're in these files:

- the comment at the bottom of `exploratory_utils.py` still says the docstring "won't be
  tested with PyTest". Not true: `--collect-only` shows the doctest item is collected and
  run; the `pytest.mark.skip` only hits the accidentally-collected function item. Worth
  fixing the comment since this PR makes that doctest live on the bridge.
- `ActivationCache.py:63` still cross-references
  ``:meth:`transformer_lens.HookedTransformer.HookedTransformer.run_with_cache` ``, which
  will dangle once HookedTransformer is removed. Could point at the bridge's
  `run_with_cache` instead.

For anyone looking at the red check: Full Code Coverage fails on
`test_optimizer_compatibility.py::test_bridge_hooked_parity_multi_step_optimization` at
0.001032 vs a 0.001 threshold, a numerics flake unrelated to this doctest-only diff.

Rest looks right to me: dropping the "Loaded pretrained model ..." lines matches boot
printing nothing, the compat-mode choices line up with the acceptance-test precedent
(test_activation_cache.py, test_evals.py), and the HF repo ids are the documented
preference over the deprecated aliases.

Austin1serb:
I have addressed the requested changes below:

1. Pinned the IOI doctest to device="cpu" to avoid the device mismatch.
2. Updated the exploratory_utils.py comment to match the current doctest behavior.
3.  Fixed the ActivationCache cross-reference to point to TransformerBridge.run_with_cache.

jlarson4:
Great work on this @Austin1serb, looks great, merging it now!

PR #1577, opened 2026-07-31
Honor native bridge weight initialization config

### LINKED ISSUE #1566: [Proposal] Honor cfg.init_weights=False in boot_native and add bridge.init_weights()
### Proposal
Base on `dev-4.x`. Gate the `initialize_native_model(...)` call in `boot` on `cfg.init_weights` (`transformer_lens/model_bridge/sources/native/__init__.py:56-62`), and add a public `TransformerBridge.init_weights()` re-init method.

### Motivation
Deprecating `HookedTransformer` moves train-from-scratch to `boot_native`, but the bridge ignores `cfg.init_weights=False` (randomizing weights anyway) and has no re-init entry point, breaking the load-into-empty-shell workflow.

### Pitch
- Gate `initialize_native_model(...)` on `cfg.init_weights`, keeping `fork_rng` for the seeded branch when init runs.
- Add `TransformerBridge.init_weights()` re-running `initialize_native_model` in place, honoring `cfg.seed`.
- Raise a clear error when `init_weights()` is called on a non-native (HF-loaded) bridge.

Acceptance:
- [ ] `boot_native(cfg)` with `init_weights=False` leaves default `nn.Linear`/`nn.Embedding` reset state
- [ ] `bridge.init_weights()` re-inits a native bridge in place, honoring `cfg.seed`
- [ ] `init_weights()` on an HF-loaded bridge raises an actionable error
- [ ] New unit tests in `tests/unit/model_bridge/test_boot_native.py` (skip + re-init paths). The file already exists, it just needs to be extended

Mirror: HT gate `transformer_lens/HookedTransformer.py:254-255`; HT `init_weights` at `HookedTransformer.py:1500`.

### Checklist

happykawayigt:
# Description

Honors `cfg.init_weights=False` in `TransformerBridge.boot_native()` and adds a public `TransformerBridge.init_weights()` method for deterministic, in-place reinitialization of TL-native bridges.

Bridge construction replaces native submodules with generalized component wrappers, so the native initializer now unwraps those components before resetting the original parameters. Calling `init_weights()` on a non-native bridge raises an actionable error pointing to `boot_native()`.

Fixes #1566

## Type of change

## Validation

- `pytest tests/unit/model_bridge/test_boot_native.py tests/unit/model_bridge/test_build_bridge_from_module.py tests/unit/model_bridge/test_native_features.py -q`: 78 passed
- `mypy .`: Success, no issues found in 421 source files
- Full `pytest tests/unit -m "not slow" -q`: 5142 passed; the only two failures were existing UTF-8 test files read with the Windows GBK locale, and both pass with `PYTHONUTF8=1`
- Black, isort, pycln, and `git diff --check`

Development note: AI-assisted implementation; I manually traced bridge component replacement, reviewed the patch, and ran the focused and repository-wide validation above.

# Checklist

### DIFF
--- a/docs/source/content/migrating_to_v3.md
+++ b/docs/source/content/migrating_to_v3.md
@@ -237,6 +237,7 @@ Weight-matrix rows return **raw** HuggingFace weights by default. `HookedTransfo
 | `model.reset_hooks(...)` | `bridge.reset_hooks(clear_contexts=..., direction=..., including_permanent=..., level=...)` | The bridge supports the same scoped hook cleanup controls. |
 | A helper typed as `HookedTransformer` | Type it as `TransformerLensModel` or `TransformerLensModelWithWeights` from `transformer_lens.model_protocol` | Use the narrower structural protocol unless the helper reads weights or advanced `ActivationCache` helpers. |
 | `HookedTransformerConfig(...)` | `TransformerBridgeConfig(...)` | Construct the bridge config with the same core fields, then pass it to `TransformerBridge.boot_native(config)` for train-from-scratch or toy models. Do not pass a legacy config object to `boot_native`. |
+| `cfg.init_weights` | `bridge.init_weights()` | `boot_native()` honors the config flag during construction; call `init_weights()` to reinitialize a TL-native bridge in place. |
 | `model.all_head_labels()` | `bridge.all_head_labels` | This is a property on the bridge, so omit the call parentheses. |
 | `model.set_tokenizer(tokenizer)` | `TransformerBridge.boot_transformers(name, tokenizer=tokenizer)` | A bridge's tokenizer is fixed when it boots. Reboot to change it; assigning `bridge.tokenizer` directly bypasses tokenizer/config wiring. |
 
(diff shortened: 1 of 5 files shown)


emerardd:
Thanks for working on this. I tested the re-init path locally and found two edge cases worth checking:

`UnembeddingBridge` adds a zero bias to the native bias-free head during bridge construction. The current test zeroes all parameters before re-init, so it cannot detect that this bias is not reinitialized. Filling parameters with a non-zero value such as `42` shows that `head.bias` remains stale unless `initialize_native_model()` explicitly zeros it.

With `init_weights=False`, the current seeded branch still enters `fork_rng`, but `cfg.seed` is never applied to `NativeModel` construction. As a result, different non-null seed values produce identical default parameters and only affect whether global RNG state is restored. The issue wording may imply that the seeded `fork_rng` should only be used when custom initialization runs.

I also exercised re-init across plain/RMS/no-norm, gated MLP, attention-only, rotary, separate-output-vocab, and all five initialization modes; those paths otherwise worked correctly.

jlarson4 (commented):
Thanks for closing this out @happykawayigt! Great resolution. Just one small note:

`HookedTransformer.init_weights()` now has a bridge equivalent, but the migration guide never mentions `init_weights`. There is a table for API table equivalents in Migrating to v3 (`migrating_to_v3.md:223`). Could you add that row for `cfg.init_weights` / `bridge.init_weights()`?

happykawayigt:
Added in 3089115: the migration table now maps `cfg.init_weights` to `bridge.init_weights()` and notes that `boot_native()` honors the config flag during construction. The updated checks are green. Thanks for catching this.

PR #1579, opened 2026-07-31
Drop optional unused aliases on hybrid architectures

### LINKED ISSUE #1578: [Proposal] Hybrid architectures warn on unresolved optional 'hook_attn' aliases when they are skipped
### Proposal 

When setup_submodules skips an optional submodule, prune that block’s hook_aliases whose target path starts at the skipped key (e.g. drop hook_attn_out → attn.hook_out when attn was skipped). Copy the aliases dict per layer so templates don’t share mutations; leave residual/MLP aliases alone. Full-attention siblings keep HT aliases; GDN native aliases are unchanged. Cover with unit tests for prune + no warning on register.

### Motivation

When loading models like Qwen-3.5 Hybrid, a warning is printed per hybrid layer, cluttering the terminal quickly (as noted in https://github.com/TransformerLensOrg/TransformerLens/pull/1547). The proposed solution would remove unused aliases since it is expected behavior for GDN layers to use their native `hook_linear_attn_out` instead.
```
UserWarning: Hook alias 'hook_attn_out' -> 'attn.hook_out' on BlockBridge(name='model.language_model.layers.2') did not resolve; this hook will not be accessible.
  getattr(module, "_register_aliases")()
```

### Pitch

Loading hybrids like Qwen3.5 spam harmless hook_attn_out did not resolve warnings on every GDN/Mamba-2 layer. Those layers never had full attn; they expose hook_linear_attn_out (and similar) instead. Treat that as expected, not a fault, so boot stays quiet without hiding real alias bugs on full-attention layers.

### Alternatives

The warning could be left as-is, as this is not a functional error and the user might want to be notified at build-time that the standard `hook_attn_out` alias will be missing, though it could be caught and reworded to reflect its non-breaking nature.

### Checklist

KYinXu:
# Description

When hybrid layers skip an optional submodule (e.g. `attn` on GDN / Mamba-2 layers), the shared `BlockBridge` template still declared HT aliases like `hook_attn_out → attn.hook_out`. `_register_aliases` then failed to resolve them and emitted a harmless warning per hybrid layer (notably when loading Qwen3.5; see also #1547).

This PR prunes those inapplicable aliases in `setup_submodules` when an optional submodule is skipped—per layer, so full-attention siblings keep their HT aliases and real unresolved-alias bugs still warn. Unit tests cover prune behavior and no warning on register. The Jacobian Lens Qwen3.5 demo cell is updated now that boot no longer expects that stderr.

Fixes #1578

## Type of change

### Screenshots

Warnings no longer appear

image

# Checklist:


transformer_lens/model_bridge/component_setup.py
```diff
@@ -150,6 +150,43 @@ def setup_submodules(
     # Clean up so architecture_adapter traversal won't find stale entries
     for name in skipped_optional:
         component.submodules.pop(name, None)
+    if skipped_optional:
+        _prune_hook_aliases_for_skipped(component, skipped_optional)
+
+
+def _prune_hook_aliases_for_skipped(component: GeneralizedComponent, skipped: list[str]) -> None:
+    """Drop hook aliases whose target path starts at a skipped optional submodule.
```
jlarson4: The prune fires for every optional skip across 12 architectures, including native SSM aliases on `nemotron_h`/`zamba2`. Can we update the docstring to be more general, and add one non-`attn` test case?

transformer_lens/model_bridge/component_setup.py
```diff
@@ -150,6 +150,43 @@ def setup_submodules(
...
+
+    to_drop: list[str] = []
+    for alias_name, target in aliases.items():
+        if isinstance(target, list):
+            kept = [t for t in target if _first_segment(t) not in skipped_set]
+            if not kept:
+                to_drop.append(alias_name)
+            elif len(kept) != len(target):
+                aliases[alias_name] = kept
+        elif _first_segment(target) in skipped_set:
```
jlarson4, line 210: On OLMo-Hybrid, `BlockBridge` retargets `hook_mlp_out` onto the optional `ln2_post` (`block.py:78`), so this branch drops it on linear-attention layers that do have an `mlp`, and removes the warning that flagged it. Can the prune keep aliases with a live fallback target?

tests/unit/test_optional_submodule.py
```diff
@@ -157,6 +158,37 @@ def test_non_optional_raises(self):
...
+        for i in range(3):
+            for alias in attn_aliases:
+                assert alias in blocks[i].hook_aliases
+        for alias in attn_aliases:
+            assert alias not in blocks[3].hook_aliases
+        # Residual / MLP aliases are independent of attn and must survive.
+        assert "hook_resid_pre" in blocks[3].hook_aliases
+        assert "hook_mlp_out" in blocks[3].hook_aliases
+
+    def test_pruned_aliases_do_not_warn_on_register(self):
```
jlarson4, line 235: Both new tests call `_register_aliases()` directly, so nothing asserts a hybrid bridge booting warning-free through `TransformerBridge.__init__`. Can you add a boot-level assertion using the config-only tiny-bridge fixture in `test_qwen3_5_adapter.py:562`?

tests/unit/test_optional_submodule.py
```diff
@@ -157,6 +158,37 @@ def test_non_optional_raises(self):
...
+        blocks = _setup_blocks(HybridModel(), AttnAdapter())
+        with warnings.catch_warnings(record=True) as caught:
+            warnings.simplefilter("always")
+            blocks[3]._register_aliases()
+        unresolved = [
```
jlarson4: `AttnAdapter` declares no `ln2`/`mlp`, so block 3 still emits two legitimate unresolved warnings. This forces the substring filter instead of asserting the list is empty. Can the fixture carry those submodules so the assertion can be unconditional?

jlarson4 (commented):
Nice, tightly-scoped fix @KYinXu! 

A couple small requests before merging, detailed below

KYinXu:
@jlarson4 just addressed changes, sorry about delay was busy with other tasks earlier in the week and had to catch up.

jlarson4:
@KYinXu No worries at all about the delay, thank you for coming back and wrapping those up. I will review shortly

jlarson4:
Great work @KYinXu merging now!

PR #1591, opened 2026-08-01
fix(model-bridge): restore native state dict round trips

### LINKED ISSUE #1587: [Bug Report] Native bridge state_dict()/load_state_dict() are not inverses
### **Base on `dev-4.x`**

**Describe the bug**

On a native `TransformerBridge`, `state_dict()` emits TL-renamed keys (`embed.weight`, `blocks.0.attn.q.weight`, …), but `load_state_dict()` only matches raw native names, so a `state_dict()` -> `load_state_dict()` round-trip silently loads nothing and `strict=True` is silently downgraded to `strict=False`. You can reload a checkpoint into a random-init model, get no error, and analyse garbage. This blocks the checkpoint save/reload workflow that `HookedTransformer` deprecation (removal at 4.0) requires.

**Code example**

```python

bridge = TransformerBridge.boot_transformers("gpt2")
import torch
with torch.no_grad():
    for p in bridge.parameters():
        p.zero_()
sd = bridge.state_dict()  # TL-renamed keys, e.g. "embed.weight"
bridge.load_state_dict(sd, strict=True)  # no-op; strict silently downgraded, no error
```

**System Info**

Round-trip must overwrite params, and `strict=True` must raise on genuine mismatch — mirroring `HookedTransformer`'s trivially-inverse `state_dict`/`load_state_dict`. In `transformer_lens/model_bridge/transformer_bridge.py`, give `load_state_dict` (`:3478-3510`) an inverse of `convert_hf_key_to_tl_key` (forward at `architecture_adapter.py:880`) before the clean↔actual mapping, and remove the silent strict downgrade at `:3507`. The forward direction is the `state_dict` override at `transformer_bridge.py:3424` (TL-key emission at `:3470`). Compatibility constraint: `transformer_lens/utilities/tracr.py` `make_tracr_transformer_bridge_state_dict` (def `:99`) emits raw native keys (`:117-122`; documented at `:106-108`) and relies on raw-name matching, so the inverse-key fix must keep that path loading.

Acceptance:
- [ ] `load_state_dict(state_dict())` overwrites params (mutate, round-trip, assert equal) — no no-op
- [ ] `strict=True` raises on mismatched state dict
- [ ] Existing tracr path (`utilities/tracr.py`, raw-native names) still loads
- [ ] New round-trip test under `tests/unit/model_bridge/`

### Checklist

mikemikimike:
## Summary

- Restore `TransformerBridge.state_dict()` round trips through the native/raw model.
- Map each TransformerLens key back to all underlying aliased raw keys before loading.
- Preserve direct clean/native key loading for Tracr-style state dicts.
- Respect `strict=True` instead of silently downgrading strict loads.
- Add regression coverage for round trips, strict missing-key failures, and raw native keys.

Fixes #1587

## Validation

- `PYTHONUTF8=1 uv run pytest tests/unit -m "not slow" -n 4 -q` — 5,165 passed, 57 skipped, 10 xfailed
- `PYTHONUTF8=1 uv run pytest tests/unit/model_bridge/test_boot_native.py -q` — 28 passed
- `uv run mypy .` — passed, 421 source files
- `uv run pycln --check ...` — passed
- `uv run isort --check-only ...` — passed
- `uv run black --check ...` — passed
- `git diff --check` — passed

The focused `tests/integration/model_bridge` run was attempted locally but exceeded the 15-minute execution limit without producing a test assertion failure; the unit suite and targeted regression suite completed successfully.


jlarson4:
Hi @mikemikimike! There were 3 different PRs all submitted around the same time, providing different solutions to issue #1587. I have opted to take #1598 as our primary solution, but I do want to keep your round-trip tests because they are very strong. Once #1598 lands, would you be willing to rebase to `dev-4.x` and trim your PR to those tests so I can merge them as is?

PR #1592, opened 2026-08-01
deprecate remaining hooked entry points

### LINKED ISSUE #1590: [Proposal] Emit DeprecationWarning on remaining Hooked* entry points
This is 4.x prep, but we should base this one to `dev`, to ensure the warnings go out in the next release.

### Proposal
Add a `DeprecationWarning` to the still-silent `Hooked*` entry points. 
- `HookedEncoder`
- `BertNextSentencePrediction`
- `HookedRootModule`
- `HookedTransformer.__init__`

### Motivation
Every `Hooked*` class is being deprecated toward `TransformerBridge` (removal at 4.0); a user who never calls the three already-warned `from_pretrained` paths currently gets zero notice.

### Pitch
- `transformer_lens/HookedEncoder.py` – `__init__` (:53) and `from_pretrained` (:369, which today emits only a non-deprecation `logging.warning` — replace/augment it with a real `DeprecationWarning`).
- `transformer_lens/BertNextSentencePrediction.py:33` – `__init__` (point to bridge NSP, `demos/BERT.ipynb`).
- `transformer_lens/HookedTransformer.py` – `__init__` **only**. Do **not** add one to `from_pretrained_no_processing` (:1469): it delegates to `from_pretrained`, which already warns (`:1318-1325`), so a second warning would double-fire.
- `transformer_lens/HookedRootModule.py:55` – `__init__`, **but** this is the base class of all four Hooked models and every subclass `__init__` calls `super().__init__()`, so an unconditional warning here would fire a second time on top of the subclass warning. Gate it on `type(self) is HookedRootModule` so it fires only on *direct* instantiation – keeping "exactly one per construction".
- Each message names the class, says removed at 4.0, points at `TransformerBridge.boot_transformers(...)`; add NO import-time warning in `__init__.py`.

Acceptance:
- [ ] Each entry point emits exactly one `DeprecationWarning` naming the class + bridge replacement
- [ ] No warning fires from `import transformer_lens`
- [ ] `pytest.warns` regression test per entry point
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

Mirror: `transformer_lens/HookedTransformer.py:1318-1325` (siblings `HookedEncoderDecoder.py:553-560`, `HookedAudioEncoder.py:366-373`).

### Checklist

msaule:
## Summary

- Emit `DeprecationWarning` from the remaining silent legacy entry points.
- Gate the `HookedRootModule` warning to direct construction, avoiding duplicate warnings from subclasses.
- Add regression coverage for one warning per constructor and a warning-free package import.

## Why

Users who instantiate these legacy classes directly otherwise receive no migration notice ahead of the 4.0 removal. Each warning points to `TransformerBridge.boot_transformers(...)`.

Closes #1590

## Validation

- `pytest tests/unit/test_deprecation_warnings.py -q` — 5 passed
- `mypy .` — 385 source files, no issues
- `pycln`, `isort`, and `black` checks passed on changed files

### DIFF
--- a/demos/Othello_GPT.ipynb
+++ b/demos/Othello_GPT.ipynb
@@ -279,7 +279,18 @@
    "cell_type": "code",
    "execution_count": 56,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "DeprecationWarning:\n",
+      "\n",
+      "HookedTransformer is deprecated and will be removed in 4.0. Use TransformerBridge.boot_transformers(...) instead, then call enable_compatibility_mode() for HookedTransformer-equivalent numerics.\n",
+      "\n"
+     ]
+    }
+   ],
    "source": [
     "import transformer_lens.utilities as utils\n",
     "\n",
--- a/tests/unit/test_deprecation_warnings.py
+++ b/tests/unit/test_deprecation_warnings.py
@@ -2,9 +2,15 @@
 
 from __future__ import annotations
 
+import importlib
+import subprocess
+import sys
 import warnings
+from pathlib import Path
 
 import pytest
+
+PROJECT_ROOT = Path(__file__).parents[2]
 
 
 def _small_config():
@@ -31,11 +37,26 @@ def _assert_single_deprecation(constructor, class_name: str) -> None:
 
 
 def test_importing_transformer_lens_emits_no_deprecation_warning():
-    with warnings.catch_warnings(record=True) as caught:
-        warnings.simplefilter("always")
-        import transformer_lens  # noqa: F401
+    code = "\n".join(
+        [
+            "import warnings",
+            "warnings.filterwarnings(",
+            "    'error',",
+            "    category=DeprecationWarning,",
+            "    module=r'^transformer_lens(?:\\.|$)',",
+            ")",
+            "import transformer_lens",
+        ]
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        cwd=PROJECT_ROOT,
+        text=True,
+        check=False,
+    )
 
-    assert not [warning for warning in caught if issubclass(warning.category, DeprecationWarning)]
+    assert result.returncode == 0, result.stderr
 
 
 def test_hooked_transformer_constructor_warns_once():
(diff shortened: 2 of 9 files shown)


transformer_lens/HookedEncoder.py
```diff
@@ -58,6 +59,12 @@ def __init__(
         **kwargs: Any,
     ):
         super().__init__()
+        warnings.warn(
```
jlarson4, line 62: `HookedEncoder.from_pretrained("bert-base-cased")` still surfaces no deprecation notices because `stacklevel=2` attributes the warning to `HookedEncoder.py` itself and Python's default `ignore::DeprecationWarning` filter then drops it. Can the warning also fire at the `from_pretrained` entry point, the way the sibling loaders do?

tests/unit/test_deprecation_warnings.py
```diff
@@ -0,0 +1,64 @@
...
+def _assert_single_deprecation(constructor, class_name: str) -> None:
+    with pytest.warns(DeprecationWarning, match=class_name) as caught:
+        constructor()
+
+    assert len(caught) == 1
+    assert "TransformerBridge.boot_transformers" in str(caught[0].message)
+    assert "4.0" in str(caught[0].message)
+
+
+def test_importing_transformer_lens_emits_no_deprecation_warning():
```
jlarson4, line 37: By the time this body runs, `transformer_lens` is already in `sys.modules`, so the import is a no-op that records nothing. I appended a real import-time `DeprecationWarning` to `__init__.py` and this test still passed. Is there a way to check import-time cleanliness that would fail?

demos/Othello_GPT.ipynb
```diff
@@ -279,7 +279,18 @@
    "cell_type": "code",
    "execution_count": 56,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
jlarson4, line 282: `Grokking_Demo.ipynb` and `No_Position_Experiment.ipynb` also construct `HookedTransformer(cfg)` in cells with empty stored outputs, and both run under `make notebook-test`. Can their outputs be re-recorded as well?

jlarson4 (commented):
Thanks for picking this up @msaule! Looks pretty good, just a couple small adjustments to request before I can merge this. Detailed below:

msaule:
thanks for the thorough review. addressed all three items:

- added a `HookedEncoder.from_pretrained` deprecation warning that reaches the caller, with a regression test that stops before any model download
- changed the import-cleanliness test to run in a fresh python process
- re-recorded the expected warning output for Grokking and No_Position

verified with `pytest tests/unit/test_deprecation_warnings.py -q` with 6 passing tests, plus the targeted No_Position notebook check.

transformer_lens/HookedEncoder.py
```diff
@@ -379,6 +386,12 @@ def from_pretrained(
         **from_pretrained_kwargs: Any,
     ) -> HookedEncoder:
         """Loads in the pretrained weights from huggingface. Currently supports loading weight from HuggingFace BertForMaskedLM. Unlike HookedTransformer, this does not yet do any preprocessing on the model."""
+        warnings.warn(
```
jlarson4, line 389: Outside pytest `from_pretrained` sits one frame below its caller, so `stacklevel=4` overshoots and CPython attributes the warning to `<sys>:0` — a plain script still sees nothing, while the sibling using `stacklevel=2` does surface (`HookedEncoderDecoder.py:555`). The depth and the `filename` assertion at `test_deprecation_warnings.py:91` have to move together.

msaule:
thanks, good catch. fixed in b2f5965. i changed the loader warning to stacklevel=2, matching HookedEncoderDecoder, and replaced the pytest-only filename assertion with a fresh-process test that checks the warning reaches the external caller. pytest tests/unit/test_deprecation_warnings.py -q passes with 6 tests.

jlarson4:
Thanks @msaule! Looks great, merging!

PR #1596, opened 2026-08-02
feat(jacobian_lens): J-space sparse decomposition

janmenjayap:
# Description

Adds **J-space sparse decomposition** to the Jacobian lens: it writes an activation (or a steering / sparse-autoencoder direction) as a `k`-sparse **nonnegative** combination of J-lens vectors `v_t = J_ℓ^T W_U[:, t]`, following Gurnee et al. (2026), "Verbalizable Representations Form a Global Workspace in Language Models" (Transformer Circuits Thread). This is the Tier-2 decomposition item tracked in #1539.

## New public surface (TransformerBridge only, matching the rest of the Jacobian lens)

- `get_sparse_decomposition(x, dictionary, k=25, *, algorithm=...)` — a model-free greedy solver returning a `JSpaceDecomposition` (`support`, nonnegative `coordinates`, `reconstruction`, `j_space_component`, `non_j_space_component`).
- `JacobianLens.lens_vector_dictionary(model, layer)` — the cached full-vocabulary J-lens dictionary for a layer.
- `JacobianLens.decompose(model, activation_or_prompt, layer, *, position=None, k=25, algorithm=...)` — decomposes either a raw `[d_model]` activation or the `blocks.{layer}.hook_out` residual at a prompt position.
- `JSpaceDecomposition` and `get_sparse_decomposition` are exported from `transformer_lens.tools.analysis`.

## Two coefficient-update rules (`algorithm=`)

- `nonnegative_orthogonal_matching_pursuit` (**default**) — an exact nonnegative least-squares re-solve on the active set (Lawson & Hanson, 1974); optimal on the selected support at the small `k` used here.
- `gradient_pursuit` — the directional update of Blumensath & Davies (2008), for faithfulness to the paper and the large-active-set regime.

Both share the same greedy orthogonal-matching-pursuit selection (Pati et al., 1993).
At vocabulary scale the per-step cost is dominated by the correlation over all atoms,
so the exact re-solve is effectively free while returning optimal coefficients on the
support — hence it is the default. The choice is disclosed here because the paper uses
gradient pursuit; both are provided.

## Two outputs that need not coincide (paper appendix)

- `coordinates` — the nonnegative pursuit coefficients (the "local J-space coordinates").
- `j_space_component` — the orthogonal projection of the activation onto the span of the selected vectors; its residual matches `swap_hooks`.

They diverge whenever a coordinate is clamped to zero by the nonnegativity constraint; both are returned.

## Docs

A new "Sparse decomposition (J-space coordinates)" section in `jacobian_lens_fitting.md`, including the paper's variance findings **explicitly caveated as measured on closed Anthropic models** (on open-weight models the shape may hold but the exact values will not necessarily transfer). A runnable demo notebook will follow on a separate `docs`-named branch, per the Tier-2 scope.

Part of #1539 (Tier 2).

## Type of change

## Checklist

## Verification run locally

Full `make test-pr` equivalent (direct binaries, single Python) plus the static gates:

- `check-format` (pycln / isort / black), repo-wide — clean.
- `mypy .` — Success, no issues in 388 source files.
- `unit` (`pytest tests/unit -m "not slow"`) — 4513 passed, 29 skipped, 10 xfailed.
- `docstring` (`pytest transformer_lens/`) — 18 passed.
- `acceptance` (`pytest tests/acceptance -m "not slow"`) — passed.
- `integration` (`pytest tests/integration -m "not slow"`) — 1145 passed. The only failure is `tests/integration/model_bridge/test_jamba_adapter.py::TestJambaGeneration::test_greedy_matches_hf`, a pre-existing device-placement issue (`index on cuda:0` vs tensors on `cpu`) in the unrelated Jamba adapter — untouched by this PR, which changes only `transformer_lens/tools/analysis/jacobian_lens*`.
- `build-docs` (Sphinx) — build succeeded; the new docs page adds no warnings.
- Real-model integration: GPT-2 `decompose` and the slow `gemma-2-2b-it` `decompose` case both passed.

CI re-runs the full suite across Python 3.10 / 3.11 / 3.12 (`compatibility-checks`).

cc: @jlarson4


transformer_lens/tools/analysis/jacobian_lens_decomposition.py
```diff
@@ -0,0 +1,219 @@
...
+    reconstruction: torch.Tensor
+    j_space_component: torch.Tensor
+    non_j_space_component: torch.Tensor
+
+
+def _nonnegative_least_squares(active_atoms: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    """Minimize ``||active_atoms @ coefficients - target||`` over ``coefficients >= 0``.
+
+    A small active-set solver: solve the unconstrained least squares, and while any
+    coefficient is negative, drop the most-negative atom and re-solve. Exact for the tiny
```
jlarson4: The active-set solver only drops atoms and never re-admits one, so it is not exact NNLS: 4 of 288 real GPT-2 decompositions returned coordinates violating the KKT dual condition, with relative violations up to 0.55. Can you complete the Lawson–Hanson re-entry step?

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
```diff
@@ -0,0 +1,219 @@
...
+    if not bool(torch.isfinite(atoms).all()):
+        raise ValueError("dictionary contains non-finite entries")
+    atom_norms = (atoms * atoms).sum(dim=1).sqrt()
+    if bool((atom_norms == 0).any()):
+        raise ValueError("dictionary contains a zero-norm atom")
+
+    residual = target.clone()
+    support: List[int] = []
+    coordinates = target.new_zeros(0)
+    for _ in range(k):
```
jlarson4, line 388: Once a coefficient is clamped to zero the residual is unchanged, so the next pick uses a stale residual. On GPT-2 layer 10 only ~9 of 25 coordinates end up nonzero. Can you reconcile k with the docs, whose example prints all k support tokens as active concepts?

tests/unit/tools/test_jacobian_lens_decomposition.py
```diff
@@ -0,0 +1,266 @@
...
+    result = get_sparse_decomposition(x, dictionary, k=4, algorithm="gradient_pursuit")
+
+    assert (result.coordinates >= 0).all()
+    assert (x - result.reconstruction).norm().item() <= x.norm().item() + 1e-6
+
+
+# --------------------------------------------------------------------------- #
+# Brute-force oracle (no external reference implementation exists)
+# --------------------------------------------------------------------------- #
+def _brute_force_best_support(x, dictionary, k):
```
jlarson4, line 638: The brute-force oracle calls `_nonnegative_least_squares` as its own inner solver, so it validates the greedy selection, but assumes the coefficient solver is correct. A wrong solver passes this test unchanged. Can the oracle use an independent reference, such as a KKT optimality check on the returned coordinates?

jlarson4 (commented):
Hey @janmenjayap! Great work on this, it is an excellent piece of functionality. Just a couple adjustments I'd like you to make before I can merge. Please let me know if you have any questions!

jlarson4:
@janmenjayap Hey, just wanted to touch base and see if you'd have a moment to wrap this PR up? Thanks!

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
janmenjayap: Fixed in https://github.com/TransformerLensOrg/TransformerLens/pull/1596/commits/4cfe8700b5b557af60512f5d3361f5f9dcc47bac. `_nonnegative_least_squares` is now the full Lawson–Hanson active-set method: a released atom can re-enter (outer dual-feasibility loop + inner feasibility corrections on the classical `3·num_active` budget). The passive set is solved in float64 with an explicit pseudoinverse rank threshold, and the result is checked against the KKT conditions (`_validate_nnls_kkt`) before returning --- an exhausted budget, an invalid line-search step, a stalled correction, or a failed KKT check raises `RuntimeError` rather than clamping and returning an unverified vector.

I reran the same class of check you reported --- **935 real GPT-2 decompositions** across all 11 source layers: **0 KKT failures, max relative dual violation 2.77e-08** (normalization: `max_j max(0, aⱼ·r) / (‖aⱼ‖·‖x‖)`), versus your up-to-0.55 on 4/288 with the old drop-only solver.

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
janmenjayap: Fixed in https://github.com/TransformerLensOrg/TransformerLens/pull/1596/commits/9ae606c89a60234c74770e28198396cfd904cc27. `k` is now an **upper bound**, not a target: greedy selection stops early once no unselected atom is materially positively correlated with the residual (under nonnegativity a non-positively-correlated atom can't reduce it). I also split the two supports the paper conflates:

- `support` = the numerically **active** set (selected atoms with strictly-positive coordinate); `coordinates` is aligned with it.
- new `selected_support` = every greedily selected atom, which defines the span for `j_space_component`.
- so `len(support) <= len(selected_support) <= k`.

The "25 slots, ~9 nonzero" mismatch is now impossible by construction --- `support` equals the active count. `jacobian_lens_fitting.md` was updated to say "up to `k` active vectors" and to distinguish active support from the projection span. Across the 935 GPT-2 decompositions: **0 dead slots**, and **322/935** have `selected > active`, so the distinction is real and common rather than a corner case.

tests/unit/tools/test_jacobian_lens_decomposition.py
janmenjayap: Fixed in https://github.com/TransformerLensOrg/TransformerLens/pull/1596/commits/4cfe8700b5b557af60512f5d3361f5f9dcc47bac. Validation no longer trusts the solver to check itself:

- `_reference_nnls`: a brute-force support-enumeration NNLS that **shares no code** with `_nonnegative_least_squares`; production and reference objectives are compared **two-sidedly** (a production objective materially below the exhaustive optimum now fails too).
- `_assert_nnls_kkt`: a direct **KKT optimality certificate** on the returned coordinates (primal feasibility, dual feasibility on zeros, stationarity on positives, complementarity), applied on rank-deficient, duplicate-column, near-collinear, boundary, zero-target, and jointly-rescaled systems.

So a wrong coefficient solver now fails. `_brute_force_best_support` stays scoped to greedy-selection quality only, since the solver it calls is independently verified by the two checks above.

janmenjayap:
Thanks for the patience, and for the thorough review! I've pushed two commits that address all three threads:

- https://github.com/TransformerLensOrg/TransformerLens/pull/1596/commits/4cfe8700b5b557af60512f5d3361f5f9dcc47bac: completes NNLS to exact Lawson–Hanson and adds independent KKT validation (comments 1 & 3)
- https://github.com/TransformerLensOrg/TransformerLens/pull/1596/commits/9ae606c89a60234c74770e28198396cfd904cc27: makes `k` an upper bound and splits active vs. selected support end-to-end (comment 2)

I've replied inline on each thread with the specifics and fresh measurements. Ready for another look.

jlarson4:
Looks great, thanks for taking this on @janmenjayap

PR #1595, opened 2026-08-02
Add native bridge state dict regression tests

ArS377:
# Description

Adds regression coverage for the native `TransformerBridge.state_dict()` / `load_state_dict()` round-trip behavior in #1587.

This PR intentionally contains tests only. It depends on #1598, which provides the production implementation selected by the maintainer. After #1598 is merged, these tests verify native round trips, strict handling of an unexpected key, and continued loading of Tracr-generated state dictionaries.

No new dependencies are required.

## Type of change

- [x] Bug-fix regression coverage

### Screenshots

Not applicable — no UI change.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [ ] New and existing unit tests pass locally with my changes (depends on #1598)
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #1598, opened 2026-08-02
fix(bridge): make native TransformerBridge state_dict()/load_state_dict() true inverses

### LINKED ISSUE #1587: [Bug Report] Native bridge state_dict()/load_state_dict() are not inverses

LightWork666:
Fixes #1587.

## The bug

On a native `TransformerBridge`, `state_dict()` returns TL-renamed keys (`embed.weight`, `blocks.0.attn.q.weight`, ...), but `load_state_dict()` only knew how to match raw native parameter names. So `bridge.load_state_dict(bridge.state_dict())` silently did nothing — no error, no warning, params just stayed whatever they were before. Worse, `strict=True` was silently downgraded to `strict=False` whenever the key counts didn't line up, so there was no way to even notice the round trip had failed.

## The fix

`load_state_dict` now builds the inverse of the TL-key renaming that `state_dict()` applies (`_tl_key_to_actual_keys`), so it can map TL-format keys back to the underlying native parameter paths before handing them to the wrapped model's own `load_state_dict`. One wrinkle: some bridge components expose the same underlying parameter through more than one attribute path — e.g. GPT-2's `q`/`k`/`v` are views into the wrapped module's combined `c_attn` weight, reachable both through a block-level shortcut and through the nested `_original_component` chain. Writing to only one of those paths leaves the model's actual forward pass untouched even though `state_dict()` looks fine, so the mapping keeps every alias for a given TL key, not just the first one found.

The silent `strict=True → False` downgrade is gone. Missing/unexpected keys are now computed properly (scoped to what the TL state dict actually needs) and raise a real `RuntimeError` under `strict=True`, matching how `HookedTransformer` already behaves.

The raw-key loading path used by `tracr` (`make_tracr_transformer_bridge_state_dict`) still works — raw native keys are matched directly before falling through to the TL-key path.

## Testing

- New tests in `tests/unit/model_bridge/test_state_dict_round_trip.py`: round-trip actually overwrites params (not a no-op), `strict=True` raises on both missing and unexpected keys, `strict=False` doesn't raise on a partial dict, the tracr raw-key path still loads, and a real GPT-2 case checking forward-pass logits match after a zero-and-reload cycle (this last one is what caught the aliasing issue above — a naive key-rename fix passes the round-trip-key-equality check but still produces different logits, because it never touches the aliased storage).
- `uv run mypy .` — clean.
- Ran the full `tests/unit/model_bridge/` suite (207 files) in isolated batches; everything passes except a handful of pre-existing `generate()`/KV-cache crashes already tracked as an upstream PyTorch/HF bug on Apple Silicon in `tests/QUARANTINES.md` — confirmed via `git stash` that those reproduce identically on unmodified `dev-4.x`, unrelated to this change.

## Note on #1595

I noticed after finishing this that #1595 is already open for the same issue, taking a related but different approach (it detects aliasing by comparing live tensor identity rather than by TL-key-name collisions, which is arguably a more principled check). I'm submitting this anyway since it was already done and transparency seemed better than not mentioning it. Happy to have the maintainers pick whichever they prefer, or close this if #1595 is the better fix.

### DIFF
--- a/tests/unit/model_bridge/test_state_dict_round_trip.py
+++ b/tests/unit/model_bridge/test_state_dict_round_trip.py
@@ -1,0 +1,131 @@
+"""Regression tests for TransformerBridge.state_dict()/load_state_dict() round-tripping (#1587).
+
+state_dict() emits TL-renamed keys (e.g. "blocks.0.attn.q.weight"), but
+load_state_dict() only matched raw native parameter names, so a
+state_dict() -> load_state_dict() round trip silently loaded nothing and
+strict=True was silently downgraded to strict=False.
+"""
+from __future__ import annotations
+
+import pytest
+import torch
+
+from transformer_lens.config import TransformerBridgeConfig
+from transformer_lens.model_bridge import TransformerBridge
+
+
+def _native_cfg(**overrides) -> TransformerBridgeConfig:
+    base = dict(
+        d_model=32,
+        d_head=16,
+        n_heads=2,
+        n_layers=2,
+        n_ctx=8,
+        d_vocab=16,
+        d_mlp=64,
+        act_fn="gelu",
+        normalization_type="LN",
+        seed=0,
+    )
+    base.update(overrides)
+    return TransformerBridgeConfig(**base)
+
+
+def test_native_round_trip_overwrites_params_not_a_noop():
+    bridge = TransformerBridge.boot_native(_native_cfg())
+
+    sd = {k: v.clone() for k, v in bridge.state_dict().items()}
+    assert sd, "state_dict() returned no TL-format keys"
+
+    with torch.no_grad():
+        for p in bridge.parameters():
+            p.zero_()
+    assert all((p == 0).all() for p in bridge.parameters())
+
+    bridge.load_state_dict(sd, strict=True)
+
+    # Compare against the snapshot directly rather than asserting "not all
+    # zero" - LayerNorm bias legitimately initializes to all-zero, so that
+    # check would pass even for a param that never got reloaded.
+    reloaded = bridge.state_dict()
+    for key, value in sd.items():
+        assert torch.equal(reloaded[key], value), f"{key} did not round-trip"
+
+
+def test_native_strict_true_raises_on_missing_key():
+    bridge = TransformerBridge.boot_native(_native_cfg())
+    sd = bridge.state_dict()
+    incomplete = dict(sd)
+    incomplete.pop(next(iter(incomplete)))
+
+    with pytest.raises(RuntimeError, match="Missing key"):
+        bridge.load_state_dict(incomplete, strict=True)
+
+
+def test_native_strict_true_raises_on_unexpected_key():
+    bridge = TransformerBridge.boot_native(_native_cfg())
+    sd = dict(bridge.state_dict())
+    sd["totally.bogus.key"] = torch.zeros(1)
+
+    with pytest.raises(RuntimeError, match="Unexpected key"):
+        bridge.load_state_dict(sd, strict=True)
+
+
+def test_native_strict_false_does_not_raise_on_partial_dict():
+    bridge = TransformerBridge.boot_native(_native_cfg())
+    sd = bridge.state_dict()
+    first_key = next(iter(sd))
+    partial = {first_key: sd[first_key]}
+
+    result = bridge.load_state_dict(partial, strict=False)
+    assert result.unexpected_keys == []
+    assert len(result.missing_keys) > 0
+
+
+def test_native_raw_keys_still_load_tracr_style():
+    """boot_native's own raw parameter names must keep loading directly,
+    mirroring tracr's make_tracr_transformer_bridge_state_dict compatibility
+    contract (utilities/tracr.py)."""
+    bridge = TransformerBridge.boot_native(_native_cfg())
+    raw_sd = {k: v.clone() for k, v in bridge.original_model.state_dict().items()}
+
+    with torch.no_grad():
+        for p in bridge.parameters():
+            p.zero_()
+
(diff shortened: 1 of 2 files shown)


transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -3494,26 +3524,47 @@ def load_state_dict(self, state_dict, strict=True, assign=False):
         """
         current_state_dict = self.original_model.state_dict()
         clean_to_actual = {}
-        actual_to_clean = {}
         for actual_key in current_state_dict.keys():
             if actual_key != "_original_component":
-                clean_key = actual_key.replace("._original_component", "")
-                clean_to_actual[clean_key] = actual_key
-                actual_to_clean[actual_key] = clean_key
+                clean_to_actual[actual_key.replace("._original_component", "")] = actual_key
```
jlarson4, line 3529: A complete raw-HF-format gpt2 checkpoint raises under `strict=True` (337 missing) even though `strict=False` restores the forward pass exactly. The clean branch maps one alias while `required_actual_keys` demands the union (lines 3546-3549). Can the strict accounting treat a required key as satisfied when any alias of its tensor is written?

tests/unit/model_bridge/test_state_dict_round_trip.py
```diff
@@ -0,0 +1,131 @@
...
+            p.zero_()
+
+    bridge.load_state_dict(raw_sd, strict=True)
+
+    reloaded_raw = bridge.original_model.state_dict()
+    for key, value in raw_sd.items():
+        assert torch.equal(reloaded_raw[key], value), f"{key} did not round-trip"
+
+
+@pytest.mark.slow
```
jlarson4, line 146: This is the only test in any of the three competing PRs that exercises real HF-key conversion, but slow is deselected by make unit-test and the MPS job only runs for main-targeted PRs. Can you add an unmarked `boot_native` test that builds a multi-alias clean-key dict at `strict=True`?

jlarson4 (commented):
Thanks for taking a look at this @LightWork666! My intention is to take this implementation as our primary solution, and keep a couple tests and other small elements from #1591 and #1595. I do have a couple change requests, please take a look when you have a moment. Your hard work on this is much appreciated

transformer_lens/model_bridge/transformer_bridge.py
LightWork666: You're right, thanks for catching this. The bug was that `required_actual_keys` demanded every alias in a TL key's group be present, but the clean-key branch only ever writes one alias per key. Since aliases of the same TL key share the underlying storage, writing any one of them already updates what forward() reads for all of them, so treating the group as satisfied by any single alias is correct. Fixed in the latest commit: `missing_keys` now checks `any(k in mapped_state_dict for k in actual_keys)` per group instead of requiring the full union. Verified against your exact repro (gpt2, clean-key dict, strict=True): 0 missing, 0 unexpected now.

tests/unit/model_bridge/test_state_dict_round_trip.py
LightWork666: Added `test_native_clean_key_dict_with_partial_aliases_does_not_raise_strict`. No HF download needed; it uses `boot_native`'s own `_original_component` wrapping, which turns out to alias internally too (not just gpt2's c_attn split), so it exercises the same bug without needing the slow tier. Also added a second slow test (`test_boot_transformers_clean_key_dict_does_not_raise_strict`) that's your exact repro on real gpt2, so there's a permanent regression test for the specific case you hit too.

jlarson4:
@LightWork666 thanks for the updates! Looks good, merging now

PR #1602, opened 2026-08-04
fix(bridge): remove orphaned convert_weights override from nanogpt adapter

### LINKED ISSUE #1601: [Bug Report] nanogpt adapter's convert_weights calls a super method that no longer exists
### **Base on `dev-4.x`**

**Describe the bug**

`NanogptArchitectureAdapter.convert_weights` (`transformer_lens/model_bridge/supported_architectures/nanogpt.py:92-103`) ends with `return super().convert_weights(remote_module)`, but `ArchitectureAdapter` defines no `convert_weights` method — it was removed from the base class in `3efbd6e` ("Cleanup (#1129)", 2025-11-15). Any call therefore raises `AttributeError`.

The failure is invisible to CI because of the `# type: ignore[misc]` on `:103`. Removing that comment and running `uv run mypy .` reports the bug directly:

```
transformer_lens/model_bridge/supported_architectures/nanogpt.py:103: error: "convert_weights" undefined in superclass  [misc]
```

This is the failure mode [AGENTS.md §10](https://github.com/TransformerLensOrg/TransformerLens/blob/dev-4.x/AGENTS.md#10-hard-rules) ("Never add `# type: ignore`") exists to prevent- the suppression hid a real defect for ~9 months.

Two further problems in the same method:

1. The `_orig_mod.` prefix strip is a no-op for module inputs. `nn.Module.state_dict()` returns a fresh dict, so the `pop`/reassign loop mutates a throwaway copy; the original `remote_module` is then passed to `super()` unchanged.
2. The method has no callers. `grep -rn "convert_weights"` over `*.py` / `*.ipynb` / `*.md` (plus a check for dynamic `getattr` dispatch) finds only the two definitions — this one and `bd3lm.py:208` (`convert_weights(self) -> dict`, a different signature that returns `{}` and is likewise never called). Both are orphans left behind by the base-class removal.

**Code example**

```python
import torch
from torch import nn

from transformer_lens.model_bridge.supported_architectures.nanogpt import (
    NanogptArchitectureAdapter,
)

    d_model=64, d_head=16, n_layers=2, n_ctx=128, n_heads=4,
    d_vocab=1000, d_mlp=256, architecture="NanoGPTForCausalLM",
)


class FakeCompiledNanoGPT(nn.Module):
    """Stand-in for a torch.compile()'d nanoGPT checkpoint."""

    def state_dict(self, *args, **kwargs):
        return {"_orig_mod.transformer.wte.weight": torch.zeros(1000, 64)}


NanogptArchitectureAdapter(cfg).convert_weights(FakeCompiledNanoGPT())
# AttributeError: 'super' object has no attribute 'convert_weights'
```

**System Info**

 * Installed from source, at `dev-4.x` (`a6e00330`)
 * OS: Linux, CPU-only
 * Python 3.12.13; torch 2.10.0; transformers 5.13.0

Not environment-dependent - the base class simply has no such method.

`NanogptArchitectureAdapter` should not carry a method that raises on every call, and no adapter should need a `# type: ignore` to stay green.

Since nothing in the Bridge system calls `convert_weights`, the minimal correct fix is to delete the orphaned override at `nanogpt.py:92-103` together with its `# type: ignore[misc]`. Removal cannot regress behaviour: every call path through it currently raises.

Deliberately **not** proposed here: re-homing the `_orig_mod.` logic into `preprocess_weights` (`architecture_adapter.py:336`). That hook runs on `self.state_dict()` inside `process_weights` (`transformer_bridge.py:608`), i.e. after loading and with TL-renamed keys, so it never observes `_orig_mod.`-prefixed checkpoint keys. Re-homing would look like a fix while doing nothing.

**Related gap (separate issue, not this one)**

`TransformerBridge` consequently has no `_orig_mod.` handling for `torch.compile()`'d nanoGPT checkpoints, while the HookedTransformer path does (`transformer_lens/pretrained/weight_conversions/nanogpt.py:14`). That is HT→Bridge drift, but adding it is a feature with its own design questions (which load path should own the strip), so it doesn't belong in a dead-code removal. Happy to file it separately if that's wanted.

**Test coverage**

`NanogptArchitectureAdapter` has no unit or integration test, but that is by design rather than an oversight: it sits in `_COVERAGE_EXEMPT_MODULES` (`tests/unit/model_bridge/supported_architectures/test_thin_subclass_adapters.py:94`) alongside `mingpt`, `native`, and `neel_solu_old`, as internal-only adapters never published on the Hub. Noting this only to be explicit that the absent test is not itself the bug being reported.

That exemption is why the defect survived: with no test exercising the adapter and the `# type: ignore` suppressing the type error, nothing was positioned to notice. Once the ignore is gone, the CI `type-check` job catches any recurrence directly.

If you'd like structural unit coverage for this adapter as part of the fix, I'm happy to add it — note that doing so also means dropping `"nanogpt"` from `_COVERAGE_EXEMPT_MODULES`, since `test_coverage_exemptions_are_still_uncovered` is self-cleaning and fails while a covered module remains listed. That is a small policy call on an intentionally-exempt adapter, so I'd rather you make it than assume it.

Acceptance:
- [ ] Orphaned `convert_weights` override removed from `nanogpt.py`, along with its `# type: ignore[misc]`
- [ ] No `# type: ignore` remains in `nanogpt.py`
- [ ] `uv run mypy .` passes (and now fails if the override is reintroduced)
- [ ] `make unit-test` passes

### Checklist

sohv:
# Description

`NanogptArchitectureAdapter.convert_weights` ended in `return super().convert_weights(remote_module)`, but `ArchitectureAdapter` has had no `convert_weights` method since `3efbd6e` ("Cleanup (#1129)", 2025-11-15). Every call therefore raised `AttributeError: 'super' object has no attribute 'convert_weights'`.

The defect was invisible to CI because of the `# type: ignore[misc]` on that line. With the ignore removed, `uv run mypy .` reports it directly:

```
transformer_lens/model_bridge/supported_architectures/nanogpt.py:103: error: "convert_weights" undefined in superclass  [misc]
```

This PR deletes the orphaned override together with its ignore, and drops the now-unused `torch` import.

**Why removal rather than repair:**

- The method had no callers. A search over `*.py` / `*.ipynb` / `*.md`, plus a check for dynamic `getattr` dispatch, finds only two definitions of `convert_weights` in the tree — this one and `bd3lm.py:208` (a different signature returning `{}`, likewise never called). Both are orphans from the base-class removal.
- Its `_orig_mod.` prefix strip was a no-op regardless. `nn.Module.state_dict()` returns a fresh dict, so the `pop`/reassign loop mutated a throwaway copy and then passed the original `remote_module` to `super()`.
- Removal cannot regress behaviour, because every path through the method already raised.

**Deliberately not re-homed into `preprocess_weights`.** That hook runs on `self.state_dict()` inside `process_weights` (`transformer_bridge.py:608`) — after loading, with TL-renamed keys — so it would never observe `_orig_mod.`-prefixed checkpoint keys. Moving the logic there would look like a fix while doing nothing.

**Regression guard:** with the `# type: ignore` gone, the CI `type-check` job fails if the override is reintroduced.

**Adjacent gap, not addressed here:** `TransformerBridge` consequently has no `_orig_mod.` handling for `torch.compile()`'d nanoGPT checkpoints, while the HookedTransformer path does (`transformer_lens/pretrained/weight_conversions/nanogpt.py:14`). That is HT→Bridge drift, but adding it is a feature with its own design question (which load path should own the strip), so it is out of scope for a dead-code removal. Happy to file it separately.

**On unit coverage:** `nanogpt` sits in `_COVERAGE_EXEMPT_MODULES` (`tests/unit/model_bridge/supported_architectures/test_thin_subclass_adapters.py:94`) as an internal-only adapter, so no test file accompanies this change. I have a structural unit suite for the adapter ready if you want it — note that adding it also requires dropping `"nanogpt"` from that exemption set, since `test_coverage_exemptions_are_still_uncovered` is self-cleaning. That is a policy call on an intentionally-exempt adapter, so I did not make it unilaterally.

Fixes #1601

## Type of change

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas — the change is a deletion; the rationale is in the commit message and above
- [ ] I have made corresponding changes to the documentation — no documented behaviour changes; the removed method was unreachable
- [ ] I have added tests that prove my fix is effective or that my feature works — see "On unit coverage" above: `nanogpt` is an exempt module, and the `# type: ignore` removal makes the CI `type-check` job the regression guard. A structural unit suite is ready if you want the exemption retired.

### DIFF
--- a/transformer_lens/model_bridge/supported_architectures/nanogpt.py
+++ b/transformer_lens/model_bridge/supported_architectures/nanogpt.py
@@ -1,7 +1,5 @@
 from typing import Any
 
-import torch
-
 from transformer_lens.conversion_utils.conversion_steps import RearrangeTensorConversion
 from transformer_lens.conversion_utils.param_processing_conversion import (
     ParamProcessingConversion,
@@ -88,16 +86,3 @@ class NanogptArchitectureAdapter(ArchitectureAdapter):
             ),  # Final layer norm
             "unembed": UnembeddingBridge(name="lm_head"),
         }
-
-    def convert_weights(self, remote_module: Any) -> dict[str, torch.Tensor]:
-        # Nanogpt models saved after torch.compile() have this unwanted prefix
-        # This is a simple way to remove it
-        unwanted_prefix = "_orig_mod."
-        state_dict: dict[str, torch.Tensor] = (
-            remote_module.state_dict() if hasattr(remote_module, "state_dict") else remote_module
-        )
-        for k, v in list(state_dict.items()):
-            if k.startswith(unwanted_prefix):
-                state_dict[k[len(unwanted_prefix) :]] = state_dict.pop(k)
-
-        return super().convert_weights(remote_module)  # type: ignore[misc]


sohv:
Sibling instance of the same pattern, from the same commit: `ArchitectureAdapter._enable_ht_attention` (`architecture_adapter.py:1076-1126`) calls `_extract_linear_ht_format` (`:1100-1102`) and `_extract_qkv_neox_style` (`:1106`), whose definitions `3efbd6e` deleted while keeping the call sites — four `# type: ignore[attr-defined]` hide it. Only the GPT-2 branch works; the split-QKV (Llama/Qwen/Gemma) and NeoX branches raise `AttributeError`.

The same commit also removed its two callers (`architecture_adapter.py:1249` and `bridge.py:1144` as of `3efbd6e^`), so it is unreachable — no user impact. That makes `:1076-1171` dead: its three remaining helpers are called only from it, and `_disable_hook_conversions` is already `pass`.

Fold the removal into #1602, separate PR, or leave it? If component-level HT-vs-bridge weight comparison is still wanted, restoring the two extractors from `3efbd6e^` would be the right direction instead.

jlarson4:
@sohv in response to your question, feel free to make that as a separate PR, I am going to merge this because it looks great!

sohv:
@jlarson4 Thank you! I will create it as a separate PR and will request merging with the `dev-4.x` branch.

PR #1603, opened 2026-08-04
Reanchoring tests and benchmarks

jlarson4:
# Description

Re-anchors TransformerBridge's entire test and benchmark suite off the live `HookedTransformer` oracle, so the upcoming `Hooked*` removal can land without deleting the proof that the bridge is correct. After this PR, no test or benchmark constructs a `HookedTransformer`. The references come from three sources instead:

1. **Frozen goldens** – reference numerics (state dicts, logits, activation snapshots, hook manifests, ablation anchors) captured from `HookedTransformer` across 6 models × 4 processing configs, hosted on the HF Hub ([`lars4776/TL-Goldens`](https://huggingface.co/datasets/lars4776/TL-Goldens)) and pinned by commit revision in `tests/goldens.py`. The gpt2 golden anchors reproduce the published Main-Demo values (3.999 → 5.453) exactly.
2. **Raw HuggingFace models** – unprocessed-numerics and training parity compare against `AutoModelForCausalLM` (fp32, eager) directly.
3. **Intrinsic contracts** – properties that need no reference model (self-patch is a logits no-op, gated hooks capture the residual copy, bwd-hook gradients equal `retain_grad` ground truth, batch equals per-row forward).

Main changes:

- **Golden infrastructure**: capture script (`scripts/capture_ht_goldens.py`), revision-pinned loader + `GoldenCell` accessor (`tests/goldens.py`), skip-gated session fixtures.
- **Benchmarks (verify_models phases)**: every benchmark takes plain reference data (`reference_logits`, `reference_hooks`, `reference_activations`, …) instead of a `reference_model`; Phase 3's HF `log_softmax` fallback promoted to the only equivalence path; Phases 5/6 retired; `--no-ht-reference` removed; memory estimates recalibrated.
- **Tests**: ~40 files migrated onto goldens/HF/intrinsic anchors; 36 HT-only test files deleted; dead HT fixtures removed from conftests. A long-skipped `adapter=None ↔ adapter` weight-processing equivalence test is resurrected as an executable spec against the frozen state dicts (1e-5, fp64-verified noise floor), and the skipped backward-hook gradient test is resurrected against autograd.
- **Product fixes found by the re-anchoring**: `get_bridge_params`/`tl_parameters()` returned a wrongly-reshaped `W_V` for joint-QKV models and omitted LN params (corrupting `SVDInterpreter` results); `ActivationCache.remove_batch_dim`/`apply_slice_to_batch_dim` mishandled broadcast cache entries on batch>1. Weight-conversion and factory signatures widened from `HookedTransformerConfig` to `TransformerLensConfig`.
- **`lit/` rewritten onto the bridge**: `HookedTransformerLIT*` → `TransformerLensLIT*` (legacy aliases kept for one transition release); model validation is protocol-based; `from_pretrained` boots a compat-mode bridge.

## Type of change

# Checklist:

- [x] New and existing unit tests pass locally with my changes (unit 5093, integration/acceptance/benchmarks 1083+ against the goldens; mypy clean)


sohv:
Drive-by note from reading this against `tests/QUARANTINES.md` — not a blocker, just something that's easy to miss in a changeset this size.

This PR removes 35 test files, and six of them are inventoried in `QUARANTINES.md`, which isn't in the changeset. That leaves 10 rows pointing at files that no longer exist (line numbers on `dev-4.x`):

| QUARANTINES.md | Deleted file | Section |
|---|---|---|
| :144, :167 | `acceptance/test_hooked_transformer.py` | whole-file quarantine + 2 dtype `skipif`s |
| :71, :72 | `acceptance/test_multi_gpu.py` | hardware (2+ / 4+ CUDA) |
| :163 | `integration/test_tensor_extraction_consistency.py` | technical debt |
| :164 | `integration/test_tokenization_methods.py` | technical debt |
| :14, :65, :66 | `unit/components/test_attention.py` | bitsandbytes, CUDA half/bf16, MPS+torch 2.8.0 regression env |
| :15 | `unit/factories/test_mlp_factory.py` | bitsandbytes |

The file's own header says it's an "Inventory of every `skip` / `skipif` / `xfail` in `tests/`", and the ⚠️ convention is that whole-module `pytestmark` skips get a row, so keeping it in sync in the same PR seems in keeping with how it's maintained.

Two of these are worth a second look beyond bookkeeping, since deleting the test also drops the documented coverage note:

- `unit/components/test_attention.py:492` is the MPS + torch 2.8.0 regression env for pytorch#161640.
- `test_attention.py:130` / `test_mlp_factory.py:40` are the only bitsandbytes-gated tests in the inventory.

If that coverage moves somewhere under the new goldens layout, updating the rows to point at the new home would keep the inventory useful; if it's intentionally dropped, deleting the rows says so explicitly.

Also dangling after this lands, both referencing `tests/acceptance/test_hooked_transformer.py` by path: the root `AGENTS.md:34` warning block, and `docs/source/content/contributing.md` (the admonition at the top and the "Quarantined tests" section).

Happy to send a follow-up PR that just syncs `QUARANTINES.md` and those two docs once this merges, if that's easier than folding it in here- I have the un-quarantining of `test_hooked_encoder.py` / `test_hooked_encoder_decoder.py` in flight anyway (they turn out to be green; the "CI test pollution" label was covering several unrelated things). Those two files aren't touched by this PR so there's no conflict either way.

PR #1604, opened 2026-08-04
fix(bridge): remove dead _enable_ht_attention and its exclusive helpers

sohv:
# Description

Follow-up to #1602, as agreed in https://github.com/TransformerLensOrg/TransformerLens/issues/1601 - same defect pattern, same origin commit, separate PR.

`ArchitectureAdapter._enable_ht_attention` dispatched on three attention layouts, but two of the three branches called methods that do not exist anywhere in the tree:

| Branch | Attention layout | Helper called | Defined? |
|---|---|---|---|
| `c_attn` | GPT-2 | `_extract_qkv_gpt2_style` | yes |
| `q_proj` / `k_proj` / `v_proj` | Llama, Mistral, Qwen, Gemma, Phi, … | `_extract_linear_ht_format` | **no** |
| `query_key_value` | GPT-NeoX, Falcon | `_extract_qkv_neox_style` | **no** |

`3efbd6e` ("Cleanup (#1129)", 2025-11-15) deleted both helper definitions while keeping their call sites:

```
$ git show 3efbd6e -- transformer_lens/model_bridge/architecture_adapter.py | grep '^-.*def _extract'
-    def _extract_qkv_neox_style(self, query_key_value, n_heads, d_model, d_head):
-    def _extract_linear_ht_format(self, linear_module, n_heads, d_head, d_model):
```

Four `# type: ignore[attr-defined]` comments suppressed the resulting mypy errors, so CI stayed green. Only the GPT-2 branch could run; the other two raised `AttributeError` on entry.

**No user was affected.** The same commit also removed both callers (`architecture_adapter.py:1249` and `bridge.py:1144` as of `3efbd6e^`), leaving the method unreachable — verified by searching `*.py` / `*.ipynb` / `*.md` and checking for dynamic `getattr` dispatch. The risk was latent: the method advertises itself as "architecture-agnostic" in its docstring, so the next person wiring it into component testing would have hit an immediate crash on every architecture except GPT-2.

## What is removed

The whole unreachable cluster — 98 lines, the tail of the file:

- `_enable_ht_attention`
- `_extract_qkv_gpt2_style`, `_extract_output_proj`, `_disable_hook_conversions` — each called **only** from `_enable_ht_attention`, so they die with it. `_disable_hook_conversions` was already a no-op (`pass`, with a docstring saying it "currently does nothing in no_processing mode").
- the four `# type: ignore[attr-defined]` comments
- `import einops`, now unused (pruned by `make format`)

`setup_component_testing` and `_wire_rotary_for_testing` are untouched — they were never part of this path.

## If you would rather keep the capability

The opposite fix is equally available: restore `_extract_linear_ht_format` and `_extract_qkv_neox_style` from `3efbd6e^` and re-wire a caller. That is a feature with its own design questions (which call path owns it, how it relates to `setup_component_testing`), so I have not assumed it — happy to send that version instead if you prefer.

## Type of change

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas — this is a pure deletion; rationale is in the commit message and above
- [ ] I have made corresponding changes to the documentation — no documented behaviour changes; the removed code was unreachable
- [ ] I have added tests that prove my fix is effective or that my feature works — nothing to test: the code is unreachable and removed. Dropping the `# type: ignore` comments leaves the CI `type-check` job as the guard against reintroduction.

### DIFF
--- a/transformer_lens/model_bridge/architecture_adapter.py
+++ b/transformer_lens/model_bridge/architecture_adapter.py
@@ -4,7 +4,6 @@ This module contains the base class for architecture adapters that map between d
 """
 from typing import Any, Dict, Optional, cast
 
-import einops
 import torch
 
 from transformer_lens.config import TransformerBridgeConfig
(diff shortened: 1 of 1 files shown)


jlarson4:
Looks great, thank you @sohv!

sohv:
Great @jlarson4 , happy to contribute!

PR #1608, opened 2026-08-05
Fix masked causal loss in TransformerBridge

### LINKED ISSUE #1607: [Bug Report] TransformerBridge causal loss ignores attention_mask
**Describe the bug**

`TransformerBridge.forward(..., return_type="loss")` and `return_type="both"` ignore the supplied `attention_mask` when computing causal LM loss.

The mask is forwarded to the underlying model, but `BridgeCore._finalize_return()` calls `self.loss_fn(logits, input_ids, ...)` without passing it. Padding transitions therefore contribute to both the loss sum and denominator.

Changing only masked padding token IDs changes the reported loss, while the correct mask-aware loss remains identical.

**Code example**

```python
import torch

    d_model=32,
    d_head=8,
    n_heads=4,
    n_layers=2,
    n_ctx=6,
    d_vocab=32,
    d_mlp=64,
    act_fn="gelu",
    normalization_type="LN",
    seed=7,
    initializer_range=0.2,
)

mask = torch.tensor([
    [1, 1, 1, 0, 0, 0],
    [1, 1, 1, 1, 1, 1],
])
inputs = {
    "pad_a": torch.tensor([
        [1, 2, 3, 0, 0, 0],
        [4, 5, 6, 7, 8, 9],
    ]),
    "pad_b": torch.tensor([
        [1, 2, 3, 31, 30, 29],
        [4, 5, 6, 7, 8, 9],
    ]),
}

for name, tokens in inputs.items():
    with torch.no_grad():
        logits = bridge(tokens, attention_mask=mask, return_type="logits")
        actual = bridge(tokens, attention_mask=mask, return_type="loss")
        expected = bridge.loss_fn(logits, tokens, attention_mask=mask)
    print(name, "forward:", actual.item(), "masked:", expected.item())
```

Output:

```text
pad_a forward: 3.550006866455078 masked: 3.8386409282684326
pad_b forward: 4.128363132476807 masked: 3.8386409282684326
```

**Expected behavior**

`return_type="loss"` and `return_type="both"` should pass the forward-pass `attention_mask` into `loss_fn()`. Masked padding token IDs must not affect the result.

The same mask should also be applied when `loss_per_token=True`.

**System Info**

- Installed from source
- Branch: `dev-4.x`
- Commit: `f17dff30`
- Windows
- Python 3.12.10
- PyTorch 2.11.0+cpu
- transformers 5.13.0

**Additional context**

`BridgeCore.loss_fn()` and `lm_cross_entropy_loss()` already accept an `attention_mask`; it is lost between `TransformerBridge.forward()` and `_finalize_return()`.

Native Bridge also exposes a related left-padding edge case: a fully masked padding query softmaxs to NaN, and those NaNs can poison later layers. Multiplying masked losses by zero is insufficient because `NaN * 0` remains NaN. A complete fix needs NaN-safe masked loss handling and finite attention patterns for fully masked Native queries.

### Checklist

emerardd:
# Description

Propagate the forward-pass `attention_mask` into TransformerBridge causal loss so padding transitions do not affect `return_type="loss"`, `return_type="both"`, or per-token loss.

The fix also:

- moves the loss mask to the logits device when needed;
- uses NaN-safe masking for ignored token losses;
- zeros fully masked Native attention pattern entries so left padding cannot poison later layers; and
- corrects the per-token loss return annotation to reflect its `seq_len - 1` shape.

Fixes #1607

## Type of change

### Screenshots

Not applicable.

# Validation

- Regression tests: `4 passed`
- Affected Bridge unit tests: `102 passed, 1 skipped`
- Changed files: pycln, isort, and Black clean
- `git diff --check`: clean
- `uv run mypy .`: success on 422 source files

# Checklist

- [x] Documentation is not required for this internal correctness fix
- [ ] The complete unit-test suite passes locally; only the affected unit-test surface was run

### DIFF
--- a/tests/unit/model_bridge/test_loss_attention_mask.py
+++ b/tests/unit/model_bridge/test_loss_attention_mask.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import pytest
 import torch
+import torch.nn.functional as F
 
 from transformer_lens.config import TransformerBridgeConfig
 from transformer_lens.model_bridge import TransformerBridge
@@ -28,6 +29,16 @@ def _bridge() -> TransformerBridge:
 
 def _extract_loss(output: torch.Tensor | tuple[torch.Tensor, torch.Tensor]) -> torch.Tensor:
     return output[1] if isinstance(output, tuple) else output
+
+
+def _manual_masked_loss(
+    logits: torch.Tensor, tokens: torch.Tensor, attention_mask: torch.Tensor
+) -> torch.Tensor:
+    transition_mask = attention_mask[:, :-1].bool() & attention_mask[:, 1:].bool()
+    return F.cross_entropy(
+        logits[:, :-1][transition_mask],
+        tokens[:, 1:][transition_mask],
+    )
 
 
 @pytest.mark.parametrize("return_type", ["loss", "both"])
@@ -59,7 +70,7 @@ def test_forward_loss_ignores_masked_padding_tokens(return_type: str) -> None:
         output = bridge(tokens, attention_mask=attention_mask, return_type=return_type)
         loss = _extract_loss(output)
         logits = bridge(tokens, attention_mask=attention_mask, return_type="logits")
-        expected = bridge.loss_fn(logits, tokens, attention_mask=attention_mask)
+        expected = _manual_masked_loss(logits, tokens, attention_mask)
 
         torch.testing.assert_close(loss, expected)
         losses.append(loss)
@@ -125,3 +136,65 @@ def test_forward_loss_is_finite_with_left_padding() -> None:
 
     assert torch.isfinite(logits).all()
     assert torch.isfinite(loss)
+
+
+@pytest.mark.parametrize("mask_kind", ["bool", "additive"])
+@pytest.mark.parametrize("mask_layout", ["key_only", "causal"])
+def test_forward_loss_accepts_equivalent_4d_attention_mask(
+    mask_kind: str, mask_layout: str
+) -> None:
+    bridge = _bridge()
+    tokens = torch.tensor(
+        [
+            [1, 2, 3, 0, 0, 0],
+            [4, 5, 6, 7, 8, 9],
+        ]
+    )
+    attention_mask = torch.tensor(
+        [
+            [1, 1, 1, 0, 0, 0],
+            [1, 1, 1, 1, 1, 1],
+        ]
+    )
+    blocked = ~attention_mask.bool()[:, None, None, :]
+    if mask_layout == "causal":
+        blocked = blocked | torch.ones(6, 6, dtype=torch.bool).triu(1)[None, None]
+    attention_mask_4d = blocked if mask_kind == "bool" else blocked.float() * -10_000.0
+
+    logits_2d, loss_2d = bridge(
+        tokens,
+        attention_mask=attention_mask,
+        return_type="both",
+    )
+    logits_4d, loss_4d = bridge(
+        tokens,
+        attention_mask=attention_mask_4d,
+        return_type="both",
+    )
+
+    torch.testing.assert_close(logits_4d, logits_2d, rtol=0, atol=0)
+    torch.testing.assert_close(loss_4d, loss_2d)
+    torch.testing.assert_close(loss_4d, _manual_masked_loss(logits_4d, tokens, attention_mask))
+
+
+def test_loss_fn_reduces_rectangular_cached_4d_attention_mask() -> None:
+    bridge = _bridge()
+    tokens = torch.tensor([[4, 5]])
+    logits = torch.zeros(1, 2, 32)
+    logits[0, 0, 5] = 2.0
+    cache_and_new_mask = torch.tensor([[0, 1, 1, 1, 1, 1]])
+    key_blocked = ~cache_and_new_mask.bool()[:, None, None, :]
+    query_positions = torch.tensor([4, 5])
+    causal = torch.arange(6)[None, None, None, :] > query_positions[None, None, :, None]
+    attention_mask_4d = key_blocked | causal
+
+    loss = bridge.loss_fn(
+        logits,
+        tokens,
+        attention_mask=attention_mask_4d,
+        per_token=True,
+    )
+
+    expected = F.cross_entropy(logits[:, 0], tokens[:, 1])
+    torch.testing.assert_close(loss, expected.reshape(1, 1))
+    assert loss.shape == (1, 1)
--- a/tests/unit/model_bridge/test_loss_attention_mask.py
+++ b/tests/unit/model_bridge/test_loss_attention_mask.py
@@ -1,0 +1,127 @@
+"""Regression tests for padding-aware TransformerBridge causal loss."""
+
+from __future__ import annotations
+
+import pytest
+import torch
+
+from transformer_lens.config import TransformerBridgeConfig
+from transformer_lens.model_bridge import TransformerBridge
+
+
+def _bridge() -> TransformerBridge:
+    cfg = TransformerBridgeConfig(
+        d_model=32,
+        d_head=8,
+        n_heads=4,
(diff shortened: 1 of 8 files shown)


transformer_lens/utilities/lm_utils.py
```diff
@@ -37,7 +37,7 @@ def lm_cross_entropy_loss(
         # Ignore token positions which are masked out or where the next token is masked out
         # (generally padding tokens)
         next_token_mask = torch.logical_and(attention_mask[:, :-1], attention_mask[:, 1:])
-        predicted_log_probs *= next_token_mask
+        predicted_log_probs = predicted_log_probs.masked_fill(~next_token_mask, 0.0)
```
jlarson4, line 46: Swapping in-place `*=` for `masked_fill` drops an implicit length guard: `*=` raised when the mask was longer than the scored tokens, but `masked_fill` broadcasts both operands and replicates the value instead. Now that the mask is forwarded, `bridge(new_tokens, attention_mask=<cache+new>, past_key_values=cache, return_type="loss", loss_per_token=True)` returns shape `(1,5)` with one value repeated five times, but previously it returned `(1,1)`, and the scalar variant raises. Could you add an explicit `attention_mask.shape[1] == tokens.shape[1]` assertion here so the mismatch fails loudly again?

transformer_lens/utilities/lm_utils.py
```diff
@@ -17,7 +17,7 @@ def lm_cross_entropy_loss(
     tokens: Int[torch.Tensor, "batch pos"],
     attention_mask: Optional[Int[torch.Tensor, "batch pos"]] = None,
     per_token: bool = False,
-) -> Union[Float[torch.Tensor, ""], Float[torch.Tensor, "batch pos"]]:
+) -> Union[Float[torch.Tensor, ""], Float[torch.Tensor, "batch pos_minus_one"]]:
```
jlarson4: `pos_minus_one` is a fresh jaxtyping name bound only by the return value, so it matches *any* length including `pos`, the exact off-by-one it exists to catch. The repo already spells this axis symbolically as `"batch pos-1"`, which binds against `pos` from the arguments and so actually catches it. Can you use that form to match?

transformer_lens/utilities/lm_utils.py
jlarson4: `lm_accuracy`'s `per_token=True` branch is annotated `Float[torch.Tensor, "batch pos"]` but returns a **bool** tensor of shape `[batch, pos-1]`. It is wrong in both dtype and shape, and its own docstring says `[batch, seq_len-1]`. It raises `BeartypeCallHintReturnViolation` today under the repo's pytest config, and it is public API via `transformer_lens.utils`, reachable from Bridge caches. Since you're already correcting the sibling function in this file, could you fix it here too?

transformer_lens/model_bridge/bridge_core.py
```diff
@@ -384,7 +387,12 @@ def _finalize_return(
                 )
             assert isinstance(logits, torch.Tensor), f"Expected logits tensor, got {type(logits)}"
             assert input_ids is not None, "input_ids required for return_type='loss'"
-            return self.loss_fn(logits, input_ids, per_token=loss_per_token)
+            return self.loss_fn(
```
jlarson4, line 451: The Bridge's native forward documents 2-D and 4-D masks as equivalent (`sources/native/model.py:344-349`), and they are. On the issue's own tiny bridge, a 4-D bool encoding of the same padding gives **bit-identical logits** to the 2-D form. But the loss path now distinguishes them: at this head the 2-D call returns the correct `3.8386409282684326` while the equivalent 4-D call raises `RuntimeError: The size of tensor a (6) must match the size of tensor b (5)` (on `dev-4.x` it returned a number). The same unvalidated forwarding is what turns the KV-cache case in `lm_utils.py` on line 40 from a number into a crash. Could you slice/reduce the mask to the scored `[batch, pos]` window before `loss_fn`?

tests/unit/model_bridge/test_loss_attention_mask.py
```diff
@@ -0,0 +1,127 @@
...
+        attention_mask=attention_mask,
+        return_type="loss",
+        loss_per_token=True,
+    )
+    next_token_mask = torch.logical_and(attention_mask[:, :-1], attention_mask[:, 1:])
+
+    assert torch.count_nonzero(loss[~next_token_mask]) == 0
+
+
+def test_forward_loss_is_finite_with_left_padding() -> None:
```
jlarson4, line 107: Every assertion in the file is bridge-vs-bridge. A masked loss with the wrong denominator or mask shift would satisfy all four tests, this function in particular is value-blind. And the `lm_utils.py:40` NaN-safety hunk has no test. As a regression check I reverted it to `*=` and all 4 tests stayed green (the native hunk alone keeps the logits finite). Could you add one anchor against an externally-computed value, staying inside the Bridge?

jlarson4 (commented):
Thanks for reporting this and putting together a fix! There are a couple tricky areas with this mask, I put a couple notes in. Let me know if you have any questions.

Additionally, in `remote_bridge.py`, there is another calcite of this helper and it still passes no `attention_mask`, so `RemoteBridge.forward(..., return_type="loss")` still has the bug. The mask is already in scope in `**kwargs`. Could you add `attention_mask=kwargs.get("attention_mask")` to the call at line 113?

transformer_lens/utilities/lm_utils.py
emerardd: Addressed in df5f46d0. `lm_cross_entropy_loss` now explicitly asserts that `attention_mask.shape == tokens.shape`, while `BridgeCore` first slices or reduces valid cache plus new masks to the scored token window. I also added mismatch and cached-window regression coverage.

transformer_lens/utilities/lm_utils.py
emerardd: Addressed in df5f46d0. The per-token loss return annotation now uses `batch pos-1`, binding the result length to the input `pos` dimension.

transformer_lens/utilities/lm_utils.py
emerardd: Addressed in df5f46d0. `lm_accuracy(per_token=True)` is now annotated as `Bool[Tensor, batch pos-1]`, and a runtime test checks both its boolean dtype and `[batch, pos-1]` shape.

transformer_lens/model_bridge/bridge_core.py
emerardd: Addressed in df5f46d0. `BridgeCore` now normalizes 2-D, 4-D bool or additive, and cache plus new masks to the scored `[batch, pos]` window before calling `loss_fn`. Tests cover key-only and full-causal 4-D masks, bit-identical 2-D or 4-D logits, equal manually anchored losses, and rectangular cached 4-D masks.

tests/unit/model_bridge/test_loss_attention_mask.py
emerardd: Addressed in df5f46d0. The Bridge regression now computes the expected value independently with `torch.nn.functional.cross_entropy` over manually selected valid transitions instead of calling `bridge.loss_fn`. I also added a direct NaN-masking regression that fails with multiplication and passes with `masked_fill`.

emerardd:
Thanks for the detailed review. I pushed df5f46d0 addressing all five inline threads. The review-body RemoteBridge issue is also fixed: `RemoteBridge.forward` now passes `attention_mask=kwargs.get("attention_mask")` into `_finalize_return`, with a cache-plus-new masked-loss regression. Validation: 40 focused tests passed, formatting checks passed, and `uv run mypy .` reports no issues in 422 source files. I left the threads unresolved for reviewer verification.

jlarson4:
Awesome, thank you @emerardd. I will do a re-review shortly.

jlarson4:
Great work on this @emerardd, no additional change requests on my end, merging now!

PR #1610, opened 2026-08-05
fix(bridge): derive position_ids from attention_mask for left-padded input

### LINKED ISSUE #1609: [Bug Report] TransformerBridge ignores attention_mask when deriving position_ids, so left-padded input silently returns wrong logits and loss
### **Describe the bug**

`TransformerBridge.forward()` does not derive `position_ids` from a supplied `attention_mask`, so left-padded input silently gets the wrong absolute positions. No error, no NaN — just wrong logits and a wrong loss. `HookedTransformer` handles this correctly, so the same code returns different numbers after migrating to the bridge.

**Not a duplicate of #1607**  - That issue is the `attention_mask` never reaching `loss_fn`, which affects loss aggregation only but the logits are correct. This is the mask never reaching `position_ids`, which corrupts the logits themselves and the loss downstream. They are independent and they compound — with #1608 applied (which fixes #1607), max |logit diff| here is still `1.361e+02`, and with this fix alone the loss is still wrong by `1.343308`. Both are needed for correct left-padded batches; measurements in "Interaction with #1607 / #1608" below.

gpt2, one prompt, mask supplied, varying the number of left pads:

| n_pad | HT loss | Bridge loss | HT drift | Bridge drift |
|---|---|---|---|---|
| 0 | 4.503170 | 4.503170 | — | — |
| 1 | 4.503169 | 13.594296 | 1.4e-06 | **9.09** |
| 2 | 4.503170 | 11.478184 | 4.8e-07 | **6.98** |
| 3 | 4.503169 | 11.154946 | 9.5e-07 | **6.65** |
| 5 | 4.503170 | 11.113400 | 4.8e-07 | **6.61** |
| 8 | 4.503169 | 10.787075 | 1.4e-06 | **6.28** |

Right padding is unaffected for both (drift ≤ 9.5e-07), which is expected — causality already protects it. The failure is specific to left padding, where every real token's absolute position shifts by the pad count.

Not model-specific. distilgpt2, `n_pad=3`: HT drift `4.8e-07`, Bridge drift **`4.417137`**, max |logit diff| **`5.791e+01`**.

The magnitude scales with how much padding a row carries. A realistic left-padded batch from the HF tokenizer, comparing each prompt alone against the same prompt inside the batch:

| prompt | pads | HT drift | Bridge drift |
|---|---|---|---|
| "The capital of France is Paris and it is beautiful" | 0 | 2.5e-06 | 0.00000 |
| "Hello world" | 8 | 4.3e-06 | **2.95003** |
| "Machine learning models require careful evaluation" | 4 | 1.8e-06 | **6.80700** |

**This also breaks `enable_compatibility_mode()`**, whose documented contract is "HookedTransformer-equivalent numerics". On unpadded input compat mode matches HT exactly (max |logit diff| `0.000e+00`). Left-padded it does not: loss `4.503170` → `11.154946`.

**Root cause**

`transformer_bridge.py:1532` derives `position_ids` only for batched *list* input:

```python
# Auto-compute attention_mask + position_ids for batched list input
# when the caller didn't supply them. Matches HF generation convention.
if (
    _is_batched_list
    and attention_mask is None
    ...
):
    ...
    position_ids = attention_mask.long().cumsum(-1) - 1
    position_ids.masked_fill_(attention_mask == 0, 1)
    kwargs["position_ids"] = position_ids
```

Pre-tokenized tensor input never reaches that branch, so HF falls back to a plain `arange` and the padding offset is never removed. Supplying `position_ids` by hand fixes it exactly — max |logit diff| drops from `1.361e+02` to `9.918e-05`.

That also makes the bridge inconsistent with itself: the same batch gives different logits depending on whether you passed strings or token IDs. Max |logit diff| between the two input forms for identical content: `4.142e+01`.

**Code example**

```python
import torch

ht = HookedTransformer.from_pretrained("gpt2", device="cpu").eval()
br = TransformerBridge.boot_transformers("gpt2", device="cpu")
br.eval()

seq = ht.to_tokens("The capital of France is", prepend_bos=False)
L, n_pad = seq.shape[1], 3
pad = ht.tokenizer.eos_token_id

left = torch.cat([torch.full((1, n_pad), pad), seq], dim=1)
mask = torch.cat([torch.zeros(1, n_pad, dtype=torch.long),
                  torch.ones(1, L, dtype=torch.long)], dim=1)

with torch.no_grad():
    print("HT     unpadded :", ht(seq, return_type="loss").item())
    print("HT     left-pad :", ht(left, attention_mask=mask, return_type="loss").item())
    print("Bridge unpadded :", br(seq, return_type="loss").item())
    print("Bridge left-pad :", br(left, attention_mask=mask, return_type="loss").item())

# HT     unpadded : 4.503170
# HT     left-pad : 4.503169   <- invariant, correct
# Bridge unpadded : 4.503170
# Bridge left-pad : 11.154946  <- silently wrong
```

**System Info**

 * Installed from source, at `dev-4.x` (`f17dff30`)
 * OS: macOS 26.5 (arm64), CPU-only
 * Python 3.12.13; torch 2.11.0; transformers 5.13.0

Reproduced on three independent paths — `boot_native`, `boot_transformers`, and `enable_compatibility_mode()` — and on two models (gpt2, distilgpt2). Deterministic across repeats (identical to 1e-9).

**Additional context**

*Scope, and an honest caveat.* Raw HuggingFace behaves the same way: `GPT2LMHeadModel(left_tok, attention_mask=mask)` without `position_ids` also drifts by `1.361e+02`, and is correct once `position_ids` are passed. So the bridge is faithfully passing HF's convention through, and one could argue this is HF's contract rather than a bridge defect.

I think it is still a bug here, for three reasons:

1. `HookedTransformer` deliberately corrects it (`pos_embed(tokens, pos_offset, attention_mask)`), and the migration guide actively pushes users from HT to the bridge. Same code, different numbers, no warning.
2. The bridge already applies exactly this correction for list input, so the two input forms disagree with each other.
3. `enable_compatibility_mode()` documents "HookedTransformer-equivalent numerics", and this breaks that contract specifically.

When an `attention_mask` is supplied, `position_ids` are absent, and the mask indicates left padding, derive positions from the mask — the same two lines the list-input branch already uses:

```python
if (
    attention_mask is not None
    and "position_ids" not in kwargs
    and not _is_inputs_embeds
    and attention_mask.ndim == 2
    and bool((attention_mask[:, 0] == 0).any())
):
    _pos = attention_mask.long().cumsum(-1) - 1
    kwargs["position_ids"] = _pos.masked_fill(attention_mask == 0, 1)
```

Locally this restores per-row parity with HT exactly (drift `0.00000` on every row of the batch above), with `tests/unit/model_bridge` + `tests/unit/test_tokenizer_padding_side.py` green (3955 passed, 27 skipped, 10 xfailed).

**Interaction with #1607 / #1608 — these compound**

Two independent bugs on the same path. Measured across four code states, gpt2, aggregate loss on the left-padded batch above (HT reference `4.814578`):

| state | batch loss | max \|logit diff\| | compat drift |
|---|---|---|---|
| `dev-4.x` today | 7.254170 | 1.361e+02 | 6.651776 |
| #1608 alone | 7.411356 | 1.361e+02 | 11.115239 |
| position fix alone | 6.688155 | **9.918e-05** | 1.343308 |
| both | **4.814578** | **9.918e-05** | **0.000001** |

They separate cleanly: the position fix corrects the *logits* (`1.361e+02` → `9.918e-05`), #1608 corrects the *loss aggregation*, and only both together give correct batched loss.

Worth flagging that **#1608 alone increases the reported drift** (6.65 → 11.12) — two errors were partially cancelling. That is not a regression in #1608; it is an argument for landing both, and it would look like one if #1608 were benchmarked on padded batches in isolation.

Also: on the Native bridge, #1608's masked-softmax fix is a prerequisite for even observing this one — without it a fully-masked query row NaNs and the left-padded logits come back `nan` rather than merely wrong.

**Deliberately not proposed here**

`generate()` with left padding. `HookedTransformer.generate` has no `attention_mask` parameter (only `**generation_kwargs`), so a mask passed there is silently swallowed and I could not construct a valid comparison. There may or may not be a problem; I have no evidence either way and would rather not guess.

Also not proposed: changing the `_is_batched_list` branch itself. It is correct as written; this is about extending the same correction to tensor input.

**Test coverage**

No test asserts left-padding invariance for the bridge's forward pass, which is why this is invisible. A regression test is cheap and needs no real checkpoint — `boot_native` shows it — though a gpt2 test is the one that would have caught the compat-mode contract break.

Acceptance:
- [ ] `position_ids` derived from `attention_mask` for tensor input, matching the existing list-input behaviour
- [ ] Left-padding invariance regression test for `forward` / `return_type="loss"`
- [ ] Compat-mode parity test against `HookedTransformer` on a left-padded batch

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/TransformerLens/issues) in the repo (**required**) — searched `left padding`, `position_ids`, `padding_side`; nothing open or closed covers this

sohv:
## Description

Fixes #1609.

`TransformerBridge.forward()` did not derive `position_ids` from a supplied `attention_mask`, so left-padded input silently got the wrong absolute positions — no error, no NaN, just wrong logits and a wrong loss.

On gpt2, one prompt, mask supplied:

| n_pad | HT loss | Bridge loss (before) | HT drift | Bridge drift |
|---|---|---|---|---|
| 0 | 4.503170 | 4.503170 | — | — |
| 1 | 4.503169 | 13.594296 | 1.4e-06 | **9.09** |
| 3 | 4.503169 | 11.154946 | 9.5e-07 | **6.65** |
| 8 | 4.503169 | 10.787075 | 1.4e-06 | **6.28** |

Right padding was never affected (drift ≤ 9.5e-07) — causality already protects it.

`transformer_bridge.py` derived `position_ids` only for batched *list* input, so pre-tokenized tensors fell through to HF's plain `arange` and the padding offset was never removed. This extends the same correction that branch already applies. An explicitly supplied `position_ids` still wins.

Two consequences this also fixes:

- The bridge was inconsistent with itself — the same batch gave different logits depending on whether it was passed as strings or token IDs (max |logit diff| `4.142e+01`).
- `enable_compatibility_mode()`, which documents "HookedTransformer-equivalent numerics", matched HT exactly on unpadded input (`0.000e+00`) but diverged on left-padded input.

### Tests

Adds `tests/integration/model_bridge/test_left_padding_positions.py`: logit invariance under both padding sides, the same property in compatibility mode, and agreement between the derived and an explicitly supplied `position_ids`.

Red-before / green-after: **9 passed** with the fix, **5 failed** without it. The right-padding cases are controls — they pass in both states, so the tests are specific to the bug rather than to padding in general.

These sit in the integration tier rather than the unit tier deliberately: left padding produces a fully masked query row, which the Native attention path turns into `NaN` until the masked-softmax fix in #1608 lands, so `boot_native` cannot express the property yet.

### Verification

- New tests: 9 passed / 5 failed without the fix
- `tests/unit/model_bridge` + `tests/unit/test_tokenizer_padding_side.py`: 3955 passed, 27 skipped, 10 xfailed
- `pycln` / `isort` / `black` clean; `mypy` clean

### Relationship to #1607 / #1608

Independent bugs on the same path that compound. This is measured across four states (gpt2, aggregate loss on a left-padded batch; HT reference `4.814578`):

| state | batch loss | max \|logit diff\| |
|---|---|---|
| `dev-4.x` | 7.254170 | 1.361e+02 |
| #1608 alone | 7.411356 | 1.361e+02 |
| this PR alone | 6.688155 | **9.918e-05** |
| both | **4.814578** | **9.918e-05** |

This PR fixes the **logits**; #1608 fixes the **loss aggregation**. Batched loss is only correct with both, so reviewing this one in isolation will still show a wrong aggregate. No file overlap, so they merge in either order.

### Type of change

## Checklist:

### DIFF
--- a/transformer_lens/model_bridge/transformer_bridge.py
+++ b/transformer_lens/model_bridge/transformer_bridge.py
@@ -1551,6 +1551,37 @@ class TransformerBridge(BridgeCore, HookIntrospectionMixin, nn.Module):
                     position_ids.masked_fill_(attention_mask == 0, 1)
                     kwargs["position_ids"] = position_ids
 
+            # Any masked-out token shifts the absolute position of every real token
+            # after it, so positions must be derived from the mask rather than left
+            # to HF's default arange. This is the same derivation HookedTransformer
+            # applies in pos_embed; without it the bridge silently returns wrong
+            # logits. An all-ones mask reduces to arange, so this is a no-op there.
+            #
+            # The mask spans any cached prefix as well as the new tokens, so it is
+            # offset back to just the tokens actually being passed — matching how
+            # AbstractAttention/PosEmbed use past_kv_pos_offset.
+            if (
+                attention_mask is not None
+                and "position_ids" not in kwargs
+                and not _is_inputs_embeds
+                and attention_mask.ndim == 2
+                and isinstance(input_ids, torch.Tensor)
+                and input_ids.ndim == 2
+                and attention_mask.shape[1] >= input_ids.shape[1]
+            ):
+                _derived = utils.get_offset_position_ids(0, attention_mask)
+                _arange = torch.arange(attention_mask.shape[1], device=_derived.device)
+                # Only intervene when the mask actually moves an attended token off
+                # its default position — i.e. some masked token precedes a real one
+                # (left padding, or an interior gap). Pure right padding and an
+                # all-ones mask already agree with arange, and injecting position_ids
+                # there would be a no-op at best and unsupported by models whose
+                # forward does not take them at worst.
+                if bool(((_derived != _arange) & (attention_mask != 0)).any()):
+                    kwargs["position_ids"] = _derived[
+                        :, attention_mask.shape[1] - input_ids.shape[1] :
+                    ]
+
             if attention_mask is not None:
                 kwargs["attention_mask"] = attention_mask
             if past_key_values is not None:
--- a/transformer_lens/model_bridge/transformer_bridge.py
+++ b/transformer_lens/model_bridge/transformer_bridge.py
@@ -1383,6 +1383,82 @@ class TransformerBridge(BridgeCore, HookIntrospectionMixin, nn.Module):
         residual = self.forward(tokens, stop_at_layer=0, attention_mask=attention_mask)
         return residual, tokens, None, attention_mask
 
+    def _accepts_derived_position_ids(self) -> bool:
+        """Whether it is safe to hand the wrapped model a mask-derived ``position_ids``.
+
+        Two families of model must be left alone, so the injection below is
+        gated on the target the same way ``output_attentions`` is in
+        :meth:`BridgeCore.run_with_cache`:
+
+        * **Fixed-signature models.** Remote-code forwards such as
+          ``LLaDAModelLM.forward`` take neither ``position_ids`` nor
+          ``**kwargs``, so passing it raises ``TypeError`` where the model
+          previously returned logits.
+        * **Models that own their position derivation.** mRoPE architectures
+          (Qwen2-VL, Qwen2.5-VL, Qwen3-VL, GLM-4V) build a 3-D temporal /
+          height / width index in ``get_rope_index``, and only while
+          ``position_ids is None``; a supplied 2-D tensor is silently expanded
+          across all three streams instead. Their derivation already scatters
+          positions onto attended slots only, so it handles left padding
+          correctly on its own and needs no help from us.
+        * **Mask-consuming positional embeddings.** OPT's
+          ``OPTLearnedPositionalEmbedding.forward`` takes the mask and derives
+          the same positions we would, so injection buys nothing — but it does
+          replace the model's own padding-slot convention with ours, which
+          shows up as a whole-tensor diff.
+
+        Non-torch drivers (vLLM, Inspect) expose no module to introspect and
+        manage positions internally, so they are excluded as well.
+        """
+        underlying = getattr(self._driver, "underlying_model", None)
+        if underlying is None:
+            return False
+
+        cached = self.__dict__.get("_derived_position_ids_ok")
+        if cached is not None and cached[0] is underlying:
+            return bool(cached[1])
+
+        def verdict() -> bool:
+            fwd_params = inspect.signature(underlying.forward).parameters
+            if "position_ids" not in fwd_params and not any(
+                p.kind is inspect.Parameter.VAR_KEYWORD for p in fwd_params.values()
+            ):
+                return False
+
+            # ``get_rope_index`` lives on the inner text model, not the
+            # ForConditionalGeneration wrapper that is usually original_model.
+            for module in (
+                underlying,
+                getattr(underlying, "model", None),
+                getattr(underlying, "language_model", None),
+            ):
+                if module is not None and hasattr(module, "get_rope_index"):
+                    return False
+
+            # Config-level backstop for mRoPE models that spell the derivation
+            # differently; the section list is what makes positions 3-D.
+            config = getattr(underlying, "config", None)
+            for candidate in (config, getattr(config, "text_config", None)):
+                scaling = getattr(candidate, "rope_scaling", None)
+                if isinstance(scaling, dict) and "mrope_section" in scaling:
+                    return False
+
+            # A positional embedding that takes the mask derives positions for
+            # itself. Only embeddings that override nn.Embedding.forward are
+            # worth inspecting, which keeps this to a handful per model.
+            for module in underlying.modules():
+                if not isinstance(module, nn.Embedding):
+                    continue
+                if type(module).forward is nn.Embedding.forward:
+                    continue
+                if "attention_mask" in inspect.signature(module.forward).parameters:
+                    return False
+            return True
+
+        accepts = verdict()
+        self.__dict__["_derived_position_ids_ok"] = (underlying, accepts)
+        return accepts
+
     def forward(
         self,
         input: Union[str, List[str], torch.Tensor],
@@ -1568,17 +1644,22 @@ class TransformerBridge(BridgeCore, HookIntrospectionMixin, nn.Module):
                 and isinstance(input_ids, torch.Tensor)
                 and input_ids.ndim == 2
                 and attention_mask.shape[1] >= input_ids.shape[1]
+                and self._accepts_derived_position_ids()
             ):
-                _derived = utils.get_offset_position_ids(0, attention_mask)
+                # .long() because callers may hand in a float 0/1 mask, and
+                # positions index an embedding table.
+                _derived = utils.get_offset_position_ids(0, attention_mask.long())
                 _arange = torch.arange(attention_mask.shape[1], device=_derived.device)
-                # Only intervene when the mask actually moves an attended token off
-                # its default position — i.e. some masked token precedes a real one
-                # (left padding, or an interior gap). Pure right padding and an
-                # all-ones mask already agree with arange, and injecting position_ids
-                # there would be a no-op at best and unsupported by models whose
-                # forward does not take them at worst.
-                if bool(((_derived != _arange) & (attention_mask != 0)).any()):
-                    kwargs["position_ids"] = _derived[
+                # Decide per row, not per batch. A row only needs the derived
+                # positions when its mask actually moves one of its attended
+                # tokens off the default position — i.e. a masked token precedes
+                # a real one (left padding, or an interior gap). Rows that are
+                # unpadded or purely right-padded keep arange verbatim, so one
+                # left-padded row in a batch cannot perturb its neighbours.
+                _needs = ((_derived != _arange) & (attention_mask != 0)).any(dim=1, keepdim=True)
+                if bool(_needs.any()):
+                    _positions = torch.where(_needs, _derived, _arange.expand_as(_derived))
+                    kwargs["position_ids"] = _positions[
                         :, attention_mask.shape[1] - input_ids.shape[1] :
                     ]
 
--- a/tests/integration/model_bridge/test_left_padding_positions.py
+++ b/tests/integration/model_bridge/test_left_padding_positions.py
@@ -1,0 +1,168 @@
+"""Regression tests for left-padding position handling in TransformerBridge.
+
+A causal LM's logits at a sequence's real token positions must not depend on how
+that sequence is padded, provided the caller supplies the matching attention_mask.
+Left padding shifts every real token's absolute position, so position_ids have to
+be derived from the mask; without that the bridge silently returns wrong logits
+and a wrong loss. See #1609.
+
+Right padding is included as a control: causality already protects it, so it was
+never affected and must stay that way.
+"""
+
+from __future__ import annotations
+
+import pytest
+import torch
+
+from transformer_lens import utilities as utils
+
+PAD_ID = 0
+
+
+def _pad(tokens: torch.Tensor, n_pad: int, side: str) -> tuple[torch.Tensor, torch.Tensor]:
+    """Pad `tokens` on `side`, returning (padded_tokens, attention_mask)."""
+    pads = torch.full((tokens.shape[0], n_pad), PAD_ID, dtype=tokens.dtype)
+    ones = torch.ones(tokens.shape, dtype=torch.long)
+    zeros = torch.zeros((tokens.shape[0], n_pad), dtype=torch.long)
+    if side == "left":
+        return torch.cat([pads, tokens], dim=1), torch.cat([zeros, ones], dim=1)
+    return torch.cat([tokens, pads], dim=1), torch.cat([ones, zeros], dim=1)
+
+
+@pytest.fixture(scope="module")
+def tokens(distilgpt2_bridge) -> torch.Tensor:
+    return distilgpt2_bridge.to_tokens("The capital of France is")
+
+
+@pytest.mark.parametrize("side", ["left", "right"])
+@pytest.mark.parametrize("n_pad", [1, 3, 5])
+def test_logits_are_invariant_to_padding(distilgpt2_bridge, tokens, side, n_pad) -> None:
+    """Padding must not change the logits at a sequence's real positions."""
+    padded, mask = _pad(tokens, n_pad, side)
+    real = slice(n_pad, None) if side == "left" else slice(None, tokens.shape[1])
+
+    with torch.no_grad():
+        baseline = distilgpt2_bridge(tokens, return_type="logits")
+        actual = distilgpt2_bridge(padded, attention_mask=mask, return_type="logits")[:, real]
+
+    assert torch.isfinite(actual).all()
+    torch.testing.assert_close(actual, baseline, rtol=1e-3, atol=1e-3)
+
+
+@pytest.mark.parametrize("side", ["left", "right"])
+def test_compat_mode_logits_are_invariant_to_padding(distilgpt2_bridge_compat, side: str) -> None:
+    """enable_compatibility_mode() promises HookedTransformer-equivalent numerics,
+    which this property is part of."""
+    tokens = distilgpt2_bridge_compat.to_tokens("The capital of France is")
+    n_pad = 3
+    padded, mask = _pad(tokens, n_pad, side)
+    real = slice(n_pad, None) if side == "left" else slice(None, tokens.shape[1])
+
+    with torch.no_grad():
+        baseline = distilgpt2_bridge_compat(tokens, return_type="logits")
+        actual = distilgpt2_bridge_compat(padded, attention_mask=mask, return_type="logits")[
+            :, real
+        ]
+
+    torch.testing.assert_close(actual, baseline, rtol=1e-3, atol=1e-3)
+
+
+def test_derived_position_ids_match_hooked_transformer(distilgpt2_bridge, tokens) -> None:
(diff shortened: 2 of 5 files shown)


transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -1551,6 +1551,20 @@ def forward(
                     position_ids.masked_fill_(attention_mask == 0, 1)
                     kwargs["position_ids"] = position_ids
 
+            # A supplied attention_mask with left padding shifts every real token's
+            # absolute position, so positions must be derived from the mask rather
+            # than left to HF's default arange. HookedTransformer does this via
+            # pos_embed; without it the bridge silently returns wrong logits.
+            if (
+                attention_mask is not None
```
jlarson4, line 1640: A manual `forward(new_token, attention_mask=<full mask>, past_key_values=cache)` with left padding returns logits, but raises `RuntimeError: The size of tensor a (19) must match the size of tensor b (10)` because the derived `position_ids` spans past+new while `input_ids` is only the new token. Can the derivation be limited to the tokens actually being passed, the way `get_offset_position_ids` does it (`utilities/tensors.py:131`), with a test covering a cached step?

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -1551,6 +1551,20 @@ def forward(
...
+                and "position_ids" not in kwargs
+                and not _is_inputs_embeds
+                and attention_mask.ndim == 2
+                and bool((attention_mask[:, 0] == 0).any())
```
jlarson4: The gate tests only column 0, so a mask with an interior gap and no leading padding still diverges from HookedTransformer. I measured `3.631e+00` on gpt2 in compat mode against a `0.000e+00` unpadded control. Would it be possible to widen this to any mask with a gap?

jlarson4 (commented):
Thanks for both discovering and resolving this bug @sohv! Great work. Just a couple review comments below, let me know if you have any questions

sohv:
Thank you @jlarson4  for bringing these issues to my attention both were real. I reproduced these issues as you described and have pushed the revised code in commit cdc2af2.

Instead of separately patching the two cases, I switched the derivation to `utils.get_offset_position_ids`, the helper `PosEmbed` and `AbstractAttention` already use. That resolves both comments at once and means the bridge shares HookedTransformer's position derivation instead of running a parallel one, which is really what it should have been doing from the start.

**Cached step.** Reproduced: `RuntimeError: The size of tensor a (27) must match the size of tensor b (14)`. Now offset by the cached prefix — `past_kv_pos_offset = attention_mask.shape[1] - input_ids.shape[1]` — so the positions are sliced back to the tokens actually being passed, matching how the helper is used elsewhere. Covered by `test_cached_step_with_left_padding`.

**Interior gap.** Reproduced: 6.752e+00 on gpt2 in compat mode against a 0.000e+00 unpadded control, with the gate correctly not firing. Rather than widen the predicate I dropped it — an all-ones mask gives `cumsum - 1 == arange`, so deriving unconditionally is a provable no-op when there's no padding, and there's no gate left to get subtly wrong. Now 0.000e+00. Covered by `test_interior_mask_gap_uses_derived_positions`, parametrised over two gap positions.

One thing you did not flag that this surfaced: my pad-position convention was also wrong. I had `masked_fill(mask == 0, 1)`; the helper uses `masked_fill(shifted < 0, 0)`, so pads inherit the previous real index rather than a constant. One of my own tests failed on the switch- it had encoded my convention instead of the contract. So I fixed the test and now `test_derived_position_ids_match_hooked_transformer` pins it to the shared helper so it can't drift again.

Verification:

- The two repros above: clean.
- Test file: 13 passed with the fix, 7 failed without it, and 4 fail against the previous commit (c37a8d8) — specifically the cached-step, both interior-gap, and convention tests, so they do guard what you reported rather than just passing.
- `tests/unit/model_bridge` + `test_tokenizer_padding_side.py` + `test_utils.py`: 4038 passed, 27 skipped, 10 xfailed.
- Stacked with #1608, the left-padded batch loss is 4.814578 against HookedTransformer's 4.814578, per-row drift 0.00000.

One correction to the PR description: the "this PR alone" row in the compounding table was 6.688155 and is now 5.396149, since pad positions inherit the previous index and that changes the still-unmasked aggregate #1607 produces. The "both" row is unchanged at 4.814578.

sohv:
CI caught a regression in my last push so this needed a third pass.

Dropping the gate entirely was an over-correction. You asked me to widen it to any mask with a gap; I widened it to any mask at all, which made the derivation fire for cases that never needed it. Right padding and all-ones masks already agree with `arange`, so injecting `position_ids` there is a no-op at best — and at worst it breaks models that don't accept the argument (`TinyLLaDAModelLM.forward() got an unexpected keyword argument 'position_ids'`) or that compute their own position streams (the four multimodal HF-parity tests, where naive positions overrode mRoPE). I confirmed those five were mine by reproducing them against that commit: the Qwen2.5-VL diff came back as 0.00014135, matching the CI number.

Fixed in 32dd94d: the derivation now fires only when the mask actually moves an attended token off its default position — some masked token precedes a real one. That is exactly left padding and interior gaps, and excludes right padding and all-ones masks.

- LLaDA right-padding test: passes
- Qwen2.5-VL adapter suite: 5 passed
- Your two findings: still fixed — the cached step returns clean logits, and interior gaps measure 0.000e+00 across three gap positions with no leading padding
- Added `test_no_position_ids_injected_when_unnecessary`, parametrised over all-ones and right-padded masks, which fails against the previous commit — so this specific over-firing can't come back
- `tests/unit/model_bridge` + the LLaDA adapter + the new file: 4070 passed, 27 skipped, 10 xfailed

One unrelated failure in that run: `test_bridge_hooked_parity_multi_step_optimization`. It doesn't pass an `attention_mask`, so this code path can't reach it, and it fails identically on pristine `dev-4.x` with no fix applied (`Step 10: 0.100004 exceeds threshold 0.100000`). CI tripped it at step 1 by ~3%, my machine trips it at step 10 by 0.004% — a knife-edge threshold rather than anything this PR touches. Happy to open a separate issue if that's useful.

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -1551,6 +1551,37 @@ def forward(
...
+            # Any masked-out token shifts the absolute position of every real token
+            # after it, so positions must be derived from the mask rather than left
+            # to HF's default arange. This is the same derivation HookedTransformer
+            # applies in pos_embed; without it the bridge silently returns wrong
+            # logits. An all-ones mask reduces to arange, so this is a no-op there.
+            #
+            # The mask spans any cached prefix as well as the new tokens, so it is
+            # offset back to just the tokens actually being passed — matching how
+            # AbstractAttention/PosEmbed use past_kv_pos_offset.
+            if (
```
jlarson4, line 1639: For mRoPE architectures a left-padded batch now overrides the model's own 3-D derivation with a naive 2-D arange. HF only computes its rope index when `position_ids is None` (`modeling_qwen2_5_vl.py:1249`) and silently expands 2-D input across all three streams (lines 822-823) and for fixed-signature remote-code models like `LLaDAModelLM` the same injection raises `TypeError`, where base returned logits. Can injection be conditional on the target actually accepting and not already owning positions, the way `output_attentions` is guarded (`bridge_core.py:1205-1218`)?

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -1551,6 +1551,37 @@ def forward(
...
+                _derived = utils.get_offset_position_ids(0, attention_mask)
+                _arange = torch.arange(attention_mask.shape[1], device=_derived.device)
+                # Only intervene when the mask actually moves an attended token off
+                # its default position — i.e. some masked token precedes a real one
+                # (left padding, or an interior gap). Pure right padding and an
+                # all-ones mask already agree with arange, and injecting position_ids
+                # there would be a no-op at best and unsupported by models whose
+                # forward does not take them at worst.
+                if bool(((_derived != _arange) & (attention_mask != 0)).any()):
+                    kwargs["position_ids"] = _derived[
```
jlarson4: The predicate is a whole-batch `.any()`, so a single left-padded row causes `position_ids` to be supplied for every row, including unpadded ones, which is what carries the mRoPE corruption into rows that had no padding at all. Is it possible to make the decision per-row?

jlarson4 (commented):
Hey @sohv thanks for the detailed breakdown of your changes! I have a couple of additional follow ups based on this work.

Narrowing the predicate removed the *tested* symptoms (right padding, all-ones) without removing the *mechanism*: the bridge still injects a `position_ids` kwarg with no check that the target model accepts or wants it. Left padding and interior gaps still fire, and both of those reach architectures that break. A couple of line-specific comments to come for resolution:

sohv:
I reproduced both on `32dd94d` and fixed them in `a2efad9`.

**Gating the injection on the target.** I added `_accepts_derived_position_ids()`, modelled on the `output_attentions` guard at `bridge_core.py:1205-1218`. It refuses to inject when the forward can't take the kwarg, and when the model derives positions itself.

| case | detection | before → after |
|---|---|---|
| fixed-signature forward (LLaDA) | no `position_ids`, no `**kwargs` | `TypeError` → passes |
| mRoPE (Qwen2/2.5/3-VL, GLM-4V) | `get_rope_index`, with `mrope_section` as backstop | `1.414e-04` → `0.000e+00` |
| OPT | `nn.Embedding` subclass whose `forward` takes `attention_mask` | `2.66e-01` → `0.000e+00` |

Two things I want to flag about the mRoPE case. The damage only shows up with an image: text-only left padding measured `0.000e+00` even on `32dd94d`, because for pure text the 2-D broadcast across the three streams happens to coincide with what `get_rope_index` computes. A text-only probe would have cleared it, which is roughly how I missed it the first time. I also checked that skipping these models isn't leaving them unfixed — `get_rope_index` already scatters positions onto attended slots only, and I measured left-pad invariance of `4.47e-08` with no injection at all. Deferring to it is the correct behaviour, not a gap.

OPT wasn't in your comment, but I found it while sweeping ten architectures and it belongs to the same family via a different mechanism, so I gated it too.

**Deciding per row.** I compute `_needs` per row now and select with `torch.where`, so unpadded and purely right-padded rows keep `arange` verbatim.

I want to be precise about what this fixed because I don't think the batch `.any()` is what carried the mRoPE corruption. Supplying *any* `position_ids` disables `get_rope_index`, so even a single-row left-padded batch broke it — only the target gate addresses that. What per-row does fix is narrower but real: the neighbour row in a mixed batch diverged by `8.223e-01`, and I localised that entirely to the pad slots, with real-token logits already exact.

The sweep also caught per-row making OPT *worse* on its own (`2.85e-01` → `4.53e-01`) before I added the gate for it, which is what convinced me the two changes are separate. "Leave this row alone" isn't the same as "give this row arange" when the model self-derives:

```
mask        [1, 1, 1, 1, 1, 1, 0, 0, 0]
OPT's own   [0, 1, 2, 3, 4, 5,-1,-1,-1]
arange      [0, 1, 2, 3, 4, 5, 6, 7, 8]
```

**Two fixes that fell out of the testing.** A float 0/1 attention mask produced float positions and crashed the embedding lookup, so I now cast with `.long()`. I also found that cached decoding under left padding was broken: prefilling through the bridge and stepping now lands `1.755e-04` from the unpadded result, against `2.844e+01` on pre-PR `f17dff3`. My original cached test prefilled with raw HF and asserted only shape and finiteness, so it passed while demonstrating nothing — I rewrote it to prefill through the bridge and assert equivalence to the unpadded result, and added a multi-step version.

**Tests.** I added 13 unit tests driving the gate with stand-in modules, and grew `test_left_padding_positions.py` to 28 tests, 15 of which are red on pre-PR code. The remaining 13 are guards rather than demonstrations — degenerate masks, 4-D masks, `inputs_embeds`, gradients through a padded forward, and `run_with_cache` matching `forward`. There's also one test each in the LLaDA and Qwen2.5-VL suites.

I ran unit, integration and acceptance together: **6419 passed, 1 failed**. The failure is `test_bridge_hooked_parity_multi_step_optimization`, which produces the identical `0.100004 exceeds threshold 0.100000` on the pre-PR bridge file, so I've left it alone. mypy is clean.

**Two things I didn't change** 
- The `_is_batched_list` injection sites at `transformer_bridge.py:1628` and `:2004` are still ungated. They're the same defect class but they predate this PR and I couldn't produce a model that actually fails there — falcon-mamba takes `**kwargs` so it survives, and HyenaDNA needs `trust_remote_code` and won't boot in CI. I'd rather not change them speculatively, but I'm happy to gate them here or in a follow-up if you'd prefer.  
- `generate()` takes no `attention_mask` and no `**kwargs`, so a caller passing an already-padded tensor has no way to mark the padding; that looks like its own issue.

jlarson4:
Excellent, thank you for the detailed writeup @sohv. These seems like excellent changes, I will do a full code review shortly

jlarson4:
Ok, full review time:

For `_is_batched_list`, I agree it is safe to leave that for a follow up.

For `generate()`, I also agree, that should be a separate issue. I had it marked on my end as something that should be listed as an issue.

Altogether, looks great in the state that it is in, merging now! Thanks for putting this together and working through it with me @sohv

sohv:
@jlarson4 Thank you! Since I'm already familiar with the generate(), can I create a separate issue and work on a PR to fix it?

jlarson4:
@sohv For sure! As soon as the issue is up I will assign you

PR #1613, opened 2026-08-06
Fix TransformerBridge loss with explicit labels

### LINKED ISSUE #1611: [Bug Report] TransformerBridge seq2seq loss ignores labels and derives decoder inputs/targets from encoder input_ids
**Describe the bug**

`TransformerBridge.forward(..., return_type="loss")` applies the shifted decoder-only causal loss to encoder-decoder models such as BART and T5.

For seq2seq models, the encoder `input_ids` and decoder `labels` are different sequences and may have different lengths. However, the current Bridge path:

1. derives `decoder_input_ids` from the encoder `input_ids`, even when `labels` are provided;
2. discards the loss already computed by the Hugging Face model;
3. recomputes a shifted causal loss against the encoder `input_ids`.

As a result:

- changing `labels` may not change either the returned Bridge loss or logits;
- the returned loss does not match Hugging Face seq2seq loss;
- source and target sequences of different lengths can raise a dimension error;
- encoder masking semantics may be incorrectly reused for decoder targets.

This is particularly risky because the returned scalar is finite and plausible when the source and target happen to have the same length, so the incorrect result can be used silently in training or interpretability experiments.

**Code example**

The following reproduction creates a tiny random BART model from config. It does not download any model or tokenizer from the Hub.

```python
import torch
from transformers import BartConfig, BartForConditionalGeneration

from transformer_lens.model_bridge.sources._bridge_builder import (
    build_bridge_from_module,
)


torch.manual_seed(1234)

config = BartConfig(
    vocab_size=32,
    d_model=16,
    encoder_layers=1,
    decoder_layers=1,
    encoder_attention_heads=2,
    decoder_attention_heads=2,
    encoder_ffn_dim=32,
    decoder_ffn_dim=32,
    max_position_embeddings=32,
    pad_token_id=0,
    bos_token_id=1,
    eos_token_id=2,
    decoder_start_token_id=2,
)

model = BartForConditionalGeneration(config).eval()
    model,
    "BartForConditionalGeneration",
    hf_config=config,
    tokenizer=None,
    model_name="tiny-random-bart",
)

source = torch.tensor([[4, 5, 6, 7, 2]])
labels_a = torch.tensor([[8, 9, 10, 2, -100]])
labels_b = torch.tensor([[11, 12, 13, 2, -100]])

with torch.no_grad():
    bridge_loss_a = bridge(source, labels=labels_a, return_type="loss")
    bridge_loss_b = bridge(source, labels=labels_b, return_type="loss")

    hf_loss_a = model(input_ids=source, labels=labels_a).loss
    hf_loss_b = model(input_ids=source, labels=labels_b).loss

    bridge_logits_a = bridge(source, labels=labels_a, return_type="logits")
    bridge_logits_b = bridge(source, labels=labels_b, return_type="logits")

    hf_logits_a = model(input_ids=source, labels=labels_a).logits
    hf_logits_b = model(input_ids=source, labels=labels_b).logits

print(f"Bridge loss A: {bridge_loss_a.item():.9f}")
print(f"Bridge loss B: {bridge_loss_b.item():.9f}")
print(
    "Bridge loss delta:",
    f"{torch.abs(bridge_loss_a - bridge_loss_b).item():.9f}",
)

print(f"HF loss A: {hf_loss_a.item():.9f}")
print(f"HF loss B: {hf_loss_b.item():.9f}")
print(
    "HF loss delta:",
    f"{torch.abs(hf_loss_a - hf_loss_b).item():.9f}",
)

print(
    "Bridge logits delta:",
    f"{torch.max(torch.abs(bridge_logits_a - bridge_logits_b)).item():.9f}",
)
print(
    "HF logits delta:",
    f"{torch.max(torch.abs(hf_logits_a - hf_logits_b)).item():.9f}",
)

short_decoder = torch.tensor([[2, 14, 15]])
short_labels = torch.tensor([[14, 15, 2]])

with torch.no_grad():
    hf_short_loss = model(
        input_ids=source,
        decoder_input_ids=short_decoder,
        labels=short_labels,
    ).loss

print(f"HF short-target loss: {hf_short_loss.item():.9f}")

with torch.no_grad():
    bridge(
        source,
        decoder_input_ids=short_decoder,
        labels=short_labels,
        return_type="loss",
    )
```

Observed output:

```text
Bridge loss A: 3.489418030
Bridge loss B: 3.489418030
Bridge loss delta: 0.000000000

HF loss A: 3.463288307
HF loss B: 3.508223534
HF loss delta: 0.044935226

Bridge logits delta: 0.000000000
HF logits delta: 0.268047512

HF short-target loss: 3.496892214

RuntimeError: Size does not match at dimension 1 expected index [1, 4, 1]
to be no larger than self [1, 2, 32] apart from dimension 2
```

The exact random loss values may vary across dependency versions, but the important properties are:

- `bridge_loss_a == bridge_loss_b`;
- `bridge_logits_a == bridge_logits_b`;
- the corresponding Hugging Face loss and logits change with the labels;
- Hugging Face supports a decoder target length different from the encoder source length, while Bridge raises a shape error.

**Expected behavior**

For an encoder-decoder model, if `labels` are supplied:

- when `decoder_input_ids` are absent, the model should derive them by shifting `labels` to the right, following the Hugging Face seq2seq contract;
- `return_type="loss"` should compute loss against `labels`;
- the seq2seq loss should not apply an additional decoder-only `logits[:, :-1]` / `labels[:, 1:]` shift;
- positions where `labels == -100` should be ignored;
- encoder and decoder sequences should be allowed to have different lengths;
- the encoder `attention_mask` should not be treated as the decoder target loss mask.

If `return_type="loss"` is requested for a seq2seq model without `labels`, the Bridge should either require labels explicitly or document another unambiguous target contract. It should not silently use encoder `input_ids` as decoder targets.

**Root cause**

The shared encoder-decoder path currently generates `decoder_input_ids` from encoder `input_ids` whenever the caller does not provide them:

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/model_bridge/transformer_bridge.py#L1563-L1582

This condition does not exclude calls that provide `labels`. Supplying `decoder_input_ids` prevents Hugging Face seq2seq models from deriving decoder inputs from the labels.

After the driver forward pass, the Bridge extracts `result.logits` but does not use `result.raw_output.loss`. It passes the encoder `input_ids` to the common return finalizer:

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/model_bridge/transformer_bridge.py#L1638-L1665

The finalizer then invokes the common shifted causal loss against those encoder tokens:

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/model_bridge/bridge_core.py#L338-L347

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/model_bridge/bridge_core.py#L350-L410

That objective is appropriate for decoder-only next-token prediction, but not for encoder-decoder teacher-forcing loss.

**Affected architectures**

The runtime reproduction above has been verified with BART.

The issue is in the shared `TransformerBridge.forward()` and loss finalization path rather than the BART adapter itself. Static inspection shows that the registered seq2seq adapters currently inherit `supports_causal_loss = True`, including:

- BART, MBART, M2M100, Marian, Pegasus, Blenderbot and LED;
- T5, MT5, legacy T5, LongT5 and SwitchTransformers;
- T5Gemma and T5Gemma2.

The centralized seq2seq architecture list is here:

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/utilities/architectures.py#L9-L26

Only BART has been exercised in the reproduction above; the remaining families are listed based on the shared code path and adapter capability flags, not individual runtime verification.

**Suggested direction**

The preferred fix would be to add an explicit seq2seq loss path rather than reuse the causal loss path.

Possible behavior:

1. Detect encoder-decoder models before causal-loss finalization.
2. If `labels` are supplied and `decoder_input_ids` are not, allow the HF model to derive decoder inputs from the labels.
3. For scalar loss, return or otherwise preserve the architecture-correct loss from the raw model output.
4. For `loss_per_token=True`, compute unreduced cross-entropy directly against `labels`, without applying another causal shift, and ignore `-100`.
5. Require `labels` for seq2seq `return_type="loss"` / `"both"` unless another explicit target API is defined.
6. Keep encoder attention masks, decoder attention masks and loss-label masks as separate concepts.

A temporary safety fix would be to mark seq2seq adapters as not supporting the current shifted causal loss and raise an explicit error. That would be safer than returning a plausible but incorrect scalar, although full seq2seq support would be more useful.

It may also be clearer long-term to distinguish loss types explicitly, for example:

```text
causal / seq2seq / unsupported
```

rather than representing all objective support through `supports_causal_loss: bool`.

**Existing test/verification gap**

I could not find a current test that compares:

```python
bridge(
    encoder_input_ids,
    labels=target_ids,
    return_type="loss",
)
```

against the corresponding raw Hugging Face seq2seq loss.

The benchmark reference named `hf_loss` also appears to be populated from a Bridge call:

https://github.com/TransformerLensOrg/TransformerLens/blob/f17dff30/transformer_lens/benchmarks/main_benchmark.py#L1038-L1043

Therefore, existing Bridge loss-equivalence results do not appear to exercise the Hugging Face `labels` contract demonstrated above.

**System Info**

- Installation: source checkout, dependencies managed with `uv`
- OS: Windows, PowerShell
- Python: 3.12.10
- PyTorch: 2.11.0+cpu
- Transformers: 5.13.0
- Model used for reproduction: randomly initialized tiny `BartForConditionalGeneration`; no Hub download

**Additional context**

The same relevant path is present on `dev-4.x` at commit `f17dff30`.

### Checklist

emerardd:
# Description

## Summary

- Add an explicit seq2seq loss path that preserves Hugging Face label-driven decoder input construction and computes loss against unshifted decoder labels.
- Honor explicit `labels` for decoder-only causal loss, including `-100`, per-token loss, and the existing attention-mask transition semantics.
- Require `labels` for encoder-decoder `return_type="loss"` and `"both"` instead of silently treating encoder inputs as decoder targets.
- Handle Hugging Face tuple outputs correctly when `labels` prepend a native loss value.
- Keep `RemoteBridge` explicit-label causal loss consistent with the local bridge.

## Root cause and impact

`TransformerBridge` forwarded `labels` to Hugging Face but discarded the native label-driven loss during finalization. It then reused the shifted causal loss helper with encoder `input_ids`, so seq2seq losses ignored decoder targets and failed when source and target lengths differed. Decoder-only models had the related issue that explicit labels were ignored in favor of input IDs.

This change makes both paths follow their intended contracts while preserving the input-ID fallback for decoder-only callers that omit labels. It also safely detects encoder-decoder models through the optional driver model so lightweight and non-torch test bridges continue to reach their architecture-specific validation paths.

Fixes #1611

## Type of change

## Validation

- Full non-slow unit suite: `5140 passed, 55 skipped, 54 deselected, 10 xfailed`.
- Tiny BART, T5, and GPT-2 label-loss regression suite: `18 passed`.
- Lightweight Bridge compatibility regressions covering diffusion and audio stubs: `9 passed`.
- `uv run mypy .`: no issues found in 422 source files.
- Full pycln, isort, Black, and `git diff --check` completed successfully.

# Checklist

### DIFF
--- a/transformer_lens/model_bridge/remote_bridge.py
+++ b/transformer_lens/model_bridge/remote_bridge.py
@@ -78,7 +78,11 @@ class RemoteBridge(BridgeCore, HookIntrospectionMixin):
         labels: Any = None,
         **kwargs: Any,
     ) -> Any:
-        """Tokenize → driver.forward → replay captures → finalize per return_type."""
+        """Tokenize → driver.forward → replay captures → finalize per return_type.
+
+        Explicit ``labels`` use shifted causal loss; remote encoder-decoder loss
+        is unsupported.
+        """
         # Early copy of _finalize_return's gate — fail before the wasted remote forward.
         self._check_loss_supported(return_type)
         self._reject_stop_at_layer(kwargs.pop("stop_at_layer", None))
--- a/transformer_lens/model_bridge/remote_bridge.py
+++ b/transformer_lens/model_bridge/remote_bridge.py
@@ -75,6 +75,7 @@ class RemoteBridge(BridgeCore, HookIntrospectionMixin):
         *,
         return_type: str | None = "logits",
         loss_per_token: bool = False,
+        labels: Any = None,
         **kwargs: Any,
     ) -> Any:
         """Tokenize → driver.forward → replay captures → finalize per return_type."""
@@ -109,12 +110,17 @@ class RemoteBridge(BridgeCore, HookIntrospectionMixin):
             return logits  # weird shape — let caller handle
         if logits is not None:
             logits = to_torch(logits)
+        if labels is not None:
+            if not isinstance(labels, TensorLike):
+                raise TypeError(f"labels must be tensor-like, got {type(labels).__name__}")
+            labels = to_torch(labels)
 
         return self._finalize_return(
             return_type,
             logits,
             kwargs.get("input_ids"),
             attention_mask=kwargs.get("attention_mask"),
+            labels=labels,
             is_audio_model=getattr(self.cfg, "is_audio_model", False),
             is_visual_model=getattr(self.cfg, "is_visual_model", False),
             loss_per_token=loss_per_token,
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ logits, activations = bridge.run_with_cache("Hello World")
 
 > Gated models (Llama, Mistral, Gemma, ...) require `HF_TOKEN` in your environment. See [Environment Variables](https://TransformerLensOrg.github.io/TransformerLens/content/getting_started.html#environment-variables) for the full list.
 
-`TransformerBridge` is the recommended 3.0 path and supports 15,000+ models across 140+ architecture families (see [`supported_models.json`](transformer_lens/tools/model_registry/data/supported_models.json) for the full inventory). By default it preserves raw HuggingFace weights – logits and activations match HF, *not* legacy `HookedTransformer` (which folds LayerNorm and centers weights by default). Call `bridge.enable_compatibility_mode()` after booting for HookedTransformer-equivalent numerics. The legacy `HookedTransformer.from_pretrained` API is still available but deprecated — see the [Migrating to TransformerLens 3](https://TransformerLensOrg.github.io/TransformerLens/content/migrating_to_v3.html) guide.
+`TransformerBridge` is the recommended path and supports 15,000+ models across 140+ architecture families (see [`supported_models.json`](transformer_lens/tools/model_registry/data/supported_models.json) for the full inventory). By default it preserves raw HuggingFace weights – logits and activations match HF, *not* legacy `HookedTransformer` (which folds LayerNorm and centers weights by default). Call `bridge.enable_compatibility_mode()` after booting for HookedTransformer-equivalent numerics. The legacy `HookedTransformer.from_pretrained` API is still available but deprecated — see the [Migrating to TransformerLens 3](https://TransformerLensOrg.github.io/TransformerLens/content/migrating_to_v3.html) guide.
 
 ## Key Tutorials
 
--- a/tests/integration/model_bridge/test_seq2seq_loss.py
+++ b/tests/integration/model_bridge/test_seq2seq_loss.py
@@ -1,0 +1,281 @@
+"""Regression tests for TransformerBridge explicit-label loss semantics."""
+
+from __future__ import annotations
+
+import pytest
+import torch
+import torch.nn.functional as F
+from transformers import (
+    BartConfig,
+    BartForConditionalGeneration,
+    GPT2Config,
+    GPT2LMHeadModel,
+    T5Config,
+    T5ForConditionalGeneration,
+)
+
+from tests.integration.model_bridge.helpers import make_tiny_pair
+from transformer_lens.model_bridge import TransformerBridge
+
+
+@pytest.fixture(scope="module")
+def tiny_gpt2_pair() -> tuple[TransformerBridge, torch.nn.Module]:
+    config = GPT2Config(
+        vocab_size=32,
+        n_embd=16,
+        n_layer=1,
+        n_head=2,
+        n_positions=16,
+        n_ctx=16,
+        bos_token_id=1,
+        eos_token_id=2,
+        pad_token_id=0,
+    )
+    return make_tiny_pair(config, "GPT2LMHeadModel", loader=GPT2LMHeadModel)
+
+
+@pytest.fixture(scope="module", params=("bart", "t5"))
+def tiny_seq2seq_pair(request) -> tuple[TransformerBridge, torch.nn.Module]:
+    if request.param == "bart":
+        config = BartConfig(
+            vocab_size=32,
+            d_model=16,
+            encoder_layers=1,
+            decoder_layers=1,
+            encoder_attention_heads=2,
+            decoder_attention_heads=2,
+            encoder_ffn_dim=32,
+            decoder_ffn_dim=32,
+            max_position_embeddings=32,
+            pad_token_id=0,
+            bos_token_id=1,
+            eos_token_id=2,
+            decoder_start_token_id=2,
+        )
+        return make_tiny_pair(
+            config,
+            "BartForConditionalGeneration",
+            loader=BartForConditionalGeneration,
+        )
+
+    config = T5Config(
+        vocab_size=32,
+        d_model=16,
+        d_kv=8,
+        d_ff=32,
+        num_layers=1,
+        num_decoder_layers=1,
+        num_heads=2,
+        pad_token_id=0,
+        eos_token_id=1,
+        decoder_start_token_id=0,
+    )
+    return make_tiny_pair(
+        config,
+        "T5ForConditionalGeneration",
+        loader=T5ForConditionalGeneration,
+    )
+
+
+def test_seq2seq_loss_and_logits_follow_labels(
+    tiny_seq2seq_pair: tuple[TransformerBridge, torch.nn.Module],
+) -> None:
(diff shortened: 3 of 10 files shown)


emerardd:
@jlarson4 The four failing jobs did not reach checkout or execute any repository code. They failed while resolving GitHub Actions dependencies with `Service Unavailable` / `Internal Server Error`:

- Full Code Coverage Test
- Notebook Checks (Main_Demo)
- Notebook Checks (LLaMA_GPU_Quantized)
- Notebook Checks (Santa_Coder)

The compatibility checks for Python 3.10, 3.11, and 3.12, as well as format, type, docstring, benchmark, and the other notebook checks, passed.

I do not have permission to re-run workflows on the upstream repository. Could a maintainer please re-run the failed jobs?

jlarson4:
@emerardd Rerunning now!

transformer_lens/model_bridge/remote_bridge.py
```diff
@@ -109,12 +110,17 @@ def forward(
             return logits  # weird shape — let caller handle
         if logits is not None:
             logits = to_torch(logits)
+        if labels is not None:
```
jlarson4, line 117: The local bridge routes enc-dec + labels to `_seq2seq_loss`, but `RemoteBridge.forward` passes `labels` straight into `_finalize_return` to `_causal_labels_loss` (shifted). No remote enc-dec driver exists today (vLLM/Inspect are decoder-only),  but a hypothetical seq2seq remote driver would silently get the wrong objective, which is the exact failure mode this PR eliminates locally. Could you add a one-line guard or docstring note?

jlarson4 (commented):
Great work on this @emerardd! A couple small considerations if you'd be willing to take a look at them

`benchmark_loss_equivalence`and the ungated Phase-1 fallback at `main_benchmark.py:1041` still call `bridge(test_text, return_type="loss")` with no labels, which now raises for encoder-decoder models. Previously it "passed", now it fails with an error. The error is better than the false pass, but it will look like a regression across all 15 seq2seq families in `verify_models`. Can you update these two call sites the new contract by suppling labels (e.g. the tokenized text as its own target)?

jlarson4:
Just a quick update that GitHub Actions are down at the moment, whenever you make changes you'll hopefully get CI response at that time. Everything was passing on the final attempt before the GitHub wide outage occurred.

emerardd:
Thanks for catching these. Addressed in c705ab10:

- Updated both `benchmark_loss_equivalence` and the Phase 1 reference capture to use a shared self-target loss helper that supplies the tokenized test text as explicit `labels`. This preserves the decoder-only causal objective while giving encoder-decoder models a valid target.
- Documented that `RemoteBridge` currently interprets explicit labels with shifted causal loss and does not support remote encoder-decoder loss.
- Added a unit test covering the benchmark's explicit-label contract.

Local validation:
- Related loss/RemoteBridge tests: 38 passed
- New benchmark regression test: 1 passed
- `uv run mypy .`: no issues in 422 source files
- pycln/isort/Black checks and `git diff --check`: passed

transformer_lens/model_bridge/remote_bridge.py
emerardd: Addressed in c705ab10 by documenting directly on `RemoteBridge.forward` that explicit labels use shifted causal loss and remote encoder-decoder loss is unsupported. I used the docstring option because the current driver protocol/config does not expose a reliable encoder-decoder capability flag, while all current remote drivers are decoder-only; an optional-attribute guard could otherwise provide false assurance if a future driver omitted the flag.

jlarson4:
Great work @emerardd, merging now!

PR #1615, opened 2026-08-06
feat(bridge): support disk device_map offload targets

### LINKED ISSUE #1280: [Proposal] Add support for `cpu`, `meta`, and `disk` to TransformerBridge `device_map`

LightWork666:
Fixes #1280 (the remaining disk-offload piece — CPU and meta-target support already landed in #1459/#1493).

## The bug

`device_map` entries targeting `"disk"` were rejected outright, with the reason on record in #1459: a smoke test found that a GPT-2 block offloaded to disk loaded successfully but crashed mid-forward on a `meta` tensor inside a Bridge-wrapped component.

Root cause: Accelerate's disk/CPU offload works by attaching a hook to whichever module directly owns a `Parameter` (e.g. `c_attn`, `c_proj`, `ln_1` — not their parent containers), and that hook only materializes real data around *that module's own* `forward()` call. Bridge components don't exclusively call the wrapped module's forward — they read its raw params directly (`self.weight`, `self.original_component.bias`, etc., via `__getattr__` or direct attribute access) to implement their own from-scratch computation. Outside the window Accelerate's hook controls, that read silently returns the `meta` placeholder instead of the real tensor — no error, just wrong (or in most cases, an immediate device-mismatch crash the moment it hits a real op).

A second, gnarlier layer specific to GPT-2-style combined-QKV architectures: `JointQKVAttentionBridge` builds its `q`/`k`/`v` sub-bridges once, at setup time, by slicing the combined `c_attn` weight into three independent tensors. That slice read happens outside of any component's own `__call__` entirely, and — unlike the general case — needs to read from a *descendant* of the object it's called with (`original_component.c_attn`, not `original_component` itself).

## The fix

- `GeneralizedComponent.__call__` (the base class every bridge component shares) now wraps the actual `forward()` call in Accelerate's own `align_module_device(original_component)` context — the same pre/post-forward hook Accelerate would have run natively. It's a no-op when the wrapped module has no offload hook (confirmed negligible, ~0.004ms/call), which covers the common case: no device_map, single device, or a GPU split with no CPU/disk offload involved.
- For the one-time QKV split, a second helper (`align_offloaded_subtree`) walks every descendant of the passed-in module and enters `align_module_device` for each one that actually has a hook — necessary because I can't assume in general which descendant a given architecture's slicing code reads from (there are 140+ supported architectures; I didn't want to bake in "materialize the whole container" as the default, since I found that actually regresses memory usage — more on that below).
- `resolve_device_map` / `_validate_device_map_values` no longer reject `"disk"` targets. `"meta"` stays rejected under `load_weights=True` (it has no real data to offload from at all, a genuinely different case from disk/cpu). The CPU+GPU mixed-map restriction is untouched — I don't have GPU hardware to verify that path, and didn't want to claim something I couldn't test (the underlying fix is probably the same mechanism, but that's for someone with GPU access to confirm as a follow-up).
- Added an `offload_folder` parameter to `boot_transformers` so callers can control where offloaded shards get written, rather than relying on HF's auto-temp-dir default.

## A design note worth flagging in review

My first pass applied the "walk every descendant" helper unconditionally at every `GeneralizedComponent.__call__`, not just the QKV setup case. It technically worked, but it meant a block-level component's outer call would eagerly materialize its *entire* subtree — every leaf's real data — for the whole duration of that call, rather than the incremental, one-leaf-at-a-time pattern Accelerate's native hooks give a plain (non-bridge) forward pass. That's a real memory regression for exactly the use case disk offload exists to serve, so I pulled it back to the minimal, single-object `align_module_device(original_component)` for the general case, and reserved the subtree walk for the one call site that actually needs it (documented inline in both places).

## Testing

- `tests/acceptance/model_bridge/test_multi_gpu_bridge.py`:
  - Two fast tests using a synthetic `nn.LayerNorm` with a directly-attached Accelerate offload hook (no model download) — confirms the failure mode (raw weight read is `meta`) and confirms `NormalizationBridge` now materializes correctly through it.
  - Two real-gpt2 tests: disk-offloaded forward pass (4 of 12 blocks on disk, including the split-QKV path) matches a non-offloaded load exactly (`atol=1e-4`), including loss; and a plain regression guard that `"disk"` no longer raises at boot.
  - Updated `test_disk_value_in_device_map_rejected` → `test_disk_value_in_device_map_passes_through`, since the old assertion is now the wrong behavior.
- `uv run mypy .` — clean, 422 files.
- Ran the full `tests/unit/model_bridge/` core suite (boot_native, architecture_adapter, qkvo conversions, generalized_components) and all 135 `supported_architectures` test files in isolated batches — everything passes, no regressions from the shared base-class change.
- Verified `generate()` + disk offload crashes with a SIGBUS on this machine — then verified the *identical* crash happens with zero offload involved at all (plain gpt2, `generate()`, no device_map), confirming via `git stash` that it's the pre-existing, already-quarantined macOS-arm64 KV-cache bug documented in `tests/QUARANTINES.md`, unrelated to this change. Not claiming `generate()` + disk offload works or doesn't on other platforms — only that this specific crash isn't caused by this PR.


transformer_lens/model_bridge/generalized_components/joint_qkv_attention.py
```diff
@@ -284,9 +285,20 @@ def set_original_component(self, original_component: torch.nn.Module) -> None:
...
+        # to build independent q/k/v slices - not a live view, so under Accelerate
+        # offload this needs the real (not meta) data materialized for this one
+        # read. The resulting slices are real, standalone tensors that stay valid
+        # afterward regardless of what Accelerate later does to original_component
+        # (unlike whatever reads original_component itself on every forward call,
+        # e.g. GeneralizedComponent.__call__ / LinearBridge for "o"/c_proj, split
+        # q/k/v specifically stay permanently resident rather than re-offloading
+        # after each forward - a deliberate, small, documented memory trade-off
+        # for combined-qkv architectures).
+        with align_offloaded_subtree(original_component):
```
jlarson4, line 298: The same setup-time raw slice this hunk guards also exists in `JointGateUpMLPBridge.set_original_component`. Can you extend the guard to that split path (`joint_gate_up_mlp.py:81`)?

transformer_lens/model_bridge/generalized_components/base.py
```diff
@@ -1,20 +1,42 @@
...
 import torch.nn as nn
+from accelerate.utils import align_module_device
 
 from transformer_lens.conversion_utils.conversion_steps.base_tensor_conversion import (
     BaseTensorConversion,
 )
 from transformer_lens.hook_points import HookPoint
 
 
+def align_offloaded_subtree(module: nn.Module) -> contextlib.ExitStack:
```
jlarson4, line 20: `align_module_device` first shipped in accelerate 1.1.0, I verified that 1.0.1 does not have it. The project floor is `accelerate>=0.23.0`, so this module-level import breaks the whole package on older installs. Please raise the base level of `accelerate` to 1.1.0 in the `pyproject` file.

transformer_lens/utilities/multi_gpu.py
```diff
@@ -204,19 +202,13 @@ def resolve_device_map(
 def _validate_device_map_values(
     device_map: Union[str, Dict[str, Union[str, int]]],
 ) -> None:
-    """Reject explicit disk values and mixed CPU+GPU targets in a user-supplied
-    device_map dict. Meta values are passed through (validated at boot against
-    load_weights)."""
+    """Reject mixed CPU+GPU targets in a user-supplied device_map dict. Disk
```
jlarson4: A map mixing `cuda` with `disk` now passes both this check and the boot gate, yet disk offload rides the same hook mechanism whose CPU+GPU mix stays rejected as unverifiable without GPU hardware. Could the mixed-map gate treat disk like cpu until that's GPU-verified?

transformer_lens/model_bridge/sources/transformers/source.py
```diff
@@ -84,12 +85,17 @@ def boot(
...
+            dispatched inference. Explicit maps may include CPU and disk targets; meta targets
+            are still rejected when ``load_weights=True`` (meta has no real data to offload from,
+            unlike disk/cpu). Mixed CPU+GPU maps are rejected too — accelerate implements that as
+            CPU offload, and the underlying materialization mechanism can't split across it and a
+            GPU-resident forward pass without further hook-routing work. Mutually exclusive with
+            ``device``.
         n_devices: Convenience: split the model across this many CUDA devices (translated to a
             ``max_memory`` dict internally). Requires CUDA with at least this many visible devices.
         max_memory: Optional per-device memory budget for HF's dispatcher.
+        offload_folder: Directory for disk-offloaded weight shards when ``device_map`` includes
```
jlarson4, line 100: `enable_compatibility_mode()` on a disk-offloaded bridge dies deep in weight folding with an opaque meta-device RuntimeError, and raw reads like `blocks[0].ln1.weight` silently return meta placeholders. Could the docs here note disk offload is forward-pass-only, or fail earlier with a clear message?

jlarson4 (commented):
Excellent diagnosis and writeup @LightWork666! The `meta` bug catch is a good one and the memory-regression note is very helpful for understanding your thought process. I have a couple of additional requests below, let me know if you have any questions.

LightWork666:
Thank you, @jlarson4! Really appreciate you taking the time to dig through this in detail. Working through the four points now, and found one more instance of the same bug myself while fixing the gate/up one (UnembeddingBridge has the identical pattern) — pushing shortly.

jlarson4:
Great work on this @LightWork666, and good find on the `UnembeddingBridge` pattern, I missed that one. Merging now

PR #1616, opened 2026-08-06
Add kurtosis-profile validation test for JacobianLens (#1539 Tier-1)

SanjidMzi:
# Description

Adds the Tier-1 kurtosis-profile check from #1539. The §4 signature is asserted as relative structure rather than absolute levels, following the cross-family measurement posted in that thread: absolute early-third kurtosis ranges from 0.2 to ~140 across published artifacts, so a fixed "≈0 early" assert only holds for one family.

Four checks, on two pinned artifacts:

1. Band rise on Qwen3.5-0.8B: interior peak minus the model's own early-third baseline is at least 0.35. The reference run measured +0.78.
2. Lens-specificity: that rise exceeds the logit-lens control's rise through the identical code path by at least 0.25. Measured gap +0.59.
3. Negative control on gpt2-small, which has no band rise (measured +0.16), so the check cannot pass vacuously.
4. Final-layer identity transport: both arms produce the model's own logits there and their kurtosis agrees exactly.

Prompts come from the wikitext-103 validation split; the published lenses were fitted on train, so the readout corpus is held out. Lens artifacts are pinned to the same neuronpedia revision as `test_jacobian_lens_oracle_parity.py`. Marked slow: downloads Qwen3.5-0.8B (~1.6 GB), runs on CPU in a few minutes. QUARANTINES.md row added.

Part of #1539 (Tier 1, kurtosis profile).

## Type of change

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ interpretability](https://distill.pub/2020/circuits/zoom-in/) of GPT-2 Style lan
 goal of mechanistic interpretability is to take a trained model and reverse engineer the algorithms
 the model learned during training from its weights.
 
-TransformerLens lets you load in 9,000+ open source language models across 50+ architecture families,
+TransformerLens lets you load in 15,000+ open source language models across 140+ architecture families,
 and exposes the internal activations of the model to you. You can cache any internal activation in the
 model, and add in functions to edit, remove or replace these activations as the model runs.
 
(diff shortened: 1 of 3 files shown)


jlarson4:
Looks good @SanjidMzi, thanks for taking this on

PR #1621, opened 2026-08-07
Fix gpt oss olmo3 parity

### LINKED ISSUE #1618: [Bug Report] `TransformerBridge` activations on gpt-oss-20b disagree with the HF model it wraps
`TransformerBridge.boot_transformers("openai/gpt-oss-20b")` loads without warnings, but its
`hook_resid_post` disagrees with the underlying HF forward from **layer 0 onwards** and becomes
*anti-correlated* by mid-stack: cosine 0.84 at layer 0, 0.04 at layer 12, and a final-layer residual
whose norm is ~200x the reference. Since the bridge wraps the same HF module it is being compared
against, both sides should be the same tensor up to bf16 noise.

## Reproduce

Records HF `hidden_states`, frees the model, then boots the bridge on the same checkpoint and compares
`hook_resid_post` for the same token ids. (`hidden_states[i + 1]` is block `i`'s output, which is what
`blocks.{i}.hook_resid_post` reports.)

```python
import torch

HF_ID, LAYERS = "openai/gpt-oss-20b", (0, 12, 23)
ids = torch.tensor([[15496, 11, 616, 1438, 318, 1757, 13, 314, 1101, 257]]).cuda()

hf = AutoModelForCausalLM.from_pretrained(HF_ID, dtype=torch.bfloat16, device_map="cuda")
with torch.no_grad():
    out = hf(ids, output_hidden_states=True)
ref = {layer: out.hidden_states[layer + 1][0].float().cpu() for layer in LAYERS}
del hf, out
torch.cuda.empty_cache()

bridge = TransformerBridge.boot_transformers(HF_ID, device="cuda", dtype=torch.bfloat16)
wanted = {f"blocks.{layer}.hook_resid_post" for layer in LAYERS}
with torch.no_grad():
    _, cache = bridge.run_with_cache(ids, names_filter=lambda n: n in wanted)

for layer in LAYERS:
    a = ref[layer].flatten()
    b = cache[f"blocks.{layer}.hook_resid_post"][0].float().cpu().flatten()
    cos = torch.dot(a, b) / (a.norm() * b.norm())
    print(f"resid_post.{layer}: cos={cos:.4f} rel={(a - b).norm() / a.norm():.4f}")
```

## Actual

```
resid_post.0:  cos=0.8449 rel=0.5946
resid_post.12: cos=0.0379 rel=1.0023
resid_post.23: cos=0.1892 rel=210.6264
```

Forward hooks on the blocks' `attn` / `mlp` submodules — raw module outputs on both sides, so no
post-norm convention is involved — go negative in the middle of the stack: attention output cosine
**−0.60** and MLP output cosine **−0.79** at layer 12.

## Expected

`cos ≈ 1.0` and a relative error at the bf16 noise floor (~1e-3), as the bridge produces on other
checkpoints. A cosine near zero, or negative, is a different quantity rather than a precision
difference.

## Root cause

Not diagnosed. Two observations narrow it: the divergence is already 0.59 relative at **layer 0**, so it
is not accumulation, and `resid_post` comes from the bridge's own cache rather than from a user hook, so
it is on the bridge's forward path.

Candidates specific to this checkpoint, in the order we would check them:

1. **MXFP4 experts.** `experts.gate_up_proj` here is not a `torch.Tensor` but a
   `triton_kernels.tensor.Tensor` holding the packed 4-bit payload plus scales. Any code that reads
   those weights, rather than calling the module, has to unwrap or dequantize first. (The legacy
   `HookedTransformer` conversion path trips over the same object — filed separately.)
2. **Attention sinks.** gpt-oss adds a learned per-head sink logit to the attention softmax; a wrapper
   that reimplements or reorders the softmax has to carry it.
3. **MoE routing.** Top-k router weights and expert combination happen inside the block; a bridge that
   re-derives them can select different experts.

## Environment

| | |
| --- | --- |
| `transformer_lens` | 3.6.0, and `main` at `69e98ab` — identical cosines to four decimals |
| `transformers` | 5.14.1 |
| `torch` | 2.13.0 |
| python | 3.11.10 |
| GPU | NVIDIA A100 80GB PCIe, driver 550.127.05 |
| checkpoint | `openai/gpt-oss-20b` (`GptOssForCausalLM`, 24 layers, MXFP4 experts, attention sinks) |

### LINKED ISSUE #1619: [Bug Report] `convert_gpt_oss_weights` raises `TypeError: 'Tensor' object is not subscriptable` on MXFP4 experts
`HookedTransformer.from_pretrained_no_processing("openai/gpt-oss-20b")` fails during weight conversion.
On an MXFP4 gpt-oss checkpoint, `experts.gate_up_proj` is a `triton_kernels.tensor.Tensor` — the packed
4-bit payload plus its scales — not a `torch.Tensor`, so the converter's `gate_up_proj[e, :, ::2]` slice
raises. The wrapper class is also *named* `Tensor`, which makes the error look like a torch bug.

## Reproduce

```python
import torch

HookedTransformer.from_pretrained_no_processing("openai/gpt-oss-20b", device="cuda", dtype=torch.bfloat16)
```

## Actual

```
  File ".../transformer_lens/loading_from_pretrained.py", line 1991, in get_pretrained_state_dict
    state_dict = convert_gpt_oss_weights(hf_model, cfg)
  File ".../transformer_lens/pretrained/weight_conversions/openai.py", line 92, in convert_gpt_oss_weights
    state_dict[f"blocks.{l}.mlp.experts.{e}.W_gate.weight"] = gate_up_proj[
                                                             ^^^^^^^^^^^^^
TypeError: 'Tensor' object is not subscriptable
```

The object being sliced, loaded through plain `transformers`:

```python
>>> from transformers import AutoModelForCausalLM
>>> import torch
>>> m = AutoModelForCausalLM.from_pretrained("openai/gpt-oss-20b", dtype=torch.bfloat16, device_map="cuda")
>>> g = m.model.layers[0].mlp.experts.gate_up_proj
>>> type(g).__module__ + "." + type(g).__name__
'_gpt_oss_triton_kernels_cuda_8a730d9.tensor.Tensor'   # triton_kernels, fetched via the `kernels` hub
>>> isinstance(g, torch.Tensor)
False
>>> g[0, :, ::2]
TypeError: 'Tensor' object is not subscriptable
>>> [a for a in ("data", "storage", "shape", "dtype") if hasattr(g, a)]
['data', 'storage', 'shape', 'dtype']
```

Note this happens with `dtype=torch.bfloat16`: the MXFP4 weights stay packed in the triton-kernels
wrapper rather than being materialized as bf16 tensors. `transformers` itself loads and runs the
checkpoint fine, so the same script that produced the object above can be used as a reference forward.

## Expected

Either a converted `HookedTransformer`, or a clear error saying MXFP4 gpt-oss checkpoints are not
supported by this path — `'Tensor' object is not subscriptable` sends the reader looking at torch.

## Suggested fix

In `convert_gpt_oss_weights`, unwrap or dequantize the expert weights before slicing. The wrapper
exposes `.storage` / `.data`, and `transformers` can also be asked for dequantized weights
(`dequantize=True` on the MXFP4 quantization config) so the converter always sees plain tensors. At
minimum, an `isinstance(gate_up_proj, torch.Tensor)` guard raising `NotImplementedError` and naming
MXFP4 would make the failure legible.

## Environment

| | |
| --- | --- |
| `transformer_lens` | 3.6.0, and `main` at `69e98ab` — same `TypeError`, same slice, no MXFP4 handling in that converter on either |
| `transformers` | 5.14.1 |
| `torch` | 2.13.0 |
| python | 3.11.10 |
| GPU | NVIDIA A100 80GB PCIe, driver 550.127.05 |
| checkpoint | `openai/gpt-oss-20b` (`GptOssForCausalLM`, MXFP4 experts) |

### LINKED ISSUE #1620: [Bug Report] `convert_olmo3_weights` zero-fills K/V weights when `num_key_value_heads == num_attention_heads`
`HookedTransformer.from_pretrained_no_processing("allenai/Olmo-3-1025-7B")` loads with 64
`Missing key for a weight matrix … filled in with an empty tensor` warnings — `_W_K` and `_W_V` for each
of the 32 layers. Attention therefore contributes **identically zero** at every layer, and the model
returns wrong activations and wrong logits with no error.

The trigger is a config that declares MHA through the GQA field (`num_key_value_heads ==
num_attention_heads`): `convert_olmo3_weights` and `TransformerBlock` use different tests for "is this
GQA?", so the converter writes `attn.W_K` while the instantiated module expects `attn._W_K`.

## Reproduce

Compares against the HF model the weights came from. The zero-weight and zero-`attn_out` facts are
prompt-independent; the cosines depend on the prompt.

```python
import torch

HF_ID, LAYERS = "allenai/Olmo-3-1025-7B", (0, 16, 31)
ids = torch.tensor([[100, 200, 300, 400]], device="cuda")

hf = AutoModelForCausalLM.from_pretrained(HF_ID, dtype=torch.bfloat16, device_map="cuda")
with torch.no_grad():
    out = hf(ids, output_hidden_states=True)
ref = {layer: out.hidden_states[layer + 1][0].float().cpu() for layer in LAYERS}
del hf, out
torch.cuda.empty_cache()

model = HookedTransformer.from_pretrained_no_processing(HF_ID, device="cuda", dtype=torch.bfloat16)
for name in ("W_Q", "W_K", "W_V"):
    zeroed = sum(1 for b in model.blocks if getattr(b.attn, name).abs().max() == 0)
    print(f"layers with an all-zero {name}: {zeroed}/{len(model.blocks)}")

with torch.no_grad():
    _, cache = model.run_with_cache(ids)
for layer in LAYERS:
    a = ref[layer].flatten()
    b = cache[f"blocks.{layer}.hook_resid_post"][0].float().cpu().flatten()
    print(
        f"layer {layer}: max |attn_out| = {cache[f'blocks.{layer}.hook_attn_out'].abs().max():.1f}, "
        f"resid_post vs HF cos = {torch.dot(a, b) / (a.norm() * b.norm()):.4f}"
    )
```

## Actual

```
WARNING:root:Missing key for a weight matrix in pretrained, filled in with an empty tensor: blocks.4.attn._W_K
WARNING:root:Missing key for a weight matrix in pretrained, filled in with an empty tensor: blocks.21.attn._W_V
... (64 such warnings: _W_K and _W_V for each of the 32 layers)

layers with an all-zero W_Q: 0/32
layers with an all-zero W_K: 32/32
layers with an all-zero W_V: 32/32
layer 0:  max |attn_out| = 0.0, resid_post vs HF cos = 0.84
layer 16: max |attn_out| = 0.0, resid_post vs HF cos = 0.28
layer 31: max |attn_out| = 0.0, resid_post vs HF cos = 0.30
```

Per-layer weight norms confirm only K/V are affected — layer 0: `|W_Q|=36.25`, `|W_K|=0`, `|W_V|=0`,
`|W_O|=69`; layer 31: `|W_Q|=69`, `|W_K|=0`, `|W_V|=0`, `|W_O|=84`.

## Expected

K/V loaded from the checkpoint, and attention output matching the HF forward. Failing that, an
exception: a silently zeroed sublayer is worse than a refusal, because every downstream number still
looks plausible.

## Root cause

Two places decide "is this GQA?" with different tests, and they disagree exactly when
`num_key_value_heads == num_attention_heads`:

- `transformer_lens/components/transformer_block.py:77`

  ```python
  attention = Attention if self.cfg.n_key_value_heads is None else GroupedQueryAttention
  ```

  `n_key_value_heads` is *set* (32), so the block is built as `GroupedQueryAttention`, whose parameters
  are `_W_K` / `_W_V` (`components/grouped_query_attention.py`).

- `transformer_lens/pretrained/weight_conversions/olmo3.py:24`

  ```python
  using_gqa = cfg.n_key_value_heads is not None and cfg.n_key_value_heads < cfg.n_heads
  gqa_uscore = "_" if using_gqa else ""
  ...
  state_dict[f"blocks.{l}.attn.{gqa_uscore}W_K"] = W_K
  ```

  `32 < 32` is false, so the converter emits `blocks.{l}.attn.W_K` / `W_V`.

`_W_K` / `_W_V` are then missing from the state dict, `fill_missing_keys` replaces them with zeros, and
the `W_K` / `W_V` entries the converter did write are dropped as unexpected.

## Suggested fix

Use the same test as the block, which is what the other converters already do — `llama.py:16`,
`phi3.py:15` and `apertus.py:26` are all `using_gqa = cfg.n_key_value_heads is not None`:

```diff
-    using_gqa = cfg.n_key_value_heads is not None and cfg.n_key_value_heads < cfg.n_heads
+    using_gqa = cfg.n_key_value_heads is not None
```

`olmo3.py` is the only converter in 3.6.0 carrying the extra `< cfg.n_heads` clause.

Separately: `fill_missing_keys` silently zero-filling an attention weight matrix is what turns a name
mismatch into wrong numbers. Raising for `W_{Q,K,V,O}` (as opposed to, say, a missing bias) would make
any future converter mismatch fail loudly.

## Who else is affected

Any checkpoint routed through `convert_olmo3_weights` whose config sets `num_key_value_heads` equal to
`num_attention_heads`. `allenai/Olmo-3-1025-7B` (32 == 32) hits it; `allenai/Olmo-3-1125-32B`
(`num_key_value_heads=8`, 40 heads) is genuinely GQA and loads correctly.

## Environment

| | |
| --- | --- |
| `transformer_lens` | 3.6.0, and `main` at `69e98ab` — same result on both |
| `transformers` | 5.14.1 |
| `torch` | 2.13.0 |
| python | 3.11.10 |
| GPU | NVIDIA A100 80GB PCIe, driver 550.127.05 |
| checkpoint | `allenai/Olmo-3-1025-7B` (`Olmo3ForCausalLM`, 32 layers, 32 heads, `num_key_value_heads=32`) |

jlarson4:
# Description

Fixes the three parity/loading bugs reported for gpt-oss-20b and Olmo-3, plus three deeper correctness bugs that verification against the real checkpoints flushed out along the way. Every fix is gated by an HF-parity test, not just "it loads."

Fixes #1618, #1619, and #1620

## #1618

Two independent bugs in the bridge's reconstructed attention (`PositionEmbeddingsAttentionBridge`):

1. **Missing attention sinks.** GPT-OSS's eager attention concatenates a learned per-head sink logit as an extra softmax column and drops it after normalization, rescaling every pattern row. The bridge's softmax omitted it entirely. On the real checkpoint this is not a small correction: some trained heads route **~99.9998% of their probability mass to the sink** (min pattern row-sum 2e-6), so an implementation without sinks computes a different function, which is why mid-stack cosines went negative.
2. **Half-width RoPE misread as partial rotary.** GPT-OSS applies *full* rotation using un-duplicated half-width cos/sin (`rotary_dim == head_dim/2`). The bridge's partial-rotary branch rotated only the first half of each head and passed the rest through. Since gpt-oss's rotation is exactly llama's rotate-half once the halves are duplicated, the fix reuses that path, gated on `rotary_dim * 2 == head_dim and rotary_pct == 1.0` so genuine partial-rotary models (Pythia, Phi) are untouched.

## #1619

- Loading by model name now detects `quant_method == "mxfp4"` in the checkpoint config and passes `Mxfp4Config(dequantize=True)` to `from_pretrained`, so the converter always sees plain tensors. A caller-supplied `quantization_config` takes precedence, and unquantized gpt-oss finetunes are untouched.
- If a pre-loaded quantized model is passed as `hf_model`, the converter now raises a `NotImplementedError` that names MXFP4, identifies the triton-kernels wrapper (whose `Tensor` class name made the original error look like a torch bug), and says how to load dequantized.

Getting past the crash exposed that **`HookedTransformer`'s gpt-oss forward had never been numerically correct** (logits vs HF: cos 0.85, top-1 agreement 0.00). Three missing pieces, all fixed:

- **Attention sinks** – new `use_attention_sinks` config flag; `AbstractAttention` carries the per-head `sinks` parameter through the softmax (same semantics as the bridge fix), and the converter maps the weights.
- **Sliding-window layers** – the config mapping had `use_local_attn: False`; HT's existing local attention has exactly HF's sliding-window semantics, so `attn_types`/`window_size` are now wired from `layer_types`/`sliding_window`.
- **YaRN rope** – the mapping only read `rope_theta`; yarn parameters are now wired, including a new `yarn_truncate` flag for gpt-oss's `truncate: False` variant and HF's default attention factor.

## #1620

- The converter's GQA predicate now matches `TransformerBlock`'s (`n_key_value_heads is not None`), so it emits `_W_K`/`_W_V` exactly when the block instantiates `GroupedQueryAttention`. The previous test that encoded the buggy contract is inverted.
- As suggested in the issue, `fill_missing_keys` now **raises** when a missing key is an attention weight matrix (`W_Q/W_K/W_V/W_O`, `_W_K/_W_V`) instead of silently zero-filling. Any future converter/component naming mismatch fails loudly. Non-attention weights keep the warn-and-fill behavior.

Verification then caught a further pre-existing bug: Olmo-3 declares **per-layer-type rope** in transformers 5.x, but the mapping read the defunct `rope_scaling` attribute, so yarn never engaged. Separately, `attn_types` without `use_local_attn` is inert, which left sliding masks unwired. Fixed via a new `yarn_global_attn_only` flag and a per-layer-type config mapping.

## Verification on real checkpoints (RTX PRO 6000, bf16 for gpt-oss / fp32 for Olmo-3)

| Test | Issue's broken state | This PR |
| --- | --- | --- |
| Bridge, gpt-oss-20b resid_post (L0/12/23) | cos 0.84 / 0.04 / 0.19 | cos 1.0000 / 1.0000 / 1.0000 |
| Bridge, gpt-oss-20b layer-0 pattern vs HF `output_attentions` | — | max diff at bf16 noise |
| HT, gpt-oss-20b | `TypeError` crash; post-crash-fix logits top-1 0.00 | loads; cos 0.9998, top-1 1.00 |
| HT, Olmo-3-1025-7B | 64 missing-key warnings, 32/32 K/V zeroed, cos 0.28–0.84 | 0 warnings, 0/32 zeroed, cos 1.0000, rel ≤ 1e-6 |

CI-runnable regression tests use tiny random checkpoints through the production load paths (sinks, sliding windows, and yarn all exist unquantized): bridge-vs-HF parity via `boot_transformers`, HT-vs-HF parity via `convert_hf_model_config`. Both at fp32 machine precision (≤ 3e-7).

## Type of change

The only intentionally breaking behavior is narrow: state dicts that are missing attention weight matrices now raise in `fill_missing_keys` instead of loading with silently zeroed attention. Any code relying on the old behavior was computing wrong numbers.

# Checklist:


jlarson4:
@hijohnnylin Thanks for bringing these GPT OSS & OLMo 3 errors to my attention. They should be resolved as of this PR, once CI passes I will bundle this into a release and get that out. Let me know if you have any questions or concerns

PR #1625, opened 2026-08-08
Fix run_with_cache BOS handling for string inputs

### LINKED ISSUE #1624: [Bug Report] TransformerBridge.run_with_cache ignores prepend_bos=False for string inputs
**Describe the bug**

`TransformerBridge.run_with_cache()` ignores `prepend_bos=False` when its input is a single string. The cached forward therefore runs on a different token sequence from a direct forward with the same arguments.

This is silent: the call succeeds, but the logits and every position-indexed cached activation include an extra BOS position.

**Code example**

On current `dev-4.x` (`7db5f8dc`):

```python
import torch

text = "Hello"

tokens = bridge.to_tokens(text, prepend_bos=False)

with torch.no_grad():
    direct_logits = bridge(text, prepend_bos=False)
    cached_logits, cache = bridge.run_with_cache(
        text,
        prepend_bos=False,
    )

print(tokens.shape)
print(direct_logits.shape)
print(cached_logits.shape)
print(cache["hook_embed"].shape)
```

Observed:

```text
torch.Size([1, 1])
torch.Size([1, 1, 50257])
torch.Size([1, 2, 50257])
torch.Size([1, 2, 768])
```

I also reproduced this without downloading a checkpoint using a tiny `TransformerBridge.boot_native()` model and an offline tokenizer:

```text
to_tokens(prepend_bos=False): [[7]]
to_tokens(prepend_bos=True):  [[0, 7]]
direct forward shape:         (1, 1, 8)
run_with_cache shape:         (1, 2, 8)
cached hook_embed shape:      (1, 2, 8)
```

The direct and cached paths should use the same token sequence when given the same string and `prepend_bos` value.

The immediate cause is the eager string tokenization in `BridgeCore.run_with_cache()`:

```python
input_ids = self.to_tokens(processed_args[0])
```

and the equivalent `kwargs["input"]` branch. These calls do not forward the `prepend_bos` value from `kwargs`. The original value is later passed to `forward()`, but by then the input is already a token tensor, so it can no longer affect tokenization.

This can shift token positions in attention patterns, residual streams, activation patching, and attribution analyses. It can also produce a double BOS for preformatted chat-template text.

A fix should make the eager tokenization path consume and forward its tokenization options, including both `prepend_bos` and `padding_side`, and add parity coverage between direct and cached forwards for `prepend_bos=True` and `False`.

**System Info**

- Installed from source
- OS: Windows
- Python: 3.12.10
- Branch/commit: `dev-4.x` at `7db5f8dc`
- Reproduction uses CPU and does not require model downloads in the `boot_native` variant

**Additional context**

This is related to the BOS contract addressed for generation in #1439, but that PR did not update `run_with_cache()`. The older #990 changed this path to call `self.to_tokens()`, but did not forward tokenization options.

### Checklist

emerardd:
# Description

## Summary

- Preserve an explicit `prepend_bos` value when `BridgeCore.run_with_cache()` eagerly tokenizes a scalar string input.
- Consume the tokenization option before forwarding the resulting token tensor, rather than leaving an ineffective `prepend_bos` argument on the tensor forward.
- Add offline native-Bridge regression coverage for direct-vs-cached parity with both `prepend_bos=True` and `False`.

## Root cause and impact

`run_with_cache()` converts scalar string inputs with `self.to_tokens(input)` before calling `forward()`, but did not pass the caller's `prepend_bos` value. `to_tokens()` therefore fell back to `cfg.default_prepend_bos`, typically `True`. Although the original argument was subsequently forwarded, the input was already a token tensor and could no longer be retokenized.

As a result, `bridge(text, prepend_bos=False)` and `bridge.run_with_cache(text, prepend_bos=False)` could run on different token sequences. The cached path silently included an extra BOS position, shifting logits and position-indexed activations used for attention analysis, attribution, and activation patching.

The fix lives in `BridgeCore`, so it preserves the shared local/remote cache contract while leaving tensor/list inputs and driver-aware device placement unchanged. This PR intentionally does not broaden into `padding_side`, whose tokenization support differs between `TransformerBridge` and `RemoteBridge` and is not required to fix the demonstrated bug.

No documentation change is required because the existing `prepend_bos` contract is already documented; this change makes `run_with_cache()` conform to it.

Fixes #1624

## Type of change

### Screenshots

N/A — code-only behavior fix.

## Validation

- Target regression: `2 passed`.
- Full affected files (`test_boot_native.py` and `test_remote_bridge.py`): `62 passed`.
- Related model-bridge unit surface selected by `run_with_cache`, tokenizer, and `prepend_bos`: `68 passed, 6 skipped`.
- Offline tiny native Bridge parity: direct logits, cached logits, and `hook_embed` all have shape `(1, 1, 8)` with `prepend_bos=False`; maximum direct-vs-cached logit difference is `0.0`.
- `mypy .`: no issues in 424 source files.
- pycln, isort, Black, and `git diff --check`: passed for the changed files.

# Checklist:

- [x] I have made corresponding changes to the documentation (no change required; existing docs already specify this behavior)
- [ ] New and existing unit tests pass locally with my changes (the affected unit-test surface passed; the complete suite was not rerun locally)

### DIFF
--- a/tests/unit/model_bridge/test_boot_native.py
+++ b/tests/unit/model_bridge/test_boot_native.py
@@ -367,6 +367,35 @@ def test_boot_native_forward_and_cache():
     assert "blocks.0.attn.hook_pattern" in cache
 
 
+@pytest.mark.parametrize("prepend_bos", [True, False])
+def test_run_with_cache_forwards_prepend_bos_for_string_input(monkeypatch, prepend_bos):
+    cfg = _cfg()
+    bridge = TransformerBridge.boot_native(cfg)
+    bridge._tokenizer = object()
+    tokenization_calls = []
+
+    def to_tokens(input, prepend_bos=None, padding_side=None):
+        tokenization_calls.append((input, prepend_bos, padding_side))
+        if prepend_bos is None:
+            prepend_bos = bridge.cfg.default_prepend_bos
+        tokens = [0, 7] if prepend_bos else [7]
+        return torch.tensor([tokens])
+
+    monkeypatch.setattr(bridge, "to_tokens", to_tokens)
+    bridge.eval()
+
+    with torch.no_grad():
+        direct_logits = bridge("hello", prepend_bos=prepend_bos)
+        cached_logits, cache = bridge.run_with_cache("hello", prepend_bos=prepend_bos)
+
+    assert tokenization_calls == [
+        ("hello", prepend_bos, None),
+        ("hello", prepend_bos, None),
+    ]
+    torch.testing.assert_close(cached_logits, direct_logits)
+    assert cache["hook_embed"].shape[1] == direct_logits.shape[1]
+
+
 def test_boot_native_does_not_load_transformers_runtime():
     # Sanity that the native path doesn't depend on HuggingFace's `transformers`
     # for the runtime work — we check that calling boot_native doesn't trigger
--- a/transformer_lens/model_bridge/bridge_core.py
+++ b/transformer_lens/model_bridge/bridge_core.py
@@ -1293,14 +1293,16 @@ class BridgeCore:
         target_device = self._input_device()
         if processed_args and isinstance(processed_args[0], str):
             assert self.tokenizer is not None, "Tokenizer must be set to pass string input."
-            input_ids = self.to_tokens(processed_args[0])
+            prepend_bos = kwargs.pop("prepend_bos", None)
+            input_ids = self.to_tokens(processed_args[0], prepend_bos=prepend_bos)
             if target_device is not None:
                 input_ids = input_ids.to(target_device)
             kwargs["input_ids"] = input_ids
             processed_args = processed_args[1:]
         elif "input" in kwargs and isinstance(kwargs["input"], str):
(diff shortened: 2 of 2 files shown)


emerardd:
The only failing check is Full Code Coverage, and both failures reproduce identically on the exact base commit (`7db5f8dc`) before this PR changes:

- `test_one_left_padded_row_does_not_perturb_its_neighbours`
  - maximum absolute difference: `9.1552734375e-05` with `atol=1e-6`
- `test_bridge_hooked_parity_multi_step_optimization`
  - step 1 mean logits difference: `0.0010319872526451945` with a `0.001` threshold

Base coverage run:
https://github.com/TransformerLensOrg/TransformerLens/actions/runs/31233076384/job/93040556397

PR coverage run:
https://github.com/TransformerLensOrg/TransformerLens/actions/runs/31254874477/job/93096680161

The values and failure locations are identical in both runs. This PR only changes scalar-string tokenization in `run_with_cache()`, while both failing tests use direct tensor forward paths.

I have therefore kept this PR focused and have not changed the unrelated numerical tolerances. Please let me know whether you would prefer the coverage job to be retried, or these base-branch failures to be handled separately.

PR #1627, opened 2026-08-08
fix(bridge): gate batched-list position_ids on the target model

### LINKED ISSUE #1626: [Bug Report] Batched list input injects position_ids into models whose forward cannot accept it
**Describe the bug**

When `forward()` or `generate()` receives a batched list of prompts, the bridge builds an `attention_mask` and `position_ids` itself so pad tokens don't contaminate the forward (`transformer_bridge.py:1737` and `:2148`). The mask is safe for any model, but the `position_ids` are handed over with no check that the target accepts them. A model whose forward takes neither `position_ids` nor `**kwargs` raises `TypeError` where it would otherwise have returned logits.

This is the same gap that was identified during review of #1610. That PR added `_accepts_derived_position_ids()` and gated the main `forward()` derivation on it, but we agreed to leave these two batched-list sites for a follow-up because no model could be found that demonstrably failed there. I've now reproduced it.

Affected models are remote-code architectures with fixed-signature forwards — `LLaDAModelLM` and `HyenaDNAForCausalLM` among the supported ones. This is narrow: of 436 HF top-level forwards I surveyed, only 9 are signature-explicit and none of those are text LMs. It is not, however, hypothetical.

**Code example**

`tests/integration/model_bridge/test_llada_adapter.py` builds a tiny model mirroring the released dense LLaDA contract, whose forward is `(input_ids, attention_mask, attention_bias, use_cache, output_attentions, output_hidden_states)` — no `position_ids`, no `**kwargs`.

```python
bridge = ...  # a bridge over that model, with a tokenizer that has a pad token
bridge(["token_5 token_7 token_9", "token_5"], return_type="logits")
```

```
TypeError: TinyLLaDAModelLM.forward() got an unexpected keyword argument 'position_ids'
kwargs attempted: ['attention_mask', 'position_ids']
```

A single unbatched string is unaffected — no `position_ids` is injected — so this is specific to the batched-list path.

**System Info**

* Installed from source (`uv sync`), branch `dev-4.x`
* macOS (Darwin arm64), Python 3.12, transformers 5.13.0

**Additional context**

The fix looks mechanical: gate both sites on the existing `_accepts_derived_position_ids()`, leaving the `attention_mask` unconditional since it's safe everywhere and withholding it would reintroduce the padding contamination the branch exists to prevent. I have this working locally with a regression test, and will open a PR.

One testing note that cost me some time and may be worth knowing: a test that spies on `original_model.forward` will defeat this gate, because the gate reads that forward's signature and a bare `(*args, **kwargs)` wrapper looks like it accepts `position_ids`. `functools.wraps` preserves the real signature.

### Checklist

sohv:
Fixes #1626.

Batched list input builds an `attention_mask` and `position_ids` itself so pad tokens don't contaminate the forward. The mask is safe for any model, but the `position_ids` were handed over unchecked — a forward taking neither `position_ids` nor `**kwargs` raises `TypeError` where it would otherwise have returned logits.

This is the gap that was raised while reviewing #1610. That PR added `_accepts_derived_position_ids()` and gated the main `forward()` derivation, but we left these two sites for a follow-up because I couldn't produce a model that demonstrably failed there. The LLaDA test harness builds a fixed-signature forward in-process, which reproduces it:

```
TypeError: TinyLLaDAModelLM.forward() got an unexpected keyword argument 'position_ids'
kwargs attempted: ['attention_mask', 'position_ids']
```

## What I changed

Both sites now call the same `_accepts_derived_position_ids()` helper, so there's no new predicate. The `attention_mask` stays unconditional -it's safe for every model, and withholding it would reintroduce the padding contamination the branch exists to prevent.

## Verification

One regression test in the LLaDA suite, red on `dev-4.x` with the `TypeError` above and green with the fix. It asserts both halves: no `position_ids` reaches the model, and the `attention_mask` still does. A single unbatched string was never affected and is covered as a control.

Full unit + integration + acceptance: **6594 passed, 1 failed**. The failure is `test_bridge_hooked_parity_multi_step_optimization`, which fails identically on unmodified code and is `--ignore`d on the macOS CI job. mypy clean.

### DIFF
--- a/tests/integration/model_bridge/test_llada_adapter.py
+++ b/tests/integration/model_bridge/test_llada_adapter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import copy
+import functools
 import gc
 import math
 import weakref
@@ -661,6 +662,54 @@ def test_left_padding_does_not_inject_unsupported_position_ids(models: TinyModel
         reference_logits = models.reference(tokens, attention_mask=attention_mask).logits
         bridge_logits = models.bridge(tokens, attention_mask=attention_mask)
     torch.testing.assert_close(bridge_logits, reference_logits, rtol=1e-5, atol=1e-6)
+
+
+def test_batched_list_input_does_not_inject_unsupported_position_ids() -> None:
+    """Batched list input builds its own attention_mask and position_ids so pad
+    tokens don't contaminate the forward (#1626). The mask is safe for any model;
+    the position_ids are not, and this forward takes neither them nor **kwargs.
+
+    A local bridge rather than the module fixture: this needs a tokenizer, and
+    attaching one to the shared instance would leak into the other tests. The
+    tokenizer is given a BOS so the path under test is reached independently of
+    BOS handling elsewhere.
+    """
+    local = _build_models()
+    tokenizer = _offline_tokenizer()
+    tokenizer.bos_token = "<bos>"
+    local.bridge.tokenizer = tokenizer
+
+    seen: dict = {}
+    original = local.bridge.original_model.forward
+
+    # functools.wraps so inspect.signature() still resolves to the real forward:
+    # the gate reads that signature, and a bare (*args, **kwargs) spy would look
+    # like it accepts position_ids and defeat the check under test.
+    @functools.wraps(original)
+    def _spy(*args, **kwargs):
+        seen.clear()
+        seen.update(kwargs)
+        return original(*args, **kwargs)
+
+    local.bridge.original_model.forward = _spy
+    try:
+        with torch.inference_mode():
+            logits = local.bridge(["token_5 token_7 token_9", "token_5"], return_type="logits")
+        batched = dict(seen)
+        with torch.inference_mode():
+            local.bridge("token_5 token_7 token_9", return_type="logits")
+        unbatched = dict(seen)
+    finally:
+        local.bridge.original_model.forward = original
+
+    assert logits.shape[0] == 2
+    assert "position_ids" not in batched
+    # The mask is still supplied — withholding it would reintroduce the padding
+    # contamination this branch exists to prevent.
+    assert "attention_mask" in batched
+    # Control: a single unbatched string never reached this branch, so the gate
+    # must not have changed anything for it either.
+    assert "position_ids" not in unbatched
 
 
 def test_run_with_cache_exposes_hooks_without_hf_output_attentions(
--- a/transformer_lens/model_bridge/transformer_bridge.py
+++ b/transformer_lens/model_bridge/transformer_bridge.py
@@ -2248,24 +2248,32 @@ class TransformerBridge(BridgeCore, HookIntrospectionMixin, nn.Module):
                                     dtype=torch.long,
                                     device=device,
                                 )
-                            if "position_ids" in forward_kwargs:
-                                forward_kwargs["position_ids"] = forward_kwargs["position_ids"][
-                                    :, -1:
-                                ]
-                            elif running_attention_mask is not None:
-                                # total_len - 1 counts pad slots, so it is wrong
-                                # for a left-padded prompt. Derive the new token's
-                                # position from the mask instead.
-                                forward_kwargs["position_ids"] = utils.get_offset_position_ids(
-                                    0, running_attention_mask.long()
-                                )[:, -1:]
-                            else:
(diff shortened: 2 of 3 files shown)


transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -2142,9 +2146,12 @@ def _generate_tokens(
                         ).to(self.cfg.device)
                         self.tokenizer.padding_side = _prev_side
                         forward_kwargs["attention_mask"] = attn_mask
-                        position_ids = attn_mask.long().cumsum(-1) - 1
-                        position_ids.masked_fill_(attn_mask == 0, 1)
-                        forward_kwargs["position_ids"] = position_ids
+                        # Same target gate as the forward() path: the mask is safe
+                        # for every model, the derived positions are not (#1626).
+                        if self._accepts_derived_position_ids():
```
jlarson4, line 2194: For a `gate=False` model that accepts `position_ids` (opt-125m), this gate makes cached steps fall into the `torch.full(total_len - 1)` fallback at line 2220, which counts pad slots, causing logit drift. Can you gate the continuation's derive/fallback injection starting at line 2213-2226 on the same helper?

tests/integration/model_bridge/test_llada_adapter.py
```diff
@@ -663,6 +664,54 @@ def test_left_padding_does_not_inject_unsupported_position_ids(models: TinyModel
     torch.testing.assert_close(bridge_logits, reference_logits, rtol=1e-5, atol=1e-6)
 
 
+def test_batched_list_input_does_not_inject_unsupported_position_ids() -> None:
```
jlarson4, line 667: The `_generate_tokens` portion of this fix is untested here and neither fixed-signature architecture can reach that site. Can you add a cached-vs-uncached batched-generate parity test on a `gate=False` model?

jlarson4 (commented):
Hi @sohv! Sorry for the delayed review on this, I have been a bit under the weather.

This looks great, the `forward` portion is perfect. Just a couple notes on the `_generate_tokens` path below.

sohv:
@jlarson4 A quick request to review this PR and merge it if you have no issues. I reviewed the PR again today but let me know if there is anything to be addressed here.

jlarson4:
> @jlarson4 A quick request to review this PR and merge it if you have no issues. I reviewed the PR again today but let me know if there is anything to be addressed here.

The review just posted! Sorry I didn't get it to you yesterday, was down with the flu.

sohv:
@jlarson4 It's totally alright and thank you for highlighting these issues. I will work on fixing them and will update the PR soon.

sohv:
**The cached-step fallback** - Gating only the prompt derivation moved the injection instead of stopping it. Every branch of that block supplies `position_ids`, so a refused model fell straight through to `torch.full(total_len - 1)`. Measured on `facebook/opt-125m`:

```
before this PR   cached steps [[9],[2]]  [[10],[3]]  [[11],[4]]   <- per row, correct
this PR as sent  cached steps [[9],[9]]  [[10],[10]] [[11],[11]]  <- pad slots counted
now              cached steps None                                <- OPT derives its own
```

Cached-vs-uncached max |logit diff| goes `9.5e-06` → `1.46e+00` → `9.5e-06`, so the fix restores the pre-PR value exactly. I gated the whole three-way block rather than each branch, so a model that owns its position derivation gets the mask alone, matching the uncached path.

**The test gap** - LLaDA raises `NotImplementedError` on `generate`, so neither fixed-signature architecture can reach `_generate_tokens` at all and that half of the fix was entirely unverified.

I have added `tests/integration/model_bridge/test_batched_generate_position_ids.py` and it is to be noted that a cached-vs-uncached parity test on decoded text would **not** have caught this - greedy argmax absorbs the drift and the strings match in both states. The tests compare logits instead and separately assert the mechanism since the fallback supplies a per-batch constant that a coarser check can miss. Two of the five are red on the previous commit; `distilgpt2` is the `gate=True` control and stays green throughout.

I also rebased onto current `dev-4.x`, since the branch was ~30 commits behind and re-ran everything afterwards so the numbers above reflect the pushed state. mypy clean. I couldn't run the full three-tier suite locally this time for environment reasons so the above is targeted verification and I'm leaning on CI for the wider check.

jlarson4:
Great work @sohv! Thank you for doing the rebase as well, that is very helpful. I am merging now

PR #1629, opened 2026-08-08
fix(tokenizer): do not strip a BOS token the tokenizer does not have

sohv:
`get_tokens_with_bos_removed` assumed the tokenizer has a `bos_token_id`. When `cfg.tokenizer_prepends_bos` goes stale the helper is reached anyway, and it then does damage in two directions.

Under right padding, which is the default, it drops the first token unconditionally. For a BERT tokenizer that silently removes `[CLS]` and returns a plausible looking wrong result. Under left padding it evaluates `(tokens == None).int()`, which is a Python bool rather than a tensor, and raises `AttributeError: 'bool' object has no attribute 'int'`.

## What I changed

With no `bos_token_id` there is nothing to remove, so the helper returns the tokens unchanged. This is correct however the config got out of sync, it fixes both symptoms, and it leaves the `boot_transformers` path untouched.

## What I did not change and why

The root cause is the reassignment test at `bridge_core.py:113`. I left it alone, and not only because it alters build and attach semantics. Correcting the flag would route `to_tokens(prepend_bos=True)` into the manual prepend branch at `transformer_bridge.py:716`, which calls `get_input_with_manually_prepended_bos(tokenizer.bos_token, ...)` and raises `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'` when there is no BOS token. The stale flag is currently masking that, so the root cause fix needs the prepend path hardened first. Happy to take that on as a follow up if you want it.

## Verification

Six unit tests, four of which are red on `dev-4.x`. The two that pass either way are controls confirming a real BOS is still removed for `distilgpt2` on both padding sides.

Reproduced with two off the shelf tokenizers, `google-bert/bert-base-cased` and `google-t5/t5-small`, both of which have `bos_token_id` set to `None`.

### DIFF
--- a/tests/unit/utilities/test_get_tokens_with_bos_removed.py
+++ b/tests/unit/utilities/test_get_tokens_with_bos_removed.py
@@ -1,0 +1,82 @@
+"""Tests for get_tokens_with_bos_removed when the tokenizer has no BOS token.
+
+Callers gate this helper on ``cfg.tokenizer_prepends_bos``. That flag is set by
+``detect_tokenizer_bos_eos``, which requires a ``bos_token_id`` — so a tokenizer
+with none should never reach here. It does when the flag is stale: a bridge built
+via ``build_bridge_from_module(tokenizer=None)`` keeps the config default of True,
+and the tokenizer setter only re-runs detection on *re*-assignment.
+
+Trusting a stale flag is not harmless. Under right padding the helper drops the
+first token unconditionally, which silently removes ``[CLS]`` from a BERT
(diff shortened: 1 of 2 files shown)


jlarson4:
Looks good @sohv, thanks for filing and tackling this issue. Feel free to file the follow-up and work on that at your convenience.

jlarson4:
@sohv It looks like @Chinmayrawat15 has already addressed that follow up in #1634.

sohv:
@jlarson4 Great and thanks for the merge! Please also take a look at #1627 and let me know if you have any issues.

Issue #1630, opened 2026-08-08
[Bug Report] Phi-3 bridge construction crashes when head_dim != hidden_size // num_attention_heads


physicsrob:
**Describe the bug**

 `Phi3ForCausalLM` configs may set `head_dim` independently of `hidden_size // num_attention_heads` -- HF sizes the fused `qkv_proj` as  `(num_attention_heads + 2 * num_key_value_heads) * head_dim`, so the fused width need not equal `3 * hidden_size`. The generic config mapper handles this correctly (`sources/transformers.py` maps an explicit `head_dim` to `cfg.d_head`), but `Phi3ArchitectureAdapter` ignores it and re-derives `d_head = d_model // n_heads` in two places: the split sizes installed in `weight_processing_conversions`, and the live QKV split in  `_split_phi3_qkv`. Bridge construction then crashes in `torch.split` for any such checkpoint.

To be clear this has probably never come up before, because as far as I know previous phi3 checkpoints all had `head_dim = hidden_size // num_attention_heads`, but I've recently released a compiler that generates phi3 checkpoints from a computation graph (https://github.com/physicsrob/torchwright), and it uses fewer attention heads when possible to reduce checkpoint size. Huggingface supports these checkpoints, and the fix here is simple, so I think it makes sense to patch.

  **Code example**

  Self-contained, tiny random weights, ~8 s on CPU:

  ```python
  import tempfile

  from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer

 cfg = AutoConfig.for_model(
      "phi3",
      vocab_size=100,
      hidden_size=32,
      intermediate_size=64,
      num_hidden_layers=2,
      max_position_embeddings=64,
      pad_token_id=0,
      bos_token_id=1,
      eos_token_id=2,
      num_attention_heads=4,
      num_key_value_heads=2,
      head_dim=6,  # explicit, != 32 // 4
  )
  path = tempfile.mkdtemp()
  AutoModelForCausalLM.from_config(cfg).save_pretrained(path)
  AutoTokenizer.from_pretrained("gpt2").save_pretrained(path)  # any tokenizer

  bridge = TransformerBridge.boot_transformers(path, device="cpu")
  print("ok, d_head =", bridge.cfg.d_head)
  ```

  On current `dev` (c64d2a4):

  ```
    File ".../transformer_lens/model_bridge/generalized_components/joint_qkv_attention.py", line 287, in set_original_component
[...]

PR #1631, opened 2026-08-08
Fix Phi-3 bridge support for explicit head_dim

physicsrob:
# Description

Fixes TransformerLensOrg/TransformerLens#1630

This is kind of an obscure bug, but technically huggingface supports `Phi3ForCausalLM` configs where `head_dim != hidden_size // num_attention_heads`. This probably has never come up before, but I recently released a compiler that creates `Phi3ForCausalLM` checkpoints from a computation graph, and it trims unused heads to save space. Huggingface supports these checkpoints, so it seems reasonable to patch TransformerLens to also support these checkpoints, and the fix is small.

Bridge construction crashes for any `Phi3ForCausalLM` config where `head_dim != hidden_size // num_attention_heads`. The generic config mapper already handles this correctly: `sources/transformers.py` maps an explicit `head_dim` to `cfg.d_head`, falling back to `d_model // n_heads` when absent. But `Phi3ArchitectureAdapter` ignored it, re-deriving `d_head = d_model // n_heads` in two places.

The fix uses `cfg.d_head` at both sites. Models without an explicit `head_dim` are unaffected: the mapper's fallback already sets `cfg.d_head = d_model // n_heads`, so the adapter computes identical sizes for every previously-working config.

**Validation on a real checkpoint** ([`physicsrob/torchwright-calculator-memorize-max-digits-2`](https://huggingface.co/physicsrob/torchwright-calculator-memorize-max-digits-2) @ `e87a8852`, fp32, CPU, transformers 5.14.1):
- the documented `boot_transformers` call succeeds with no tokenizer accommodation; `cfg.d_head == 64`
- bridge logits match an independently-loaded HF eager forward with max abs diff exactly 0.0 (argmax equal at every position)
- greedy generation for `"12*34\n"` completes to `408`
- `run_with_cache` returns finite tensors for all 16 layers, with `hook_z` shaped `[1, 7, 57, 64]`

## Tests

- `tests/unit/model_bridge/supported_architectures/test_phi3_adapter.py` -- new `TestPhi3ExplicitHeadDim` class with `d_head` deliberately decoupled from `d_model // n_heads` (MHA: `d_model=16, n_heads=3, d_head=4`; GQA: `n_heads=4, n_kv_heads=2, d_head=3`), covering the `weight_processing_conversions` split sizes and the live `_split_phi3_qkv` path including bias handling. 5 of the 6 tests fail on the unfixed adapter; the sixth pins derived geometry unchanged.
- `tests/unit/model_bridge/test_config_mapping.py` -- new: pins explicit-`head_dim` precedence in `map_default_transformer_lens_config`, plus the absent and `None` fallbacks. Nothing in the suite previously discriminated the two branches -- every synthetic config that set `head_dim` chose a value equal to `hidden_size // num_attention_heads`.
- `tests/integration/model_bridge/test_phi3_tiny.py` -- new: network-free tiny Phi-3 bridges (real weights, explicit `head_dim`, MHA + GQA); asserts construction succeeds, split Q/K/V weights are bitwise slices of the fused matrix, logits match an independent HF eager forward, and cached per-head activations use the explicit `head_dim`. All 8 crash on the unfixed adapter.

## Type of change

No docs changes needed -- the fix makes the already-documented `boot_transformers` behavior hold for these configs.

# Checklist:

### DIFF
--- a/tests/integration/model_bridge/test_phi3_tiny.py
+++ b/tests/integration/model_bridge/test_phi3_tiny.py
@@ -1,0 +1,131 @@
+"""Tiny-random Phi-3 bridge tests for configs with an explicit ``head_dim``.
+
+Valid Phi-3 configs may set ``head_dim`` independently of
+``hidden_size // num_attention_heads`` (the fused QKV width then differs from
+``3 * hidden_size``). The adapter previously re-derived ``d_head`` from
+``d_model // n_heads`` and crashed splitting the fused matrix during bridge
+construction. Uses tiny programmatic configs with real weights — no network
+access or weight downloads. Fixture pattern mirrors
+tests/integration/model_bridge/test_granite_moe_hybrid_adapter.py.
+"""
+
+import pytest
+import torch
+from transformers import AutoConfig, AutoModelForCausalLM
+
+from transformer_lens.model_bridge.bridge import TransformerBridge
+from transformer_lens.model_bridge.sources import build_bridge_config_from_hf
+from transformer_lens.model_bridge.supported_architectures.phi3 import (
+    Phi3ArchitectureAdapter,
+)
+
+FP32_TOL = 1e-5
+
+# head_dim is decoupled from hidden_size // num_attention_heads in both:
+# MHA derives 32 // 3 = 10 vs explicit 8; GQA derives 32 // 4 = 8 vs explicit 6.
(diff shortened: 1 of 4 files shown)


jlarson4:
Looks great @physicsrob, thank you for filing and resolving this issue! Merging now

PR #1633, opened 2026-08-09
fix(bridge): restore native stop_at_layer and input_to_embed

### LINKED ISSUE #1632: [Bug Report] Native TransformerBridge stop_at_layer never stops, causing input_to_embed() to return logits
**Describe the bug**

`TransformerBridge.boot_native()` does not honor `stop_at_layer` in a direct forward pass. As a consequence, `input_to_embed()` returns final vocabulary logits instead of the residual stream entering block 0.

This is silent: the calls succeed, but every requested stop point produces the full forward output. If `d_vocab == d_model`, even a shape check cannot distinguish the logits from the promised residual stream.

The root cause is `BlockBridge._extract_layer_idx()`. Native Bridge blocks are named `layers.0`, `layers.1`, and so on, while the parser only recognizes `\.layers.N` with a preceding dot:

```python
re.search(r"blocks\.(\d+)", self.name)
or re.search(r"\.h\.(\d+)", self.name)
or re.search(r"\.layers\.(\d+)", self.name)
```

For a top-level name such as `layers.0`, layer extraction returns `None`, so `_check_stop_at_layer()` never raises `StopAtLayerException`.

**Code example**

```python
import torch

    d_model=32,
    d_head=16,
    n_heads=2,
    n_layers=3,
    n_ctx=8,
    d_vocab=16,
    d_mlp=64,
    act_fn="gelu",
    normalization_type="LN",
    seed=0,
)

print([block.name for block in bridge.blocks])
for stop_at_layer in (0, 2, -1):
    print(stop_at_layer, tuple(bridge(tokens, stop_at_layer=stop_at_layer).shape))

residual, _, _, _ = bridge.input_to_embed(tokens)
print("input_to_embed", tuple(residual.shape))
```

Observed:

```text
['layers.0', 'layers.1', 'layers.2']
0  (1, 3, 16)
2  (1, 3, 16)
-1 (1, 3, 16)
input_to_embed (1, 3, 16)
```

All four results have the vocabulary dimension (`d_vocab=16`), showing that they are final logits. `stop_at_layer=0` and `input_to_embed()` should instead return the pre-block residual with shape `(1, 3, 32)`; other stop points should likewise return an intermediate residual with final dimension `d_model=32`.

`run_with_cache(tokens, stop_at_layer=...)` can appear correct because it installs a separate stop hook on `blocks.{k}.hook_in`. That path bypasses the broken block-name parser, so existing cache tests do not cover direct `forward(..., stop_at_layer=...)` or the `input_to_embed()` helper that depends on it.

**System Info**

* Installed from source on current `dev-4.x`
* Windows 11, CPU
* Python 3.12.10
* Reproduction uses `TransformerBridge.boot_native()` and requires no model download

**Additional context**

A boundary-aware parser such as `(?:^|\.)layers\.(\d+)` would recognize both top-level native names (`layers.0`) and nested HF-style names (`model.layers.0`). The `h` pattern should use the same boundary form so a future top-level `h.0` layout does not reproduce the bug.

Regression coverage should directly exercise `boot_native()` rather than relying only on `run_with_cache()`:

* direct `forward(..., stop_at_layer=0)`
* a positive intermediate layer and a negative layer index
* `input_to_embed()` returning a `d_model` residual
* resuming the returned residual with `start_at_layer=0` and matching a complete forward pass

### Checklist

emerardd:
# Description

Fixes #1632.

Native `TransformerBridge` blocks are named `layers.N`, but the block layer-index parser only recognized `layers.N` when it had a preceding dot. As a result, direct `forward(..., stop_at_layer=...)` calls never stopped and `input_to_embed()` returned final vocabulary logits instead of the residual stream entering block 0.

This change:

- makes the layer-name parser boundary-aware for `blocks.N`, `h.N`, and `layers.N`, including top-level native names;
- normalizes negative `stop_at_layer` indices before passing the stop point to blocks;
- lets the native model accept `inputs_embeds`, which is required to resume a residual with `start_at_layer`;
- adds direct stopping and `input_to_embed()` round-trip regression coverage for an offline `boot_native()` model.

## Type of change

### Screenshots

Not applicable.

# Verification

- Before the fix, the four new regression cases failed: direct stops returned logits with the `d_vocab` dimension, and residual re-entry raised because the native model required `input_ids`.
- `uv run pytest tests/unit/model_bridge/test_boot_native.py` — 35 passed
- Focused direct-stop and round-trip regression selection — 4 passed, 31 deselected
- `uv run pytest tests/integration/model_bridge/test_bridge_input_to_embed.py tests/integration/model_bridge/test_bridge_stop_at_layer.py::test_manual_hooks_stop_at_layer_compat_with_processing` — 5 passed
- `uv run mypy .` — success, 424 source files
- `pycln`, `isort`, and `black` checks on the four changed files — passed
- `git diff --check` — passed

The complete unit-test suite was not rerun; only the affected unit and integration surface was run locally.

# Checklist:

- [x] I have made corresponding changes to the documentation (not applicable; behavior now matches the documented API contract)
- [ ] New and existing unit tests pass locally with my changes (only the affected test surface was run)

### DIFF
--- a/tests/unit/model_bridge/test_boot_native.py
+++ b/tests/unit/model_bridge/test_boot_native.py
@@ -75,6 +75,42 @@ def test_boot_native_returns_bridge_over_native_model():
     assert isinstance(bridge.original_model, NativeModel)
 
 
+@pytest.mark.parametrize("stop_at_layer", [0, 2, -1])
+def test_boot_native_direct_stop_matches_cached_stop(stop_at_layer: int):
+    bridge = TransformerBridge.boot_native(_cfg(n_layers=3))
+    bridge.eval()
+    tokens = torch.tensor([[1, 2, 3]])
+
+    with torch.no_grad():
+        expected, _ = bridge.run_with_cache(tokens, stop_at_layer=stop_at_layer)
+        actual = bridge(tokens, stop_at_layer=stop_at_layer)
+
+    assert actual.shape == (1, 3, bridge.cfg.d_model)
+    torch.testing.assert_close(actual, expected)
+
+
+def test_boot_native_input_to_embed_round_trip():
+    bridge = TransformerBridge.boot_native(_cfg(n_layers=3))
+    bridge.eval()
+    tokens = torch.tensor([[1, 2, 3]])
+
+    with torch.no_grad():
+        expected = bridge(tokens)
+        residual, returned_tokens, shortformer_pos_embed, attention_mask = bridge.input_to_embed(
+            tokens
+        )
+        actual = bridge(
+            residual,
+            start_at_layer=0,
+            attention_mask=attention_mask,
+        )
+
+    assert residual.shape == (1, 3, bridge.cfg.d_model)
+    assert torch.equal(returned_tokens, tokens)
+    assert shortformer_pos_embed is None
+    torch.testing.assert_close(actual, expected)
+
+
 def test_native_state_dict_round_trip_restores_parameters():
     bridge = TransformerBridge.boot_native(_cfg())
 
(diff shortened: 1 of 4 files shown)


jlarson4:
Thanks for filing and resolving this issue @emerardd, looks great, merging as is.

PR #1635, opened 2026-08-09
fix(bridge): stop_at_layer never stops on native Bridge

### LINKED ISSUE #1632: [Bug Report] Native TransformerBridge stop_at_layer never stops, causing input_to_embed() to return logits

LightWork666:
Fixes #1632

## What was broken

`stop_at_layer` silently did nothing on native Bridge — every stop value returned full logits instead of the residual stream. Two root causes:

**1. Regex mismatch in `_extract_layer_idx`** (`block.py`)

Native Bridge blocks are named `layers.0`, `layers.1`, etc. (top-level). The existing pattern `\.layers\.(\d+)` requires a preceding dot, so it matched `model.layers.0` (LLaMA-style) but returned `None` for `layers.0`. Since `_check_stop_at_layer` only fires when the index is known, the exception was never raised.

**2. Negative index stored raw** (`transformer_bridge.py`)

`block._stop_at_layer_idx = stop_at_layer` stored `-1` directly. `_check_stop_at_layer` then compared `layer_idx == -1` (e.g. `2 == -1`) which never matched. `bridge_core.py` already normalizes negative indices in its own stop path — this brings `transformer_bridge.py` in line.

## Changes

- `block.py`: `r"\.layers\.(\d+)"` → `r"(?:^|\.)layers\.(\d+)"` — matches both top-level and nested naming
- `transformer_bridge.py`: normalize negative `stop_at_layer` before stamping it on blocks
- `tests/unit/model_bridge/test_native_stop_at_layer.py`: 7 new unit tests covering stop values 0, 1, 2, -1, value-distinctness between stops, and `input_to_embed` returning `d_model` not `d_vocab`

## Test plan

- [ ] `uv run pytest tests/unit/model_bridge/test_native_stop_at_layer.py -v` — 7/7 pass
- [ ] `uv run pytest tests/unit/model_bridge/generalized_components/ -q` — 184/184 pass
- [ ] `uv run pytest tests/unit/model_bridge/ -q -k native` — all pass


jlarson4:
@LightWork666 I am going to close this, but thank you for taking a look at it! #1633 was already filed in response to #1632  by the issues creator. It passed my review without any comments and is merged now.

PR #1638, opened 2026-08-10
Fix TransformerBridge temporary hook cleanup

### LINKED ISSUE #1636: [Bug Report] TransformerBridge temporary hook helpers remove pre-existing non-permanent hooks
**Describe the bug**

`TransformerBridge` temporary hook helpers can remove non-permanent hooks that were installed before the helper call.

For example, `run_with_cache()` adds its own cache hook and then cleans up with an unscoped `HookPoint.remove_hooks(dir=...)`. Because no hook level or exact handle is supplied, cleanup removes every non-permanent hook in that direction on the touched HookPoint, including a user's pre-existing intervention or recording hook.

The call succeeds and its returned cache is correct, but later forwards silently run without the user's hook. The same unscoped cleanup pattern is present in `BridgeCore.hooks()` and `BridgeCore.run_with_hooks()`, and it also breaks nested hook-context isolation.

**Code example**

```python
import torch

    d_model=32,
    d_head=16,
    n_heads=2,
    n_layers=2,
    n_ctx=8,
    d_vocab=16,
    d_mlp=64,
    act_fn="gelu",
    normalization_type="LN",
    seed=0,
)
bridge = TransformerBridge.boot_native(cfg).eval()
hook_name = "blocks.0.hook_out"
hook_point = bridge.hook_dict[hook_name]
events = []


def existing_hook(value, hook):
    events.append("existing")
    return value


hook_point.add_hook(existing_hook, dir="fwd")

with torch.no_grad():
    bridge.run_with_cache(tokens, names_filter=hook_name)

handles_after_cache = len(hook_point.fwd_hooks)
events_after_cache = len(events)

with torch.no_grad():
    bridge(tokens)

print(f"handles_after_cache={handles_after_cache}")
print(f"hook_fired_during_cache={events_after_cache > 0}")
print(f"hook_fired_after_cache={len(events) > events_after_cache}")
```

Observed on current `dev-4.x` at `0d1259ad`:

```text
handles_after_cache=0
hook_fired_during_cache=True
hook_fired_after_cache=False
```

Expected:

```text
handles_after_cache=1
hook_fired_during_cache=True
hook_fired_after_cache=True
```

The existing hook should survive. Only hooks created by `run_with_cache()` should be removed when that call exits.

The root cause is the unscoped cleanup in the shared Bridge helpers:

- `BridgeCore.hooks()` calls `hook_point.remove_hooks(dir=direction)`.
- `BridgeCore.run_with_hooks()` calls the same unscoped removal.
- `BridgeCore.run_with_cache()` calls `hp.remove_hooks(dir="both" if incl_bwd else "fwd")`.

`HookPoint.remove_hooks()` already supports filtering by `level`. The legacy `HookedRootModule.hooks()` increments `context_level`, tags hooks added by the current context with that level, and removes only that level in `finally`. Bridge currently exposes level-aware reset machinery but does not use scoped levels for these temporary helpers.

This is not only an internal handle-count discrepancy. In a separate offline control, a pre-existing hook zeroed `blocks.0.hook_out`. The cached call reflected the intervention, but cleanup deleted the hook, and the next forward exactly matched the clean model:

```text
clean_vs_intervened_max       = 25.8200740814209
cached_vs_intervened_max      = 0.0
hook_count_after_cache        = 0
after_cache_vs_clean_max      = 0.0
after_cache_vs_intervened_max = 25.8200740814209
```

The same lifecycle audit also reproduced deletion of pre-existing hooks through `hooks()` and `run_with_hooks()`, deletion of an outer hook when an inner context exits, deletion on exception cleanup, and deletion of unrelated backward hooks.

**System Info**

- Installed from a source checkout
- Branch/commit: `dev-4.x` at `0d1259ad`
- Windows 11, CPU
- Python 3.12.10
- PyTorch 2.11.0+cpu
- Transformers 5.13.0
- The minimal reproduction uses `TransformerBridge.boot_native()` and requires no tokenizer or model download

**Additional context**

- #113 and merged PR #200 established permanent hooks and context-level isolation for the legacy hook helpers. The current behavior is an incomplete Bridge migration of those scoped semantics, not a request to make these pre-existing hooks permanent.
- #1538 added level-aware Bridge reset support, but the temporary helper cleanup remains unscoped.
- Closed #1323 reported the inverse backward-cleanup failure (temporary hooks leaking after context exit). This reproduction instead shows over-broad cleanup deleting hooks owned by callers or outer contexts, including forward hooks and `run_with_cache()`; it therefore needs scoped ownership rather than simply more cleanup.

A fix should give each temporary Bridge helper invocation a scoped hook level, add its own cache/stop/user/gradient hooks at that level, and remove only that level in exception-safe cleanup. Regression coverage should include pre-existing forward and backward hooks, nested contexts, exception paths, permanent hooks, aliases resolving to the same HookPoint, and the existing `reset_hooks_end=False` behavior.

### Checklist

emerardd:
# Description

## Summary

- Give TransformerBridge temporary hook helpers scoped context-level ownership.
- Tag user, cache, gradient-cache, and stop hooks with the current helper level, then remove only that level during cleanup.
- Preserve pre-existing, outer-context, backward, and permanent hooks across `hooks()`, `run_with_hooks()`, and `run_with_cache()`.
- Delay `run_with_cache()` hook attachment until input preprocessing is complete, and make partial hook installation exception-safe.
- Add offline regression coverage for ordinary and nested contexts, exception cleanup, backward caching, aliases, permanent hooks, and `reset_hooks_end=False`.

## Root cause and impact

BridgeCore previously tracked only each touched HookPoint and direction, then called unscoped `remove_hooks(dir=...)` when a temporary helper exited. That removed every non-permanent hook in the selected direction, including hooks installed before the helper call or owned by an outer context.

This could silently disable an intervention or recording hook after an otherwise successful cache/helper call. Later experiments would continue running without the intended hook and without an exception.

The fix mirrors the established `HookedRootModule` lifecycle contract: each helper invocation receives a context level, every temporary hook it owns is tagged with that level, and exception-safe cleanup removes only that level. Public `reset_hooks_end=False` behavior remains unchanged.

Fixes #1636

## Type of change

### Screenshots

N/A — code-only hook lifecycle fix.

## Validation

- HookPoint, hook-orchestration, and native-Bridge unit surface: `89 passed`.
- Affected cache/stop/backward cross-component surface: `9 passed, 17 deselected`.
- `mypy .`: no issues in 424 source files.
- pycln, isort, Black, and `git diff --check`: passed.
- The complete test suite was not rerun locally.

# Checklist:

- [x] I have made corresponding changes to the documentation (no documentation change required; this restores the existing temporary-hook contract)
- [ ] New and existing unit tests pass locally with my changes (the affected unit-test surface passed; the complete unit suite was not rerun)

### DIFF
--- a/tests/unit/model_bridge/supported_architectures/test_hook_orchestration_parity.py
+++ b/tests/unit/model_bridge/supported_architectures/test_hook_orchestration_parity.py
@@ -31,6 +31,14 @@ def _noop(value, hook=None):
     return value
 
 
+def _record(events, label):
+    def hook(value, hook=None):
+        events.append(label)
+        return value
+
+    return hook
+
+
 class TestResetHooks:
     def test_removes_forward_and_backward_from_every_point(self):
         """Default direction='both' clears bwd hooks even on non-io points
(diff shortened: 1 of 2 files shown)


emerardd:
CI update: all code-related checks on head ede493b7 are green, including compatibility checks on Python 3.10/3.11/3.12, full coverage, format, type, docstring, and benchmark checks. The two remaining failed notebook jobs (Attribution_Patching_Demo and Activation_Patching_in_TL_Demo) failed while downloading unchanged Hugging Face models with HTTP 429 Too Many Requests (gpt2 and NeelNanda/Attn_Only_2L512W_C4_Code); the later NameErrors are cascading from those load failures. This PR does not modify either notebook. Marking this ready for review.

jlarson4:
Great work on this @emerardd! Approved and merging as is.

Issue #1639, opened 2026-08-11
[Bug Report] BLOOM `TransformerBridge` maps `hook_attn_out` / `hook_mlp_out` to residual-added states, breaking HookedTransformer residual identities


hijohnnylin:
**Versions:** transformer-lens 3.7.0, also reproduced on 3.6.0; transformers 5.14.1; torch 2.13.0; CPU; float32

## Summary

On `bigscience/bloom-560m`, the `TransformerBridge` compatibility hooks:

```text
blocks.{layer}.hook_attn_out
blocks.{layer}.hook_mlp_out
```

contain residual-added states rather than additive attention and MLP contributions.

In `HookedTransformer`, these hook names represent the tensors added to the residual stream, so the following identities hold:

```text
resid_pre + attn_out = resid_mid
resid_mid + mlp_out  = resid_post
```

On the BLOOM bridge, the aliases instead expose the Hugging Face attention and MLP module outputs. BLOOM performs its residual additions inside those modules, so for this checkpoint:

```text
hook_attn_out = resid_mid
hook_mlp_out  = resid_post
```

Using those values as additive contributions double-counts an entire residual stream.

The converted `HookedTransformer` implementation is unaffected.

## Scope of the issue

This issue is specifically about the legacy TransformerLens-compatible aliases:

```text
blocks.{layer}.hook_attn_out
blocks.{layer}.hook_mlp_out
```

A canonical architecture-shaped hook such as `blocks.{layer}.attn.hook_out` may intentionally expose the literal output of the underlying Hugging Face submodule.

However, aliases named `hook_attn_out` and `hook_mlp_out` should preserve the established `HookedTransformer` semantics associated with those names: the additive components of the residual-stream decomposition.

## Root cause

BLOOM performs the residual addition inside each sublayer:

```python
class BloomAttention(nn.Module):
    def forward(self, hidden_states, residual, ...):
        ...
        output_tensor = self.dense(context_layer)
        output_tensor = dropout_add(
            output_tensor,
            residual,
            self.hidden_dropout,
            self.training,
        )
        return output_tensor, ...

class BloomMLP(nn.Module):
    def forward(self, hidden_states, residual):
        ...
        intermediate_output = self.dense_4h_to_h(hidden_states)
        output = dropout_add(
            intermediate_output,
            residual,
            self.hidden_dropout,
            self.training,
        )
        return output
```

The general relationship is:

```text
module output = residual argument + dropout(projected sublayer output)
```

[...]

PR #1649, opened 2026-08-11
fix(bridge): respect use_parallel_residual in the GPTNeoX adapter

### LINKED ISSUE #1644: [Bug Report] GPTNeoX bridge ignores `use_parallel_residual`, dropping `hook_resid_mid` on sequential checkpoints (all RedPajama-INCITE models)
**Describe the bug**

The GPTNeoX adapter hardcodes the parallel-residual wiring. `neox.py` sets `self.cfg.parallel_attn_mlp = True` unconditionally and builds `ParallelBlockBridge` unconditionally, ignoring HF's `use_parallel_residual`.

GPTNeoX ships both wirings. `GPTNeoXConfig.use_parallel_residual` defaults to `True` (Pythia), but `False` is a supported, real configuration, and HF branches on it:

```python
# transformers/models/gpt_neox/modeling_gpt_neox.py — GPTNeoXLayer.forward
if self.use_parallel_residual:
    mlp_output = self.mlp(self.post_attention_layernorm(hidden_states))
    hidden_states = mlp_output + attn_output + hidden_states
else:
    attn_output = attn_output + hidden_states              # <- a real resid_mid
    mlp_output = self.mlp(self.post_attention_layernorm(attn_output))
    hidden_states = mlp_output + attn_output
```

On the sequential branch there *is* a distinct post-attention residual, and `post_attention_layernorm` reads it — so `ln2.hook_in` is exactly `resid_mid`. But `ParallelBlockBridge` pops the `hook_resid_mid` alias by design, so the hook is silently unavailable on precisely the checkpoints that have one. `cfg.parallel_attn_mlp` also reports `True` for a model that is not parallel.

The failure is silent: logits are correct (the bridge delegates to HF's own block), so nothing raises. Only the hook surface and the config flag are wrong.

**This affects registered models, not just hypothetical configs.** Every RedPajama-INCITE checkpoint in [`supported_models.json`](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/transformer_lens/tools/model_registry/data/supported_models.json) is `GPTNeoXForCausalLM` with `use_parallel_residual: false`:

| Model | `use_parallel_residual` |
|---|---|
| `togethercomputer/RedPajama-INCITE-7B-Base` | `False` |
| `togethercomputer/RedPajama-INCITE-7B-Chat` | `False` |
| `togethercomputer/RedPajama-INCITE-7B-Instruct` | `False` |
| `togethercomputer/RedPajama-INCITE-Base-3B-v1` | `False` |
| `togethercomputer/RedPajama-INCITE-Chat-3B-v1` | `False` |
| `togethercomputer/RedPajama-INCITE-Instruct-3B-v1` | `False` |

(Read from each repo's `config.json`.) On all six, `blocks.{i}.hook_resid_mid` is missing today, so resid-mid activation patching, attribution and SAE work are silently unavailable on models the registry claims to support. Pythia is unaffected — it is genuinely parallel.

`StableLmArchitectureAdapter` and `FalconArchitectureAdapter` already guard exactly this — `stablelm.py:121` and `falcon.py:148` both select `ParallelBlockBridge if <flag> else BlockBridge`. GPTNeoX looks like an oversight rather than an intentional difference.

The same hardcode exists on the HookedTransformer side: `loading_from_pretrained.py:535` sets `"parallel_attn_mlp": True` in the `GPTNeoXForCausalLM` branch.

**Code example**

Seeded tiny, no hub access:

```python
import copy
import torch
from transformers import AutoModelForCausalLM, GPTNeoXConfig

cfg = GPTNeoXConfig(
    vocab_size=128, hidden_size=64, intermediate_size=128, num_hidden_layers=2,
    num_attention_heads=4, max_position_embeddings=128,
    use_parallel_residual=False,          # sequential wiring
)
cfg._attn_implementation = "eager"
torch.manual_seed(42)
hf = AutoModelForCausalLM.from_config(cfg).eval()
    hf, "GPTNeoXForCausalLM", hf_config=copy.deepcopy(cfg), tokenizer=None, device="cpu"
).eval()

print(type(bridge.blocks[0]).__name__)      # ParallelBlockBridge  (expected BlockBridge)
print(bridge.cfg.parallel_attn_mlp)         # True                 (expected False)

_, cache = bridge.run_with_cache(torch.randint(3, 128, (1, 10)))
print("blocks.0.hook_resid_mid" in cache)   # False                (expected True)
```

Logit parity against HF is exact (`0.0`) in both wirings, which is why this has gone unnoticed.

**System Info**

- Installed from source, branch `dev`
- macOS (Darwin arm64), Python 3.12, transformers 5.14.1, torch CPU, float32

**Additional context**

The root cause of the flag not reaching the adapter is that `use_parallel_residual` is not in `_HF_PASSTHROUGH_ATTRS` (`model_bridge/sources/_bridge_builder.py`), which is how Falcon's `parallel_attn` reaches its adapter. `sources/transformers.py:236` does derive `tl_config.parallel_attn_mlp` from it, but `neox.py` then overwrites that value.

Reading the resolved `parallel_attn_mlp` instead is not a safe substitute: `TransformerBridgeConfig.parallel_attn_mlp` defaults to `False`, whereas HF's GPTNeoX default is `True`, so a hand-constructed config (`boot_native`, and `tests/unit/model_bridge/supported_architectures/test_neox_adapter.py`) would silently flip to sequential. The HF field name with HF's default is the correct source, mirroring Falcon.

Separately, no test asserted the identity `ParallelBlockBridge` documents in its own docstring (`block.py:407`, `output = resid_pre + attn_out + mlp_out`) for any of the seven adapters that use it — which is the same class of gap #1639 was. I verified the identity does hold for all seven; the missing coverage is what let this wiring bug stay invisible.

- `neox.py` selects the block class from `use_parallel_residual`, mirroring `stablelm.py:121` / `falcon.py:148`, and stops overwriting `cfg.parallel_attn_mlp`.
- Add `use_parallel_residual` to `_HF_PASSTHROUGH_ATTRS` so the adapter can see it.
- Mirror on the HT side at `loading_from_pretrained.py:535` (AGENTS.md §2).

Acceptance:
- [ ] A sequential GPTNeoX bridge builds `BlockBridge`, reports `cfg.parallel_attn_mlp is False`, and exposes `hook_resid_mid`
- [ ] `resid_pre + attn_out == resid_mid` and `resid_mid + mlp_out == resid_post` on the sequential variant
- [ ] Parallel (Pythia/default) behaviour unchanged, including with a hand-built `TransformerBridgeConfig`
- [ ] `convert_hf_model_config` returns `parallel_attn_mlp` matching `use_parallel_residual`
- [ ] Regression coverage for the `ParallelBlockBridge` identity across the seven adapters that use it
- [ ] `make unit-test`, `make integration-test` and `uv run mypy .` clean

### Checklist

Chinmayrawat15:
# Description

Fixes #1644.

The GPTNeoX adapter hardcoded the parallel-residual wiring — `neox.py` set `cfg.parallel_attn_mlp = True` and built `ParallelBlockBridge` unconditionally, ignoring HF's `use_parallel_residual`. GPTNeoX ships both wirings, and on the sequential branch HF computes a genuine post-attention residual that `post_attention_layernorm` reads:

```python
# transformers/models/gpt_neox/modeling_gpt_neox.py — GPTNeoXLayer.forward
else:
    attn_output = attn_output + hidden_states              # <- a real resid_mid
    mlp_output = self.mlp(self.post_attention_layernorm(attn_output))
    hidden_states = mlp_output + attn_output
```

`ParallelBlockBridge` pops the `hook_resid_mid` alias by design, so that hook was silently missing on exactly the checkpoints that have one, and `cfg.parallel_attn_mlp` reported `True` for a model that is not parallel. Logits stayed correct (the bridge delegates to HF's own block), so nothing raised — which is why it went unnoticed.

This is not hypothetical. All six RedPajama-INCITE checkpoints in `supported_models.json` are `GPTNeoXForCausalLM` with `use_parallel_residual: false`, so resid-mid patching, attribution and SAE work were silently unavailable on registered models:

| Model | `use_parallel_residual` |
|---|---|
| `togethercomputer/RedPajama-INCITE-7B-Base` | `False` |
| `togethercomputer/RedPajama-INCITE-7B-Chat` | `False` |
| `togethercomputer/RedPajama-INCITE-7B-Instruct` | `False` |
| `togethercomputer/RedPajama-INCITE-Base-3B-v1` | `False` |
| `togethercomputer/RedPajama-INCITE-Chat-3B-v1` | `False` |
| `togethercomputer/RedPajama-INCITE-Instruct-3B-v1` | `False` |

Pythia is genuinely parallel and is unaffected.

## The fix

`StableLmArchitectureAdapter` and `FalconArchitectureAdapter` already guard this exact case, so the change mirrors them rather than inventing a pattern:

- **`neox.py`** — `block_cls = ParallelBlockBridge if use_parallel_residual else BlockBridge`, matching `stablelm.py:121` / `falcon.py:148`, and `cfg.parallel_attn_mlp` now follows the flag instead of being overwritten.
- **`_bridge_builder.py`** — added `use_parallel_residual` to `_HF_PASSTHROUGH_ATTRS`, which is how Falcon's `parallel_attn` already reaches its adapter.
- **`loading_from_pretrained.py`** — same hardcode existed on the HookedTransformer side; mirrored per AGENTS.md §2.

Reading the already-resolved `cfg.parallel_attn_mlp` instead looked simpler but is wrong: `TransformerBridgeConfig.parallel_attn_mlp` defaults to `False` while HF's GPTNeoX default is `True`, so a hand-built config (`boot_native`, and the existing `test_neox_adapter.py`) would silently flip to sequential. Using the HF field name with HF's default keeps the parallel path byte-identical — I confirmed the existing NeoX adapter unit tests pass unchanged.

## Test coverage

`ParallelBlockBridge` documents its identity in its own docstring (`block.py:407`, `output = resid_pre + attn_out + mlp_out`), but nothing asserted it for any of the seven adapters that use it — the same class of gap as #1639, and what let this wiring bug stay invisible.

New `tests/integration/model_bridge/test_parallel_residual_identities.py` covers all seven (`gptj`, `codegen`, `phi`, `cohere`, `neox`, `stablelm`, `falcon`) and exercises the three config-switchable families in both wirings. Fixtures are seeded tinies built from local HF configs, so there is no hub access and both wirings are reachable from one checkpoint shape — 20 tests in ~8s.

It is a real regression guard, not a rubber stamp: reverting the `neox.py` + `_bridge_builder.py` change fails exactly `test_sequential_variant_exposes_resid_mid[neox]` and `test_sequential_variant_decomposition[neox]`, with the other 18 still passing. Reverting the `loading_from_pretrained.py` change fails the new `convert_hf_model_config` case.

The new file is a sibling to `test_residual_decomposition_identities.py` (which explicitly scoped these architectures out); I updated that docstring to cross-reference rather than moving anyone's tests.

## Type of change

# Validation

`make test-pr` clean on macOS arm64 / Python 3.12 / transformers 5.14.1:

| Tier | Result |
|---|---|
| `make unit-test` | 4550 passed, 36 skipped, 44 deselected, 10 xfailed |
| `make docstring-test` | 18 passed, 25 skipped |
| `make acceptance-test` | 150 passed, 96 skipped, 11 deselected |
| `make integration-test` | 1182 passed, 23 skipped, 191 deselected, 1 xfailed |

- `uv run pytest tests/integration/model_bridge/test_parallel_residual_identities.py` — 20 passed in 8s
- `make check-format` clean; `uv run mypy .` — Success, no issues in 385 source files

# Noted but not included

`transformer_lens/benchmarks/hook_registration.py` still carries a BLOOM carve-out (lines ~246-255 and ~592-599) whose comment — "hook_attn_out and hook_mlp_out capture attn+residual instead of just attn" — was made false by #1640/#1642. It is unreachable today (both call sites in `main_benchmark.py:401-402` pass no `reference_activations`, so the functions return before it), so it is dead code with a misleading comment rather than a live suppression. Left out to keep this PR to one story — happy to send it separately.

# Checklist:

- [x] I have made corresponding changes to the documentation (behaviour is covered by tests; no doc pages describe this wiring)

### DIFF
--- a/tests/integration/model_bridge/test_parallel_residual_identities.py
+++ b/tests/integration/model_bridge/test_parallel_residual_identities.py
@@ -101,16 +101,19 @@ def _run(hf_architecture: str, config_cls: Any, config_kwargs: dict[str, Any]):
     if key not in _CACHE:
         config = config_cls(**config_kwargs)
         config._attn_implementation = "eager"
-        torch.manual_seed(42)
-        hf = AutoModelForCausalLM.from_config(config).eval()
-        bridge = build_bridge_from_module(
-            hf, hf_architecture, hf_config=copy.deepcopy(config), tokenizer=None, device="cpu"
-        ).eval()
+        # Fork the RNG so seeding here stays deterministic on a cache miss
+        # without leaking state into later unseeded tests on this worker.
+        with torch.random.fork_rng(devices=[]):
+            torch.manual_seed(42)
+            hf = AutoModelForCausalLM.from_config(config).eval()
+            bridge = build_bridge_from_module(
+                hf, hf_architecture, hf_config=copy.deepcopy(config), tokenizer=None, device="cpu"
+            ).eval()
 
-        torch.manual_seed(0)
-        tokens = torch.randint(3, config_kwargs["vocab_size"], (1, 10))
-        with torch.no_grad():
-            logits, cache = bridge.run_with_cache(tokens)
+            torch.manual_seed(0)
+            tokens = torch.randint(3, config_kwargs["vocab_size"], (1, 10))
+            with torch.no_grad():
+                logits, cache = bridge.run_with_cache(tokens)
         assert torch.isfinite(logits).all(), f"{hf_architecture} produced non-finite logits"
         _CACHE[key] = (bridge, cache)
     return _CACHE[key]
--- a/tests/integration/model_bridge/test_parallel_residual_identities.py
+++ b/tests/integration/model_bridge/test_parallel_residual_identities.py
@@ -1,0 +1,179 @@
+"""Residual-stream decomposition identity for parallel-residual architectures.
+
+Companion to ``test_residual_decomposition_identities.py``, which covers the
+sequential wiring and scopes these architectures out. A parallel block has no
+``hook_resid_mid`` — attention and MLP both read ``resid_pre`` and both write
+into ``resid_post``, so the identity ``ParallelBlockBridge`` documents is:
+
+    resid_post == resid_pre + attn_out + mlp_out
+
+#1639 showed how quietly an alias can end up pointing at a residual-added state
+instead of an additive contribution. That identity held for every
+``ParallelBlockBridge`` adapter but nothing asserted it.
+
+GPTNeoX, StableLM and Falcon pick their block class from the config
+(``use_parallel_residual`` / ``parallel_attn``), so each is exercised in both
+wirings — the sequential cases are the regression guard for a NeoX adapter that
+hardcoded the parallel block and dropped ``hook_resid_mid`` on checkpoints that
+genuinely have one.
+
+Fixtures are seeded tinies built from local HF configs, so no hub access is
+needed and both wirings are reachable from one checkpoint shape. Bridge-vs-HF
+logit parity is deliberately not asserted here — that contract belongs to the
+per-adapter tests (e.g. ``test_cohere_adapter.py``).
+"""
+
+import copy
+from typing import Any
+
+import pytest
+import torch
+from transformers import (
+    AutoModelForCausalLM,
+    CodeGenConfig,
+    CohereConfig,
+    FalconConfig,
+    GPTJConfig,
+    GPTNeoXConfig,
+    PhiConfig,
+    StableLmConfig,
+)
+
+from transformer_lens.model_bridge.sources import build_bridge_from_module
+
+_SHARED = dict(vocab_size=128, pad_token_id=0, bos_token_id=1, eos_token_id=2)
+# Multi-head (no GQA); GPTNeoX splits a fused QKV and needs n_kv == n_heads.
+_MHA = dict(
+    hidden_size=64,
+    intermediate_size=128,
+    num_hidden_layers=2,
+    num_attention_heads=4,
+    max_position_embeddings=128,
+    **_SHARED,
+)
+_GQA = dict(_MHA, num_key_value_heads=2)
+_GPTJ = dict(n_embd=64, n_layer=2, n_head=4, n_positions=128, rotary_dim=16, **_SHARED)
+_FALCON = dict(
+    hidden_size=64,
+    num_hidden_layers=2,
+    num_attention_heads=4,
+    max_position_embeddings=128,
+    new_decoder_architecture=False,
+    **_SHARED,
+)
+
+PARALLEL_CASES = [
+    pytest.param("GPTJForCausalLM", GPTJConfig, _GPTJ, id="gptj"),
+    pytest.param("CodeGenForCausalLM", CodeGenConfig, _GPTJ, id="codegen"),
+    pytest.param("PhiForCausalLM", PhiConfig, _GQA, id="phi"),
+    pytest.param("CohereForCausalLM", CohereConfig, _GQA, id="cohere"),
+    pytest.param(
+        "GPTNeoXForCausalLM", GPTNeoXConfig, dict(_MHA, use_parallel_residual=True), id="neox"
+    ),
+    pytest.param(
+        "StableLmForCausalLM", StableLmConfig, dict(_GQA, use_parallel_residual=True), id="stablelm"
+    ),
+    pytest.param("FalconForCausalLM", FalconConfig, dict(_FALCON, parallel_attn=True), id="falcon"),
+]
+
+SEQUENTIAL_CASES = [
+    pytest.param(
+        "GPTNeoXForCausalLM", GPTNeoXConfig, dict(_MHA, use_parallel_residual=False), id="neox"
+    ),
+    pytest.param(
+        "StableLmForCausalLM",
+        StableLmConfig,
+        dict(_GQA, use_parallel_residual=False),
+        id="stablelm",
+    ),
+    pytest.param(
+        "FalconForCausalLM", FalconConfig, dict(_FALCON, parallel_attn=False), id="falcon"
+    ),
+]
+
+# Built once per (architecture, wiring); the tests only read from them.
+_CACHE: dict[tuple[str, str], tuple[Any, dict[str, torch.Tensor]]] = {}
+
+
+def _run(hf_architecture: str, config_cls: Any, config_kwargs: dict[str, Any]):
+    """Boot a seeded tiny bridge for this architecture and cache one forward."""
+    key = (hf_architecture, repr(sorted(config_kwargs.items())))
+    if key not in _CACHE:
+        config = config_cls(**config_kwargs)
+        config._attn_implementation = "eager"
+        torch.manual_seed(42)
+        hf = AutoModelForCausalLM.from_config(config).eval()
+        bridge = build_bridge_from_module(
+            hf, hf_architecture, hf_config=copy.deepcopy(config), tokenizer=None, device="cpu"
+        ).eval()
+
+        torch.manual_seed(0)
+        tokens = torch.randint(3, config_kwargs["vocab_size"], (1, 10))
+        with torch.no_grad():
+            logits, cache = bridge.run_with_cache(tokens)
+        assert torch.isfinite(logits).all(), f"{hf_architecture} produced non-finite logits"
+        _CACHE[key] = (bridge, cache)
+    return _CACHE[key]
+
+
+@pytest.mark.parametrize("hf_architecture,config_cls,config_kwargs", PARALLEL_CASES)
+def test_parallel_block_omits_resid_mid(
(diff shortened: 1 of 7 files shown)


transformer_lens/model_bridge/supported_architectures/neox.py
```diff
@@ -44,7 +45,13 @@ def __init__(self, cfg: Any) -> None:
         self.cfg.final_rms = False
         self.cfg.gated_mlp = False
         self.cfg.attn_only = False
-        self.cfg.parallel_attn_mlp = True
+
+        # GPTNeoX ships both parallel (Pythia, HF's default) and sequential
+        # variants. Hardcoding parallel drops hook_resid_mid on sequential
+        # checkpoints that genuinely have a post-attention residual.
+        use_parallel_residual = getattr(cfg, "use_parallel_residual", True)
```
jlarson4: `TransformerBridgeConfig` has no `use_parallel_residual` field, so on the `build_bridge_from_module(..., tl_config=...)` path this fallback fires and a config with `parallel_attn_mlp=False` still gets `ParallelBlockBridge` with no `hook_resid_mid`. Can you make the fallback `getattr(cfg, "use_parallel_residual", getattr(cfg, "parallel_attn_mlp", True))` and add a unit case asserting `parallel_attn_mlp=False` leads to a `BlockBridge` being created? A plain `parallel_attn_mlp` read won't work due to its constructor default being `False` rather than HF's `True`, as your PR description notes.

tests/integration/model_bridge/test_parallel_residual_identities.py
```diff
@@ -0,0 +1,179 @@
...
+_CACHE: dict[tuple[str, str], tuple[Any, dict[str, torch.Tensor]]] = {}
+
+
+def _run(hf_architecture: str, config_cls: Any, config_kwargs: dict[str, Any]):
+    """Boot a seeded tiny bridge for this architecture and cache one forward."""
+    key = (hf_architecture, repr(sorted(config_kwargs.items())))
+    if key not in _CACHE:
+        config = config_cls(**config_kwargs)
+        config._attn_implementation = "eager"
+        torch.manual_seed(42)
```
jlarson4: `manual_seed(42)`/`manual_seed(0)` runs only when cache misses and leaks RNG state to later unseeded tests on the same worker. `boot_native` wraps its seeding in `torch.random.fork_rng(devices=[])` for exactly this reason (bridge.py:360), the same two-line wrapper works here.

jlarson4 (commented):
Thanks for tackling this @Chinmayrawat15, just a couple small change requests. Let me know if you have any questions!

Chinmayrawat15:
Thanks @jlarson4 — both were real, fixed in a4e0ba6d and pinned in e06b3949.

**Fallback on the `tl_config=` path.** Applied as suggested. One consequence worth your call: with `use_parallel_residual` absent, the chain reduces to a plain `parallel_attn_mlp` read, and `_make_cfg` in `test_neox_adapter.py` never set that flag — so it defaulted to `False` and `test_bridge_types` started failing. Rather than special-case the adapter I made the fixture state `parallel_attn_mlp=True` explicitly, since a caller-supplied NeoX config is otherwise indistinguishable from one asking for sequential.

That does change behaviour for hand-built configs: previously any `TransformerBridgeConfig` got `ParallelBlockBridge` regardless, now an unset (i.e. `False`) `parallel_attn_mlp` yields `BlockBridge`. HF-booted models are unaffected. Happy to invert it if you'd rather the adapter keep defaulting to parallel when the flag was never touched.

Added both requested cases — `test_sequential_config_builds_plain_block_bridge` and `test_hf_use_parallel_residual_overrides_config_default`.

**RNG leak.** Good catch, wrapped in `torch.random.fork_rng(devices=[])` per `bridge.py:360`. Verified rather than assumed: RNG is preserved across `_run()` on both the cache-miss and cache-hit paths, and is *not* preserved with the wrapper removed.

**One thing that fell out of verifying this.** Re-checking the guards against `upstream/dev`, the fallback makes the `_HF_PASSTHROUGH_ATTRS` entry redundant — reverting it left every test green, because `sources/transformers.py:236` already derives `parallel_attn_mlp` and the fallback picks it up. That only holds while `TransformerBridgeConfig` keeps defaulting `parallel_attn_mlp` to `False`; if that default ever flips, sequential NeoX breaks silently. I kept the passthrough and added `test_hf_use_parallel_residual_reaches_the_bridge_config` so it is no longer resting on two defaults agreeing — but say the word if you'd prefer the smaller diff and I'll drop it.

Guard checked against `upstream/dev` rather than a stash this time: reverting `neox.py` fails 3 tests, reverting the passthrough fails 2.

jlarson4:
Looks good @Chinmayrawat15! Thanks for covering all the edits. Merging now

PR #1660, opened 2026-08-12
fix(bridge): preserve split-component views under load_state_dict(assign=True)

### LINKED ISSUE #1637: [Bug Report] TransformerBridge.load_state_dict(assign=True) desyncs split QKV/gate-up components from the combined weight they view into
### **Base on `dev-4.x`**

**Describe the bug**

`JointQKVAttentionBridge`/`JointGateUpMLPBridge` build their split `q`/`k`/`v` (or `gate`/`in`) sub-bridges once, at `set_original_component()`, via `torch.tensor_split` on the combined weight (`c_attn`, `gate_up_proj`) — `tensor_split` returns views, and wrapping one in `nn.Parameter` preserves the view (confirmed: `q.original_component.weight.untyped_storage().data_ptr() == c_attn.weight.untyped_storage().data_ptr()` is `True` right after construction, for both the QKV and gate/up split paths).

`bridge.load_state_dict(state_dict, assign=True)` breaks that view. PyTorch's `assign=True` *replaces* each target parameter with the tensor from `state_dict` rather than copying into existing storage (`assign=False`, the default, does an in-place `.copy_()` and correctly preserves the view — not a bug there). Once a split sub-bridge's parameter is replaced, it's no longer connected to the combined weight's storage at all — the combined weight keeps whatever it had before the call.

The bridge's own forward pass is unaffected, since `LinearBridge.forward()` reads `self.original_component` (the split module) directly, never `c_attn` itself. The break only surfaces if something reads the wrapped model's own state independent of the bridge — most concretely, `bridge.original_model.state_dict()` (what `save_pretrained()` would export).

**Code example**

```python
import torch

sd = {k: v.clone() for k, v in bridge.state_dict().items()}
mutated = dict(sd)
mutated["blocks.0.attn.q.weight"] = sd["blocks.0.attn.q.weight"] + 100.0

bridge.load_state_dict(mutated, strict=True, assign=True)

# Bridge itself: correct.
print(torch.equal(bridge.blocks[0].attn.q.original_component.weight, mutated["blocks.0.attn.q.weight"]))
# -> True

# The wrapped model's own state, independent of the bridge: silently stale.
raw_sd = bridge.original_model.state_dict()
c_attn_w = raw_sd["transformer.h.0.attn._original_component.c_attn._original_component.weight"]
d_model = bridge.cfg.d_model
print(torch.allclose(c_attn_w[:, :d_model].T, mutated["blocks.0.attn.q.weight"]))  # updated? -> False
print(torch.allclose(c_attn_w[:, :d_model].T, sd["blocks.0.attn.q.weight"]))       # stale?   -> True
```

No error, no warning — `bridge.original_model.state_dict()` (and by extension `bridge.original_model.save_pretrained(...)`) silently contains pre-`assign=True` data for every split component's slice of the combined weight, while the bridge itself reports (and computes with) the correct value. Anyone exporting a bridge-loaded checkpoint after an `assign=True` load gets a corrupted file with no indication anything went wrong.

Confirmed on both split paths that build sub-components this way: `JointQKVAttentionBridge` (`c_attn` → q/k/v, real gpt2 checkpoint) and `JointGateUpMLPBridge` (`gate_up_proj` → gate/up, synthetic module).

**System Info**

- Installed from source, `dev-4.x`
- OS: macOS (arm64), CPU-only
- Reproduces with `boot_transformers("gpt2")`; underlying mechanism confirmed directly on `JointGateUpMLPBridge` too

**Additional context**

Not a regression from #1615/#1587 — `assign=False` (the default `load_state_dict()` path, and everything those PRs tested) is unaffected; verified the combined weight stays correctly in sync in that case. This is specifically about the `assign=True` parameter PyTorch's own `nn.Module.load_state_dict` exposes and this bridge passes through untouched.

Three directions, roughly in order of how invasive they are — flagging rather than picking, since the right one probably depends on how much `assign=True` (memory-efficient loading, e.g. onto `meta` tensors) actually matters for split-component architectures specifically:

1. **Disallow `assign=True` when the target has any split (`JointQKVAttentionBridge`/`JointGateUpMLPBridge`-style) components** — raise a clear error instead of silently desyncing. Simplest, but removes the memory-efficiency benefit of `assign=True` for these architectures specifically.
2. **After `original_model.load_state_dict(..., assign=True)` returns, re-sync the combined weight's slice from each split component's (possibly-replaced) parameter.** Keeps `assign=True` working, but is extra bookkeeping specific to every split-component type, and needs to run even when the split components *weren't* in the state dict being loaded (e.g. a partial load under `strict=False`).
3. **Route `assign=True` state-dict entries for split components through `.data.copy_()` explicitly, bypassing PyTorch's own replace-semantics for just this subset of keys**, so the view relationship survives regardless of the caller's `assign` choice. Keeps behavior uniform without a post-hoc resync pass, but means `assign=True` isn't *actually* honored (still doing a copy under the hood) for these specific keys — worth documenting if so.

Acceptance:
- [ ] `bridge.load_state_dict(sd, assign=True)` does not leave `bridge.original_model.state_dict()` stale for any split-component key (QKV and gate/up)
- [ ] `assign=False` (default) behavior and existing #1587/#1615 tests are unaffected
- [ ] New regression test using `boot_native` or a synthetic split component (no real-model download needed) exercising `assign=True` specifically

### Checklist

### LINKED ISSUE #1725: [Bug Report] TransformerBridge.load_state_dict(assign=True) desyncs tied embed/unembed weights, silently breaking forward pass
**Describe the bug**

GPT-2 (and any architecture with `tie_word_embeddings`) ties its embedding and unembedding weights -- `transformer.wte.weight` and `lm_head.weight` share the same underlying storage even though, in the bridge-wrapped model, they are already separate `nn.Parameter` objects pointing at that shared storage (confirmed via `untyped_storage().data_ptr()` equality at construction, before any load -- structurally the same "different object, shared storage" pattern as #1637's split QKV components, just full-size rather than a partial view).

`load_state_dict(state_dict, assign=True)` breaks this the same way it broke split components in #1637: PyTorch's `assign=True` replaces each target parameter object independently per key, so loading only one of the two tied keys (e.g. just `embed.weight`, a completely ordinary partial/fine-tuned checkpoint) leaves the other holding stale pre-load data with no error or warning.

This is worse than #1637 in one respect: there, the bridge's own forward pass stayed correct and only the raw export (`original_model.state_dict()` / `save_pretrained()`) went stale. Here, `UnembeddingBridge.forward()` reads `original_component.weight` directly -- the now-desynced, stale `lm_head` weight -- so **the bridge's own live forward pass silently uses the wrong weights**, not just its export.

Also distinct from #1637 mechanically: this is not a partial-view case, so the `_is_view_backed` detection added in #1660 does not (and structurally cannot) catch it -- both tensors are full-size, `_is_view_backed` returns `False` for both.

**Code example**

```python
import torch

bridge.eval()
tokens = torch.randint(0, 1000, (1, 6))
with torch.no_grad():
    logits_before = bridge(tokens).clone()

sd = {k: v.clone() for k, v in bridge.state_dict().items()}
new_embed = sd["embed.weight"] + 3.0
bridge.load_state_dict({"embed.weight": new_embed}, strict=False, assign=True)

with torch.no_grad():
    logits_after = bridge(tokens).clone()

print(torch.allclose(logits_before, logits_after))  # -> True: the load was a silent no-op on the forward pass
print(torch.allclose(bridge.unembed.original_component.weight, sd["embed.weight"]))  # -> True: unembed is still the OLD embed value
```

No error, no warning. The embedding matrix was successfully replaced, but every downstream computation through the unembedding silently keeps using the pre-load weights, because the tie desynced.

Separately, even a *complete* round-trip load (both `embed.weight` and `unembed.weight` present with consistent values, so no data loss) still silently breaks the tie at the object level -- doubling memory for that weight matrix and decoupling their gradients for any future training, with no error either.

`assign=False` (the default) is unaffected in both cases -- verified the tie and forward pass stay correct.

**System Info**

- Installed from source, `dev-4.x`, on top of #1660
- OS: macOS (arm64), CPU-only
- Reproduces with `boot_transformers("gpt2")`; mechanism applies to any architecture with `tie_word_embeddings=True`

Same family of fix as #1637 (option 3 + option 1: route `assign=True` for these keys through explicit in-place `.copy_()`, raise a clear error on shape mismatch), but the *detection* needs to change -- `_is_view_backed`'s "smaller than its own storage" check is specific to partial views and doesn't fire for a full-size tensor that merely shares storage with another same-size tensor. Detecting this case needs something like: for each incoming key, check whether any *other* current parameter/buffer shares storage with this key's current target (a storage-identity index built from `original_model.state_dict(keep_vars=True)`), not just a self-contained size check.

Acceptance:
- [ ] Loading only `embed.weight` (or only `unembed.weight`) with `assign=True` does not leave the other stale in either `original_model.state_dict()` or the bridge's own forward pass
- [ ] A complete tied-pair load with `assign=True` does not silently break the storage tie
- [ ] `assign=False` and existing #1587/#1615/#1637 tests are unaffected
- [ ] New regression test on real gpt2 (tied by default) exercising both the partial-load and full-round-trip cases

### Checklist

LightWork666:
Fixes #1637.

`JointQKVAttentionBridge`/`JointGateUpMLPBridge` build their split sub-components once, at `set_original_component()`, via `torch.tensor_split` on the combined weight (`c_attn`, `gate_up_proj`). These are true storage-sharing views, confirmed via `untyped_storage().data_ptr()` equality.

`load_state_dict(state_dict, assign=True)` broke that view: PyTorch's `assign=True` replaces each target parameter object rather than copying into existing storage. The bridge's own forward pass stayed correct (`LinearBridge.forward` reads the split component directly), but `original_model.state_dict()` -- what `save_pretrained()` exports -- silently kept stale data for the combined weight, with no error or warning.

**Fix** (per jlarson4's direction on the issue -- option 3 + option 1 from the three listed there): for `assign=True` calls, detect view-backed targets structurally (`tensor.numel() * tensor.element_size() < tensor.untyped_storage().nbytes()` -- more reliable than `Tensor._base`, which nn.Parameter wrapping doesn't preserve here) and route them through an explicit in-place `.data.copy_()` instead of letting PyTorch replace the parameter object. If a shape mismatch makes that copy unsafe, raise a clear `RuntimeError` rather than failing opaquely or silently corrupting data.

`assign=False` (the default) was already correct and is untouched.

**Testing**:
- New `_is_view_backed` unit test (synthetic `torch.tensor_split`, no model download)
- New `boot_native` test confirming ordinary (non-split) `assign=True` round-trips unchanged
- New real-gpt2 test mirroring the issue's exact repro: confirms `original_model.state_dict()` is no longer stale after `assign=True`
- New real-gpt2 test for the shape-mismatch error path
- Full existing `test_state_dict_round_trip.py` suite (#1587), `generalized_components` split-bridge tests, and `gpt2`/`gpt_bigcode`/`falcon`/`openai_gpt` adapter tests all still pass (198 tests, no regressions)
- `uv run mypy .` clean

Acceptance criteria from the issue are all met.

### DIFF
--- a/tests/unit/model_bridge/test_state_dict_round_trip.py
+++ b/tests/unit/model_bridge/test_state_dict_round_trip.py
@@ -4,6 +4,11 @@ state_dict() emits TL-renamed keys (e.g. "blocks.0.attn.q.weight"), but
 load_state_dict() only matched raw native parameter names, so a
 state_dict() -> load_state_dict() round trip silently loaded nothing and
 strict=True was silently downgraded to strict=False.
+
+Also covers "copy-split staleness" (#1637): load_state_dict(..., assign=True)
+used to replace view-backed split-component parameters (e.g. gpt2's q/k/v,
+which are torch.tensor_split views into the combined c_attn weight) wholesale,
+desyncing them from the combined weight they share storage with.
 """
 from __future__ import annotations
 
@@ -12,6 +17,7 @@ import torch
 
 from transformer_lens.config import TransformerBridgeConfig
 from transformer_lens.model_bridge import TransformerBridge
+from transformer_lens.model_bridge.transformer_bridge import _is_view_backed
 
 
 def _native_cfg(**overrides) -> TransformerBridgeConfig:
@@ -143,6 +149,38 @@ def test_native_clean_key_dict_with_partial_aliases_does_not_raise_strict():
             ), f"{actual_key} (alias of {clean_key}) did not round-trip"
 
 
+def test_is_view_backed_detects_tensor_split_views():
+    """Core detection helper for #1637: a torch.tensor_split view shares storage
+    with a larger source tensor, even after nn.Parameter wrapping (which does not
+    reliably preserve Tensor._base view tracking, so this must not rely on it)."""
+    combined = torch.nn.Parameter(torch.randn(12, 4))
+    a, b, c = torch.tensor_split(combined, 3, dim=0)
+    split_param = torch.nn.Parameter(a)
+    assert split_param.untyped_storage().data_ptr() == combined.untyped_storage().data_ptr()
+    assert _is_view_backed(split_param)
+
+    assert not _is_view_backed(combined)
+    assert not _is_view_backed(torch.nn.Parameter(torch.randn(4, 4)))
+
+
+def test_native_assign_true_round_trip_no_split_components():
+    """boot_native's components are independent parameters (no split/view
+    components), so assign=True should take the ordinary passthrough path and
+    round-trip exactly like assign=False does."""
+    bridge = TransformerBridge.boot_native(_native_cfg())
+
+    sd = {k: v.clone() for k, v in bridge.state_dict().items()}
+    with torch.no_grad():
+        for p in bridge.parameters():
+            p.zero_()
+
+    bridge.load_state_dict(sd, strict=True, assign=True)
+
+    reloaded = bridge.state_dict()
+    for key, value in sd.items():
+        assert torch.equal(reloaded[key], value), f"{key} did not round-trip under assign=True"
+
+
 @pytest.mark.slow
 def test_boot_transformers_round_trip_matches_forward_pass():
     """GPT-2's Conv1D-combined attention makes the bridge's q/k/v components
@@ -194,3 +232,42 @@ def test_boot_transformers_clean_key_dict_does_not_raise_strict():
     result = bridge.load_state_dict(clean_sd, strict=True)
     assert result.missing_keys == []
     assert result.unexpected_keys == []
+
+
+@pytest.mark.slow
+def test_boot_transformers_assign_true_does_not_leave_combined_weight_stale():
+    """#1637 repro: assign=True on a split QKV component used to replace the
+    parameter object instead of copying into it, breaking the view relationship
+    with c_attn -- the bridge itself read the new value, but
+    original_model.state_dict() (what save_pretrained() exports) silently kept
+    the pre-load data for the combined weight."""
+    bridge = TransformerBridge.boot_transformers("gpt2", device="cpu")
+
+    sd = {k: v.clone() for k, v in bridge.state_dict().items()}
+    mutated = dict(sd)
+    mutated["blocks.0.attn.q.weight"] = sd["blocks.0.attn.q.weight"] + 100.0
+
+    bridge.load_state_dict(mutated, strict=True, assign=True)
+
+    assert torch.equal(
+        bridge.blocks[0].attn.q.original_component.weight, mutated["blocks.0.attn.q.weight"]
+    )
+
+    raw_sd = bridge.original_model.state_dict()
+    c_attn_w = raw_sd["transformer.h.0.attn._original_component.c_attn._original_component.weight"]
+    d_model = bridge.cfg.d_model
+    assert torch.allclose(c_attn_w[:, :d_model].T, mutated["blocks.0.attn.q.weight"])
+
+
+@pytest.mark.slow
+def test_boot_transformers_assign_true_shape_mismatch_raises_clear_error():
+    """A view-backed split component can only be loaded under assign=True via an
+    in-place copy, which requires a matching shape -- fail loudly instead of a
+    confusing error surfacing from deep inside copy_, or silently corrupting data."""
+    bridge = TransformerBridge.boot_transformers("gpt2", device="cpu")
+
+    sd = dict(bridge.state_dict())
+    sd["blocks.0.attn.q.weight"] = sd["blocks.0.attn.q.weight"][:-1]
+
+    with pytest.raises(RuntimeError, match="view sharing storage"):
+        bridge.load_state_dict(sd, strict=True, assign=True)
--- a/tests/unit/model_bridge/test_state_dict_round_trip.py
+++ b/tests/unit/model_bridge/test_state_dict_round_trip.py
@@ -6,18 +6,21 @@ state_dict() -> load_state_dict() round trip silently loaded nothing and
 strict=True was silently downgraded to strict=False.
 
 Also covers "copy-split staleness" (#1637): load_state_dict(..., assign=True)
-used to replace view-backed split-component parameters (e.g. gpt2's q/k/v,
-which are torch.tensor_split views into the combined c_attn weight) wholesale,
-desyncing them from the combined weight they share storage with.
+used to replace parameters that share underlying storage with something else
+(a split QKV/gate-up component's view into its combined weight, the combined
+weight itself, or a tied pair like embed/unembed, #1725) wholesale, desyncing
+them from whatever they share storage with.
 """
 from __future__ import annotations
 
 import pytest
 import torch
+from transformers import GPT2Config, GPT2LMHeadModel, Phi3Config, Phi3ForCausalLM
 
 from transformer_lens.config import TransformerBridgeConfig
 from transformer_lens.model_bridge import TransformerBridge
-from transformer_lens.model_bridge.transformer_bridge import _is_view_backed
+from transformer_lens.model_bridge.sources import build_bridge_from_module
+from transformer_lens.model_bridge.transformer_bridge import _storage_group_keys
 
 
 def _native_cfg(**overrides) -> TransformerBridgeConfig:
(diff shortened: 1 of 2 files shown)


transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -3964,7 +3976,42 @@ def load_state_dict(self, state_dict, strict=True, assign=False):
...
+        # keep the stale pre-load data. Route those specific keys through an
+        # explicit in-place .data.copy_() instead, regardless of the caller's
+        # assign=True, so the view relationship survives.
+        passthrough_items = {}
+        for key, value in mapped_state_dict.items():
+            target = current_state_dict.get(key)
+            if target is None or not _is_view_backed(target):
+                passthrough_items[key] = value
+                continue
+            if tuple(target.shape) != tuple(value.shape):
```
jlarson4: Only shape is guarded, an fp16 (or MPS-resident) `q.weight` under `assign=True` silently copy-converts into the fp32 CPU view while passthrough keys adopt the incoming dtype/device, yielding a silently mixed-dtype or split-device model from one call. Can you extend this guard to `target.dtype != value.dtype or target.device != value.device`, with the message naming which property mismatched?

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -3964,7 +3976,42 @@ def load_state_dict(self, state_dict, strict=True, assign=False):
...
+                raise RuntimeError(
+                    f"Cannot load {key!r} with assign=True: the current parameter is a "
+                    "view sharing storage with another parameter (e.g. a split QKV/"
+                    "gate-up component's view into a combined weight), so it can only "
+                    "be loaded via an in-place copy, which requires a matching shape. "
+                    f"Got {tuple(value.shape)}, expected {tuple(target.shape)}."
+                )
+            with torch.no_grad():
+                target.data.copy_(value)
```
jlarson4, line 4484: The loop copies as it validates, so the first mismatch raises with every earlier view-backed key already written and everything else unapplied. PyTorch's loader applies & then aggregates, but in our use case a first pass over `mapped_state_dict` collecting all mismatches before any `copy_` is trivial since the dict is fully materialized.

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -3964,7 +3976,42 @@ def load_state_dict(self, state_dict, strict=True, assign=False):
...
+        # (LinearBridge.forward reads the split component directly), but the
+        # combined weight -- and anything reading original_model's state
+        # independent of the bridge, e.g. save_pretrained() -- would silently
+        # keep the stale pre-load data. Route those specific keys through an
+        # explicit in-place .data.copy_() instead, regardless of the caller's
+        # assign=True, so the view relationship survives.
+        passthrough_items = {}
+        for key, value in mapped_state_dict.items():
+            target = current_state_dict.get(key)
+            if target is None or not _is_view_backed(target):
```
jlarson4: A raw-native-key load of `c_attn`'s actual key takes this passthrough branch, torch replaces the parameter object, and the split views are orphaned: `q<->c_attn` storage sharing breaks, bridge forward reads stale values while `save_pretrained()` exports the new ones. A storage-group check (route non-view keys whose storage is shared by a view-backed sibling through `copy_` too) would close that gap here.

tests/unit/model_bridge/test_state_dict_round_trip.py
```diff
@@ -194,3 +232,42 @@ def test_boot_transformers_clean_key_dict_does_not_raise_strict():
     result = bridge.load_state_dict(clean_sd, strict=True)
     assert result.missing_keys == []
     assert result.unexpected_keys == []
+
+
+@pytest.mark.slow
+def test_boot_transformers_assign_true_does_not_leave_combined_weight_stale():
```
jlarson4, line 465: Can you also add a fast test on a tiny `from_config` model via `build_bridge_from_module`? A tiny Phi-3/GLM config would cover the `JointGateUpMLPBridge` path and a tiny `GPT2Config` the QKV path, giving us coverage in CI (slow tests don't run in CI because the runners aren't large enough).

tests/unit/model_bridge/test_state_dict_round_trip.py
```diff
@@ -143,6 +149,38 @@ def test_native_clean_key_dict_with_partial_aliases_does_not_raise_strict():
...
+    a, b, c = torch.tensor_split(combined, 3, dim=0)
+    split_param = torch.nn.Parameter(a)
+    assert split_param.untyped_storage().data_ptr() == combined.untyped_storage().data_ptr()
+    assert _is_view_backed(split_param)
+
+    assert not _is_view_backed(combined)
+    assert not _is_view_backed(torch.nn.Parameter(torch.randn(4, 4)))
+
+
+def test_native_assign_true_round_trip_no_split_components():
```
jlarson4, line 234: This test catches dropped-key/no-op loads, but its equality asserts hold whether the passthrough branch assigns or copies, so nothing anywhere pins that non-view keys keep true assign semantics which is the point of not running everything through `copy_`. One extra assert like `bridge.state_dict()[key].data_ptr() == sd[key].data_ptr()` would cover that.

jlarson4 (commented):
Great solution to this @LightWork666! Just a couple comments, let me know if you have any questions

koriyoshi2041 (commented):
One additional `assign=True` case appears to remain unsafe on head `b4cea229`: a view-backed target on `meta` accepts `target.data.copy_(cpu_value)` without raising, but remains a meta tensor. The proposed path then removes that key from `passthrough_items`, so the load can report success while never materializing the split parameter.

Minimal PyTorch repro:

```
combined = torch.nn.Parameter(torch.empty(12, 4, device="meta"))
view = torch.nn.Parameter(torch.tensor_split(combined, 3, dim=0)[0])
with torch.no_grad():
    view.data.copy_(torch.ones(4, 4))
assert view.is_meta  # still True; copy_ raised no error
```

This matters because meta initialization is one of the main reasons callers select `assign=True`. Alongside the existing dtype/device guard request, I think the view-backed branch should explicitly reject a meta target unless it has a path that reconstructs the shared storage group on the incoming device. A fast synthetic test should assert the call cannot silently succeed while the split remains meta.

LightWork666:
Apologies for the long delay in getting back to this, and thanks for the thorough reviews.

Rewrote the detection to be storage-group-based rather than per-tensor: any key whose current tensor shares underlying storage with another key in the model (not just partial views) now routes through the copy_ path under assign=True. This addresses your point about the raw c_attn key write orphaning its views -- that's now caught -- and as a side effect it also fixes tied embed/unembed desyncing (filed separately as #1725, since it's a distinct enough failure mode to warrant its own report, but the same generalized fix here resolves it too).

Also addressed:
- dtype/device guard added alongside the existing shape guard
- validation now happens for every key before any copy_ is applied, instead of copy-as-it-validates
- meta tensors are excluded from the storage-group check -- turns out every meta tensor reports `data_ptr() == 0`, which would have falsely grouped every unrelated offloaded parameter together. A meta target now takes the ordinary assign=True materialization path instead, which is the correct way to load real weights onto it anyway.
- new fast (no-download) tests via `build_bridge_from_module` with tiny GPT2/Phi-3 configs covering the QKV and gate/up paths in CI, plus tests for each of the above

All existing tests plus the new ones pass, mypy clean.

jlarson4:
No worries on the delay @LightWork666, thank you for wrapping it up! Everything looks good here, and this covers both of your issues quite nicely. Merging as is, thank you!

LightWork666:
Thank you, appreciate you taking the time to look it over again and merge it. And thanks again to @koriyoshi2041 for catching the meta-tensor case, that one would've been a nasty silent failure to leave in.

Issue #1655, opened 2026-08-12
[Bug Report] TransformerBridge is not composable as an nn.Module: parent traversal misses live parameters


emerardd:
**Describe the bug**

`TransformerBridge` behaves like a standard `torch.nn.Module` when used directly, but it is not composable as a child of another `nn.Module`.

`TransformerBridge.__init__()` stores the live source model with:

```python
self.__dict__["original_model"] = model
```

This intentionally bypasses `nn.Module.__setattr__()` and keeps `original_model` outside the registered `_modules` tree. Bridge then restores several root-level operations through explicit overrides, including `parameters()`, `named_parameters()`, `to()`, `train()`, `state_dict()`, and `load_state_dict()`.

Those overrides make direct calls on the Bridge work, but a parent module does not use them for ordinary recursive traversal:

- parameter discovery walks registered modules and parameters;
- `parent.to(...)`, `.half()`, and `.bfloat16()` recurse through `_apply()`;
- `parent.requires_grad_(...)`, optimizers, and `zero_grad()` use the parent's recursively discovered parameters;
- parent checkpointing passes a shared `destination` and `prefix` through recursive `state_dict()` calls and ignores a child's replacement return object.

The Bridge component tree re-registers only the source modules represented by the adapter mapping. Unmapped parts of the hidden source model are still used by forward, but are invisible to the parent module. As a result, parent optimizers and lifecycle operations can silently omit live parameters, parent dtype conversion can create a mixed-dtype model that crashes, and nested checkpoint round-trips fail.

**Code example**

This reproduction constructs tiny random models from config only and performs no downloads:

```python
import torch
from transformers import BertConfig, BertForMaskedLM

from transformer_lens.model_bridge.sources import build_bridge_from_module


class Parent(torch.nn.Module):
    def __init__(self, bridge):
        super().__init__()
        self.bridge = bridge

    def forward(self, tokens):
        return self.bridge(tokens)


def make_bridge():
    config = BertConfig(
        vocab_size=32,
        hidden_size=16,
        intermediate_size=32,
        num_hidden_layers=2,
        num_attention_heads=4,
        max_position_embeddings=16,
    )
    return build_bridge_from_module(
        BertForMaskedLM(config).eval(),
        "BertForMaskedLM",
        hf_config=config,
        dtype=torch.float32,
        device="cpu",
        model_name="tiny-parent-module-repro",
    )


bridge = make_bridge()
parent = Parent(bridge)

[...]

PR #1661, opened 2026-08-13
Fix recursive TransformerBridge state dict composition

emerardd:
# Description

This is the state-dict contract half of #1655, following the scope agreed in the maintainer discussion.

When a `TransformerBridge` is nested inside another `nn.Module`, PyTorch supplies a shared `destination` and a child `prefix` while recursively collecting state. The Bridge override instead delegated to the raw Hugging Face model and returned a different mapping, so parent checkpoints did not follow normal `nn.Module` composition semantics. Switching to standard recursive traversal also exposed two existing asymmetries: joint QKV/gate-up save hooks removed registered child keys without load-side handling, and the first forward registered `_ln1_module`, changing the module tree after a checkpoint had been saved.

This PR:

- preserves the existing no-argument `bridge.state_dict()` TransformerLens key contract;
- uses standard `nn.Module.state_dict()` semantics whenever `destination` or `prefix` is supplied;
- adds load-side handling for the intentionally filtered joint QKV and gate-up children;
- wires the attention layer-norm execution reference during construction as a non-owning reference, keeping the registration tree stable across forward;
- adds download-free regression coverage for shared destinations and prefixes, nested strict round-trips, pre/post-forward registration stability, real GPT-2 joint QKV composition, and filtered-child save/load symmetry with value restoration.

Part of #1655. The BERT adapter mappings, container-buffer ownership, and broader architecture traversal sweep remain isolated to the follow-up PR requested by the maintainer.

## Type of change

# Validation

- Full local suite: `5266 passed, 55 skipped, 54 deselected, 10 xfailed` in 14m21s
- Focused state-dict/native surface: `51 passed`
- `mypy .`: success across 431 source files

# Checklist

### DIFF
--- a/tests/unit/model_bridge/test_state_dict_composition.py
+++ b/tests/unit/model_bridge/test_state_dict_composition.py
@@ -108,12 +108,19 @@ def test_nested_joint_qkv_bridge_strict_round_trip() -> None:
     )
     parent = _parent_with_bridge(bridge)
 
-    checkpoint = parent.state_dict()
+    checkpoint = {key: value.clone() for key, value in parent.state_dict().items()}
     assert not any(".qkv." in key for key in checkpoint)
+    with torch.no_grad():
+        for parameter in parent.parameters():
+            parameter.zero_()
+
     result = parent.load_state_dict(checkpoint, strict=True)
 
     assert result.missing_keys == []
     assert result.unexpected_keys == []
+    reloaded = parent.state_dict()
+    for key, value in checkpoint.items():
+        assert torch.equal(reloaded[key], value), f"{key} did not round-trip"
 
 
 def _filtered_joint_component(kind: str) -> torch.nn.Module:
@@ -144,6 +151,7 @@ def _filtered_joint_component(kind: str) -> torch.nn.Module:
 @pytest.mark.parametrize("filtered_child_name", ["qkv", "gate_up"])
 def test_filtered_joint_component_strict_round_trip(filtered_child_name: str) -> None:
     component = _filtered_joint_component(filtered_child_name)
+    filtered_child = component.get_submodule(filtered_child_name)
     checkpoint = {key: value.clone() for key, value in component.state_dict().items()}
 
     assert checkpoint
@@ -159,3 +167,5 @@ def test_filtered_joint_component_strict_round_trip(filtered_child_name: str) ->
     reloaded = component.state_dict()
     for key, value in checkpoint.items():
         assert torch.equal(reloaded[key], value), f"{key} did not round-trip"
+    for parameter in filtered_child.parameters():
+        assert torch.count_nonzero(parameter) == 0
--- a/tests/unit/model_bridge/test_state_dict_composition.py
+++ b/tests/unit/model_bridge/test_state_dict_composition.py
@@ -1,0 +1,161 @@
+"""Regression tests for recursive TransformerBridge checkpoint composition (#1655)."""
+
+from __future__ import annotations
+
+from collections import OrderedDict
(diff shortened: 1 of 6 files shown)


transformer_lens/model_bridge/generalized_components/joint_qkv_attention.py
```diff
@@ -143,6 +146,25 @@ def _filter_qkv_state_dict(
...
+        unexpected_keys: list[str],
+        error_msgs: list[str],
+    ) -> None:
+        """Supply the intentionally filtered child during recursive loads."""
+        del local_metadata, strict, missing_keys, unexpected_keys, error_msgs
+        qkv = module._modules.get("qkv")
+        if qkv is None:
+            return
+        for key, value in qkv.state_dict(prefix=f"{prefix}qkv.").items():
+            state_dict.setdefault(key, value)
```
jlarson4, line 170: The docstring says "supply the intentionally filtered child", but what it supplies is the *current* combined weight, not the checkpoint's. In production that's harmless because the same weight arrives via the un-filtered `attn._original_component.<hf_name>.weight` path. Where that is absent, though, it injects stale data: in `test_filtered_joint_component_strict_round_trip`, after zeroing and reloading, `qkv.weight` ends at `0.0` while `q`/`k`/`v` restore correctly, and the assertions can't see it because the checkpoint has no `qkv.*` keys to compare. Can you note in the docstring that this exists to satisfy strict key-matching (values come from the HF-path entry), and add a `qkv`-value assertion to that test so the stale-injection case is tested?

tests/unit/model_bridge/test_state_dict_composition.py
```diff
@@ -0,0 +1,161 @@
...
+        eos_token_id=2,
+    )
+    bridge = build_bridge_from_module(
+        GPT2LMHeadModel(cfg),
+        architecture="GPT2LMHeadModel",
+        hf_config=cfg,
+    )
+    parent = _parent_with_bridge(bridge)
+
+    checkpoint = parent.state_dict()
```
jlarson4: `parent.state_dict()` without `.clone()` means the checkpoint tensors are the live parameters, so the load is a self-copy and the assertions cover only missing/unexpected keys. It would pass even if nothing were restored. Could you clone, zero, and add the `torch.equal` loop the synthetic sibling uses?

jlarson4 (commented):
Thanks for putting this together @emerardd, looks great! Just a small docstring note and a test logistic comment, everything in the actual solution looks great.

transformer_lens/model_bridge/generalized_components/joint_qkv_attention.py
emerardd: Updated both filtered-child docstrings to clarify that the pre-hook inserts the current combined values only for strict key matching, while production checkpoint values arrive through the unfiltered Hugging Face `_original_component` path. The synthetic test now explicitly asserts that the filtered qkv/gate-up child remains zero when that HF path is absent. The real GPT-2 test separately verifies production restoration by cloning, zeroing, strictly loading, and comparing every recursive state-dict value.

tests/unit/model_bridge/test_state_dict_composition.py
emerardd: Updated as suggested. The GPT-2 parent checkpoint now clones every tensor, zeros all parent parameters, strictly reloads, and compares every reloaded state-dict value against the snapshot. This makes the test verify actual restoration rather than a live-tensor self-copy.

jlarson4:
Thanks for updating this @emerardd! This is approved, I will merge once it passes CI

PR #1664, opened 2026-08-13
Expose BERT token-type embeddings through TransformerBridge

LarryHu0217:
# Description

Expose BERT token-type (segment) embeddings as the bridge-native `token_type_embed.hook_out` hook.

The BERT adapter now maps `bert.embeddings.token_type_embeddings` through an `EmbeddingBridge`. The regression test boots `bert-base-cased`, captures the new hook with `run_with_cache`, and compares it directly with Hugging Face's token-type embedding output.

Fixes #1662

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Validation

- `uv run pytest tests/unit/model_bridge/supported_architectures/test_bert_adapter.py -q --tb=short -o addopts=`: 22 passed
- `make unit-test`: 5,260 passed, 56 skipped, 54 deselected, 10 xfailed
- `make check-format`: passed
- `uv run mypy .`: no issues in 431 source files
- `make docstring-test`: 17 passed, 24 skipped
- `make acceptance-test`: 199 passed, 72 skipped, 11 deselected
- `make integration-test`: 1,016 passed; the two `google/gemma-2-2b-it` cases could not run locally because this machine is not authenticated for that gated Hugging Face model (401)

# Checklist

- [x] The change is self-explanatory and does not need additional inline comments
- [x] No documentation change is needed for this bridge-native hook mapping
- [x] My changes generate no new warnings
- [x] I have added tests that prove the feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

PR #1671, opened 2026-08-14
Fix TransformerBridge adapter traversal coverage

### LINKED ISSUE #1655: [Bug Report] TransformerBridge is not composable as an nn.Module: parent traversal misses live parameters

emerardd:
# Description

This is the adapter-coverage half of #1655, following the two-PR, compatibility-first split agreed in the maintainer discussion. PR #1661 addressed recursive state-dict composition; this PR completes parent parameter and buffer traversal without changing the existing direct `bridge.state_dict()` TransformerLens-key contract.

The original BERT reproduction exposed adapter gaps that left live forward parameters outside the registered Bridge tree. Since `token_type_embed` has now landed on `dev-4.x` via #1664, this PR adds the remaining BERT embedding LayerNorm and MLM transform dense mappings, and covers pooler plus NSP-only and combined MLM/NSP variants using HookedEncoder-aligned names. A broader download-free structural sweep also found and maps the AST classifier LayerNorm and the bare ViT pooler.

Some Hugging Face models also own state directly on container modules rather than on mapped leaf modules; BART's root `final_logits_bias` buffer is one example. The Bridge now registers non-owning state-owner views only for parameter and buffer identities not already owned by the adapter tree. These views synchronize replacements made by `_apply()` and `load_state_dict(assign=True)` back to the original containers, so parent `.to()`, freezing, optimizers, and recursive loading continue to operate on the objects used by forward without registering the complete source model twice.

Regression coverage constructs 15 tiny random models entirely from config, spanning joint QKV, split QKV/RoPE, GQA, BERT MLM/NSP task heads, encoder-decoder, MoE, vision, and audio. It asserts that the source model, direct Bridge traversal, and parent traversal expose identical parameter and buffer identity sets, and separately covers container-buffer dtype conversion and assign-loading.

Fixes #1655

## Type of change

# Validation

- Full non-slow unit surface: all collected tests passed across the original run and an exact retry of 51 tests initially blocked by a local `basetemp` setup error (`5219 + 51` passed; `55 skipped`, `54 deselected`, `10 xfailed`)
- Focused adapter and parent-traversal surface: `82 passed`
- `mypy .`: success across 431 source files

# Checklist:


transformer_lens/model_bridge/component_setup.py
```diff
@@ -20,6 +20,45 @@
...
+    def __init__(self, original_container: nn.Module) -> None:
+        super().__init__()
+        self.__dict__["_original_container"] = original_container
+
+    def _sync_original_container(self) -> None:
+        original_container = self.__dict__["_original_container"]
+        original_container._parameters.update(self._parameters)
+        original_container._buffers.update(self._buffers)
+
+    def _apply(self, fn: Any, recurse: bool = True) -> "_ContainerStateOwner":
```
jlarson4, line 42: `_sync_original_container` pushes the owner to the original container using tensors captured at registration, but `TransformerBridge.load_state_dict(..., assign=True)` replaces the tensor on the container and bypasses the bridge tree, leaving `self` stale. The next `_apply` writes the pre-load tensor back over the loaded one, so loaded weights have zero effect on forward. This hits BART's `final_logits_bias` and Hubert's `masked_spec_embed`.

Can you make the original container the single source of truth? Add a `_refresh_from_original_container()` that re-reads the container's `_parameters`/`_buffers` for the names it owns and call it at the top of `_apply`, and resync the owners after the assign load in `TransformerBridge.load_state_dict` (transformer_bridge.py:3971) so traversal isn't stale in between? Please add a guard test covering `bridge.load_state_dict(..., assign=True)` followed by `.cpu()`.

transformer_lens/model_bridge/supported_architectures/bert.py
```diff
@@ -145,7 +151,15 @@ def prepare_model(self, hf_model: Any) -> None:
         BertForNextSentencePrediction has cls.seq_relationship (NSP head)
         and no MLM-specific LayerNorm.
         """
-        if hasattr(hf_model, "cls") and hasattr(hf_model.cls, "seq_relationship"):
-            # NSP model — swap head components
+        if getattr(getattr(hf_model, "bert", None), "pooler", None) is not None:
+            self.components["pooler"] = LinearBridge(name="bert.pooler.dense")
```
jlarson4, line 155: `bridge.state_dict()` now returns `bert.bert.pooler.dense.dense.weight` where we previously returned `bert.pooler.dense.weight`. The anti-doubling guard in `_normalize_bridge_key_to_hf` only skips when the TL name is the last segment of its HF path, and `pooler` sits mid-path in `bert.pooler.dense`, so the segment gets re-expanded. This is the same issue you worked around on the ViT side with `vision_pooler`.

Can you widen the guard at transformer_bridge.py:3795 to skip whenever the TL name is any segment of its own HF name (`tl_name in component.name.split(".")`)? That yields a clean `pooler.weight`, keeps the HookedEncoder-aligned name, and would let ViT drop the `vision_pooler` workaround. A test pinning the emitted key names for BERT/ViT/AST would stop this recurring.

jlarson4 (commented):
Thanks for putting this together! The traversal invariant you added is great. A couple quick comments:

transformer_lens/model_bridge/component_setup.py
emerardd: Addressed in 96c73a0b. _ContainerStateOwner now refreshes its owned parameters and buffers from the original container before _apply, and TransformerBridge.load_state_dict(assign=True) refreshes all owners immediately after the HF load. I also added direct assign-then-cpu regression coverage for both BART final_logits_bias and HuBERT masked_spec_embed.

transformer_lens/model_bridge/supported_architectures/bert.py
emerardd: Addressed in 96c73a0b. _normalize_bridge_key_to_hf now skips remapping whenever the TL name is already any segment of the HF component path. ViT now uses pooler directly instead of the vision_pooler workaround, and regression tests pin the emitted BERT, ViT, and AST key names. Local validation: the expanded affected suite passed 111 tests; the exact reviewer regression cases passed 5 tests with 17 deselected; mypy passed all 431 source files; pycln, isort, Black, and git diff --check were clean.

jlarson4:
Thanks for the updates @emerardd! No additional comments, the code is ready to go. Merging now.

Issue #1646, opened 2026-08-11
[Bug Report] Gemma 4 MLPs omit `hook_pre`, `hook_pre_linear`, and `hook_post`


hijohnnylin:
## Summary

`Gemma3ArchitectureAdapter` maps each block's MLP with `self._gated_mlp()`, which creates a `GatedMLPBridge` and supplies the standard gated-MLP aliases:

```text
hook_pre        -> gate.hook_out
hook_pre_linear -> in.hook_out
hook_post       -> out.hook_in
```

`Gemma4ArchitectureAdapter` maps the same `gate_proj` / `up_proj` / `down_proj` structure to a bare `GeneralizedComponent`. As a result, Gemma 4 exposes only the MLP boundary hooks (`hook_in` and `hook_out`); the three neuron-basis hooks are absent from `hook_dict`.

This is a mapping gap rather than an architectural limitation: Hugging Face's `Gemma4TextMLP` is an ordinary gated MLP, and the TransformerLens adapter already identifies all three projection submodules.

## Versions

- `transformer-lens==3.7.0`
- `transformers==5.15.0`
- Python 3.12

I also checked the `v3.7.1` source. The Gemma 4 MLP mapping is unchanged.

## Reproduction

This is a config-only reproduction; it does not load model weights.

`transformers==5.15.0` also exposes a separate Gemma 4 heterogeneous-config crash in TransformerLens. The two-line opt-in below is used only to get past that independent crash and inspect the adapter mapping. It is not a proposed runtime workaround.

```bash
pip install "transformer-lens==3.7.1" "transformers==5.15.0"
python gemma4_mlp_hooks.py
```

```python
# gemma4_mlp_hooks.py
from transformers import AutoConfig
from transformer_lens.config import TransformerBridgeConfig
from transformer_lens.factories.architecture_adapter_factory import (
    ArchitectureAdapterFactory,
)
from transformer_lens.model_bridge.sources.transformers import (
    determine_architecture_from_hf_config,
    map_default_transformer_lens_config,
)

for model_id in ("google/gemma-3-270m", "google/gemma-4-E2B"):
    hf_config = AutoConfig.from_pretrained(model_id)

    # Isolate the adapter-mapping question from the separate heterogeneous-config crash.
    for config_obj in (hf_config, hf_config.get_text_config()):
        config_obj.allow_global_per_layer_attribute_access = True

    mapped = map_default_transformer_lens_config(hf_config)
[...]

PR #1672, opened 2026-08-14
Fix Gemma 4 gated MLP hook aliases

LarryHu0217:
# Description
Gemma 4 dense MLPs use the same gate/up/down projection structure as other gated MLP architectures, but the adapter mapped them as a bare GeneralizedComponent. This prevented the standard hook_pre, hook_pre_linear, and hook_post aliases from being registered.\n\nThis change uses the shared gated-MLP bridge for Gemma 4 and adds unit and integration coverage that verifies the aliases map to the expected projections and are available through run_with_cache across the PLE/KV-sharing, dense K=V, and MoE fixtures.\n\nFixes #1646\n\n## Type of change\n\n- [x] Bug fix (non-breaking change which fixes an issue)\n\n## Testing\n\n- Unit adapter test: 10 passed\n- Gemma 4 integration test: 15 passed\n- Full non-slow unit suite: 5266 passed, 56 skipped, 10 xfailed\n- Format check passed\n- mypy passed for 431 source files\n\n# Checklist\n\n- [x] The implementation is self-explanatory and does not require additional comments\n- [x] No documentation change is required for restoring the existing hook interface\n- [x] My changes generate no new warnings\n- [x] I have added tests that prove the fix is effective\n- [x] New and existing unit tests pass locally with my changes\n- [x] I have not rewritten tests relating to key interfaces

### DIFF
--- a/tests/integration/model_bridge/test_gemma4_bridge.py
+++ b/tests/integration/model_bridge/test_gemma4_bridge.py
@@ -98,3 +98,16 @@ def test_run_with_cache_text_only(bridge):
         logits, cache = bridge.run_with_cache(IDS)
     assert torch.isfinite(logits).all()
     assert len(cache) > 0
+
+
+def test_gated_mlp_hooks_are_available(bridge):
+    names = {
+        "blocks.0.mlp.hook_pre",
+        "blocks.0.mlp.hook_pre_linear",
+        "blocks.0.mlp.hook_post",
+    }
+    assert names <= bridge.hook_dict.keys()
+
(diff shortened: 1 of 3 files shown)


jlarson4:
Hi @LarryHu0217! Sorry I didn't see your message on the issue before you posted this. The issue is for `3.x`, and has already been resolved on `dev`, the issue will be closed at the next release. The fix on `3.x` will be ported to `4.x` directly to avoid conflicts from a parallel implementation. Sorry for the confusion

PR #1676, opened 2026-08-15
feat(jacobian_lens): add occupancy and fraction-of-variance

janmenjayap:
# Description

Adds **occupancy** and **fraction-of-variance** to the Jacobian lens: two J-space profiling statistics that sit on top of the sparse decomposition (#1596), following Gurnee et al. (2026), "Verbalizable Representations Form a Global Workspace in Language Models" (Transformer Circuits Thread). They quantify two of the paper's qualitative claims: that "only a small number of J-lens vectors are strongly active at a time" (occupancy) and that the J-space is "never more than about 10%" of activation variance, "a median of 6 to 7%" for concept vectors (fraction of variance).

This is the Tier-A follow-up the maintainer explicitly deferred from #1596; it builds directly on that PR's greedy selection and cached J-lens dictionary. **Part of #1539 (Tier-A follow-up to #1596).** No new dependencies.

## New public surface (TransformerBridge only, matching the rest of the Jacobian lens)

Model-free core in `transformer_lens/tools/analysis/jacobian_lens_decomposition.py`:

- `estimate_occupancy(x, dictionary, *, max_atoms=25, num_control_dictionaries=32, seed=0)` returns a `JSpaceOccupancy` (`occupancy`, `marginal_captured_variance`, `control_captured_variance`, `support`).

On `JacobianLens` (additive; existing methods unchanged):

- `occupancy(model, activation_or_prompt, layer, *, position=None, max_atoms=25, num_control_dictionaries=32, seed=0)` resolves a raw `[d_model]` activation or a prompt position exactly as `decompose`, then calls `estimate_occupancy` on the cached full-vocabulary dictionary.
- `fraction_of_variance(model, prompts, layers=None, *, k=25, skip_first=16, positions=None, show_progress=False)` profiles a prompt corpus and returns a `JSpaceVarianceProfile` (`layers`, `median`, `pooled`, `per_position`).

`JSpaceOccupancy`, `JSpaceVarianceProfile`, and `estimate_occupancy` are exported from `transformer_lens.tools.analysis`.

## Occupancy: max-separation versus a random-dictionary control

`estimate_occupancy` runs the same greedy selection as `get_sparse_decomposition` up to `max_atoms`, recording the per-step captured variance `||Pi_S x||² / ||x||²` (orthogonal projection onto the selected span). It runs the same greedy on `num_control_dictionaries` random unit-norm dictionaries of equal size and averages their curves. The occupancy is the step of **maximum separation** between the real and control *cumulative* captured variance: the point past which further vectors add no more than random directions would. It is **deterministic** (seeded) and needs **no threshold**.

An earlier per-step quantile-cutoff variant was calibration-sensitive (planted structure wanted `q ≤ 0.75`, a clean noise floor wanted `q ≥ 0.9`, with no single robust value). The max-separation rule recovers planted sparsity exactly on orthonormal dictionaries (`k=1` gives 1, `k=4` gives 4, `k=7` gives 7) with no tuning, so `quantile` was dropped.

## Fraction of variance: the span-projection operationalization

For each `(layer, position)` at or past `skip_first` (mirroring the fit's early-position skip), `fraction_of_variance` records `||j_space_component||² / ||activation||²` and reports, per layer, the `median` of those fractions and the `pooled` ratio `Σ||j_space||² / Σ||activation||²`. The numerator is the `selected_support` **span projection** (the paper's appendix operationalization), **not** the nonnegative `reconstruction`; the two coincide only when every selected atom stays active. This is named explicitly in the docstring and docs so the number is not misread. A layer that samples no positions (every prompt shorter than `skip_first`, or only zero-norm activations) yields `NaN` `median` and `pooled` values and an empty `per_position`. This is a documented, tested contract.

## Tests

- **Model-free unit** (`test_jacobian_lens_decomposition.py`): planted-sparsity recovery (`occupancy == k`), ordering by planted density, determinism, and input validation.
- **Wrapper unit** (`test_jacobian_lens.py`): occupancy raw and prompt paths, `fraction_of_variance` median and pooled over a corpus, unfitted-layer and empty-corpus rejection, and the no-sample `NaN` contract.
- **gpt2-small integration** (`test_jacobian_lens.py`): occupancy is a small positive integer with `[0, 1]`-bounded cumulative captured-variance curves and is seed-deterministic; `fraction_of_variance` median and pooled land in `[0, 1]`, including the `positions=` override path over a multi-prompt corpus.

## Honesty note

The paper's quantitative occupancy and variance figures are **closed-model** results (Sonnet, Haiku, Opus). Tests and docs assert **shape** (occupancy is a small positive integer; the variance fraction is a small ratio in `[0, 1]`), not the paper's exact numbers, which are not expected to transfer to open weights.

## Docs

A new "Occupancy and fraction of variance" subsection in the "Sparse decomposition" section of `jacobian_lens_fitting.md`, with runnable snippets and the open-weight, shape-only caveat.

## Type of change

### Screenshots

Not applicable. This adds an analysis API plus a docs page, with no visual output.

# Checklist:

## Verification run locally

The change is confined to `transformer_lens/tools/analysis/jacobian_lens*`, so the local run was scoped to the Jacobian-lens surface; CI re-runs the full suite across Python 3.10, 3.11, and 3.12.

- `check-format`: clean under the CI-pinned **black 23.12.1** plus isort (profile=black) on all changed files.
- `mypy`: Success, no issues in the two analysis modules.
- **unit plus integration plus docstring (not slow)**: one pytest invocation over `tests/unit/tools/test_jacobian_lens.py`, `test_jacobian_lens_decomposition.py`, `tests/integration/test_jacobian_lens.py`, and (docstring tier, `--doctest-modules --doctest-plus`) both analysis modules: **246 passed, 4 deselected** (the deselected are the slow `gemma-2-2b-it` cases).
- `build-docs` (Sphinx): builds clean (exit 0) with notebooks excluded; the new page and API entries render. The only warnings touching the new dataclasses are the pre-existing `duplicate object description` re-export pattern (identical to `JSpaceDecomposition`; `-W` is disabled in `conf.py` for exactly this), so no new warning class is introduced. A full local `build-docs` additionally needs `pandoc` for the demo notebooks, absent on this box and present in CI.
- Real-model integration: GPT-2 `occupancy` returns a small positive integer with `[0, 1]`-bounded captured-variance curves, and `fraction_of_variance` returns per-layer `median` and `pooled` ratios in `[0, 1]`.

### DIFF
--- a/transformer_lens/tools/analysis/jacobian_lens.py
+++ b/transformer_lens/tools/analysis/jacobian_lens.py
@@ -1036,11 +1036,12 @@ class JacobianLens:
 
         Args:
             model: A raw ``TransformerBridge``.
-            prompts: A prompt, or a sequence of prompts (strings and/or ``[1, seq]`` token tensors).
+            prompts: A prompt, or a sequence of prompts. Each token tensor must represent exactly
+                one prompt and have shape ``[1, seq]``.
             layers: Source layers to profile; defaults to all fitted ``source_layers``.
             k: Number of J-lens vectors per decomposition.
-            skip_first: Skip positions before this index (mirrors the fit's early-position skip);
-                ignored when ``positions`` is given.
+            skip_first: Non-negative index before which positions are skipped (mirrors the fit's
+                early-position skip); not used for sampling when ``positions`` is given.
             positions: Explicit positions to sample instead of ``skip_first`` onward.
             show_progress: Show a tqdm progress bar over prompts.
 
@@ -1048,9 +1049,12 @@ class JacobianLens:
             A :class:`JSpaceVarianceProfile`.
 
         Raises:
-            ValueError: On an invalid model, an unfitted layer, or an empty corpus.
+            ValueError: On an invalid model, an unfitted layer, an empty corpus, a negative
+                ``skip_first``, or a token tensor that does not have shape ``[1, seq]``.
         """
         self.validate_model(model)
+        if skip_first < 0:
+            raise ValueError(f"skip_first must be non-negative, got {skip_first}")
         if layers is None:
             resolved_layers = list(self.source_layers)
         else:
@@ -1078,6 +1082,11 @@ class JacobianLens:
 
         for prompt in tqdm(prompt_list, desc="J-space variance", disable=not show_progress):
             tokens = model.to_tokens(prompt) if isinstance(prompt, str) else prompt
+            if tokens.ndim != 2 or tokens.shape[0] != 1:
+                raise ValueError(
+                    "fraction_of_variance expects each tokenized prompt to have shape "
+                    f"[1, seq], got {tuple(tokens.shape)}"
+                )
             _, cache = model.run_with_cache(tokens, names_filter=lambda name: name in wanted_hooks)
             seq_len = tokens.shape[1]
             sampled = (
--- a/transformer_lens/tools/analysis/jacobian_lens.py
+++ b/transformer_lens/tools/analysis/jacobian_lens.py
@@ -73,6 +73,9 @@ from tqdm.auto import tqdm
 from transformer_lens.tools.analysis.jacobian_lens_decomposition import (
     DEFAULT_K,
     JSpaceDecomposition,
+    JSpaceOccupancy,
+    JSpaceVarianceProfile,
+    estimate_occupancy,
     get_sparse_decomposition,
 )
 from transformer_lens.utilities.hf_utils import call_hf_with_retry
@@ -899,6 +902,27 @@ class JacobianLens:
             RuntimeError: If the default nonnegative least-squares solver cannot validate its
                 result against the KKT conditions.
         """
+        activation, resolved_layer = self._resolve_activation(
+            model, activation_or_prompt, layer, position
+        )
+        dictionary = self.lens_vector_dictionary(model, resolved_layer)
+        return get_sparse_decomposition(
+            activation.float().to(dictionary.device), dictionary, k, algorithm=algorithm
+        )
+
+    def _resolve_activation(
+        self,
+        model: Any,
+        activation_or_prompt: Union[torch.Tensor, str],
+        layer: int,
+        position: Optional[int],
+    ) -> Tuple[torch.Tensor, int]:
+        """Validate the model and layer, then resolve either a raw ``[d_model]`` activation or a
+        prompt plus ``position`` to the activation vector to analyse.
+
+        Returns ``(activation, resolved_layer)``. Shared by :meth:`decompose` and
+        :meth:`occupancy` so both accept the same input forms with identical validation.
+        """
         self.validate_model(model)
         resolved_layer = _normalize_layer(layer, model.cfg.n_layers)
         if resolved_layer not in self.jacobians:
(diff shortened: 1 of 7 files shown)


transformer_lens/tools/analysis/jacobian_lens_decomposition.py
```diff
@@ -447,3 +447,157 @@ def get_sparse_decomposition(
...
+    if x.ndim != 1 or x.shape[0] != d_model:
+        raise ValueError(f"x must be 1-D of length d_model={d_model}, got shape {tuple(x.shape)}")
+    if not 1 <= max_atoms <= num_atoms:
+        raise ValueError(f"max_atoms must be between 1 and num_atoms={num_atoms}, got {max_atoms}")
+    if num_control_dictionaries < 1:
+        raise ValueError(
+            f"num_control_dictionaries must be at least 1, got {num_control_dictionaries}"
+        )
+
+    target = x.float()
```
jlarson4, line 559: `x` is never validated beyond shape. A zero-norm or NaN activation divides by `||x||^2 = 0` (or propagates NaN) and returns `occupancy=1` with all-NaN curves as if it were a real estimate. One of our other Jacobian Lens analysis tools, `get_sparse_decomposition`, rejects non-finite `x`, and the wrapper passes raw user activations straight through.

Can we mirror this solution for `estimate_occupancy`? We will want to validate `x` like `get_sparse_decomposition` does,  raising `ValueError` on non-finite entries, and also on a zero-norm target rather than dividing by zero.

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
```diff
@@ -447,3 +447,157 @@ def get_sparse_decomposition(
...
+    control_captured_variance: torch.Tensor
+    support: torch.Tensor
+
+
+def _greedy_captured_variance_gains(
+    atoms: torch.Tensor, atom_norms: torch.Tensor, target: torch.Tensor, max_atoms: int
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Greedily select up to ``max_atoms`` atoms and return the per-step captured-variance gains.
+
+    Selection matches :func:`get_sparse_decomposition`: at each step the atom most correlated
```
jlarson4: Can we rephrase this docstring (and in the `estimate_occupancy` docstring and `jacobian_lens_fitting.md` lines 323–324) to say the estimators share the per-step correlation *rule* but run on different residuals?

This greedy recurses with no early stop on the unconstrained span-projection residual, while `get_sparse_decomposition` recurses on the NNLS residual and stops early. The selected supports diverge once a nonnegativity clamp binds, so `support` here need not match `decompose(...).selected_support`.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -938,10 +962,160 @@ def decompose(
...
+        wanted_hooks = set(hook_names.values())
+        dictionaries = {
+            layer: self.lens_vector_dictionary(model, layer) for layer in resolved_layers
+        }
+        fractions: Dict[int, List[float]] = {layer: [] for layer in resolved_layers}
+        pooled_j_space: Dict[int, float] = {layer: 0.0 for layer in resolved_layers}
+        pooled_total: Dict[int, float] = {layer: 0.0 for layer in resolved_layers}
+
+        for prompt in tqdm(prompt_list, desc="J-space variance", disable=not show_progress):
+            tokens = model.to_tokens(prompt) if isinstance(prompt, str) else prompt
```
jlarson4, line 1084: The token-tensor shape is never validated: a `[2, seq]` batch runs the full forward but `cache[hook][0]` reads only row 0, so the profile silently covers one prompt while appearing to cover the batch, and a 1-D token tensor fails with a bare `IndexError`.

Can we apply a similar check to the one used in `_resolve_activation` to each prompt here? (`tokens.ndim != 2 or tokens.shape[0] != 1` generate a `ValueError` in that function)

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -938,10 +962,160 @@ def decompose(
...
+            _, cache = model.run_with_cache(tokens, names_filter=lambda name: name in wanted_hooks)
+            seq_len = tokens.shape[1]
+            sampled = (
+                list(range(skip_first, seq_len))
```
jlarson4, line 1093: Could we raise a `ValueError` when `skip_first < 0`? `range(skip_first, seq_len)` with a negative `skip_first` yields negative indices that wrap around, sampling trailing positions twice and silently skewing the median and pooled statistics.

jlarson4 (commented):
Hi @janmenjayap, I appreciate the thoroughness of this PR! Looks great, just a couple small comments on some edge cases that will need resolution before I merge

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
janmenjayap: Fixed in b1a7628a.

`estimate_occupancy` now rejects, before any projection math:

- non-finite entries in `x` or `dictionary` (mirroring `get_sparse_decomposition`),
- zero-norm targets (`x @ x == 0`),
- complex inputs (silently dropping imaginary parts was another silent-corruption path),
- non-finite or zero atom norms in the dictionary.

Each path has a focused regression with an explicit message match so a downstream `pinv`/`argmax` exception cannot satisfy the test. The full decomposition + import unit surface passes (182 tests).

transformer_lens/tools/analysis/jacobian_lens_decomposition.py
janmenjayap: Fixed in b1a7628a (same commit as Comment 1).

The `_greedy_captured_variance_gains` docstring, the `estimate_occupancy` docstring, and `jacobian_lens_fitting.md` now state that occupancy runs an unconstrained span-projection residual recurrence for exactly `max_atoms` steps with no early stop, while `get_sparse_decomposition` recurses on the NNLS residual and may stop early; so their supports can diverge once a nonnegativity clamp binds. The shared per-step correlation rule is still noted.

transformer_lens/tools/analysis/jacobian_lens.py
janmenjayap: Fixed in 4e6f9928.

Each token tensor is now checked for `ndim == 2 and shape[0] == 1` before `run_with_cache`, raising `ValueError` with the expected shape and the actual shape. Parameterized regressions cover the three failure modes: missing batch dimension (`[seq]`), multiple prompts (`[2, seq]`), and extra dimension (`[1, 1, seq]`).

transformer_lens/tools/analysis/jacobian_lens.py
janmenjayap: Fixed in 4e6f9928 (same commit as Comment 3).

`skip_first < 0` now raises `ValueError` unconditionally, including when explicit `positions` override the default sampling, so the contract is consistent regardless of whether the value is actually used for indexing. Parameterized regressions cover both the default-sampling and explicit-positions paths.

jlarson4:
Thanks for addressing all my comments @janmenjayap, looks great I am merging now!

PR #1678, opened 2026-08-15
Fix batchless accumulated residual normalization

### LINKED ISSUE #1677: [Bug Report] ActivationCache.accumulated_resid(apply_ln=True) crashes for a batchless scalar position
**Describe the bug**

`ActivationCache.accumulated_resid(..., apply_ln=True)` raises an `IndexError` when the cache no longer has a batch dimension and `pos_slice` selects a scalar position.

Selecting one batch element with `apply_slice_to_batch_dim(int)` is a supported operation and sets `cache.has_batch_dim=False`. `accumulated_resid(..., apply_ln=False)` continues to work on that cache, but enabling final-LayerNorm recomputation crashes when the position is also selected with an integer.

The equivalent operation with the batch dimension retained works correctly.

**Code example**

This reproduces using a tiny randomly initialized native Bridge and does not download a model:

```python
import torch

torch.manual_seed(0)

    n_layers=2,
    d_model=16,
    n_ctx=8,
    d_head=4,
    n_heads=4,
    d_vocab=32,
    act_fn="gelu",
)
model = TransformerBridge.boot_native(cfg)

tokens = torch.tensor(
    [
        [1, 2, 3, 4],
        [5, 6, 7, 8],
        [9, 10, 11, 12],
    ]
)

_, cache = model.run_with_cache(tokens)

# Control: retaining the batch dimension works.
full = cache.accumulated_resid(layer=-1, pos_slice=-1, apply_ln=True)
print(full.shape)
# torch.Size([3, 3, 16])

# Select one batch element and remove the batch dimension.
batchless = cache.apply_slice_to_batch_dim(1)
print(batchless.has_batch_dim)
# False

# Control: the batchless scalar-position result works without LN recomputation.
raw = batchless.accumulated_resid(layer=-1, pos_slice=-1, apply_ln=False)
print(raw.shape)
# torch.Size([3, 16])

# This crashes.
normalized = batchless.accumulated_resid(layer=-1, pos_slice=-1, apply_ln=True)
```

The final call raises:

```text
Traceback (most recent call last):
  File ".../transformer_lens/ActivationCache.py", line 519, in accumulated_resid
    components = self.apply_ln_to_stack(
  File ".../transformer_lens/ActivationCache.py", line 1414, in apply_ln_to_stack
    out = out.squeeze(1)
IndexError: Dimension out of range
(expected to be in range of [-1, 0], but got 1)
```

Expected behavior:

- The call should return a finite tensor with shape `[components, d_model]`.
- Its values should match the selected batch row of the equivalent result computed before removing the batch dimension.

For the example above, the expected shape is `torch.Size([3, 16])`.

**Root cause**

`accumulated_resid()` calls `apply_ln_to_stack()` without forwarding the cache's explicit batch-dimension state:

```python
components = self.apply_ln_to_stack(
    components,
    layer,
    pos_slice=pos_slice,
    mlp_input=mlp_input,
    recompute_ln=recompute_ln,
)
```

The final-LayerNorm recomputation path then infers whether a position dimension exists only from the tensor rank:

```python
had_pos_dim = residual_stack.ndim == 4
```

After selecting a single batch element and a scalar position, the residual stack has shape `[components, d_model]`. Each component therefore has shape `[d_model]`. The current code only inserts a temporary position dimension when `x.ndim == 2`, so the one-dimensional result later reaches `out.squeeze(1)` even though dimension 1 does not exist.

The shape handling should use the explicit `has_batch_dim` contract instead of relying only on rank inference. A batchless scalar component can be temporarily normalized as `[1, 1, d_model]`, then restored to `[d_model]`.

Regression coverage should compare the batchless result against the corresponding row of the batched result, rather than only asserting that no exception is raised.

Useful cases to cover include:

- batch dimension present versus removed with `apply_slice_to_batch_dim(int)`
- scalar, sliced, and unsliced positions
- `apply_ln=True` and `apply_ln=False`
- final-layer recomputation and cached-scale paths
- LayerNorm and RMSNorm configurations

**System Info**

- Installed from source
- Base: `dev-4.x` at `8b2ab80a7c42804cb58d8186862476459e07f911`
- OS: Windows
- Python: 3.12.10
- PyTorch: 2.11.0+cpu
- Transformers: 5.13.0

**Additional context**

The controls indicate that this is specific to the final-LayerNorm recomputation shape handling:

- retaining the batch dimension and using the same scalar position works;
- removing the batch dimension works when `apply_ln=False`;
- the cache contents and batch-selection operation are otherwise valid.

Closed issues #523 and #1076/#1077 concern residual normalization semantics and unfolded/fresh LayerNorm behavior, rather than this batchless scalar-position shape error.

Searches across open and closed issues and pull requests found no matching report using:

- `ActivationCache batchless`
- `apply_ln_to_stack has_batch_dim`
- `accumulated_resid Dimension out of range`
- `remove_batch_dim apply_ln`
- `accumulated_resid apply_ln batch dimension`

### Checklist

emerardd:
# Description

Fix `ActivationCache.accumulated_resid(..., apply_ln=True)` for caches whose batch dimension has been removed before selecting a scalar position.

The final-normalization recomputation path previously inferred batch and position semantics only from tensor rank. A batchless scalar residual stack has shape `[components, d_model]`, so each component reached `ln_final` as a one-dimensional tensor and then failed when the code attempted `squeeze(1)`.

`accumulated_resid()` now forwards the cache's explicit `has_batch_dim` state. The recomputation path temporarily restores the structural batch and position dimensions expected by `ln_final`, then reshapes each result back to its original component shape.

The new download-free regression matrix uses tiny native LN and RMS models. It compares batchless outputs against the corresponding row of the batched result across scalar, sliced, and unsliced positions, with and without normalization, for both cached-scale and final-normalization recomputation paths.

Fixes #1677

## Type of change

# Validation

- `66 passed` across the new unit regression matrix and the existing `tests/acceptance/test_activation_cache.py` surface
- `mypy .`: success across 434 source files

# Checklist:

No documentation text changes are required because this restores the existing batchless `ActivationCache` shape contract without changing the public API.

### DIFF
--- a/tests/unit/test_activation_cache.py
+++ b/tests/unit/test_activation_cache.py
@@ -1,0 +1,63 @@
+import pytest
+import torch
+
+from transformer_lens import ActivationCache
+from transformer_lens.config import TransformerBridgeConfig
+from transformer_lens.model_bridge import TransformerBridge
+
+
+@pytest.fixture(scope="module", params=["LN", "RMS"])
+def activation_cache(request: pytest.FixtureRequest) -> ActivationCache:
+    cfg = TransformerBridgeConfig(
+        n_layers=2,
+        d_model=16,
+        n_ctx=8,
+        d_head=4,
+        n_heads=4,
+        d_vocab=32,
+        act_fn="gelu",
+        normalization_type=request.param,
+    )
+    with torch.random.fork_rng(devices=[]):
+        torch.manual_seed(0)
+        model = TransformerBridge.boot_native(cfg)
+    tokens = torch.tensor(
+        [
+            [1, 2, 3, 4],
+            [5, 6, 7, 8],
+            [9, 10, 11, 12],
+        ]
+    )
+    _, cache = model.run_with_cache(tokens)
+    return cache
+
+
+@pytest.mark.parametrize("layer", [1, -1], ids=["cached-scale", "recomputed-final-ln"])
+@pytest.mark.parametrize(
+    "pos_slice", [None, (1, 3), -1], ids=["all-positions", "position-slice", "scalar-position"]
+)
+@pytest.mark.parametrize("apply_ln", [False, True], ids=["raw", "normalized"])
+def test_batchless_accumulated_resid_matches_batched_row(
+    activation_cache: ActivationCache,
+    layer: int,
+    pos_slice: tuple[int, int] | int | None,
+    apply_ln: bool,
+) -> None:
+    batch_index = 1
+    batched = activation_cache.accumulated_resid(
+        layer=layer,
+        pos_slice=pos_slice,
+        apply_ln=apply_ln,
+    )
+
+    batchless_cache = activation_cache.apply_slice_to_batch_dim(batch_index)
+    assert not batchless_cache.has_batch_dim
+    batchless = batchless_cache.accumulated_resid(
+        layer=layer,
+        pos_slice=pos_slice,
+        apply_ln=apply_ln,
+    )
+
+    expected = batched[:, batch_index]
+    assert batchless.shape == expected.shape
+    torch.testing.assert_close(batchless, expected)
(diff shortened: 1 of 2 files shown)


jlarson4:
Great work on this @emerardd, perfect solution for the problem, great testing. Merging now

PR #1693, opened 2026-08-19
Fix TransformerBridge parameter counting

### LINKED ISSUE #1691: [Bug Report] n_params_total does not mirror HookedTransformer — split and fused QKV are both counted
**Describe the bug**

`TransformerBridge.n_params_total` documents itself as mirroring `HookedTransformer.n_params_total`, but reports a substantially larger number. On gpt2:

| | value |
|---|---|
| `bridge.n_params_total` (compatibility mode) | 184,348,753 |
| `HookedTransformer.n_params_total` | 163,049,041 |

The property is used for reporting and rough memory estimates, so an inflated value is quietly misleading rather than obviously broken.

The gap comes from the bridge registering both the fused `attn.qkv` weight and the split `attn.q` / `attn.k` / `attn.v` views as real parameters, so the same underlying tensor is counted twice per layer, plus a synthetic unembedding bias.

**Code example**

```python
import torch

print(br.n_params_total, ht.n_params_total)   # 184348753 vs 163049041
```

**System Info**

Installed from source on `dev`; macOS arm64, CPU, fp32, Python 3.12, transformers 5.13.0.

`n_params_total` should count each underlying parameter once. De-duplicate by storage identity (`id(param)` / `param.data_ptr()`) when summing, so split views of a fused weight are not double-counted, at `transformer_lens/model_bridge/transformer_bridge.py:384`.

Please also check the property's docstring claims while you are there: it advertises comparability both with `HookedTransformer.n_params_total` and with HuggingFace's `model.num_parameters()`, and the latter differs again (gpt2 reports 124,439,808 because it excludes the tied unembedding). Whatever the property settles on, the docstring should state the counting convention explicitly rather than naming two references that disagree with each other.

Acceptance:
- [ ] `bridge.n_params_total` counts each underlying tensor once (no fused/split double-count)
- [ ] The value matches `HookedTransformer.n_params_total` for a joint-QKV model (gpt2) and a split-QKV model, or the docstring states precisely how and why it differs
- [ ] Test under `tests/unit/model_bridge/` pinning the count against the de-duplicated sum
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

**Additional context**

### Checklist

nightcityblade:
# Description

Make `TransformerBridge.n_params_total` preserve the wrapped model's ordinary PyTorch parameter count from before bridge instrumentation. This counts tied parameters once and excludes bridge-created split projection views and synthetic analysis tensors, so fused/GQA layouts and RoPE's fabricated `pos_embed.W_pos` cannot inflate the result.

The property docstring now states this convention and its intentional difference from `HookedTransformer.n_params_total` and `tl_parameters()`. Regression coverage uses both GPT-2 and the CI-cached `trl-internal-testing/tiny-Qwen2ForCausalLM-2.5`, with explicit checks for nonzero GQA K/V tensors and the synthetic RoPE placeholder.

Fixes #1691

Dependencies: N/A.

Validation:

- `uv run pytest tests/unit/model_bridge/test_n_params_total.py tests/unit/model_bridge/test_get_params_util.py -q` — 10 passed

## Type of change

- [ ] New feature (N/A)
- [ ] Breaking change (N/A)
- [x] This change requires a documentation update (included in the property docstring)

### Screenshots

N/A — this change has no user-interface output.

# Checklist:

- [x] New and existing relevant tests pass locally with my changes

### DIFF
--- a/tests/unit/model_bridge/test_n_params_total.py
+++ b/tests/unit/model_bridge/test_n_params_total.py
@@ -1,0 +1,27 @@
+"""Tests for ``TransformerBridge.n_params_total`` on real model layouts."""
+
+import pytest
+import torch
+from transformers import AutoModelForCausalLM
+
+from transformer_lens.model_bridge import TransformerBridge
+
+
+@pytest.mark.parametrize("model_name", ["gpt2", "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"])
+def test_n_params_total_matches_uninstrumented_model(model_name: str) -> None:
+    hf_model = AutoModelForCausalLM.from_pretrained(
+        model_name, dtype=torch.float32, attn_implementation="eager"
+    )
+    expected = sum(parameter.numel() for parameter in hf_model.parameters())
+    bridge = TransformerBridge.boot_transformers(model_name, hf_model=hf_model)
+    bridge.enable_compatibility_mode()
+    assert bridge.n_params_total == expected
+
+    if "Qwen2" in model_name:
+        tl_parameters = bridge.tl_parameters()
+        for name in ("W_K", "W_V"):
+            weight = tl_parameters[f"blocks.0.attn.{name}"]
+            assert weight.shape[0] == bridge.cfg.n_heads
+            assert torch.count_nonzero(weight)
+        assert not torch.count_nonzero(tl_parameters["pos_embed.W_pos"])
+        assert bridge.n_params_total < sum(p.numel() for p in tl_parameters.values())
--- a/transformer_lens/model_bridge/bridge.py
+++ b/transformer_lens/model_bridge/bridge.py
@@ -198,6 +198,7 @@ class TransformerBridge(HookIntrospectionMixin, nn.Module):
             tokenizer: The tokenizer to use (required)
         """
         super().__init__()
+        self._n_params_total = sum(parameter.numel() for parameter in model.parameters())
         self.__dict__["original_model"] = model
         self.adapter = adapter
         self.cfg = adapter.cfg
@@ -843,18 +844,17 @@ class TransformerBridge(HookIntrospectionMixin, nn.Module):
 
     @property
     def n_params_total(self) -> int:
-        """Total number of parameters in the model, including embeddings, biases,
-        and layer norm weights.
+        """Number of parameters in the wrapped model before bridge instrumentation.
 
-        Mirrors :attr:`HookedTransformer.n_params_total`. Use this when you want
-        the actual parameter count for memory budgeting, comparison with
-        HuggingFace's ``model.num_parameters()``, or alignment with reported
-        model sizes in papers (e.g. the Pythia suite).
+        This follows PyTorch's parameter iteration semantics, counting tied
+        parameters once. Bridge-created split views and synthetic zero tensors
+        are excluded, so the result can differ from
+        :attr:`HookedTransformer.n_params_total` and :meth:`tl_parameters`.
 
         Returns:
-            int: ``sum(p.numel() for p in self.parameters())``
+            int: Parameter count of the uninstrumented wrapped model.
         """
-        return sum(p.numel() for p in self.parameters())
+        return self._n_params_total
 
     def clear_hook_registry(self) -> None:
         """Clear the hook registry and force re-initialization."""


transformer_lens/model_bridge/bridge.py
```diff
@@ -792,18 +792,19 @@ def hook_dict(self) -> dict[str, HookPoint]:
...
+        attention projections through their split Q/K/V representation rather
+        than counting both representations. It can therefore differ from
+        HuggingFace's ``model.num_parameters()``.
 
         Returns:
-            int: ``sum(p.numel() for p in self.parameters())``
+            int: Number of elements across TransformerLens-form parameters.
         """
-        return sum(p.numel() for p in self.parameters())
+        return sum(parameter.numel() for parameter in self.tl_parameters().values())
```
jlarson4: The docstring's new claim "Mirrors `HookedTransformer.n_params_total` by counting the TransformerLens-form parameters exposed by `tl_parameters()`" is not true for GQA/RoPE models, and the property is worse than before the fix for it. 

This needs either: a fix to `get_bridge_params`'s GQA reshape + RoPE pos_embed handling so `tl_parameters()` itself is trustworthy, or an `n_params_total` implementation that doesn't route through a best-effort/zero-filling utility.

tests/unit/model_bridge/test_n_params_total.py
jlarson4, line 1: Issue #1691 asked to verify "a joint-QKV model (gpt2) and a split-QKV model." gpt2 alone can't distinguish "fixed" from "fixed only for MHA + learned-position architectures." A second case on a GQA + RoPE model would have caught the regression above immediately.

I'd recommend testing with `trl-internal-testing/tiny-Qwen2ForCausalLM-2.5`, it is already cached in CI (`.github/workflows/checks.yml`) and already loaded elsewhere in this repo's integration tier via `tests/tiny_checkpoints.py`'s `TINY_CHECKPOINTS["qwen2"]`, so it costs nothing extra offline/CI-wise. Reproduced directly against this PR's actual head (`e7e46703`, not just the full-size model): `bridge.n_params_total` = **2,842,625** vs. the real HF param count of **2,428,632** (+17.0%), with the same two mechanisms confirmed, `tl_parameters()['blocks.0.attn.W_K']` is all-zero (shape `[4, 8, 2]`), and `tl_parameters()['pos_embed.W_pos']` is a fabricated all-zero `[32768, 8]` tensor.

One caveat: a bare `assert bridge.n_params_total == expected` may produce unreliable results in that there are two offsetting bugs (RoPE overcounts, GQA undercounts). It could be easy to only fix the more obvious one (the `pos_embed` try/except in `get_params_util.py` is easy to spot) and stop once the number gets closer but still doesn't match, without realizing there's a second independent cause. Worth either asserting on the individual `W_K`/`W_V`/`pos_embed.W_pos` keys directly (not just the summed total), or at least noting in the test both mechanisms so a future maintainer isn't left to rediscover them.

Existing precedent to build on: `tests/unit/model_bridge/test_get_params_util.py` already has a `TestGQAExpansion` class (~line 263) covering GQA in `get_bridge_params`, but via a synthetic `Mock()`-based bridge with hand-picked tensor shapes rather than a real HF config, which is presumably why it didn't already catch this: a mock built by the same author's mental model of the code is liable to encode the same shape assumption the implementation makes, rather than the real HF weight layout. A same-file test using an actual tiny checkpoint would be more resistant to that. Separately, `tests/integration/model_bridge/test_bridge_qk_ov_vs_hooked_gqa.py` already exercises GQA correctness against full `Qwen/Qwen2-0.5B`, marked `@pytest.mark.slow` specifically because it isn't CI-cached.

jlarson4 (commented):
Great start on this @nightcityblade! Can we also add a verification test for a split-QKV model? From my local testing of this branch on a split-QKV, GQA, RoPE model, the new implementation is still wrong for that model type. Additional details below on the line and file specific comments

transformer_lens/model_bridge/bridge.py
nightcityblade: Addressed in e0aebba by taking `n_params_total` off `tl_parameters()` entirely. The bridge records the wrapped model's normal PyTorch parameter count before instrumentation, so fused/split bridge views and synthetic analysis tensors do not affect it. The docstring now states that convention and the intentional difference from HookedTransformer/TL-form counts.

tests/unit/model_bridge/test_n_params_total.py
nightcityblade: Added the recommended `trl-internal-testing/tiny-Qwen2ForCausalLM-2.5` case alongside GPT-2. The test explicitly checks both real GQA `W_K`/`W_V` tensors are nonzero and expanded to query-head layout by `tl_parameters()`, and that `pos_embed.W_pos` is the fabricated all-zero RoPE placeholder; it then verifies neither affects `n_params_total`. On this checkpoint the result is 2,428,632 versus a TL-analysis sum of 2,842,681. Focused result: 10 passed; `make check-format` and `uv run mypy .` pass.

jlarson4:
This one looks good! merging now

PR #1694, opened 2026-08-19
Fix compatibility attention mask sentinel

### LINKED ISSUE #1692: [Bug Report] hook_attn_scores masks with finfo.min instead of -inf, so isinf-based mask detection silently finds nothing
**Describe the bug**

Cached `hook_attn_scores` uses `torch.finfo(float32).min` (`-3.403e38`) for causally-masked positions, where `HookedTransformer` uses `-inf`. Unmasked values agree exactly (`0.0` max abs difference on gpt2), so the tensors are numerically equivalent for softmax purposes, but any analysis that locates the causal mask with `torch.isinf(scores)` or `scores == -torch.inf` silently matches nothing on the bridge, and the code carries on with an empty mask.

Measured on gpt2, `blocks.0.attn.hook_attn_scores`:

| | min value | `torch.isinf(...).any()` |
|---|---|---|
| HookedTransformer | `-inf` | True |
| TransformerBridge (compat mode) | `-3.403e+38` | False |

**Code example**

```python
import torch

_, ch = ht.run_with_cache("The capital of France is", names_filter=["blocks.0.attn.hook_attn_scores"])
_, cb = br.run_with_cache("The capital of France is", names_filter=["blocks.0.attn.hook_attn_scores"])
s_ht, s_br = ch["blocks.0.attn.hook_attn_scores"], cb["blocks.0.attn.hook_attn_scores"]

print(torch.isinf(s_ht).any(), torch.isinf(s_br).any())   # True False
```

**System Info**

Installed from source on `dev`; macOS arm64, CPU, fp32, Python 3.12, transformers 5.13.0.

Compatibility mode should present the HookedTransformer sentinel (`-inf`) in cached `hook_attn_scores`, so that mask-detection code ported from `HookedTransformer` keeps working.

Note the sentinel is HuggingFace's, not the bridge's invention so the fix belongs at the point where the scores are exposed to the hook, not in the attention maths. If changing the cached value is judged too invasive (e.g. it would break bf16/fp16 paths or a fused kernel), the fallback is to document the sentinel in the migration guide's numerics section and in the `hook_attn_scores` docstring, since the current behaviour is a silent no-match rather than a visible difference.

Acceptance:
- [ ] Cached `hook_attn_scores` under compatibility mode uses `-inf` for masked positions, or the deviation is documented in the migration guide and the hook's docstring
- [ ] Unmasked score values remain exactly equal to HookedTransformer (currently `0.0` difference)
- [ ] Test under `tests/unit/model_bridge/` asserting the chosen sentinel
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

### Checklist

nightcityblade:
# Description

In compatibility mode, normalize fully assembled attention masks from `torch.finfo(dtype).min` to `-inf` before exposing `hook_attn_scores`. Normalizing after causal and Hugging Face masks are combined covers GPT-2's direct 4D-mask path while leaving native-mode and unmasked score values unchanged.

Fixes #1692

Dependencies: N/A.

Validation:

- `uv run pytest tests/integration/model_bridge/test_attention_score_sentinel.py tests/unit/model_bridge/generalized_components/test_joint_qkv_attention.py -q` — 10 passed

## Type of change

- [ ] New feature (non-breaking change which adds functionality) — N/A
- [ ] Breaking change (fix or feature which changes existing behavior) — N/A
- [ ] This change requires a documentation update — N/A

### Screenshots

N/A.

# Checklist:

- [ ] I have made corresponding changes to the documentation — N/A; no public API changed
- [x] New and existing relevant tests pass locally with my changes

### DIFF
--- a/tests/integration/model_bridge/test_attention_score_sentinel.py
+++ b/tests/integration/model_bridge/test_attention_score_sentinel.py
@@ -1,0 +1,67 @@
+"""Compatibility-mode attention-score sentinel regression coverage."""
+
+import torch
+
+SCORES = "blocks.0.attn.hook_attn_scores"
+PATTERN = "blocks.0.attn.hook_pattern"
+
+
+def test_gpt2_compatibility_scores_use_negative_infinity(
+    gpt2_bridge_compat, gpt2_hooked_processed
+) -> None:
+    """GPT-2's direct HF mask is normalized before the compatibility hook."""
+    tokens = gpt2_hooked_processed.to_tokens("The capital of France is")
+    _, bridge_cache = gpt2_bridge_compat.run_with_cache(tokens, names_filter=[SCORES])
+    _, hooked_cache = gpt2_hooked_processed.run_with_cache(tokens, names_filter=[SCORES])
+
+    bridge_scores, hooked_scores = bridge_cache[SCORES], hooked_cache[SCORES]
+    causal_mask = torch.isneginf(hooked_scores)
+    assert causal_mask.any()
+    assert torch.isneginf(bridge_scores[causal_mask]).all()
+    torch.testing.assert_close(
+        bridge_scores[~causal_mask], hooked_scores[~causal_mask], rtol=0, atol=0
+    )
+
+
+def test_gpt2_left_padding_uses_negative_infinity_and_finite_patterns(
+    gpt2_bridge_compat,
+) -> None:
+    """Fully masked pad queries are zeroed after softmax in compatibility mode."""
+    long = gpt2_bridge_compat.to_tokens("The capital of France is")
+    short = gpt2_bridge_compat.to_tokens("Paris")
+    n_pad = long.shape[1] - short.shape[1]
+    padded_short = torch.cat([torch.zeros_like(long[:, :n_pad]), short], dim=1)
+    tokens = torch.cat([long, padded_short], dim=0)
+    attention_mask = torch.cat(
+        [
+            torch.ones_like(long),
+            torch.cat([torch.zeros_like(long[:, :n_pad]), torch.ones_like(short)], dim=1),
+        ],
+        dim=0,
+    )
+
+    _, cache = gpt2_bridge_compat.run_with_cache(
+        tokens, attention_mask=attention_mask, names_filter=[SCORES, PATTERN]
+    )
+    scores, pattern = cache[SCORES], cache[PATTERN]
+    key_padding = ~attention_mask.bool()[:, None, None, :]
+    causal = torch.triu(torch.ones(long.shape[1], long.shape[1], dtype=torch.bool), diagonal=1)[
+        None, None
+    ]
+    masked = (key_padding | causal).expand_as(scores)
+
+    assert torch.isneginf(scores[masked]).all()
+    assert torch.isfinite(pattern).all()
+
+
+def test_gpt2_mixed_dtype_mask_is_normalized_before_addition(gpt2_bridge_compat) -> None:
+    """A lower-precision HF mask sentinel must survive score upcasting."""
+    scores = torch.zeros(1, 1, 2, 2, dtype=torch.float32)
+    attention_mask = torch.zeros_like(scores, dtype=torch.float16)
+    attention_mask[..., 0, 1] = torch.finfo(torch.float16).min
+
+    actual = gpt2_bridge_compat.blocks[0].attn._apply_reconstruct_attention_mask(
+        scores, attention_mask, seq_len=2
+    )
+
+    assert torch.isneginf(actual[..., 0, 1]).all()
--- a/transformer_lens/model_bridge/generalized_components/attention.py
+++ b/transformer_lens/model_bridge/generalized_components/attention.py
@@ -518,9 +518,24 @@ class AttentionBridge(GeneralizedComponent):
             attn_weights = torch.nn.functional.softmax(attn_scores, dim=-1)
             if target_dtype is not None:
                 attn_weights = attn_weights.to(target_dtype)
+        attn_weights = self._scrub_compatibility_pattern_nans(attn_weights)
         attn_weights = self._apply_attn_dropout(attn_weights)
         attn_weights = self.hook_pattern(attn_weights)
         return attn_weights
+
+    def _scrub_compatibility_pattern_nans(self, pattern: torch.Tensor) -> torch.Tensor:
+        """Match HookedTransformer for fully masked attention rows."""
+        if self.compatibility_mode:
+            pattern = torch.where(torch.isnan(pattern), torch.zeros_like(pattern), pattern)
+        return pattern
+
+    def _normalize_compatibility_mask_sentinel(self, attention_mask: torch.Tensor) -> torch.Tensor:
+        """Normalize additive mask sentinels before dtype conversion or addition."""
+        if self.compatibility_mode and attention_mask.is_floating_point():
+            attention_mask = attention_mask.masked_fill(
+                attention_mask <= torch.finfo(attention_mask.dtype).min, -torch.inf
+            )
+        return attention_mask
 
     def _reshape_attn_output(
         self,
@@ -558,6 +573,7 @@ class AttentionBridge(GeneralizedComponent):
         if q_seq_len is None:
             q_seq_len = seq_len
         min_dtype = torch.finfo(attn_scores.dtype).min
+        mask_value = -torch.inf if self.compatibility_mode else min_dtype
         use_direct_hf_mask = attention_mask is not None and attention_mask.ndim >= 4
         # Bidirectional attention (encoders) and cross-attention have no causal
         # structure, so only synthesize the triangular mask for causal self-attention.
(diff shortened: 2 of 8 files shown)


transformer_lens/model_bridge/generalized_components/attention.py
```diff
@@ -557,7 +557,7 @@ def _apply_reconstruct_attention_mask(
         """
         if q_seq_len is None:
             q_seq_len = seq_len
-        min_dtype = torch.finfo(attn_scores.dtype).min
+        min_dtype = -torch.inf if self.compatibility_mode else torch.finfo(attn_scores.dtype).min
```
jlarson4: The acceptance criterion "Cached `hook_attn_scores` under compatibility mode uses `-inf` for masked positions" is not met for `gpt2`. This path is not exercised by `gpt2`.

tests/unit/model_bridge/generalized_components/test_joint_qkv_attention.py
```diff
@@ -440,6 +440,26 @@ def __init__(self):
...
+    def test_compatibility_mode_exposes_causal_mask_as_negative_infinity(self):
+        """Compatibility hooks should use HookedTransformer's causal-mask sentinel."""
+
+        class TestConfig:
+            n_heads = 2
+            d_model = 4
+
+        bridge = JointQKVAttentionBridge(name="qkv", config=TestConfig())
+        captured = []
+        native_scores = bridge._apply_reconstruct_attention_mask(torch.zeros(1, 2, 3, 3), None, 3)
```
jlarson4: Calling `_apply_reconstruct_attention_mask` directly with a hand-constructed `attention_mask=None` skips the branch (`use_direct_hf_mask`) that we are trying to exercise. A test that goes through `_reconstruct_attention` (or better, a full `boot_transformers("gpt2")` + `run_with_cache` integration test matching the issue's repro) would have caught this, and is a stronger test that we can add.

jlarson4 (commented):
Hi @nightcityblade! You're on the right track here, but this doesn't actually resolve the issue for `gpt2` as reproduced in the issue. This path is not exercised by GPT-2, so the results of the repro script from the issue are the same with and without this fix. 

I'd recommend normalizing the sentinel once, after the mask is fully assembled, rather than doing it in the two construction sites at `_apply_reconstruct_attention_mask` and under `if self.compatibility_mode:`. Let's remap `attn_scores <= torch.finfo(attn_scores.dtype).min` to `-inf` regardless of which branch produced them. That covers the direct-HF-mask passthrough case along with the two already-fixed branches, and matches how `HookedTransformer` behaves (it builds its own mask from scratch with the `-inf` sentinel rather than ever incorporating an externally-provided finfo.min mask. For reference, see `transformer_lens/components/abstract_attention.py:135`, `self.register_buffer("IGNORE", torch.tensor(-torch.inf))`).

transformer_lens/model_bridge/generalized_components/attention.py
nightcityblade: Addressed in f3bff25. Compatibility mode now normalizes the sentinel once, after the causal mask and any direct HF/padding mask have been fully assembled. Values `<= torch.finfo(attn_scores.dtype).min` are remapped to `-inf`, so GPT-2's direct 4D HF-mask path is covered while native mode remains unchanged.

tests/unit/model_bridge/generalized_components/test_joint_qkv_attention.py
nightcityblade: Replaced the hand-constructed unit test with a full GPT-2 `run_with_cache` integration test matching the issue's path. It derives the causal mask from HookedTransformer, asserts every corresponding bridge score is `-inf`, and checks all unmasked scores bit-for-bit (`rtol=0`, `atol=0`). Focused result: 10 passed; `make check-format` and `uv run mypy .` also pass.

transformer_lens/model_bridge/generalized_components/attention.py
```diff
@@ -571,27 +571,28 @@ def _apply_reconstruct_attention_mask(
...
+                    torch.zeros((), dtype=attn_scores.dtype, device=attn_scores.device),
+                    torch.full((), min_dtype, dtype=attn_scores.dtype, device=attn_scores.device),
+                )
+            else:
+                attention_mask = attention_mask.to(dtype=attn_scores.dtype)
+            attn_scores = attn_scores + attention_mask
 
-        return attn_scores + attention_mask
+        if self.compatibility_mode:
+            attn_scores = attn_scores.masked_fill(attn_scores <= min_dtype, -torch.inf)
```
jlarson4: In a left-padded batch, a pad query's visible keys are all masked, so the remap makes the whole row `-inf`, softmax returns NaN, and the NaN reaches real positions through K/V from layer 1 on. The red `[left]` cases in the CI's `test_batched_string_padding` and `test_left_padding_positions`. HookedTransformer pairs this sentinel with a post-softmax NaN scrub (`abstract_attention.py:361`) that the bridge doesn't have. Please add the compatibility-gated equivalent after softmax and before `hook_pattern` in `_softmax_dropout_pattern` (`attention.py:514/518`) and at the two inline softmaxes (`position_embeddings_attention.py:618/622`, `llada.py:126`).

transformer_lens/model_bridge/generalized_components/attention.py
```diff
@@ -571,27 +571,28 @@ def _apply_reconstruct_attention_mask(
...
+                    attention_mask,
+                    torch.zeros((), dtype=attn_scores.dtype, device=attn_scores.device),
+                    torch.full((), min_dtype, dtype=attn_scores.dtype, device=attn_scores.device),
+                )
+            else:
+                attention_mask = attention_mask.to(dtype=attn_scores.dtype)
+            attn_scores = attn_scores + attention_mask
 
-        return attn_scores + attention_mask
+        if self.compatibility_mode:
```
jlarson4: The remap only reaches this helper's callers; Falcon (`alibi_joint_qkv_attention.py:121`), BLOOM, MPT (`mpt_alibi_attention.py:96`), and DeepSeek-MLA build their masks inline and keep `finfo.min`, so `torch.isinf` mask detection works on GPT-2/Llama-family models and silently returns empty on those. Please compat-gate their inline mask fills too (same one-line pattern as here).

transformer_lens/model_bridge/generalized_components/attention.py
jlarson4: Additionally here, `attn_scores <= min_dtype` compares against fp32-min whenever scores are upcast (`codegen_attention.py`, `reorder_and_upcast_attn`) while HF built the mask at the model dtype, so fp16/bf16 masked entries never reach the threshold and keep the old sentinel; in same-dtype fp16 the sum rounds above the threshold for scores ≥ ~16. Please remap the additive mask in its own dtype before it's added (masked entries there are exactly `finfo(mask.dtype).min`), or state that the `-inf` guarantee is fp32-only.

tests/integration/model_bridge/test_attention_score_sentinel.py
jlarson4, line 1: A single unpadded prompt never produces a fully-masked row, so this test stays green while the padding suites go red. Please add a left-padded batch case asserting `-inf` at masked score positions and `torch.isfinite(pattern).all()`.

tests/integration/model_bridge/test_attention_score_sentinel.py
```diff
@@ -0,0 +1,22 @@
...
+) -> None:
+    """GPT-2's direct HF mask is normalized before the compatibility hook."""
+    tokens = gpt2_hooked_processed.to_tokens("The capital of France is")
+    _, bridge_cache = gpt2_bridge_compat.run_with_cache(tokens, names_filter=[SCORES])
+    _, hooked_cache = gpt2_hooked_processed.run_with_cache(tokens, names_filter=[SCORES])
+
+    bridge_scores, hooked_scores = bridge_cache[SCORES], hooked_cache[SCORES]
+    causal_mask = torch.isneginf(hooked_scores)
+    assert causal_mask.any()
+    assert torch.isneginf(bridge_scores[causal_mask]).all()
```
jlarson4, line 20: `test_bridge_cache_behavior.py:131-133` and `:151` still say the bridge masks with `torch.finfo(dtype).min`, which this change makes false in compat mode. Please update those two notes to the new contract.

jlarson4 (commented):
Hey @nightcityblade! 

In reviewing this, I realized the scope of my issue was too narrow. There are some related components that are now failing that I missed in my documentation, detailed below. Please let me know if you have any questions. The first comment should lead you to the fix for the CI failures that are currently blocking.

transformer_lens/model_bridge/generalized_components/attention.py
nightcityblade: Addressed in 5714c6a. Floating additive masks are normalized in their original dtype before conversion and addition, so fp16/bf16 sentinels survive fp32 score upcasting. Added a mixed-dtype regression case.

tests/integration/model_bridge/test_attention_score_sentinel.py
nightcityblade: Updated in 5714c6a. Both stale notes now describe compatibility mode's negative-infinity contract rather than the old finfo-min behavior.

transformer_lens/model_bridge/generalized_components/attention.py
nightcityblade: Addressed in 5714c6a. Additive mask sentinels are now normalized for the shared ALiBi path used by Falcon/BLOOM and for DeepSeek MLA; MPT's boolean mask fill now uses negative infinity in compatibility mode.

transformer_lens/model_bridge/generalized_components/attention.py
nightcityblade: Addressed in 5714c6a. Compatibility mode now applies the same post-softmax NaN-to-zero behavior as HookedTransformer in the shared attention path and in the position-embedding and LLaDA inline softmax paths, before hook_pattern.

tests/integration/model_bridge/test_attention_score_sentinel.py
nightcityblade: Added in 5714c6a. The regression suite now includes a left-padded GPT-2 batch and asserts negative infinity at masked score positions plus finite values across hook_pattern.

nightcityblade:
Addressed the expanded compatibility-mode scope in 5714c6a:

- normalize additive mask sentinels before dtype conversion/addition, including the Falcon/BLOOM ALiBi, MPT, and DeepSeek MLA paths
- scrub fully masked-row softmax NaNs before `hook_pattern` in the shared, position-embedding, and LLaDA paths
- add left-padding and mixed-dtype regression coverage, and update the stale cache-test notes

Validation:
- 7 focused GPT-2 sentinel/padding tests passed, including both previously failing CI cases
- 227 affected attention/adapter unit tests passed
- `make check-format` passed
- `uv run mypy .` passed (388 source files)
- `git diff --check` passed

nightcityblade:
CI follow-up: the only failing job (`Full Code Coverage Test`) completed 6,759 tests and then hit a single unrelated tolerance failure in `tests/unit/test_direct_path_patching.py::TestCorrectness::test_correctness_against_actual_ln_forward` (difference `1.12e-03` versus the `1e-03` threshold). The changed model-bridge tests and every other CI job passed.

I attempted to rerun the failed job, but GitHub requires repository admin permission. Could a maintainer please rerun it?

jlarson4:
Hi @nightcityblade, thanks for tackling these updates so swiftly! I am rerunning the job now, there is an independent bug in the CI right now that is causing intermittent failures. When I get a resolution to `dev`, I will make sure to incorporate it into your branch as well if it fails this re-run

nightcityblade:
Thank you — I appreciate the rerun and the context on the intermittent CI issue. I’ll keep an eye on the result.

jlarson4:
CI passed, merging now!

PR #1699, opened 2026-08-19
Fix OLMo 2 attention input state leakage

### LINKED ISSUE #1687: [Bug Report] OLMo 2: use_split_qkv_input / use_attn_in corrupt the forward pass, non-deterministically, across repeated runs
**Describe the bug**

On OLMo 2 bridges, enabling `set_use_split_qkv_input(True)` (or `set_use_attn_in(True)`) makes the forward pass **wrong and non-reproducible**: repeating the *same* `run_with_cache` call on the same tokens returns different logits each time, drifting further from the correct values with each run. The error is the same order of magnitude as the signal, so results look plausible while being meaningless, and re-running does not reproduce them.

Measured on `hf-internal-testing/tiny-random-Olmo2ForCausalLM` (compatibility mode, fresh bridge, setter called before any forward), comparing against a reference bridge with the flag off — reference logit scale is **25.68**:

| run | max abs logit delta vs reference | `blocks.1.attn.hook_q_input − blocks.1.hook_resid_pre` |
|---|---|---|
| 1 | 30.49 | 0.0000 |
| 2 | 29.33 | 56.03 |
| 3 | 24.21 | 60.91 |
| 4 | 26.92 | 63.14 |

A `distilgpt2` control run through the identical sequence is flat at `1.81e-05` on all runs. The `hook_q_input` column shows the split-QKV residual copy diverging from `hook_resid_pre`, which it should equal.

**Code example**

```python
import torch

M = "hf-internal-testing/tiny-random-Olmo2ForCausalLM"
toks = torch.tensor([[1, 2, 3, 4, 5]])

reference = TransformerBridge.boot_transformers(M, device="cpu")
reference.enable_compatibility_mode()
expected = reference(toks, return_type="logits").clone()

bridge = TransformerBridge.boot_transformers(M, device="cpu")
bridge.set_use_split_qkv_input(True)

for i in range(4):
    bridge.run_with_cache(toks)
    delta = (bridge(toks, return_type="logits") - expected).abs().max().item()
    print(i, delta)   # 30.49, 29.33, 24.21, 26.92 — different every run
```

**System Info**

Installed from source on `dev-4.x`; macOS arm64, CPU, fp32, Python 3.12, transformers 5.13.0. Not architecture-specific to the tiny fixture — it is the OLMo 2 adapter path.

Two invariants are violated and both should hold:

1. **Determinism** – repeating an identical forward on an unchanged model must return identical activations and logits. State must not accumulate across forwards.
2. **Correctness of the split fork** – with `use_split_qkv_input`, each of `hook_q_input` / `hook_k_input` / `hook_v_input` must equal the residual entering the block (`hook_resid_pre`) before any patching, as it does on pre-norm architectures (verified `0.0` on gpt2/distilgpt2).

Start at `_propagate_attention_flag` (`transformer_lens/model_bridge/transformer_bridge.py:3621`, called from `set_use_split_qkv_input` at `:3608` and `set_use_attn_in` at `:3623`) and at the OLMo 2 adapter's attention wiring. The drift pattern points at per-forward state that is written but never reset between runs. Note OLMo 2 is post-norm, so the fork point differs from the pre-norm families where this path is exercised by tests.

Acceptance:
- [ ] Repeating an identical `run_with_cache` on an OLMo 2 bridge with `use_split_qkv_input=True` returns identical logits and activations across runs
- [ ] `hook_q_input` / `hook_k_input` / `hook_v_input` equal `hook_resid_pre` on a fresh run, or the setter refuses with `NotImplementedError`
- [ ] Same for `set_use_attn_in(True)`, which shows the same signature
- [ ] Regression test under `tests/unit/model_bridge/` (or `tests/integration/model_bridge/`) that runs the same input **at least twice** and asserts equality — a single-run test cannot catch this
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

**Additional context**

### Checklist

original4422:
# Description

OLMo 2 is post-norm, so its `ln1` capture runs after attention. The specialized attention reconstruction path did not clear `_captured_pre_ln_residual`, causing the following forward to feed the previous run's captured tensor into `hook_q_input` / `hook_k_input` / `hook_v_input` or `hook_attn_in`.

This change makes `BlockBridge` own the lifetime of that transient capture: it clears stale state at block entry and releases the current capture in a `finally` block. Repeated identical forwards are deterministic again, and the forked attention inputs match `hook_resid_pre`.

The regression uses a tiny local `Olmo2Config` model, so it exercises the real post-norm adapter path without a Hub download. It covers both `set_use_split_qkv_input(True)` and `set_use_attn_in(True)` across two consecutive forwards.

Fixes #1687

## Type of change

### Screenshots

N/A — no user-interface change.

# Checklist

- [x] I have commented the state-lifetime helper where its purpose is not obvious
- [x] Documentation is unchanged because the public API and intended behavior are unchanged
- [x] I have added tests that prove the fix is effective

## Validation

- `uv run pytest tests/integration/model_bridge/test_olmo2_attention_input_state.py -q` — 2 passed
- Related split-QKV, Mistral attention-input, and OLMo 2 hook semantics tests — 18 passed
- `make unit-test` — 5,028 passed, 36 skipped, 44 deselected, 8 xfailed

### DIFF
--- a/tests/integration/model_bridge/test_olmo2_attention_input_state.py
+++ b/tests/integration/model_bridge/test_olmo2_attention_input_state.py
@@ -1,0 +1,73 @@
+"""OLMo 2 attention-input fork regression tests using a tiny local HF model."""
+
+import pytest
+import torch
+from transformers import AutoModelForCausalLM
+from transformers.models.olmo2 import Olmo2Config
+
+from transformer_lens.model_bridge import TransformerBridge
+from transformer_lens.model_bridge.sources._bridge_builder import (
+    build_bridge_config_from_hf,
+)
+from transformer_lens.model_bridge.supported_architectures.olmo2 import (
+    Olmo2ArchitectureAdapter,
+)
+
+
+class _Tokenizer:
+    pass
+
+
+def _tiny_olmo2_bridge() -> TransformerBridge:
+    torch.manual_seed(0)
+    config = Olmo2Config(
+        vocab_size=64,
+        hidden_size=32,
+        intermediate_size=64,
+        num_hidden_layers=2,
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        max_position_embeddings=32,
+    )
+    config.architectures = ["Olmo2ForCausalLM"]
+    hf_model = AutoModelForCausalLM.from_config(config).to(torch.float32).eval()
+    bridge_config = build_bridge_config_from_hf(
+        hf_model.config, "Olmo2ForCausalLM", "olmo2-tiny", torch.float32
+    )
+    return TransformerBridge(
+        hf_model, Olmo2ArchitectureAdapter(bridge_config), tokenizer=_Tokenizer()
+    )
+
+
+@pytest.mark.parametrize(
+    ("setter_name", "input_hook_names"),
+    [
+        (
+            "set_use_split_qkv_input",
+            (
+                "blocks.1.attn.hook_q_input",
+                "blocks.1.attn.hook_k_input",
+                "blocks.1.attn.hook_v_input",
+            ),
+        ),
+        ("set_use_attn_in", ("blocks.1.attn.hook_attn_in",)),
+    ],
+)
+def test_attention_input_fork_does_not_leak_state_between_forwards(
+    setter_name: str, input_hook_names: tuple[str, ...]
+) -> None:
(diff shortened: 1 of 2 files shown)


jlarson4:
Looks good! Merging

PR #1701, opened 2026-08-19
Fix batched TransformerBridge padding semantics

### LINKED ISSUE #1690: [Bug Report] Batched string-list input pads on the opposite side from HookedTransformer, and to_tokens(padding_side=...) is ignored
**Describe the bug**

Two related padding defects:

1. Opposite padding side on string lists – Passing a list of unequal-length strings to a compatibility-mode bridge produces logits that differ from `HookedTransformer` by 22.49 on gpt2 (logit scale ~126). Both models compute internally-correct activations, but at mirrored positions: HT right-pads, the bridge left-pads. Feeding pre-tokenized tensors instead gives an exact `0.0` match, which isolates the divergence to the string-list path. Every position index into the cache or logits for a padded row is therefore wrong after migration, `logits[0, -1]` is the last real token on the bridge and a pad token on HT.

2. `padding_side` is inert in `to_tokens` – `to_tokens(strs, padding_side="left")` and `padding_side="right"` return identical tensors; the tokenizer's own `padding_side` wins. The same argument on `HookedTransformer.to_tokens` genuinely changes the output. This also means `bridge.to_tokens(list)` cannot be used to reconstruct the indices that `bridge(list)` actually used.

**Code example**

```python
import torch

batch = ["The quick brown fox", "Hello there, world! This is a longer sentence."]

print((ht(batch) - br(batch)).abs().max())                                  # 22.4894
print((ht(ht.to_tokens(batch)) - br(br.to_tokens(batch))).abs().max())      # 0.0

left  = br.to_tokens(batch, padding_side="left")
right = br.to_tokens(batch, padding_side="right")
print(torch.equal(left, right))                                             # True  <- ignored
print(torch.equal(ht.to_tokens(batch, padding_side="left"),
                  ht.to_tokens(batch, padding_side="right")))               # False <- honored
```

**System Info**

Installed from source on `dev`; macOS arm64, CPU, fp32, Python 3.12, transformers 5.13.0. Reproduced on gpt2, distilgpt2, and `roneneldan/TinyStories-1M`.

The two defects are coupled, details of my investigation below.

1. `to_tokens` must honor its own `padding_side` argument – It resolves the value
(`transformer_lens/model_bridge/transformer_bridge.py:760-761`, defaulting from the tokenizer)
and then never passes it to the `self.tokenizer(...)` call that follows, so the tokenizer's own
attribute decides. Threading the resolved value into that call fixes `to_tokens`,
`to_str_tokens` (`:812`), `get_token_position` (`:884`), and the forward paths (`:1455`,
`:1612`) in one place.

2. The batched-list forward should right-pad, matching `HookedTransformer`. – The forward
currently forces left-padding for any multi-item list (`:1735-1745`), which is where it diverges
from HT. Three pieces of evidence say right-padding is the correct default here:

- HT forces left-padding only in `generate()` (`transformer_lens/HookedTransformer.py:2007-2009`,
  "Force left-padding for batched generation"), not in its forward. Left-padding is a
  generation requirement, it puts the last real token at index `-1` for every row, and the
  bridge's `generate` should keep doing it. Extending it to `forward` is the deviation.
- The stated rationale does not hold for a causal forward. The code comment says left-padding
  is forced "so pad tokens don't contaminate attention or position embeddings", but with causal
  attention right-padding cannot contaminate anything: pads sit after the real tokens, so no
  real position can attend to them. Measured on gpt2 with a ragged batch, right-padded input
  reproduces the single-sequence logits for the real tokens exactly (0.0) both with and
  without an attention mask. Left-padding is what creates the need for mask and `position_ids`
  correction, because pads then precede the real tokens and shift every position.
- Left-padding is currently unavoidable and undiscoverable. Because the gate is
  `if _is_batched_list and padding_side is None`, setting `tokenizer.padding_side = "right"` is
  overridden (still left-pads), while passing `padding_side="right"` explicitly falls into the
  other branch, where defect 1 discards it. Neither route reaches right-padding, and
  `to_tokens` reports a layout the forward did not use.

Recommended shape of the fix: honor `padding_side` in `to_tokens`, drop the forced left-padding
from the batched-list forward so it follows the tokenizer/argument like every other path, and keep
the explicit left-padding in `generate`. Retain the auto-computed `attention_mask`/`position_ids` as they remain necessary whenever a caller does choose left-padding.

Acceptance:
- [ ] `to_tokens(padding_side="left"/"right")` produces different, correct tensors
- [ ] `bridge.to_tokens(list)` matches the token layout the string-list forward actually feeds the model
- [ ] Compatibility-mode logits for a ragged string list match `HookedTransformer` (or, if left-padding is kept, the deviation is documented in the migration guide's numerics section)
- [ ] `generate` still left-pads batched input (its correctness depends on it) — do not change that path
- [ ] Equal-length lists and single strings remain unaffected (currently exact)
- [ ] Setting `tokenizer.padding_side` is respected by the batched-list forward rather than overridden
- [ ] Test under `tests/unit/model_bridge/` covering ragged batches for both padding sides, asserting real-token activations match the single-sequence run
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

### Checklist

original4422:
# Description

Ragged string lists took two inconsistent tokenization paths in `TransformerBridge`:

- `to_tokens(..., padding_side=...)` resolved the requested side but never passed it to the tokenizer.
- ordinary string-list forward silently forced left padding, so its internal layout could differ from the public `to_tokens` result and from `HookedTransformer`'s default right-padding behavior.

This change passes the resolved side through tokenization and side-dependent BOS/mask utilities without mutating the shared tokenizer, and lets ordinary batched forward follow the explicit argument or tokenizer setting. Left padding, encoder-decoder models, and non-causal models still receive an automatic attention mask; causal right padding remains unmasked so the full logits match `HookedTransformer` and the equivalent pre-tokenized call. The independent forced-left-padding paths in `generate()` and `generate_stream()` are unchanged.

Fixes #1690

## Type of change

- [ ] Breaking change (fix or feature which would cause existing functionality not to work as expected)

### Screenshots

N/A — behavioral code fix with regression tests.

## Validation

- Initial regression before the fix: 2 failed / 2 passed, covering the ignored explicit side and public-vs-forward layout mismatch.
- CI follow-up regression: the legacy batched cache acceptance assertions failed under right padding; they now locate each row's last real token from its attention mask.
- Full-tensor compatibility regression: RED with a 3.0087 max difference between text-list and pre-tokenized forwarding; GREEN at exactly 0.0. The independent `HookedTransformer` comparison is also exactly 0.0.
- Focused tokenization, utility, cache, left-padding, mask, position-id, and generation selection: 77 passed.
- Initial `make unit-test`: 5,058 passed, 36 skipped, 44 deselected, 8 existing xfailed, 0 failed.
- Follow-up unit selection: 5,057 passed with 10 SoLU setup errors caused solely by offline mode; those exact 10 pre-existing tests passed after disabling offline mode.
- `make check-format`: passed.
- `uv run mypy .`: no issues in 388 source files.
- `git diff --check`: passed.

# Checklist:

- [x] I have made corresponding changes to the documentation (no documentation change required; existing API docs already describe the intended behavior)

### DIFF
--- a/tests/acceptance/model_bridge/test_run_with_cache_batch.py
+++ b/tests/acceptance/model_bridge/test_run_with_cache_batch.py
@@ -7,6 +7,19 @@ tests guard against that regression.
 
 import torch
 
+from transformer_lens.utilities import get_attention_mask
+
+
+def _last_real_token_positions(model, prompts: list[str]) -> torch.Tensor:
+    tokens = model.to_tokens(prompts)
+    attention_mask = get_attention_mask(
+        model.tokenizer,
+        tokens,
+        prepend_bos=getattr(model.cfg, "default_prepend_bos", True),
+    )
+    positions = torch.arange(tokens.shape[1], device=tokens.device).expand_as(tokens)
+    return positions.masked_fill(attention_mask == 0, -1).max(dim=1).values
+
 
 def test_run_with_cache_batch_matches_individual(gpt2_bridge):
     """Batched run_with_cache logits at the last real token should match per-prompt runs."""
@@ -23,9 +36,9 @@ def test_run_with_cache_batch_matches_individual(gpt2_bridge):
 
     # Batched run
     batched_logits, _ = gpt2_bridge.run_with_cache(prompts)
-    # With left-padding forced internally, position -1 is the last real token
-    for i in range(len(prompts)):
-        batched_last = batched_logits[i, -1, :]
+    last_real_positions = _last_real_token_positions(gpt2_bridge, prompts)
+    for i, position in enumerate(last_real_positions):
+        batched_last = batched_logits[i, position, :]
         assert torch.allclose(
             individual_logits[i], batched_last, atol=1e-4
         ), f"Prompt {i} logit mismatch between individual and batched run_with_cache"
@@ -54,11 +67,11 @@ def test_run_with_hooks_batch_matches_individual(gpt2_bridge):
 
     # Batched run
     captured_batched = []
+    last_real_positions = _last_real_token_positions(gpt2_bridge, prompts)
 
     def capture_batched(tensor, hook):
-        # For left-padded batch, last real token is at position -1 for all
-        for i in range(tensor.shape[0]):
-            captured_batched.append(tensor[i, -1, :].detach().clone())
+        for i, position in enumerate(last_real_positions):
+            captured_batched.append(tensor[i, position, :].detach().clone())
 
     gpt2_bridge.run_with_hooks(
         prompts,
--- a/tests/unit/model_bridge/test_batched_string_padding.py
+++ b/tests/unit/model_bridge/test_batched_string_padding.py
@@ -1,0 +1,115 @@
+"""Padding behavior for ragged string lists passed to TransformerBridge."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+import pytest
+import torch
+
+from transformer_lens.utilities import get_attention_mask
+
+PROMPTS = ["The quick brown fox", "Hello there, world! This is a longer sentence."]
+RESID_PRE = "blocks.0.hook_resid_pre"
+
+
+@pytest.fixture(scope="module")
+def compatibility_bridge():
+    """A cached model used for token-layout and real-token numerical checks."""
+    from transformer_lens.model_bridge import TransformerBridge
+
+    bridge = TransformerBridge.boot_transformers("distilgpt2", device="cpu", dtype=torch.float32)
+    bridge.enable_compatibility_mode()
+    bridge.eval()
+    return bridge
+
+
+def test_to_tokens_honors_explicit_padding_side(compatibility_bridge) -> None:
+    """A per-call padding override must win without mutating the shared tokenizer."""
+    tokenizer = compatibility_bridge.tokenizer
+    original_side = tokenizer.padding_side
+    tokenizer.padding_side = "right"
+    try:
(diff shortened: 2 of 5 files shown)


tandede:
I reproduced the four failing checks. The failures are deterministic: `tests/acceptance/model_bridge/test_run_with_cache_batch.py` still indexes `-1` after the default batched layout changes to right padding, so the short row is read at a pad position in both the cache and hook tests.

I prepared a complete patch in tandede/TransformerLens@c797856f02a95b75f6afdc6c20dbe4fc6c28eb52. In addition to updating those acceptance tests to index each row at its actual last-token position, it passes the resolved `padding_side` directly to the tokenizer call (supported by the current `transformers>=5.9.0` requirement) rather than temporarily mutating shared tokenizer state. The same resolved side is passed to BOS removal and attention-mask construction. Causal right-padded forward remains unmasked to preserve full HookedTransformer compatibility, while left-padded, encoder-decoder, and non-causal inputs still receive the required mask.

On the issue reproduction with GPT-2, the string-list max logit difference changes from `22.489370346069336` to `0.0`; pre-tokenized parity remains `0.0`. Validation completed locally: all 10 focused regressions passed, the full pre-existing unit suite passed (`5054 passed`), the full acceptance suite passed (`209 passed`), `make check-format` passed, and `uv run mypy .` reported no issues in 388 source files.

Please feel free to cherry-pick `c797856f` if this approach is useful.

jlarson4:
Looks great @original4422! I will merge as is. Thanks for the additional input @tandede.

PR #1708, opened 2026-08-20
Fix OLMo 2 attention input fork state

### LINKED ISSUE #1687: [Bug Report] OLMo 2: use_split_qkv_input / use_attn_in corrupt the forward pass, non-deterministically, across repeated runs

Excelius-Wang:
# Description

OLMo 2 uses post-norm blocks, so `ln1` runs after attention. `BlockBridge` was wiring every `ln1` as the pre-attention capture for `use_split_qkv_input` and `use_attn_in`. The captured post-attention tensor then leaked into the next forward and replaced the current residual input, causing logits and activations to drift across identical runs.

Skip the pre-`ln1` capture for blocks already marked as reading the residual directly, and add regression coverage for both attention input fork modes. The tests verify that the fork input matches `hook_resid_pre`, outputs remain baseline-equivalent, and repeated forwards are bitwise deterministic.

Fixes #1687

## Type of change

## Testing

- `pytest tests/integration/model_bridge/test_olmo2_hook_semantics.py -q` — 9 passed
- `pytest tests/unit/model_bridge/supported_architectures/test_olmo2_adapter.py tests/unit/model_bridge/supported_architectures/test_olmo3_adapter.py tests/unit/model_bridge/supported_architectures/test_olmo_adapter.py tests/unit/model_bridge/supported_architectures/test_olmo_hybrid_adapter.py tests/unit/model_bridge/supported_architectures/test_flex_olmo_adapter.py -q` — 62 passed
- `pytest tests/unit/model_bridge/compatibility/test_split_qkv.py -q` — 7 passed
- `make unit-test` — 5029 passed, 35 skipped, 44 deselected, 8 xfailed
- `uv run mypy .`
- `make check-format`

# Checklist:

### DIFF
--- a/tests/integration/model_bridge/test_olmo2_hook_semantics.py
+++ b/tests/integration/model_bridge/test_olmo2_hook_semantics.py
@@ -191,3 +191,40 @@ def test_compatibility_mode_preserves_residual_semantics() -> None:
             cache[f"blocks.{layer}.hook_resid_post"],
             cache[f"blocks.{layer}.hook_resid_mid"] + cache[f"blocks.{layer}.hook_mlp_out"],
         )
+
+
+@pytest.mark.parametrize("use_split_qkv", [True, False], ids=["split_qkv", "attn_in"])
+def test_attention_input_forks_are_stateless(use_split_qkv: bool) -> None:
+    """OLMo 2 has no pre-attention norm, so input forks must read resid_pre
+    directly and must not retain the preceding forward's post-norm output.
+    """
+    bridge = TransformerBridge.boot_transformers(MODEL, device="cpu", dtype=torch.float32)
+    tokens = torch.tensor([[1, 2, 3, 4, 5]])
+
+    with torch.no_grad():
+        expected = bridge(tokens).clone()
+
+    if use_split_qkv:
+        bridge.set_use_split_qkv_input(True)
+        hook_name = "blocks.1.attn.hook_q_input"
+    else:
+        bridge.set_use_attn_in(True)
+        hook_name = "blocks.1.attn.hook_attn_in"
+
+    try:
+        with torch.no_grad():
+            first_logits, first_cache = bridge.run_with_cache(tokens)
+            second_logits, second_cache = bridge.run_with_cache(tokens)
+    finally:
+        if use_split_qkv:
+            bridge.set_use_split_qkv_input(False)
+        else:
+            bridge.set_use_attn_in(False)
+
+    for cache in (first_cache, second_cache):
+        expected_input = cache["blocks.1.hook_resid_pre"].unsqueeze(2)
+        torch.testing.assert_close(cache[hook_name], expected_input.expand_as(cache[hook_name]))
+
+    torch.testing.assert_close(first_logits, expected, atol=1e-4, rtol=1e-5)
+    torch.testing.assert_close(second_logits, expected, atol=1e-4, rtol=1e-5)
+    torch.testing.assert_close(second_logits, first_logits, atol=0, rtol=0)
--- a/transformer_lens/model_bridge/generalized_components/block.py
+++ b/transformer_lens/model_bridge/generalized_components/block.py
@@ -135,6 +135,8 @@ class BlockBridge(GeneralizedComponent):
         attn = self.submodules.get("attn") if self.submodules else None
         if (
             ln1 is not None
+            # Post-norm blocks reuse ln1 for the attention output, not its input.
+            and not self.mlp_reads_resid_directly
             and isinstance(attn, AttentionBridge)
             and getattr(attn, "supports_split_qkv_fork", False)
             and getattr(ln1, "original_component", None) is not None


koriyoshi2041 (commented):
#1699 already merged the regression fix for #1687. Its block-entry clear and `finally` cleanup make the attention fork fall back to the current `resid_pre` before attention runs, while also bounding the lifetime of the transient capture for every architecture.

This alternative is therefore redundant after `a5894f5`, and `mlp_reads_resid_directly` is a less direct proxy for whether `ln1` is pre- or post-attention. The added OLMo 2 assertions are useful, but the implementation should not land on top of the merged lifetime fix. I would close this unless the tests demonstrate a failure on current `main`.

Excelius-Wang:
Thanks for catching this. My branch was based on c444ae1, before #1699 merged as a5894f5.

I agree that the implementation is now redundant, and that mlp_reads_resid_directly is not the right architectural signal for determining whether ln1 is pre- or post-attention. #1699 already covers both attention-input fork modes and the cross-forward state lifetime.

I’ll verify the additional OLMo 2 assertions against the current dev branch without this implementation change. If they pass, I’ll close this PR; if they expose any remaining gap, I’ll propose the coverage separately as a tests-only change.

Excelius-Wang:
Verified against current dev at 5b23538 with only the additional OLMo 2 assertions applied and no implementation changes. Both split-QKV and attn-in cases pass, including the resid_pre input, baseline-equivalence, and repeated-forward determinism checks.

This confirms that #1699 fully covers the regression, so I’m closing this PR as superseded. Thanks again for the review.

jlarson4:
@Excelius-Wang Thank you for taking the time to do the do-diligence work on existing fix, and thank you @koriyoshi2041 for letting him know the fix was already implemented. Apologies for forgetting to close the issue earlier when I merged that PR.

PR #1709, opened 2026-08-20
Fix direct TransformerBridge config flag assignment

### LINKED ISSUE #1689: [Bug Report] Assigning cfg.use_attn_result / use_hook_mlp_in / use_split_qkv_input directly is silently ignored
**Describe the bug**

Setting an attention/MLP fork flag directly on the config `bridge.cfg.use_hook_mlp_in = True` is accepted, but has no effect. The corresponding hook never fires and caches nothing. Only the `set_use_*` setters work because they additionally propagate the flag to the block components, which hold their own config object rather than sharing `bridge.cfg`.

Measured on gpt2 (compatibility mode):

```
bridge.cfg.use_hook_mlp_in = True   -> run_with_cache(names_filter=["blocks.0.hook_mlp_in"]) == 0 entries
bridge.set_use_hook_mlp_in(True)    -> run_with_cache(names_filter=["blocks.0.hook_mlp_in"]) == 1 entry
```

Direct config assignment is a natural thing to try (it is how the flags read in `cfg`), and on `HookedTransformer` the block reads the shared config so it works there.

**Code example**

```python

bridge.cfg.use_hook_mlp_in = True                       # accepted, no error
_, cache = bridge.run_with_cache("hello", names_filter=["blocks.0.hook_mlp_in"])
print(len(cache))                                       # 0   <- silently not enabled

bridge.set_use_hook_mlp_in(True)                        # supported route
_, cache = bridge.run_with_cache("hello", names_filter=["blocks.0.hook_mlp_in"])
print(len(cache))                                       # 1
```

**System Info**

Installed from source; macOS arm64, CPU, fp32, Python 3.12, transformers 5.13.0. **Reproduces identically on `dev`, measured on with byte-identical results.

Ideally make the assignment work, alternatively make it impossible to get wrong silently:

- Preferred: have the blocks read the flag from the bridge's config rather than a private copy, so assignment and setter agree. The setters (`set_use_split_qkv_input` at `transformer_lens/model_bridge/transformer_bridge.py:3608`, `set_use_attn_in` at `:3623`, `set_use_hook_mlp_in` at `:3641`) each call `_propagate_attention_flag` precisely because the copies diverge — that propagation step is the workaround for the underlying split.
- Alternative (smaller): make these fields properties on the bridge config whose setter performs the propagation, so `cfg.use_hook_mlp_in = True` routes through the same path as the setter.

Whichever route, the migration guide's wording needs to match: it currently suggests direct `cfg` assignment is honored "when blocks share the bridge's `cfg`", a condition that is never true on a constructed bridge. That clause should be removed or corrected.

Acceptance:
- [ ] `bridge.cfg.use_hook_mlp_in = True` either enables the hook or raises/warns; it must not be a silent no-op
- [ ] Same for `use_attn_result`, `use_attn_in`, `use_split_qkv_input`
- [ ] Mutual-exclusivity and unsupported-architecture errors from the setters still fire on the assignment path
- [ ] The migration guide no longer implies direct assignment works
- [ ] Test under `tests/unit/model_bridge/` asserting assignment and setter produce identical hook behaviour
- [ ] `make unit-test` passes
- [ ] `uv run mypy .` passes

**Additional context**

### Checklist

emerardd:
# Description

Fixes #1689.

Direct assignment to `bridge.cfg.use_attn_result`, `use_attn_in`, `use_hook_mlp_in`, or `use_split_qkv_input` previously updated only the top-level config and silently left the constructed block components unchanged.

This change:

- binds a live `TransformerBridgeConfig` to its bridge through a weak reference;
- routes direct assignments for the four managed flags through the existing public setters, preserving propagation, mutual-exclusivity checks, and unsupported-architecture validation;
- keeps standalone/deep-copied configs unbound and rebinds deep-copied or deserialized bridges to their own config;
- updates the migration guide to document direct-assignment parity;
- adds regression coverage for hook behavior, validation, and deepcopy isolation.

## Type of change

# Checklist:

## Testing

- Complete unit-test suite passes locally.
- Direct-assignment regression tests pass for all four managed flags.
- Formatting and `uv run mypy .` pass.


koriyoshi2041 (approved):
Verified at `f10b46ea41`. Direct assignments for all four managed flags take the same setter path without recursion, and deepcopy leaves standalone configs unbound while rebinding copied bridges. The focused regression passes locally (10/10 on Python 3.13); the CI-only left-padding failure also passes twice at this head and is outside the changed surface.

transformer_lens/model_bridge/bridge.py
```diff
@@ -4966,7 +4972,7 @@ def set_use_split_qkv_input(self, use_split_qkv_input: bool):
                     "Call set_use_attn_in(False) before enabling use_split_qkv_input."
                 )
             self._validate_attention_fork_supported("use_split_qkv_input")
-        self.cfg.use_split_qkv_input = use_split_qkv_input
+        self.cfg._set_bridge_managed_hook_flag("use_split_qkv_input", use_split_qkv_input)
```
jlarson4, line 4975: This setter's propagation step (`_propagate_attention_flag`, line 5034) still writes cloned configs with a plain `setattr` under `except Exception: pass`. That `setattr` is no longer a pure field write, if the target config is itself bound, it re-enters *that* bridge's `set_*`, which can raise `ValueError`/`NotImplementedError`, and the bare `except` swallows it. With `b2 = copy.deepcopy(b1)` and `use_attn_in` set on `b1`, `b2.set_use_split_qkv_input(True)` cached `blocks.0.attn.hook_q_input` on `dev` and now caches nothing, while `b2.cfg.use_split_qkv_input` still reports `True`.

Can we make propagation a pure data write so it can never dispatch? It'll need to be done at line 5034 and in `block_cfg.use_hook_mlp_in` at line 5009?

transformer_lens/model_bridge/bridge.py
```diff
@@ -241,6 +241,12 @@ def __init__(self, model: nn.Module, adapter: ArchitectureAdapter, tokenizer: An
         # train() recurses, so this stamps the wrappers with the model's mode.
         original_model.train(original_model.training)
         self.train(original_model.training)
+        self.cfg._bind_bridge(self)
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        """Restore runtime config routing after deepcopy or deserialization."""
+        super().__setstate__(state)
+        self.cfg._bind_bridge(self)
```
jlarson4, line 249: `__setstate__` always calls `self.cfg._bind_bridge(self)`, and `_bind_bridge` overwrites the single `_bridge_ref`. `copy.copy(bridge)` shares the same `cfg`, so it re-points the original's binding at the copy. Once the copy is collected the weakref is dead and direct assignment on the original silently degrades to a plain field write.

Let's only claim an unbound config — bind when `_bridge_ref` is absent, `None`, or dead. Deep-copied configs arrive with `_bridge_ref = None`, so `test_deepcopied_bridge_rebinds_its_config` still passes.

tests/unit/model_bridge/test_config_flag_assignment.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.parametrize(
+    ("flag_name", "hook_name"),
+    [
+        ("use_attn_result", "blocks.0.attn.hook_result"),
+        ("use_attn_in", "blocks.0.attn.hook_attn_in"),
+        ("use_hook_mlp_in", "blocks.0.hook_mlp_in"),
+        ("use_split_qkv_input", "blocks.0.attn.hook_q_input"),
+    ],
+)
+def test_direct_assignment_matches_setter_hook_behavior(flag_name: str, hook_name: str) -> None:
```
jlarson4: The fixture is a 1-layer GPT-2 whose `block.attn.config` is `bridge.cfg`, so `use_attn_result`, `use_attn_in`, and `use_split_qkv_input` already worked by direct assignment before this change. With the production files reverted to `dev`, those three parametrizations still pass; only `use_hook_mlp_in` fails. The architectures the fix actually repairs for those flags have no coverage.

Can we add a cloned-config fixture (a locally-built tiny Llama or Mistral, no download needed) and parametrize the test over both fixtures, so the three attention flags are exercised properly?

jlarson4 (commented):
Thanks for putting this together, @emerardd, routing assignment through the setters via a weakref, and keeping copies unbound with `__getstate__`/`__setstate__` is a clean way to solve this issue. A couple small comments, let me know if you have any questions.

transformer_lens/model_bridge/bridge.py
emerardd: Addressed in fe789df4. Propagation now goes through a dedicated pure-data helper: TransformerBridgeConfig instances use _set_bridge_managed_hook_flag, and other config objects use object.__setattr__, so propagation cannot dispatch through another live Bridge. I applied this to both the attention flags and use_hook_mlp_in, and narrowed the swallowed exceptions to AttributeError and TypeError. I also added regression tests that deliberately bind cloned attention and MLP configs to another Bridge and verify that the other Bridge is not mutated.

transformer_lens/model_bridge/bridge.py
emerardd: Addressed in fe789df4. _bind_bridge now claims the config only when _bridge_ref is absent, None, or dead. I added a shallow-copy regression test that confirms copy.copy shares the config without replacing the original live binding, then deletes and collects the copy before verifying that direct assignment still dispatches through the original Bridge. The existing deepcopy rebinding test remains in place and passes.

tests/unit/model_bridge/test_config_flag_assignment.py
emerardd: Addressed in fe789df4. I added a locally constructed tiny LlamaForCausalLM fixture with no model download and parametrized the four direct-assignment cases over both tiny GPT-2 and tiny Llama. The fixture explicitly verifies that GPT-2 uses the shared attention config path while Llama uses a cloned attention config, so all three attention flags now exercise the repaired propagation path. The updated test file passes 17 tests, and the broader no-download related surface passes 34 tests.

transformer_lens/config/transformer_bridge_config.py
```diff
@@ -204,9 +217,39 @@ def __init__(
...
+        return self.__dict__.copy()
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        """Restore an unbound config copy."""
+        self.__dict__.update(state)
+        object.__setattr__(self, "_bridge_ref", None)
+
+    def _bind_bridge(self, bridge: Any) -> None:
+        """Bind runtime hook-flag assignments to a constructed Bridge."""
+        bridge_ref = getattr(self, "_bridge_ref", None)
```
jlarson4, line 244: This guard fixes the shallow-copy case, but `_bind_bridge` is also called from `TransformerBridge.__init__`. When two bridges are built from one shared config object the second never binds, so `b2.cfg.use_attn_in = True` configures **b1**. `b2`'s hook stays off and an unrelated live bridge is reconfigured unintentionally. Reachable through the public `TransformerBridge(model, adapter)` constructor when a config is reused (`boot_native` / `build_bridge_from_module` deep-copy, so they are unaffected).

Can we add a warning that is displayed when declining to bind because another *live* bridge already owns the config?

transformer_lens/model_bridge/bridge.py
```diff
@@ -5025,8 +5039,8 @@ def _propagate_attention_flag(self, flag_name: str, value: bool) -> None:
             attn_cfg = getattr(attn, "config", None)
             if attn_cfg is not None and attn_cfg is not self.cfg:
                 try:
-                    setattr(attn_cfg, flag_name, value)
-                except Exception:
+                    self._write_propagated_hook_flag(attn_cfg, flag_name, value)
+                except (AttributeError, TypeError):
```
jlarson4, line 5043: The comment says frozen/immutable configs are skipped and "won't honor the flag", but `_write_propagated_hook_flag` writes non-`TransformerBridgeConfig` configs with `object.__setattr__`, which bypasses frozen dataclasses, meaning a frozen config is written.

Can we update the comment to reflect the new state?

jlarson4 (commented):
Thanks for addressing these updates @emerardd! one small comment related to config reuse, and one documentation request, and then this should be good to go.

transformer_lens/config/transformer_bridge_config.py
emerardd: Addressed in 5d066b73. _bind_bridge now warns when a different live TransformerBridge already owns the config, while preserving idempotent binding to the same Bridge and rebinding when the reference is absent or dead. The warning explains that managed-flag assignments will continue to configure the existing Bridge and uses stacklevel=3 so the public-constructor call site is surfaced. I added a regression through TransformerBridge(model, adapter, tokenizer=None) using two independent local NativeModel instances and adapters that share one config, and updated the shallow-copy test to expect the same warning. The focused file passes 18 tests; the broader no-download related surface passes 35 tests, and mypy passes all 388 source files.

transformer_lens/model_bridge/bridge.py
emerardd: Updated in 5d066b73. The comment no longer claims that frozen or immutable configs are skipped. It now documents the actual exception case: some config-like objects may reject the attribute even when their custom __setattr__ is bypassed. This is a comment-only correction; the pure-data propagation behavior is unchanged.

jlarson4:
Thanks for the quick turn around @emerardd! Assuming CI passes, this is approved and will be merged!

PR #1715, opened 2026-08-21
Preserve input dtype in GeneralizedComponent

### LINKED ISSUE #1712: [Bug Report] `GeneralizedComponent.forward` casts activations to the first floating parameter dtype
## Describe the bug

`GeneralizedComponent.forward` infers a target compute dtype from the first floating-point parameter returned by the wrapped module and casts the component input to that dtype.

A parameter's storage dtype is not necessarily the module's activation or compute dtype. Quantized modules can store weights or scales in FP8 formats, and mixed-precision Transformers modules can deliberately retain a small subset of parameters in FP32. As a result, the current rule can cast activations to FP32, an FP8 weight dtype, or even an exponent-only scale dtype.

On `deepseek-ai/DeepSeek-V4-Flash-0731`, a model loaded directly through Transformers completes and returns BF16 logits. Passing that preloaded model to `TransformerBridge.boot_transformers` causes a forward pass to fail with:

```text
RuntimeError: Promotion for Float8 Types is not supported, attempted to promote BFloat16 and Float8_e4m3fn
```

## Code example

```python
import torch

hf_id = "deepseek-ai/DeepSeek-V4-Flash-0731"
tokens = torch.tensor([[671, 6102, 294, 8760, 344]], device="cuda")

    hf_id,
    dtype=torch.bfloat16,
    device_map="cuda",
    attn_implementation="eager",
)

# Run the direct Transformers control before wrapping the model.
with torch.no_grad():
    hf_output = hf_model(tokens)
print(hf_output.logits.shape, hf_output.logits.dtype)
# torch.Size([1, 5, 129280]) torch.bfloat16

# Passing a preloaded model avoids TransformerLens's separate post-load
# cast_floating_params_to_dtype path and isolates GeneralizedComponent.
    hf_id,
    dtype=torch.bfloat16,
    hf_model=hf_model,
)

with torch.no_grad():
    model(tokens)

# RuntimeError: Promotion for Float8 Types is not supported,
# attempted to promote BFloat16 and Float8_e4m3fn
```

## Expected behavior

Wrapping a Hugging Face component for hooks should preserve the dtypes that the original model passes across that component boundary, unless a specific adapter explicitly documents and implements a required conversion. A generic hook wrapper should not infer a compute dtype from arbitrary parameter storage.

## Root cause

`GeneralizedComponent.forward` currently contains logic equivalent to:

```python
target_dtype = None
for p in original_component.parameters():
    if not p.dtype.is_floating_point:
        continue
    target_dtype = p.dtype
    break

if target_dtype is not None and input.dtype != target_dtype:
    input = input.to(target_dtype)
```

The accompanying assumption is that quantized weights are stored in integer dtypes. That does not hold for FP8/MX quantization, where weights and scales may use floating dtypes such as `torch.float8_e4m3fn` and `torch.float8_e8m0fnu`. It also does not hold for ordinary mixed-precision modules that keep selected parameters in FP32.

The DeepSeek-V4 adapter wraps `self_attn`, `o_a_proj`, and `experts` with `GeneralizedComponent`, so the generic rule is applied at all three boundaries.

## What the wrapper selects

Instrumenting the loop in layer 0 produced:

| component | wrapped module | first floating parameter | cast applied to activations |
| --- | --- | --- | --- |
| `self_attn` | `DeepseekV4Attention` | `sinks` - `torch.float32` | BF16 -> **FP32** |
| `o_a_proj` | `FP8GroupedLinear` | `weight` - `torch.float8_e4m3fn` | FP32 -> **FP8** |
| `experts` | `FP8Experts` | `gate_up_proj_scale_inv` - `torch.float8_e8m0fnu` | BF16 -> **FP8 scale dtype** |

`DeepseekV4Attention.sinks` is deliberately retained in FP32, but that does not mean the whole attention module expects FP32 activations. The generic wrapper nevertheless casts the attention input to FP32.

The nested `o_a_proj` wrapper then finds its FP8 weight and casts its input to `float8_e4m3fn`. The grouped projection preserves that input dtype in its output; `o_b_proj` passes the tensor through without undoing the cast. The decoder layer's mHC mix subsequently combines the FP8 attention output with BF16 tensors and raises the reported promotion error.

The `experts` result illustrates the same underlying problem even though execution fails earlier: `float8_e8m0fnu` is a quantization-scale storage format, not a valid activation-compute choice for the expert boundary.

## Tested alternatives

Using the same load and five-token prompt:

| rule for `target_dtype` | result |
| --- | --- |
| current rule: first floating parameter | `RuntimeError: Promotion for Float8 Types is not supported` |
| skip one-byte floats, but keep other inferred casts | `RuntimeError: expected scalar type BFloat16 but found Float` |
| no generic cast | forward succeeds; logits are `(1, 5, 129280)` / `bfloat16` |

Skipping only FP8 parameters is therefore insufficient: the FP32 cast selected from `sinks` still changes the native module boundary and reaches a kernel that rejects it.

## Suggested fix

Remove dtype inference and coercion from the generic `GeneralizedComponent` boundary. After applying the input hook, pass the resulting tensor directly to `original_component` in the dtype supplied by the original Hugging Face call path.

Using `self.config.dtype` would be safer than using an arbitrary parameter, but it would still impose one global dtype on modules whose native execution is intentionally mixed precision. In this reproduction, preserving the original input dtype is the only tested rule that succeeds.

If a particular model adapter or component genuinely requires an input conversion, that conversion should be explicit and local to that adapter/component rather than inferred from recursive parameter order.

Useful regression tests would verify that:

1. Wrapping a component with an FP32 auxiliary parameter does not change a BF16 input.
2. Wrapping a component with an FP8 weight or scale does not cast its input to FP8.
3. A no-op `GeneralizedComponent` wrapper is dtype- and output-equivalent to the unwrapped Hugging Face component when hooks do not modify values.

## System info

- `transformer_lens==3.7.2`
- `transformers==5.15.0`
- `torch==2.13.0+cu130`
- `kernels==0.16.0`
- `accelerate==1.14.0`
- NVIDIA B200 (183 GiB), single device

## Checklist

- [x] I searched the existing TransformerLens issues and did not find this exact failure.
- [x] I included a direct Transformers control and isolated the wrapper path by passing a preloaded model.
- [x] I tested the current rule, a narrow-FP8 exclusion, and removal of the generic cast.

koriyoshi2041:
## Description

`GeneralizedComponent.forward` currently infers a compute dtype from the first floating-point parameter in the wrapped module. That parameter can be an FP32 auxiliary value or FP8 quantization storage, so the wrapper can change a valid mixed-precision activation before the original component sees it.

This removes the generic parameter-based cast and preserves the dtype produced by `hook_in`. Model-specific conversions can remain explicit in their adapters. The regression covers both positional and keyword input paths with a BF16 activation and an FP32 auxiliary parameter.

Fixes #1712

## Type of change

## Checklist

- [x] The generic hook wrapper preserves the original component boundary dtype
- [x] Tests cover positional and keyword input paths
- [x] No public API or documentation changes are required

## Validation

- `uv run pytest tests/unit/model_bridge/generalized_components/test_base.py -q` — 29 passed
- `uv run black --check transformer_lens/model_bridge/generalized_components/base.py tests/unit/model_bridge/generalized_components/test_base.py`
- `uv run isort --check-only transformer_lens/model_bridge/generalized_components/base.py tests/unit/model_bridge/generalized_components/test_base.py`
- `uv run mypy transformer_lens/model_bridge/generalized_components/base.py`
- `git diff --check`


tests/unit/model_bridge/generalized_components/test_base.py
```diff
@@ -377,5 +377,31 @@ def test_complex_object_attributes(self):
         assert component.complex_attr["nested"]["deep"] == [1, 2, 3]
 
 
+@pytest.mark.parametrize("use_keyword", [False, True])
```
jlarson4, line 386: The test pins the FP32-auxiliary case, but the failure reported in the issue was `Promotion for Float8 Types is not supported`, from an FP8 weight and an FP8 scale. Nothing here would catch a future change that re-introduced inference for FP8 specifically.

Let's add a parametrization with a `torch.float8_e4m3fn` parameter asserting the BF16 input is unchanged.

jlarson4 (commented):
@koriyoshi2041 Thanks for tackling this issue! The fix looks great, there are just a couple other locations that we will want to also incorporate this. 

Identical logic still runs in:

- `generalized_components/attention.py:800-835` (AttentionBridge.forward, three call sites)
- `generalized_components/position_embeddings_attention.py:355-363` (PositionEmbeddingsAttentionBridge.forward)

A worse variant also still runs in:

- `generalized_components/moe.py:284-307`
- `generalized_components/unembedding.py:89-101`

Can we extend the fix to these locations as well? As a follow up PR, if you're interested, it may be worth while to build out a helper function for handling this specific problem, so it does not need to be mirror fixed across 5 locations in the future.

koriyoshi2041:
Thanks for the pointers — I extended the fix at the current head to all four locations you listed. I kept the helper extraction out of this PR, and added direct mixed-dtype regression coverage for the specialized attention, position-embedding attention, MoE, and unembedding forward paths.

Local verification: 57 focused tests, Black/isort, targeted mypy, and `git diff --check` all pass. Hosted compatibility/format/type/docstring/benchmark/notebook checks are green; full coverage is still running.

jlarson4:
Hi @koriyoshi2041, just pushed a CI update to your branch that should help stabilize Full Code Coverage, there was a flaw in dev that was causing failures. Everything looks good, once this run passes I'll merge, thanks for your patience!

PR #1716, opened 2026-08-21
do not overwrite storage dtypes

### LINKED ISSUE #1713: [Bug Report] `boot_transformers` casts quantizer-owned scale parameters and breaks MXFP4 checkpoints
## Describe the bug

When `TransformerBridge.boot_transformers` loads a quantized Hugging Face model itself, its post-load dtype normalization casts every floating-point parameter to the requested `dtype`. That includes floating-point parameters owned by the quantizer, such as UE8M0 scale tensors.

On `deepseek-ai/DeepSeek-V4-Flash-0731`, loading with `dtype=torch.bfloat16` changes the MXFP4 expert scales from `torch.float8_e8m0fnu` to `torch.bfloat16`. The packed `int8` expert weights are left unchanged, so the weight/scale pair no longer describes the same quantization format. The first MoE layer then fails with:

```text
AssertionError: K mismatch: A has K=4096, B has K=2048
```

The same checkpoint and prompt complete normally when the model is loaded and called directly through Transformers in the same environment.

## Code example

```python
import gc

import torch

hf_id = "deepseek-ai/DeepSeek-V4-Flash-0731"
tokens = torch.tensor([[671, 6102, 294, 8760, 344]], device="cuda")

# Control: the Transformers model completes normally.
    hf_id,
    dtype=torch.bfloat16,
    device_map="cuda",
    attn_implementation="eager",
)
with torch.no_grad():
    hf_output = hf_model(tokens)
print(hf_output.logits.shape, hf_output.logits.dtype)

del hf_output, hf_model
gc.collect()
torch.cuda.empty_cache()

# TransformerLens's own load path mutates the scale dtypes and then fails.
    hf_id,
    dtype=torch.bfloat16,
    device_map="cuda",
)
with torch.no_grad():
    model(tokens)

# AssertionError: K mismatch: A has K=4096, B has K=2048
```

Using `device="cuda"` instead of `device_map="cuda"` produces the same failure.

## Expected behavior

The requested compute dtype should not overwrite storage dtypes selected by an active quantizer. Loading a supported Hugging Face checkpoint through `TransformerBridge.boot_transformers` should preserve quantizer-owned weight/scale pairs and should not make a forward pass fail when the corresponding Transformers model succeeds.

## Root cause

In `transformer_lens/model_bridge/sources/transformers.py`, the branch that loads the Hugging Face model ends with:

```python
cast_floating_params_to_dtype(hf_model, dtype)
```

`transformer_lens/utilities/multi_gpu.py::cast_floating_params_to_dtype` uses floating-point status as its ownership test:

```python
for param in module.parameters(recurse=False):
    if not param.is_floating_point() or param.dtype == dtype:
        continue
    ...
    param.data = param.data.to(dtype=dtype)
```

That condition also matches FP8 weights and scale tensors. In this specific checkpoint, the relevant MXFP4 expert weights are packed `int8`, so they remain unchanged, while their `float8_e8m0fnu` scale parameters are converted to BF16.

Transformers' fine-grained FP8 integration passes these tensors to the loaded `finegrained-fp8` grouped-matmul implementation. The kernel dispatcher recognizes the MX path only when the packed weight and scale satisfy `is_mxfp`, including the UE8M0 scale-dtype check. Once TransformerLens changes the scale dtype to BF16, the dispatcher falls through to the block-FP8 path. That path expects an unpacked K dimension and raises the observed assertion because the MXFP4 weight stores 4096 codes in 2048 packed bytes.

## Observed parameter changes

At the first expert bank, the tensor shapes are unchanged; only the quantization-scale dtypes differ:

| tensor | direct Transformers load | `boot_transformers(dtype=bfloat16)` |
| --- | --- | --- |
| `gate_up_proj` | `(256, 4096, 2048)` `int8` | `(256, 4096, 2048)` `int8` |
| `gate_up_proj_scale_inv` | `(256, 4096, 128)` `float8_e8m0fnu` | `(256, 4096, 128)` **`bfloat16`** |
| `down_proj` | `(256, 4096, 1024)` `int8` | `(256, 4096, 1024)` `int8` |
| `down_proj_scale_inv` | `(256, 4096, 64)` `float8_e8m0fnu` | `(256, 4096, 64)` **`bfloat16`** |

## Suggested fix

Do not apply a blanket post-load cast to parameters owned by an active Hugging Face quantizer.

The safest behavior would be to preserve the parameter dtypes returned by `from_pretrained` whenever the model has an active quantization configuration/quantizer. A more targeted implementation could skip only modules or parameters claimed by that quantizer while retaining dtype normalization for ordinary, unquantized modules in mixed checkpoints.

Skipping one-byte floating dtypes (`param.dtype.itemsize == 1`, or an explicit FP8 dtype set) would prevent this particular corruption, but it is best treated as a defensive guard rather than the primary ownership test.

A regression test could construct a small module containing a packed integer weight and a floating FP8 scale parameter, run the bridge's dtype-normalization path, and assert that the quantizer-owned scale dtype is preserved. An integration test should also verify that wrapping a quantized Hugging Face model does not change its quantization metadata or parameter dtypes.

## Partial workaround / isolation

Preloading the Hugging Face model and passing it through `hf_model=` bypasses the post-load call to `cast_floating_params_to_dtype`:

```python
    hf_id,
    dtype=torch.bfloat16,
    device_map="cuda",
    attn_implementation="eager",
)
    hf_id,
    dtype=torch.bfloat16,
    hf_model=hf_model,
)
```

This isolates and avoids this particular mutation. On this checkpoint, TransformerLens 3.7.2 then exposes a separate `GeneralizedComponent.forward` dtype-casting bug, so this is not a complete end-to-end workaround.

## System info

- `transformer_lens==3.7.2`
- `transformers==5.15.0`
- `torch==2.13.0+cu130`
- `kernels==0.16.0`
- `accelerate==1.14.0`
- NVIDIA B200 (183 GiB), single device

## Checklist

- [x] I searched the existing TransformerLens issues and did not find this exact failure.
- [x] I included a reproduction, a direct Transformers control, and the observed parameter-dtype diff.

MdSadiqMd:
# Description

When `TransformerBridge.boot_transformers` loads a quantized HuggingFace model, its post-load dtype normalization via `cast_floating_params_to_dtype` casts every floating-point parameter to the requested dtype. This includes floating-point parameters owned by the quantizer, such as FP8 scale tensors (e.g., `float8_e8m0fnu`).

On MXFP4 checkpoints like `deepseek-ai/DeepSeek-V4-Flash-0731`, loading with `dtype=torch.bfloat16` changes the expert scales from `torch.float8_e8m0fnu` to `torch.bfloat16`. The packed int8 expert weights are left unchanged, so the weight/scale pair no longer describes the same quantization format, causing forward pass failures.

**Fix implemented at two levels:**
1. **Call-site skip**: Skip `cast_floating_params_to_dtype` entirely when the model has an active `quantization_config`
2. **Function guard**: Skip one-byte floats (FP8 dtypes) in `cast_floating_params_to_dtype` as a defensive guard

Fixes https://github.com/TransformerLensOrg/TransformerLens/issues/1713

## Type of change

# Checklist:


transformer_lens/model_bridge/sources/transformers.py
```diff
@@ -847,7 +847,12 @@ def boot(
         # Cast params to dtype; preserve float32 buffers (e.g., RotaryEmbedding.inv_freq).
         # Use module-level alignment so Accelerate can temporarily materialize offloaded
         # parameters before we touch them.
-        cast_floating_params_to_dtype(hf_model, dtype)
+        # Skip dtype normalization entirely when model has an active quantizer: the
+        # quantizer owns specific dtypes (e.g., FP8 scales) that must not be overwritten.
+        from transformer_lens.utilities.quantization import quantization_method
+
+        if quantization_method(hf_model.config) is None:
```
jlarson4: There are no tests covering this guard. Can we extract the guard-plus-cast into a small helper (e.g. `maybe_cast_floating_params(hf_model, dtype)`) and add a unit test on a stub module? We will want the stub's `.config` to carry a `quantization_config` and the params should be untouched when quantized, cast when not.

tests/integration/model_bridge/test_bridge_integration.py
```diff
@@ -796,5 +796,68 @@ def test_TransformerBridge_gemma2_forward():
     assert hasattr(bridge.rotary_emb, "hook_sin"), "rotary_emb should have hook_sin"
 
 
+class TestCastFloatingParamsSkipsQuantizerOwnedScales:
```
jlarson4: None of the three tests boots or loads anything. The two `quantization_method` tests duplicate `test_capture_handles_both_config_shapes` / `test_unquantized_config_captures_none` in `tests/unit/test_loading_from_pretrained_utilities.py`, and `test_cast_skips_fp8_scales_even_if_called` duplicates `test_mixed_module_casts_selectively` in the new unit file. The class docstring also says the boot call-site layer is covered, but no test exercises boot.

Lets remove this class, it doesn't add any additional coverage.

tests/unit/utilities/test_multi_gpu_unit.py
```diff
@@ -129,3 +133,74 @@ def test_cpu_device_is_returned_unchanged(self):
...
+
+    def test_skips_non_floating_point_params(self):
+        """Integer params (packed quantized weights) are left untouched."""
+        model = nn.Linear(4, 4, bias=False)
+        model.weight = nn.Parameter(torch.zeros(4, 4, dtype=torch.int8), requires_grad=False)
+        cast_floating_params_to_dtype(model, torch.bfloat16)
+        assert model.weight.dtype == torch.int8
+
+    @pytest.mark.parametrize(
+        "fp8_dtype",
```
jlarson4, line 171: The fp8 cases don't cover `float8_e8m0fnu`, which is the dtype the issue reported and the docstring names.

Can we add it as `pytest.param(torch.float8_e8m0fnu, marks=pytest.mark.skipif(not hasattr(torch, "float8_e8m0fnu"), reason="torch < 2.7"))`?

tests/unit/utilities/test_multi_gpu_unit.py
```diff
@@ -129,3 +133,74 @@ def test_cpu_device_is_returned_unchanged(self):
         cfg = _cuda_cfg(n_layers=62, n_devices=8)
         result = get_device_for_block_index(30, cfg, device="cpu")
         assert result.type == "cpu"
+
+
+class TestCastFloatingParamsToDtype:
+    """Regression tests for cast_floating_params_to_dtype.
+
+    Issue: #<MXFP4> — the function was casting quantizer-owned FP8 scale tensors
```
jlarson4: The docstring reads `Issue: #<MXFP4>`, can we replace that with the real reference, #1713?

jlarson4 (commented):
Hey @MdSadiqMd, thanks for the fix! A couple notes on the test coverage

jlarson4:
@MdSadiqMd I missed this the first time around, could you please rebase this to `dev`?

MdSadiqMd:
sure

jlarson4:
Hi @MdSadiqMd, just pushed a CI update to your branch that should help stabilize Full Code Coverage, there was a flaw in dev that was causing failures. Everything looks good, once this run passes I'll merge, thanks for your patience!

PR #1721, opened 2026-08-24
feat(projection_kernel): add attention-head subspace affinity

### LINKED ISSUE #1720: [Proposal] Projection-kernel affinity for attention-head weight subspaces
### Proposal

Add a TransformerBridge-first tool for measuring overlap between the residual-stream
subspaces read or written by attention-head weights. For orthonormal bases `U` and `V`,

`PK(S, T) = ||U.T @ V||_F^2 = sum_i cos(theta_i)^2`,

where `theta_i` are principal angles. The Projection Kernel (PK) ignores singular-value
magnitude and is invariant to basis changes inside either subspace. It complements
Composition Score (CS): PK measures overlapping support, while CS measures scale-sensitive
composition of full linear maps under an isotropic-input assumption.

The work should separate:

1. a small, maintained numerical and TransformerBridge API; and
2. a versioned research artifact for GPT-2 labels, baselines, plots, reproduction, and
   causal validation.

### Motivation

TransformerLens exposes Q/K/V/O weights and implements Q-, K-, and V-composition, but it
has no principal-angle or subspace-overlap analysis. Several results motivate this gap:

- Elhage et al. introduced weight composition for head-to-head paths.
- Merullo et al. found causally important one- to three-dimensional communication
  channels by decomposing QK and OV matrices.
- Yamagiwa, Takase, and Shimodaira report that PK over individual Q/K/V/O subspaces more
  clearly recovers several known GPT-2-small relationships.
- MAPS shows that static weight scores can predict head outputs and causal ablations, but
  also that heads are multifunctional and context dependent.

The PK paper studies one model, does not directly test whether PK predicts path-specific
causal effects, and explicitly says PK does not identify head function. Work on
representation similarity likewise warns that geometric and functional similarity can
dissociate. TransformerLens can provide the primitive and test its limits with existing
hooks and direct path patching.

#### Stable core API

Names are open to maintainer preference. A minimal model API could be:

```python
from transformer_lens.tools.analysis import attention_head_subspace_affinity

result = attention_head_subspace_affinity(
    bridge, source_role="O", target_role="Q", layer_order="forward"
)

result.scores       # [source_layer, source_head, target_layer, target_head]
result.normalized   # PK / d_head for equal-rank heads
result.valid_mask
result.top_pairs(20)
```

The numerical layer should remain model independent:

- `orthonormal_subspace(matrix, *, rank=None, rtol=None)` extracts a documented
  numerical column-space basis and measured rank;
- `projection_kernel(basis_a, basis_b)` returns raw PK, projector-normalized PK, and
  principal-angle cosines.

Explicit basis extraction avoids hiding rank decisions inside each score. For ranks `r`
and `s`, raw PK lies in `[0, min(r, s)]`. Projector normalization is
`PK / sqrt(r*s)`, the cosine between projection matrices; at equal rank this is the
paper's `PK / d_head`.

The first model wrapper should expose OQ, OK, and OV without assuming equal source and
target head counts.

#### Weight-layout and architecture contract

The paper uses left-multiplication matrices. TransformerLens basis-generating matrices are:

| Role | Matrix | Per-head shape |
|---|---|---|
| Q | `W_Q[layer, head]` | `[d_model, d_head]` |
| K | `W_K[layer, kv_head]` | `[d_model, d_head]` |
| V | `W_V[layer, kv_head]` | `[d_model, d_head]` |
| O | `W_O[layer, head].T` | `[d_model, d_head]` |

The O transpose is essential. Results carry original layer indices and separate
source/target head metadata.

For grouped-query attention, K/V retain native KV-head identity rather than being silently
repeated to query-head labels. OQ has query-head × query-head axes; OK and OV have
query-head × KV-head axes. An expanded view can be added later for comparison with CS,
but duplicated K/V views must not be treated as independent heads.

Hybrid models include only attention layers and preserve original block indices.
Architectures without standard readable Q/K/V/O projections, including MLA and opaque
native-forward adapters, raise a precise unsupported-architecture error.

#### Numerical contract

- Validate floating, finite, two-dimensional matrices with a shared ambient dimension.
- Use reduced SVD for rank measurement and basis extraction; never materialize a
  `[d_model, d_model]` projector.
- Use a documented dtype- and shape-aware rank tolerance; permit caller-pinned rank.
- Clamp overlap singular values only for roundoff before `acos`.
- Preserve device and document accumulation/return dtype for low-precision inputs.
- Precompute each head basis once before batched pair scoring.

For independent random `m`-planes in `R^d`, the paper gives:

- mean `m^2 / d`;
- variance `2*m^2*(d-m)^2 / (d^2*(d-1)*(d+2))`.

Expose these, if useful, only as descriptive moments. The normal approximation is not an
automatic p-value; trained head pairs are dependent and anisotropic.

#### Research reproduction and extension

The research artifact should pin model revision, processing mode, dtype, seed, pair
construction, and annotation data. It should not be imported by the library.

##### Study A — exact reproduction

Using raw GPT-2-small weights and the paper's 36 annotated heads across seven classes,
reproduce:

- 9,504 earlier-to-later pairs for each of OQ, OK, and OV;
- top-20 wiring diagrams and named edges;
- Table 1 PR-AUC against CS, Simple-CS, Linear CKA, and Procrustes;
- 10,296 non-identical layer-ordered QQ/KK/VV/OO pairs and Table 2;
- inlet/outlet rankings, especially L4H7;
- the random-subspace moment simulation; and
- Appendix K's raw-versus-processed comparison.

Table 1 targets are:

| Method | OQ | OK | OV | Mean |
|---|---:|---:|---:|---:|
| PK | 0.446 | 0.451 | 0.289 | 0.395 |
| CS | 0.227 | 0.352 | 0.215 | 0.265 |
| Simple-CS | 0.245 | 0.359 | 0.305 | 0.303 |
| Linear CKA | 0.337 | 0.323 | 0.409 | 0.356 |
| Procrustes | 0.328 | 0.269 | 0.428 | 0.342 |

This is not a blanket “PK wins” target: PK leads OQ and OK, while CKA and Procrustes beat
it on OV. Table 2 reports mean PR-AUC 0.047 and ROC-AUC 0.809 for PK, versus 0.038 and
0.787 for the strongest reported PR-AUC baseline, Procrustes.

##### Study B — causal-validity extension

Test:

> Does static O→Q/K/V subspace overlap predict path-specific causal communication on
> held-out prompts better than scale-sensitive or generic weight-similarity baselines?

Use clean/corrupted paired prompts and direct path patching from source-head output to a
target head's Q, K, or V input in at least two domains:

1. IOI for inhibition/name-mover and duplicate/inhibition relationships;
2. repeated-token induction for previous-token/induction relationships.

For every valid pair, record normalized absolute logit-difference recovery and uncertainty
across prompts. Evaluate Spearman correlation, top-k enrichment, PR-AUC at a preregistered
effect threshold, bootstrap confidence intervals, and paired bootstrap metric differences.

Baselines are PK, CS, Simple-CS, Linear CKA, Procrustes, source-head effect, target-head
effect, layer distance, weight norm, and seeded random ranking. Endpoint effects are
essential controls: otherwise a metric may merely rank globally important heads.

Primary hypotheses:

- **H1:** raw-weight GPT-2 results reproduce within declared numerical/ranking tolerances.
- **H2:** PK adds held-out predictive value beyond endpoint effects and layer distance for
  OQ and OK.
- **H3:** validity differs by role pairing; PK is not expected to dominate OV.
- **H4:** PK rankings are more stable than CS under behavior-preserving TransformerLens
  preprocessing.

Falsifiers include failure to reproduce Table 1, no held-out association after controls,
or equal performance from random norm-matched subspaces. Report negative results rather
than redefining the target.

##### PR 1 — numerical primitive

- basis extraction, PK, principal angles, normalization, and null moments;
- model-free unit tests and API docs;
- no model wrapper, plotting, labels, or baseline bundle.

##### PR 2 — TransformerBridge wrapper

- OQ/OK/OV extraction, native MHA/GQA metadata, hybrid handling, masks, and ranking;
- synthetic tests plus cached-model Bridge integration tests;
- explicit unsupported-layout errors.

##### Research artifact/report

- exact GPT-2 reproduction and causal-validity study;
- baselines local to the artifact unless maintainers request promotion;
- machine-readable results and limitations.

#### Correctness and test plan

1. Known-angle constructions recover exact principal-angle cosines.
2. Cosine, overlap-Frobenius, and projector definitions agree on small fixtures.
3. Scores are invariant to orthogonal rebasing and nonsingular right transforms before
   basis extraction.
4. Identical, nested, orthogonal, unequal-rank, and deficient spaces have explicit values.
5. Seeded Monte Carlo moments match analytic formulas within sampling error.
6. Synthetic attention weights prove Q/K/V and `W_O.T` orientation.
7. MHA, native GQA, hybrid indices, masks, and tie ordering are pinned.
8. Selected Bridge scores match independent Torch calculations.
9. Unit tests load no models; integration tests reuse read-only cached Bridge fixtures.
10. Paper headline values live in the versioned reproduction, not brittle library CI.

#### Risks and interpretation

- **Similarity is not causality:** high PK only proposes a candidate relationship.
- **Full head spaces may be too coarse:** narrow causal channels can be diluted by
  rank-`d_head` PK; truncated variants are follow-ups, not post-hoc fixes.
- **Subspace interventions can mislead:** causal work needs controls for dormant-path
  activation described by Makelov et al.
- **Preprocessing changes the estimand:** raw Bridge weights match the primary paper.
- **Rank is a choice:** tolerance, spectrum, and dtype must be recorded.
- **GQA creates pseudoreplication:** repeated K/V views are not independent evidence.
- **The random null is idealized:** it is not calibrated significance for trained heads.

### Alternatives

- Keep CS: retain it as a baseline; it measures scale-sensitive composition.
- Use decomposed CS: valuable for narrow high-gain channels, but not basis-invariant
  whole-subspace overlap.
- Use CKA/Procrustes: important alternatives that outperform PK in some paper settings.
- Flatten and cosine: depends on arbitrary head-basis choices.
- Use activation similarity: complementary and data conditional, not the same estimand.
- Keep this notebook-only: simpler, but provides no tested cross-architecture primitive.

### Additional context

- Primary paper: Hiroaki Yamagiwa, Yusuke Takase, and Hidetoshi Shimodaira,
  “Measuring Affinity between Attention-Head Weight Subspaces via the Projection Kernel,”
  arXiv:2601.10266v1 (2026): https://arxiv.org/abs/2601.10266.
- https://github.com/ymgw55/proj-kernel-attn-subspace contained only a README and figure
  images, with no source or license, when checked on 2026-08-24.
- Related but non-duplicate TransformerLens work includes merged Composition Score PR #39,
  SVDInterpreter issue #256 / merged PR #311, SVD-semantics issue #341, Bridge analysis
  PR #1237, GQA circuit fix PR #1593, and gated-Qwen analysis fix PR #1653. None implements
  PK, principal angles, or projector-overlap head affinity.
- Bridge GQA weight-circuit fix PR #1593 is merged. This proposal still preserves native
  KV-head identity rather than counting expanded views independently.
- No new runtime dependency is required.

### Checklist

- [x] No matching implementation or pull request was found on 2026-08-24.
- [x] Orientation, pair counts, benchmarks, annotations, and null moments were checked.
- [x] Stable API and paper reproduction are separated.
- [x] The proposal includes falsifiable causal validation.
- [x] No dependency or external code artifact is required.

Status: implemented in #1721

janmenjayap:
# Description

Implements #1720.

Adds a basis-invariant Projection Kernel analysis surface in two reviewable commits:

1. model-independent reduced-SVD basis extraction, principal angles, raw and normalized PK,
   numerical-rank metadata, roundoff-bound handling, and random-subspace moments;
2. a TransformerBridge OQ/OK/OV head-affinity wrapper with native GQA KV-head identity,
   hybrid-layer indices, forward/all layer masks, bounded-memory tiled scoring, and ranking.

The implementation promotes fp16/bf16 inputs to fp32 for stable SVDs, preserves float64,
detaches model weights from autograd, and reports role/layer/head context for malformed or
rank-deficient weights. The accompanying guide documents orientation, rank semantics,
limitations, GQA behavior, and the distinction from Composition Score, with runnable numerical
and GPT-2 TransformerBridge examples.

## Validation

- Focused PK suite: 67 passed.
- `make check-format`: passed.
- `uv run mypy .`: passed (392 source files).
- `uv run build-docs`: passed with no PK-specific warnings.
- `uv build`: source distribution and wheel built successfully.
- Local PR test surfaces (`make test-pr` did not exit successfully because of the three
  integration failures described below):
  - unit: 5,363 passed;
  - docstring: 18 passed;
  - acceptance: 209 passed;
  - integration: 1,394 passed, including all PK tests; two gated Gemma loads failed because the
    local HF token lacks access, and one existing GraniteMoeHybrid MPS parity test reproduced the
    investigated PyTorch 2.10 MPS numerical divergence.
- GitHub CI: all applicable compatibility, coverage, formatting, typing, docstring, benchmark,
  and notebook checks passed.

## Checklist

- [x] I have read and understood the contribution guidelines.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation (if appropriate).
- [x] My changes generate no new warnings.
- [x] No changelog file is required by this repository; the user-facing change is documented in
  `docs/source/content/projection_kernel.md`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

### DIFF
--- a/transformer_lens/tools/analysis/__init__.py
+++ b/transformer_lens/tools/analysis/__init__.py
@@ -12,6 +12,7 @@ Tools:
     - jacobian_lens: The Jacobian lens (J-lens) — per-layer causal transport to
       the output vocabulary basis, with loading of published lens artifacts,
       native fitting, readouts, interventions, and J-space sparse decomposition.
+    - projection_kernel: Basis-invariant subspace overlap and principal angles.
 """
 
 from transformer_lens.tools.analysis.direct_logit_attribution import (
@@ -33,6 +34,14 @@ from transformer_lens.tools.analysis.jacobian_lens_decomposition import (
     estimate_occupancy,
     get_sparse_decomposition,
 )
+from transformer_lens.tools.analysis.projection_kernel import (
+    ProjectionKernelResult,
+    RandomSubspaceReference,
+    SubspaceBasis,
+    orthonormal_subspace,
+    projection_kernel,
+    random_projection_kernel_moments,
+)
 
 __all__ = [
     "DirectLogitAttribution",
@@ -41,9 +50,15 @@ __all__ = [
     "JSpaceVarianceProfile",
     "JacobianLens",
     "JacobianLensReadout",
+    "ProjectionKernelResult",
+    "RandomSubspaceReference",
+    "SubspaceBasis",
     "direct_logit_attribution",
     "estimate_occupancy",
     "get_act_patch_direct_path",
     "get_act_patch_direct_path_all_sources",
     "get_sparse_decomposition",
+    "orthonormal_subspace",
+    "projection_kernel",
+    "random_projection_kernel_moments",
 ]
--- a/transformer_lens/tools/analysis/__init__.py
+++ b/transformer_lens/tools/analysis/__init__.py
@@ -49,8 +49,8 @@ from transformer_lens.tools.analysis.projection_kernel import (
 )
 
 __all__ = [
-    "DirectLogitAttribution",
     "AttentionHeadRef",
+    "DirectLogitAttribution",
     "HeadAffinityPair",
     "HeadAffinityResult",
     "JSpaceDecomposition",
--- a/transformer_lens/tools/analysis/__init__.py
+++ b/transformer_lens/tools/analysis/__init__.py
@@ -12,7 +12,8 @@ Tools:
     - jacobian_lens: The Jacobian lens (J-lens) — per-layer causal transport to
       the output vocabulary basis, with loading of published lens artifacts,
       native fitting, readouts, interventions, and J-space sparse decomposition.
-    - projection_kernel: Basis-invariant subspace overlap and principal angles.
+    - projection_kernel: Basis-invariant subspace overlap and TransformerBridge
+      attention-head OQ/OK/OV affinity.
 """
 
 from transformer_lens.tools.analysis.direct_logit_attribution import (
@@ -35,9 +36,13 @@ from transformer_lens.tools.analysis.jacobian_lens_decomposition import (
     get_sparse_decomposition,
 )
 from transformer_lens.tools.analysis.projection_kernel import (
+    AttentionHeadRef,
+    HeadAffinityPair,
+    HeadAffinityResult,
     ProjectionKernelResult,
     RandomSubspaceReference,
     SubspaceBasis,
+    attention_head_subspace_affinity,
     orthonormal_subspace,
     projection_kernel,
     random_projection_kernel_moments,
@@ -45,6 +50,9 @@ from transformer_lens.tools.analysis.projection_kernel import (
 
 __all__ = [
     "DirectLogitAttribution",
+    "AttentionHeadRef",
+    "HeadAffinityPair",
+    "HeadAffinityResult",
     "JSpaceDecomposition",
     "JSpaceOccupancy",
     "JSpaceVarianceProfile",
@@ -53,6 +61,7 @@ __all__ = [
     "ProjectionKernelResult",
     "RandomSubspaceReference",
     "SubspaceBasis",
+    "attention_head_subspace_affinity",
     "direct_logit_attribution",
     "estimate_occupancy",
     "get_act_patch_direct_path",
--- a/transformer_lens/tools/analysis/projection_kernel.py
+++ b/transformer_lens/tools/analysis/projection_kernel.py
@@ -13,6 +13,7 @@ from numbers import Real
 from typing import Any, List, Literal, Optional, Sequence, Tuple, cast
 
 import torch
+from jaxtyping import Bool, Float, Int
 
 AttentionRole = Literal["Q", "K", "V", "O"]
 LayerOrder = Literal["forward", "all"]
@@ -35,8 +36,8 @@ class SubspaceBasis:
         input_shape: Shape of the matrix from which the basis was extracted.
     """
 
-    basis: torch.Tensor
-    singular_values: torch.Tensor
+    basis: Float[torch.Tensor, "ambient rank"]
+    singular_values: Float[torch.Tensor, "spectrum"]
     rank: int
     measured_rank: int
     rtol: float
@@ -53,10 +54,10 @@ class SubspaceBasis:
 class ProjectionKernelResult:
     """Projection Kernel score and its principal-angle decomposition."""
 
-    score: torch.Tensor
-    normalized: torch.Tensor
-    cosines: torch.Tensor
-    angles: torch.Tensor
+    score: Float[torch.Tensor, ""]
+    normalized: Float[torch.Tensor, ""]
+    cosines: Float[torch.Tensor, "principal_angle"]
+    angles: Float[torch.Tensor, "principal_angle"]
     rank_a: int
     rank_b: int
     ambient_dim: int
@@ -104,19 +105,23 @@ class HeadAffinityResult:
     Score tensors have shape
     ``[source_layer, source_head, target_layer, target_head]``. Layer index
     tuples map tensor positions to original model block numbers.
+
+    ``source_ranks`` and ``target_ranks`` are measured numerical ranks for each
+    head before optional truncation. Scalar ``source_rank`` and ``target_rank``
+    are the retained basis widths used for their respective roles.
     """
 
-    scores: torch.Tensor
-    normalized: torch.Tensor
-    valid_mask: torch.Tensor
+    scores: Float[torch.Tensor, "source_layer source_head target_layer target_head"]
+    normalized: Float[torch.Tensor, "source_layer source_head target_layer target_head"]
+    valid_mask: Bool[torch.Tensor, "source_layer source_head target_layer target_head"]
     source_role: AttentionRole
     target_role: AttentionRole
     source_layer_indices: Tuple[int, ...]
     target_layer_indices: Tuple[int, ...]
     source_head_kind: HeadKind
     target_head_kind: HeadKind
-    source_ranks: torch.Tensor
-    target_ranks: torch.Tensor
+    source_ranks: Int[torch.Tensor, "source_layer source_head"]
+    target_ranks: Int[torch.Tensor, "target_layer target_head"]
     source_rank: int
     target_rank: int
     rank: Optional[int]
@@ -203,7 +208,7 @@ def _validate_rank(rank: Optional[int], max_rank: int) -> Optional[int]:
 
 
 def orthonormal_subspace(
-    matrix: torch.Tensor,
+    matrix: Float[torch.Tensor, "ambient width"],
     *,
     rank: Optional[int] = None,
     rtol: Optional[float] = None,
--- a/transformer_lens/tools/analysis/projection_kernel.py
+++ b/transformer_lens/tools/analysis/projection_kernel.py
@@ -10,14 +10,30 @@ from __future__ import annotations
 import math
 from dataclasses import dataclass
 from numbers import Real
-from typing import Optional, Tuple
+from typing import Any, List, Literal, Optional, Sequence, Tuple, cast
 
 import torch
+
+AttentionRole = Literal["Q", "K", "V", "O"]
+LayerOrder = Literal["forward", "all"]
+HeadKind = Literal["query", "kv"]
+
+_PAIRWISE_TEMP_BYTES = 64 * 1024 * 1024
 
 
 @dataclass(frozen=True)
 class SubspaceBasis:
-    """An explicitly ranked orthonormal basis extracted from a matrix."""
+    """An explicitly ranked orthonormal basis extracted from a matrix.
+
+    Attributes:
+        basis: Orthonormal column-space basis, ``[ambient_dim, rank]``.
+        singular_values: All reduced-SVD singular values, in descending order.
+        rank: Number of retained basis directions.
+        measured_rank: Numerical rank before optional caller truncation.
+        rtol: Effective relative rank tolerance.
+        threshold: Absolute singular-value threshold used for rank measurement.
+        input_shape: Shape of the matrix from which the basis was extracted.
+    """
 
     basis: torch.Tensor
     singular_values: torch.Tensor
@@ -54,6 +70,103 @@ class RandomSubspaceReference:
     rank: int
     mean: float
     variance: float
+
+
+@dataclass(frozen=True)
+class AttentionHeadRef:
+    """Structured identity for one attention-head weight subspace."""
+
+    layer: int
+    head: int
+    role: AttentionRole
+    kind: HeadKind
+
+    @property
+    def label(self) -> str:
+        """Return the conventional TransformerLens layer/head label."""
+        return f"L{self.layer}H{self.head}"
+
+
+@dataclass(frozen=True)
+class HeadAffinityPair:
+    """One ranked source-target head pair."""
+
+    source: AttentionHeadRef
+    target: AttentionHeadRef
+    score: float
+    normalized: float
+
+
+@dataclass(frozen=True)
+class HeadAffinityResult:
+    """Projection Kernel affinities between two attention-head roles.
+
+    Score tensors have shape
+    ``[source_layer, source_head, target_layer, target_head]``. Layer index
+    tuples map tensor positions to original model block numbers.
+    """
+
+    scores: torch.Tensor
+    normalized: torch.Tensor
+    valid_mask: torch.Tensor
+    source_role: AttentionRole
+    target_role: AttentionRole
+    source_layer_indices: Tuple[int, ...]
+    target_layer_indices: Tuple[int, ...]
+    source_head_kind: HeadKind
+    target_head_kind: HeadKind
+    source_ranks: torch.Tensor
+    target_ranks: torch.Tensor
+    source_rank: int
+    target_rank: int
+    rank: Optional[int]
+    rtol: float
+
+    def top_pairs(self, k: int = 20, *, normalized: bool = False) -> List[HeadAffinityPair]:
+        """Return the highest-scoring valid pairs with deterministic tie order."""
+        if isinstance(k, bool) or not isinstance(k, int) or k < 1:
+            raise ValueError(f"k must be a positive integer, got {k!r}")
+
+        selected_scores = self.normalized if normalized else self.scores
+        entries: List[HeadAffinityPair] = []
+        for source_layer, source_head, target_layer, target_head in (
+            torch.nonzero(self.valid_mask, as_tuple=False).cpu().tolist()
+        ):
+            source = AttentionHeadRef(
+                layer=self.source_layer_indices[source_layer],
+                head=source_head,
+                role=self.source_role,
+                kind=self.source_head_kind,
+            )
+            target = AttentionHeadRef(
+                layer=self.target_layer_indices[target_layer],
+                head=target_head,
+                role=self.target_role,
+                kind=self.target_head_kind,
+            )
+            entries.append(
+                HeadAffinityPair(
+                    source=source,
+                    target=target,
+                    score=float(self.scores[source_layer, source_head, target_layer, target_head]),
+                    normalized=float(
+                        self.normalized[source_layer, source_head, target_layer, target_head]
+                    ),
+                )
+            )
+
+        def sort_key(pair: HeadAffinityPair) -> Tuple[float, int, int, int, int]:
+            value = pair.normalized if normalized else pair.score
+            return (
+                -value,
+                pair.source.layer,
+                pair.source.head,
+                pair.target.layer,
+                pair.target.head,
+            )
+
+        entries.sort(key=sort_key)
+        return entries[:k]
 
 
 def _compute_dtype(dtype: torch.dtype) -> torch.dtype:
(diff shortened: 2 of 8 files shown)


transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+        first_identity = torch.eye(first.shape[1], dtype=dtype, device=first.device)
+        second_identity = torch.eye(second.shape[1], dtype=dtype, device=second.device)
+        if not torch.allclose(first.T @ first, first_identity, rtol=tolerance, atol=tolerance):
+            raise ValueError("subspace_a basis columns must be orthonormal")
+        if not torch.allclose(second.T @ second, second_identity, rtol=tolerance, atol=tolerance):
+            raise ValueError("subspace_b basis columns must be orthonormal")
+
+    overlap = first.T @ second
+    score = _clamp_projection_scores(overlap.square().sum(), min(subspace_a.rank, subspace_b.rank))
+    cosines = torch.linalg.svdvals(overlap)
```
jlarson4: `torch.linalg.svdvals` has no MPS kernel, so `projection_kernel` raises NotImplementedError on MPS tensors while `orthonormal_subspace` survives via the CPU fallback. Let's route the small overlap matrix through `.cpu()` or use `sqrt(eigvalsh(overlapᵀoverlap))`, whichever you prefer.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+        raise ValueError(f"matrix dimensions must be non-empty, got shape {tuple(matrix.shape)}")
+    if not torch.is_floating_point(matrix):
+        raise ValueError(f"matrix must have a real floating-point dtype, got {matrix.dtype}")
+    if not bool(torch.isfinite(matrix).all()):
+        raise ValueError("matrix must contain only finite values")
+
+    input_shape = (matrix.shape[0], matrix.shape[1])
+    requested_rank = _validate_rank(rank, min(input_shape))
+    compute_dtype = _compute_dtype(matrix.dtype)
+    effective_rtol = _validate_rtol(rtol, input_shape, compute_dtype)
```
jlarson4: The default rank tolerance derives from the compute dtype, so a genuinely rank-deficient fp16/bf16 head passes the full-column-rank gate with a quantization-noise basis direction. Can we derive `rtol` from the storage dtype's eps (or warn at `_extract_bases` for low-precision inputs)? 

It would also be nice to add a low-precision case to the rank guard test.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+    flat_target = target_bases.reshape(layer_count * target_heads, target_ambient, target_rank)
+    scores = _pairwise_projection_kernel(flat_source, flat_target).reshape(
+        layer_count, source_heads, layer_count, target_heads
+    )
+    scores = _clamp_projection_scores(scores, min(source_rank, target_rank))
+    normalized = scores / math.sqrt(source_rank * target_rank)
+    if validated_layer_order == "forward":
+        layer_tensor = torch.tensor(layer_indices, device=result_device)
+        layer_mask = layer_tensor[:, None] < layer_tensor[None, :]
+        valid_mask = layer_mask[:, None, :, None].expand_as(scores)
```
jlarson4: In the `forward` branch `valid_mask` is a stride-0 broadcast view, so a user write to one entry flips its aliased copies; the `all` branch returns a materialized tensor. Add `.contiguous()`.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+                subspace = orthonormal_subspace(
+                    matrix[head].to(dtype=dtype), rank=selected_rank, rtol=rtol
+                )
+            except ValueError as error:
+                raise ValueError(
+                    f"Could not extract role {role} at layer {layer}, head {head}: {error}"
+                ) from error
+            head_bases.append(subspace.basis.to(device=device))
+            head_ranks.append(subspace.measured_rank)
+        layer_bases.append(torch.stack(head_bases))
```
jlarson4, line 584: The basis stack is fully materialized per role (~2.15 GB/role at 32L/32H/4096) while only the overlap temp is tiled, and `top_pairs` builds every valid pair before taking k. A docs note on the practical model-size ceiling now, with stack tiling + `torch.topk` as a follow-up, would cover it.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+        second_identity = torch.eye(second.shape[1], dtype=dtype, device=second.device)
+        if not torch.allclose(first.T @ first, first_identity, rtol=tolerance, atol=tolerance):
+            raise ValueError("subspace_a basis columns must be orthonormal")
+        if not torch.allclose(second.T @ second, second_identity, rtol=tolerance, atol=tolerance):
+            raise ValueError("subspace_b basis columns must be orthonormal")
+
+    overlap = first.T @ second
+    score = _clamp_projection_scores(overlap.square().sum(), min(subspace_a.rank, subspace_b.rank))
+    cosines = torch.linalg.svdvals(overlap)
+    angles = torch.acos(cosines.clamp(min=0.0, max=1.0))
```
jlarson4: With `check_orthonormal=False` and a hand-built basis, a cosine of 1.3 stays visible in `cosines` but clamps to angle 0.0, while `score` raises on the analogous violation. Apply the same tolerance-then-raise rule to the acos clamp, or document that `angles` assumes orthonormal inputs.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+
+
+def _read_role_matrix(model: Any, block: Any, layer: int, role: AttentionRole) -> torch.Tensor:
+    attn = getattr(block, "attn", None)
+    if attn is None:
+        raise ValueError(f"attention block {layer} does not expose an attn component")
+    attribute = f"W_{role}"
+    try:
+        matrix = getattr(attn, attribute)
+    except (AttributeError, RuntimeError, ValueError) as error:
```
jlarson4, line 518: `require_readable_weight` raises NotImplementedError to mean "unsupported", but the `except (AttributeError, RuntimeError, ValueError)` here catches it via its RuntimeError base and converts it to ValueError. Let NotImplementedError re-raise with the added role/layer context.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+    scores: torch.Tensor
+    normalized: torch.Tensor
+    valid_mask: torch.Tensor
+    source_role: AttentionRole
+    target_role: AttentionRole
+    source_layer_indices: Tuple[int, ...]
+    target_layer_indices: Tuple[int, ...]
+    source_head_kind: HeadKind
+    target_head_kind: HeadKind
+    source_ranks: torch.Tensor
```
jlarson4: `source_ranks`/`target_ranks` hold measured rank while the bases carry selected rank, and the docstring documents neither. Please add two lines naming which rank each field carries.

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+        basis: Orthonormal column-space basis, ``[ambient_dim, rank]``.
+        singular_values: All reduced-SVD singular values, in descending order.
+        rank: Number of retained basis directions.
+        measured_rank: Numerical rank before optional caller truncation.
+        rtol: Effective relative rank tolerance.
+        threshold: Absolute singular-value threshold used for rank measurement.
+        input_shape: Shape of the matrix from which the basis was extracted.
+    """
+
+    basis: torch.Tensor
```
jlarson4: Sibling modules annotate tensors with `jaxtyping.Float` shape strings; this module is bare `torch.Tensor` throughout, so runtime shape-checking enforces nothing here. Should we align with its siblings or add a one-line comment recording that the deviation is deliberate?

transformer_lens/tools/analysis/projection_kernel.py
```diff
@@ -0,0 +1,685 @@
...
+            for dimension in value.input_shape
+        )
+    ):
+        raise ValueError(f"{name}.input_shape must contain two positive dimensions")
+    if singular_values.shape[0] != min(value.input_shape):
+        raise ValueError(f"{name}.singular_values length must equal min(input_shape)")
+    if (
+        isinstance(value.rank, bool)
+        or not isinstance(value.rank, int)
+        or value.rank < 1
```
jlarson4, line 309: #1720's plan lists "deficient" among the explicit-value cases and validates "moments" plural. Please add an exact-value rank-deficient PK case and a variance assertion.

transformer_lens/tools/analysis/__init__.py
```diff
@@ -33,17 +35,39 @@
...
+    SubspaceBasis,
+    attention_head_subspace_affinity,
+    orthonormal_subspace,
+    projection_kernel,
+    random_projection_kernel_moments,
+)
 
 __all__ = [
     "DirectLogitAttribution",
+    "AttentionHeadRef",
```
jlarson4, line 52: The new entries broke `__all__`'s alphabetical ordering, please restore alphabetical ordering

jlarson4 (commented):
Excellent work here @janmenjayap! This is a solid implementation of your plan from #1720. Just a couple comments to tighten things up before merging

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). MPS overlap SVD now runs on CPU and returns singular values to the original device. The MPS regression verifies score, normalized score, cosines, and angles remain on MPS.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). The default is now `max(max(shape) * compute_eps, storage_eps)`, using the least-precise participating storage dtype in the wrapper. This avoids quantization-noise rank inflation without producing unusable tolerances for realistic FP16/BF16 dimensions. Added realistic full-rank, deficient-rank, and mixed-dtype tests.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). The expanded mask is now materialized with `.contiguous()`. The regression verifies contiguity and that mutating one element does not alter another.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Addressed in [`bf62a8e1`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/bf62a8e1c4352242cdb77e92e2784c457d4800b8). The guide now documents basis-stack memory, the 2.15 GB-per-role example, quadratic pair enumeration, and `top_pairs()` allocation behavior. Tiled extraction and tensor top-k ranking are explicitly left as follow-ups.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). Principal cosines now use tolerance-then-raise validation for `[0, 1]`: tiny numerical drift is clamped, while material violations raise. Tests cover both cases.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). Unsupported weights now re-raise `NotImplementedError` with role/layer context and preserve the original exception as `__cause__`.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Addressed in [`bf62a8e1`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/bf62a8e1c4352242cdb77e92e2784c457d4800b8). The API now distinguishes per-head measured ranks from the retained scalar source and target basis widths, both in the dataclass and public guide.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Addressed in [`bf62a8e1`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/bf62a8e1c4352242cdb77e92e2784c457d4800b8). Public tensor inputs and tensor-carrying result fields now use named `Float`, `Int`, and `Bool` dimensions, following sibling modules without introducing another runtime decorator. Runtime tests cover non-tensor, dimensional, and dtype violations.

transformer_lens/tools/analysis/projection_kernel.py
janmenjayap: Fixed in [`f4801cba`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/f4801cbaee659af985f216eefb561a33cbc56470). Added an exact rank-deficient case with known score/principal angle and extended the seeded random-subspace test to validate variance as well as mean.

transformer_lens/tools/analysis/__init__.py
janmenjayap: Fixed in [`bf62a8e1`](https://github.com/TransformerLensOrg/TransformerLens/pull/1721/changes/bf62a8e1c4352242cdb77e92e2784c457d4800b8). `__all__` is alphabetized again, with a regression test preserving that invariant.

jlarson4:
Great work on this @janmenjayap, looks great, merging now.

PR #1722, opened 2026-08-25
Preserve alias hook replacements across no-op callbacks

koriyoshi2041:
## Problem

When a hook runs under multiple alias names, each callback sees the previous callback's replacement. However, if a later alias callback returns `None`, `HookPoint.full_hook` returns that final `None` and silently discards an earlier replacement. This affects both forward activations and backward gradients.

## Fix

Track whether any alias callback changed the value and return the accumulated result after all aliases have run. The existing all-no-op behavior remains unchanged.

## Test

- `uv run pytest tests/unit/test_hook_points.py -q` — 31 passed
- `uv run black --check transformer_lens/hook_points.py tests/unit/test_hook_points.py`
- `uv run isort --check-only transformer_lens/hook_points.py tests/unit/test_hook_points.py`
- `uv run mypy transformer_lens` — 388 source files clean
- `git diff --check`

The regressions cover forward replacements, conversion reversion, and backward gradient replacements.

## Risk

Low. The change is limited to alias dispatch and only changes the returned value when an earlier alias callback supplied a replacement. Canonical hooks and all-`None` alias chains retain their current behavior.

### DIFF
--- a/tests/unit/test_hook_points.py
+++ b/tests/unit/test_hook_points.py
@@ -211,6 +211,84 @@ def test_reshape_functionality_hook_returns_none_integration():
     assert torch.equal(result, test_input)
 
 
+def test_alias_hook_preserves_earlier_replacement_when_later_alias_returns_none():
+    """A later no-op alias must not discard an earlier alias replacement."""
+    import torch
+
+    hook_point = HookPoint()
+    seen = []
+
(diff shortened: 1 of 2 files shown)


jlarson4:
Looks great! Thanks @koriyoshi2041 for putting this together

PR #1723, opened 2026-08-25
feat(backward_lens): add vocabulary readout of GPT-2 MLP gradient factors

### LINKED ISSUE #1686: [Proposal] Backward Lens: vocabulary readout of multilayer perceptron gradient factors
### Proposal

Add a tested Backward Lens analysis tool for `TransformerBridge`. The first implementation will expose the low-rank factors of the two multilayer perceptron weight gradients and project every residual-width factor into vocabulary space.

For a standard two-matrix multilayer perceptron, the tool will keep the paper's two interpretations separate:

1. **First matrix, imprint:** its gradient is spanned by forward inputs `x_i`. Vocabulary projection asks what information from the prompt is written into the first matrix.
2. **Second matrix, shift:** its gradient is spanned by backward Vector Jacobian product signals `delta_i`. Vocabulary projection asks which token directions the update shifts the second matrix toward or away from.

The first version will support one-token next-token losses, non-gated GPT-2 multilayer perceptrons, and raw models loaded with `TransformerBridge.boot_transformers`.

### Motivation

TransformerLens provides forward vocabulary readouts and the Jacobian Lens, but it has no maintained tool for interpreting parameter gradients in vocabulary space. Backward Lens provides this missing view without training an auxiliary model or loading an artifact.

The method has an exact correctness invariant. For each token position, a linear layer's weight gradient is an outer product of its forward input and its output gradient. Summing those factors reconstructs the full weight gradient. This makes the implementation independently testable without copying the authors' unlicensed research repository.

The feature also fits the current TransformerLens direction. The Jacobian Lens established raw Bridge validation, autograd capture, vocabulary readout, and analysis modules under `transformer_lens/tools/analysis/`.

### Pitch

Add `BackwardLens` as a self-contained analysis tool under
`transformer_lens/tools/analysis/`. It will use the existing TransformerBridge linear-layer
hooks and PyTorch autograd support while keeping gradient interpretation separate from model
editing.

#### Proposed interface

Names are open to maintainer preference.

```python
from transformer_lens.tools.analysis import BackwardLens

model = TransformerBridge.boot_transformers("openai-community/gpt2", device="cpu")
lens = BackwardLens(model)

result = lens.analyze(
    "The Eiffel Tower is in the city of",
    target_token=" Paris",
    layers=[8, 9, 10],
)

result.first_matrix.forward_inputs
result.first_matrix.vocabulary_logits
result.second_matrix.backward_signals
result.second_matrix.top_tokens(k=5)
result.second_matrix.bottom_tokens(k=5)
result.max_reconstruction_error
```

`target_token` must encode to exactly one token. The loss is the negative log likelihood of that token under the logits at the final prompt position. Empty prompts, multiple-token targets, invalid layers, gated multilayer perceptrons, processed weights, and unsupported drivers raise clear errors.

#### Mathematical contract

For GPT-2 and TransformerLens weight orientation:

- `W_in` has shape `[d_model, d_mlp]` and computes `pre = x @ W_in`.
- `W_out` has shape `[d_mlp, d_model]` and computes `out = hidden @ W_out`.

For token positions `i`:

- `grad_W_in = sum_i outer(x_i, grad_pre_i)`.
- `grad_W_out = sum_i outer(hidden_i, grad_out_i)`.

The first matrix is interpreted through `x_i`, which is residual-width. The second matrix is interpreted through `grad_out_i`, which is also residual-width. The other factor controls each vector's coefficient in the full gradient.

The vocabulary projection follows the paper's logit lens. For each residual-width vector, apply the model's final normalization with statistics recomputed on that vector, then apply the unembedding. This is not the cached-scale linear attribution path.

For the second matrix, the raw backward signal commonly ranks the edit target near the bottom. Gradient descent subtracts the gradient. The result therefore exposes both top and bottom rankings and documents their signs. It does not assert that the target must appear in the raw signal's top tokens.

#### Implementation design

1. Require a raw causal `TransformerBridge` loaded through the transformers driver.
2. Register forward captures on `blocks.{layer}.mlp.in.hook_in`, `blocks.{layer}.mlp.in.hook_out`, `blocks.{layer}.mlp.out.hook_in`, and `blocks.{layer}.mlp.out.hook_out`.
3. Run one forward pass and compute the one-token next-token loss.
4. Use `torch.autograd.grad` to obtain gradients for the captured linear outputs and the corresponding original weight parameters. Do not call `.backward()`, mutate `.grad`, or freeze the selected weights.
5. Reconstruct each weight gradient from its token-position outer products in float32 and report absolute and relative error.
6. Detach residual-width factors, run the standard final-normalization and unembedding projection, and return raw vocabulary logits plus top and bottom token helpers.
7. Restore all temporary hooks and preserve parameter gradient state even when an exception occurs.

The first implementation should not use backward hooks on normalization points and should assert that no native-path fallback warning is emitted.

#### Validation plan

1. **Exact factorization:** synthetic linear layers reconstruct both weight gradients within a stated tolerance.
2. **Orientation:** asymmetric dimensions catch transposed outer products and swapped factors.
3. **Projection parity:** vocabulary logits match a direct call through the model's final normalization and unembedding on each detached factor.
4. **Sign handling:** a controlled example verifies raw top and bottom rankings and the sign of the gradient-descent update.
5. **Target contract:** zero-token and multiple-token targets are rejected.
6. **State safety:** parameter `.grad` values, `requires_grad` flags, hooks, and model weights are unchanged after success or failure.
7. **Bridge integration:** `gpt2-small` returns finite factors and reconstruction errors below tolerance without native-path fallback warnings.
8. **Paper-level sanity report:** report target ranks for a small fixed prompt set, but do not make a brittle semantic rank threshold a required unit test.

#### Initial pull request scope

- `transformer_lens/tools/analysis/backward_lens.py`
- public exports for `BackwardLens` and its result dataclasses
- model-free unit tests for factorization, orientation, sign, and validation
- `gpt2-small` TransformerBridge integration tests
- one demonstration notebook
- one documentation section

#### Non-goals

- gated multilayer perceptrons
- multiple-token or batched editing targets
- applying optimizer steps or editing model weights
- CounterFact replication
- Jacobian-space update tracing
- other model families
- external repository code as a dependency or oracle

These belong in a separate follow-up issue after the core contract is merged.

### Alternatives

- **Continue using ad hoc autograd scripts:** this can recover the factors, but it leaves users
  responsible for tensor orientation, target-loss semantics, normalization, hook cleanup, and
  gradient-state safety. It also provides no shared result contract or regression tests.
- **Add the implementation to `jacobian_lens.py`:** the Jacobian Lens represents causal transport
  between residual streams, while Backward Lens analyzes the gradient induced by a particular loss.
  They share infrastructure but answer different questions, so separate modules are clearer.
- **Use `run_with_cache(incl_bwd=True)`:** this is useful as a future independent cross-check, but
  the initial implementation needs gradients for selected intermediate tensors and weight
  parameters from one explicitly defined scalar loss. `torch.autograd.grad` provides that contract
  without mutating `.grad`.
- **Depend on the authors' repository:** the repository is a research demonstration with no
  declared license or maintained library interface. The implementation should instead follow the
  published equations and use the exact outer-product reconstruction as its primary oracle.
- **Support every multilayer perceptron architecture immediately:** gated architectures require a
  different factorization contract because their gate and up projections interact through a
  product and activation function. Starting with GPT-2 avoids presenting an incorrect universal
  abstraction.

### Additional context

- Katz, Belinkov, Geva, and Wolf, "Backward Lens: Projecting Language Model Gradients into the Vocabulary Space," Empirical Methods in Natural Language Processing 2024, Best Paper, arXiv:2402.12865.
- Authors' research demonstration: `shacharKZ/BackwardLens`. The repository has no declared license, so no code will be copied, vendored, or installed as a test dependency.
- The existing Jacobian Lens provides a nearby TransformerBridge precedent for raw-model
  validation, autograd capture, vocabulary readout, public result dataclasses, integration tests,
  and documentation under `transformer_lens/tools/analysis/`.
- A separate follow-up proposal will cover CounterFact replication, causal update tracing,
  Jacobian-space comparisons, editing experiments, multiple-token targets, gated architectures,
  and broader model coverage.

### Checklist

- [x] No exact Backward Lens or gradient-vocabulary issue or pull request was found in a fresh search.
- [x] The first pull request is limited to a testable GPT-2 vertical slice.
- [x] The first-matrix imprint and second-matrix shift interpretations are separated.
- [x] The raw backward-signal ranking and gradient-descent sign are documented correctly.
- [x] The correctness oracle is independent of unlicensed external code.

janmenjayap:
# Description

This PR implements **Backward Lens** as a self-contained analysis tool for raw GPT-2 models loaded
through `TransformerBridge`.

Backward Lens interprets the low-rank factors of MLP weight gradients by projecting residual-width
factors into vocabulary space. It preserves the two interpretations from Katz et al. (2024):

1. **First matrix — imprint:** the gradient is spanned by forward inputs `x_i`, describing
   information written into the first MLP matrix.
2. **Second matrix — shift:** the gradient is spanned by backward VJP signals `delta_i`, describing
   vocabulary directions toward or away from which the second MLP matrix is updated.

For every selected layer, the implementation reconstructs the independently obtained weight
gradients using:

- `grad_W_in = sum_i outer(x_i, grad_pre_i)`
- `grad_W_out = sum_i outer(hidden_i, grad_out_i)`

Reconstruction errors and numerical ranks provide independently testable correctness invariants.

## Motivation and context

TransformerLens already supports forward vocabulary readouts and the Jacobian Lens, but it did not
provide a maintained tool for interpreting parameter gradients in vocabulary space.

This feature adds that missing analysis surface without training an auxiliary model, loading a
learned artifact, applying optimizer steps, or depending on the authors' research repository.

## Implementation details

- Adds `BackwardLens` under `transformer_lens/tools/analysis/`.
- Adds public, detached result dataclasses and package exports.
- Uses one forward pass and one `torch.autograd.grad` call.
- Does not call `.backward()`, mutate parameter `.grad`, or freeze selected parameters.
- Captures aligned position-wise factors for both GPT-2 MLP projections.
- Reconstructs gradients in float32 and reports absolute and relative errors.
- Projects residual-width factors through the live final normalization and unembedding.
- Recomputes normalization statistics independently for every projected factor.
- Supports raw and explicit opt-in Normalized Logit Lens projections.
- Retains original factor norms and marks exact-zero factors.
- Exposes signed top/bottom token rankings and target-rank diagnostics.
- Documents the distinction between raw gradients and the negative update direction used by
  gradient descent.
- Preserves weights, gradients, `requires_grad`, existing hooks, training mode, and RNG state.
- Removes temporary hooks after both successful and failing analyses.

## Supported scope

This initial implementation supports:

- Raw, unprocessed GPT-2 `TransformerBridge` models.
- Models loaded through the transformers Bridge driver.
- One unbatched prompt.
- One single-token next-token target.
- Dense, non-gated GPT-2 MLPs.
- User-selected layers.

The following are intentionally deferred to separate follow-up issues:

- Gated MLPs.
- Additional model families.
- Batched prompts.
- Multiple-token targets.
- Applying optimizer steps or editing model weights.
- CounterFact replication.
- Jacobian-space update tracing.
- Paper-scale editing benchmarks.

## Documentation and demonstration

This PR adds:

- A reference page covering the mathematical contract, tensor shapes, API, normalization, signs,
  restrictions, state safety, and troubleshooting.
- An executed GPT-2-small demonstration notebook showing:
  - Exact gradient reconstruction.
  - Numerical rank and reconstruction error.
  - Layer-by-position vocabulary directions.
  - VJP norms and target ranks.
  - Raw versus normalized projections.
  - Gradient-versus-update sign semantics.
  - Exact and near-zero signal handling.
  - Contribution-sorted cumulative reconstruction.
- CI and Makefile registration for notebook validation.

The notebook figures and tables were manually inspected for correctness and legibility.

## Dependencies

## Validation

Observed local validation:

- Unit tests: **38 passed**
- Integration tests: **23 passed**
- Notebook nbval: **10 passed**
- `uv lock --check`: passed
- `git diff --check upstream/dev...HEAD`: passed
- Notebook: 10 executed code cells with no saved errors
- Branch rebased onto current `upstream/dev`

Focused validation passed in the established `transformer-lens` Conda environment.

Exact locked-`.venv` pytest execution was blocked by a local PyTorch/CUDA runtime symbol mismatch.
The documentation builder parsed the new Backward Lens page, but a complete local docs build was not
obtained because the environment lacked Pandoc and the build encountered model-registry/docstring
errors. Repository-wide checks remain subject to PR CI.

Fixes #1686

## Type of change

#### Vocabulary-facing gradient factors 
image

#### Reconstruction and rank diagnostics 
image

### Target-token ranks: raw versus Normalized Logit Lens
image

# Checklist

- [x] My changes generate no new Backward Lens-specific warnings
  - Focused Backward Lens unit tests pass; the complete repository-wide suite is pending CI.


jlarson4:
Hey @janmenjayap! This looks great on an initial review. I will probably not have a full review for you until early next week, just want to make sure I have time to give this feature a deep, thorough review. Thank you for putting this together, keep an eye out for the review.

janmenjayap:
Thanks so much! Glad to hear the initial review looks good. No rush at all. I really appreciate you taking the time to give it a thorough review. I will keep an eye out for your feedback this week

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+) -> BackwardLensMatrixResult:
+    if projected_factor not in ("forward_inputs", "output_gradients"):
+        raise ValueError("projected_factor must be 'forward_inputs' or 'output_gradients'")
+    if not isinstance(include_normalized_logits, bool):
+        raise TypeError("include_normalized_logits must be a bool")
+    rows = (
+        factors.forward_inputs if projected_factor == "forward_inputs" else factors.output_gradients
+    )
+    norms, zero_mask, normalized_rows = _factor_norms_and_normalized_rows(rows)
+    raw_logits = _project_residual_factors(model, rows)
```
jlarson4, line 480: Every matrix result retains a full `[position, d_vocab]` float32 logit tensor (two per layer when `normalized=True`) with no top-k retention or opt-out, so result size is dominated by `layers x 2 x (1+normalized) x positions x 50257 x 4` bytes. `jacobian_lens.readout` solved this: top-k values/indices by default, full logits behind `return_full_logits=True`, can we mirror that pattern?

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+    requested_layers = _validate_requested_layers(model, layers)
+    projections = _get_gpt2_mlp_projections(model, requested_layers)
+    if not isinstance(prompt, str):
+        raise TypeError("prompt must be a string")
+    if prompt == "":
+        raise ValueError("prompt must not be empty")
+    if not isinstance(target_token, str):
+        raise TypeError("target_token must be a string")
+
+    prompt_tokens = model.to_tokens(prompt)
```
jlarson4: `model.to_tokens(prompt)` uses the Bridge default `truncate=True`, so a prompt past `n_ctx` is silently clipped while `result.prompt` still echoes the full string. Can we tokenize with `truncate=False` and raise a ValueError naming the token count and `n_ctx`, and add a case to `test_capture_rejects_invalid_analysis_inputs`?

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+        )
+    target_token_id = int(target_tokens.item())
+    input_device = next(model.original_model.parameters()).device
+    prompt_tokens = prompt_tokens.to(input_device)
+    weights = [
+        projection.original_component.weight
+        for layer in requested_layers
+        for projection in projections[layer]
+    ]
+    with _preserve_model_rng(model), torch.enable_grad():
```
jlarson4, line 766: `torch.enable_grad()` does not defeat an enclosing `torch.inference_mode()`, so `analyze()` called under it dies with a low-level autograd RuntimeError. Check `torch.is_inference_mode_enabled()` up front and raise a ValueError naming inference_mode, matching the module's other misuse errors.

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+        raise ValueError(
+            f"factors must have width d_model={model.cfg.d_model}; got {factors.shape[-1]}"
+        )
+    unembed_weight = model.W_U
+    if not isinstance(unembed_weight, torch.Tensor) or unembed_weight.ndim != 2:
+        raise ValueError("the GPT-2 Bridge must expose a rank-2 unembedding weight")
+    batched = (
+        factors.detach().to(device=unembed_weight.device, dtype=unembed_weight.dtype).unsqueeze(0)
+    )
+    logits = model.unembed(model.ln_final(batched)).squeeze(0)
```
jlarson4, line 431: `_project_residual_factors` moves rows to `W_U.device` and then calls `ln_final`, so a device_map-dispatched bridge with `ln_final` elsewhere raises an uncontrolled device-mismatch RuntimeError. Reject `cfg.n_devices > 1` in `_require_raw_gpt2_bridge` with a clear message, or verify and document co-location for supported splits.

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+            ``"out_in"`` for ``torch.nn.Linear`` storage.
+
+    Returns:
+        Detached factors, the independent gradient, its reconstruction, and
+        reconstruction errors, all on CPU with float32 tensor values.
+    """
+    _validate_floating_matrix("forward_inputs", forward_inputs)
+    _validate_floating_matrix("output_gradients", output_gradients)
+    _validate_floating_matrix("weight_gradient", weight_gradient)
+    if weight_layout not in ("in_out", "out_in"):
```
jlarson4, line 320: The concrete parameter annotations here (and on `_rank_vocabulary_logits`) let beartype intercept before these manual checks run under the repo's pytest instrumentation, so the curated error messages are never exercised in CI and the tests were widened to accept `BeartypeCallHintParamViolation` instead. Widen the parameters to `Any` at the boundary (the pattern `model: Any` already uses in this file), keep the manual checks, and revert the test widening.

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+    if getattr(model, "_weights_processed", False):
+        raise ValueError(
+            "Backward Lens requires original GPT-2 weights; this Bridge processed its weights"
+        )
+    if not isinstance(model.adapter, GPT2ArchitectureAdapter):
+        raise ValueError(
+            "Backward Lens currently supports the GPT2ArchitectureAdapter only; "
+            f"got {type(model.adapter).__name__}"
+        )
+    if bool(getattr(model.cfg, "gated_mlp", False)):
```
jlarson4, line 580: The architecture-support gates here raise ValueError where `direct_logit_attribution` raises NotImplementedError for the equivalent not-yet-supported case. Switch the two scope gates to NotImplementedError, and add tests for the untested structural guards in `_get_gpt2_mlp_projections` (per-layer `.gate`, non-Conv1D component, wrong shape/dtype).

transformer_lens/tools/analysis/backward_lens.py
```diff
@@ -0,0 +1,755 @@
...
+    """Detached factors and reconstruction for one linear weight gradient.
+
+    ``forward_inputs`` and ``output_gradients`` have shapes ``[position, in]``
+    and ``[position, out]``. ``output_gradients`` and ``weight_gradient`` preserve
+    the raw ``d(loss)/d(tensor)`` sign; they are not negated into update directions.
+    Gradient tensors use the requested storage layout. All tensors are cloned to
+    CPU in float32 so the result owns no autograd graph.
+    """
+
+    forward_inputs: torch.Tensor
```
jlarson4: Tensor fields and parameters are bare `torch.Tensor` with shapes in prose, while the sibling analysis tools annotate with jaxtyping `Float[...]` shape strings — and #1721 just adopted that convention. Aligning here keeps the two additions consistent and gives runtime shape checking for free.

docs/source/content/backward_lens.md
```diff
@@ -0,0 +1,176 @@
...
+ff2_update_tokens = ff2.bottom_tokens(model.tokenizer, k=5)
+target_ranks = ff2.gradient_descent_target_ranks(
+    result.target_token_id,
+    normalized=True,
+)
+```
+
+`target_token` must encode to exactly one token without a beginning-of-sequence token.
+For GPT-2 tokenization, a leading space is often significant. The prompt is tokenized
+normally, including its prepended BOS token; all position-indexed factors and readouts
```
jlarson4: "Including its prepended BOS token" and the result docstring's "position zero is the prepended BOS" hold only when `cfg.default_prepend_bos` is True, which `analyze()` neither checks nor records. Qualify both statements (or derive them from `prompt_token_ids`), and add a test asserting `prompt_token_ids[0]` is the BOS id under the default configuration.

docs/source/content/backward_lens.md
```diff
@@ -0,0 +1,176 @@
...
+### The two GPT-2 MLP matrices
+
+The two projections expose different residual-width factors:
+
+| Result | Weight shape | Projected factor | Shape before vocabulary projection |
+|---|---:|---|---:|
+| `input_projection` (FF1 / `c_fc`) | `[d_model, d_mlp]` | Forward input $x_i$ | `[position, d_model]` |
+| `output_projection` (FF2 / `c_proj`) | `[d_mlp, d_model]` | Backward signal $\delta_i$ | `[position, d_model]` |
+
+The FF1 readout therefore describes the residual-stream directions entering the MLP.
```
jlarson4: `forward_inputs` are the post-`ln_2` vectors (the c_fc input), not the raw residual stream. Reword to "the layer-normalized residual state entering the MLP (post-ln_2, including its gain and bias)".

docs/source/content/backward_lens.md
```diff
@@ -0,0 +1,176 @@
...
+| Symptom | Cause and resolution |
+|---|---|
+| Raw-Bridge or processed-weight error | Reboot with `TransformerBridge.boot_transformers(...)`; do not enable compatibility mode or process weights. |
+| Target encodes to zero or multiple tokens | Choose text that maps to one GPT-2 token without BOS; check leading whitespace. |
+| Duplicate or out-of-range layer error | Pass a non-empty sequence of unique indices in `[0, model.cfg.n_layers)`. |
+| Normalized logits were not requested | Call `analyze(..., normalized=True)` before using `logits(normalized=True)` or normalized ranks. |
+| FF2 ranking appears sign-reversed | Remember that results are raw loss gradients and gradient descent subtracts them; inspect bottom tokens or ascending target ranks. |
+| Results change when custom hooks are installed | Existing hooks are intentionally respected; remove them to analyze the unmodified model computation. |
+
+## References
```
jlarson4, line 187: This page never links `demos/Backward_Lens_Demo.ipynb`; `jacobian_lens_fitting.md` links its demo directly. Add the link in References.

tests/unit/tools/test_backward_lens.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+def test_matrix_result_rankings_decoding_and_target_rank_conventions() -> None:
+    logits = torch.tensor([[3.0, 1.0, -2.0, 0.0], [0.0, -1.0, 4.0, 2.0]])
+    result = _matrix_result(logits)
+
+    assert result.top(k=2).indices.tolist() == [[0, 1], [2, 3]]
+    assert result.bottom(k=2).indices.tolist() == [[2, 3], [1, 0]]
+    assert result.target_ranks(2, largest=True).tolist() == [3, 0]
+    assert result.gradient_descent_target_ranks(2).tolist() == [0, 3]
```
jlarson4, line 398: `gradient_descent_target_ranks` is checked against hand-picked synthetic logits, but no test applies an actual small update and confirms the target logit moves in the documented direction. Can we add one test applying `-η·grad` to a weight copy and asserting the target logit increases?

tests/unit/tools/test_backward_lens.py
```diff
@@ -0,0 +1,377 @@
...
+
+def test_public_backward_lens_symbols_are_exported() -> None:
+    from transformer_lens.tools.analysis import (
+        BackwardLens,
+        BackwardLensLayerResult,
+        BackwardLensMatrixResult,
+        BackwardLensResult,
+    )
+
+    assert BackwardLens.__name__ == "BackwardLens"
```
jlarson4: `assert BackwardLens.__name__ == "BackwardLens"` cannot fail once the import succeeded. Drop the `__name__` asserts and keep the import as the smoke test.

tests/integration/test_backward_lens.py
```diff
@@ -0,0 +1,536 @@
...
+
+    mlp = gpt2_bridge.blocks[0].mlp
+    input_hook_point = getattr(mlp, "in").hook_in
+    output_hook_point = mlp.out.hook_out
+
+    def scale_input(tensor, hook=None):
+        return tensor * 0.5
+
+    def shift_output(tensor, hook=None):
+        return tensor + 0.01
```
jlarson4: The `+0.01` shift on `mlp.out.hook_out` cannot distinguish capturing the pre-edit from the post-edit tensor, so the output-projection half of this test exercises nothing. Use a multiplicative edit (`tensor * 3.0`) like the input side does.

demos/Backward_Lens_Demo.ipynb
```diff
@@ -0,0 +1,1749 @@
...
...abRVH+j1BkuhoaGIjY1FlSpVlMqrVq2KK1euZNsuJSVF6ea1H24bTURERERE9DH6OLJalOjFBkvZyUwy7ezslMrt7e1zTED9/f1hY2OjOLKbYkxERERERJSdzGRV04N0Q6+TVXNzcwBAXFycUnlcXJzinCpTp05FTEyM4ggNDc3TOImIiIiIqPDRLlXNuvsuaUevpwF7eHjA2NhY6R5AAPDkyROlG9t+yNTUNN+2hiYiIiIiIiLd0+uRVRMTE7Rr1w4BAQGKhcovXrzAqVOn0KVLF4mjIyIiIiKiwozTgKUl6cjqgwcPcPr0aURHRwN4d9/Vhw8fok6dOqhTpw4AYP78+WjcuDE+++wzNGzYEJs3b0bdunUxYMAAKUMnIiIiIqJCjhssSUvSkdVXr17h/PnzCA4OxtChQxEdHY3z588rbbVcpUoV3LhxA7Vq1UJISAjGjh2LkydPwtjYWMLIiYiIiIiosJNr+Yd0Q9KR1caNG6Nx48Yfrefh4YGZM2fmQ0RERERERETvCJmAkGmWfGp6X1bKnl5vsERERERERCQVocU0YCarusNklYiIiIiISAU55JBpOK2X04B1R693AyYiIiIiIqKiiSOrREREREREKmTejEbTNqQbTFaJiIiIiIhUkMvkkGm4wRKnAesOk1UiIiIiIiIVuGZVWkxWiYiIiIiIVGCyKi0mq0RERERERCpwzaq0mKwSERERERGpIEcGZMjQuA3pBm9dQ0RERERERHqHI6tEREREREQqCAgtpgGLPIqm6GGySkREREREpAJvXSMtJqtEREREREQqvFuzqtnKSa5Z1R0mq0RERERERCppvhswOLKqM0xWiYiIiIiIVJCLDGi6J+27NqQLRSpZDTyxDsWKWUkdBhERaWyk1AEQkY5lzJstdQiUDzLSZVKHkCv6eJ/Vx48fY/Xq1di/fz/atGmDX375Jcf6ixYtwqpVq7KUGxoa4vbt2zAyMsKDBw/QoUOHLHW2bduGBg0a6Cx2TRWpZJWIiIiIiKigevToEdq1a4ehQ4fCyckJL168+GgbHx8feHt7K5V169YN7u7uMDJ6lw6mpKTg0aNHOHfuHBwdHRX1XF1ddRq/ppisEhERERERqSCQAaHhNGCRhxsslSpVCg8ePIBMJsOFCxfUauPg4AAHBwfF49u3b+POnTuYPTvr7IbSpUvD2dlZZ/Hm
```
jlarson4, line 1674: Cell 18's table displays exact integer target ranks from the same call whose cell-12 comment says ranks "can move by one across numeric backends when vocabulary logits are nearly tied", and nbval strict-compares that output. Either measure the margin for these three examples and note it, or give the rank columns the same percentile/ignore treatment cell 12 got.

jlarson4 (commented):
Hi @janmenjayap! Review is here, thank you for your patience. I verified the factorization against finite differences independent of autograd, confirmed per-position alignment, and reproduced the paper's result from your own executed notebook. I have a couple comments, please let me know when you have a chance to look over them.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in de06271d. `BackwardLens.analyze` now defaults to `top_k=10` and `return_full_logits=False`. Each matrix always retains largest and smallest rankings plus the analyzed target's exact ranks; full raw and normalized vocabulary tensors are retained only when explicitly requested. The accessors enforce the retained `top_k`, and coverage compares bounded rankings and opted-in full logits with direct projections.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 83425875. Prompt tokenization now uses `truncate=False`, followed by an explicit check against `model.cfg.n_ctx` before the forward pass. Overlong prompts raise a curated `ValueError` containing both the observed token count and configured context limit, with an integration case added to the invalid-input coverage.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 83425875. Capture now checks `torch.is_inference_mode_enabled()` before installing hooks or running the model and raises a `ValueError` that names `inference_mode`. An integration test covers this path so callers do not receive the lower-level autograd failure.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 83425875. `_require_raw_gpt2_bridge` now rejects `cfg.n_devices > 1` with a message explaining that the projection path requires the relevant modules to be co-located. The integration test confirms this rejection occurs before projection validation.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 4be30480. The manually validated parameters of `_build_linear_gradient_factors` and `_rank_vocabulary_logits` are typed as `Any` at the boundary and narrowed only after their explicit checks. The corresponding unit expectations use the intended `TypeError` and `ValueError` contracts rather than accepting `BeartypeCallHintParamViolation`.

docs/source/content/backward_lens.md
janmenjayap: Addressed in b6b1f419. The FF1 text now describes the factor as "the layer-normalized residual state entering the MLP (post-`ln_2`, including its gain and bias)" rather than the raw residual stream.

docs/source/content/backward_lens.md
janmenjayap: Addressed in b6b1f419. The guide and `BackwardLensResult` docstring now state that a prepended BOS is present only when the model/tokenization configuration requests one, with `prompt_token_ids` treated as the position-alignment source of truth. The default GPT-2 integration fixture also asserts that its first prompt token is the tokenizer BOS id.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 4be30480. Dataclass fields, trusted internal tensors, and return values now use `jaxtyping` `Float`, `Int`, and `Bool` annotations with shape strings. The boundaries whose curated runtime errors are part of the API remain `Any` until manual validation completes.

transformer_lens/tools/analysis/backward_lens.py
janmenjayap: Addressed in 83425875. Unsupported adapter-family and gated-MLP scope checks now raise `NotImplementedError`, while malformed GPT-2 instances continue to raise `ValueError`. Added coverage exercises a per-layer gate, a non-`Conv1D` wrapped component, invalid weight shape, and non-floating weight dtype using temporary fixture-safe mutations.

docs/source/content/backward_lens.md
janmenjayap: Addressed in b6b1f419. The References section now links directly to `demos/Backward_Lens_Demo.ipynb`, following the Jacobian Lens documentation pattern.

demos/Backward_Lens_Demo.ipynb
janmenjayap: Addressed in de06271d. The predeclared-example table now displays raw and normalized target-rank percentiles rounded to one decimal place instead of backend-sensitive exact integer ranks. Exact ranks remain available internally for the normalization-outcome comparison, while the executed table output uses the stable percentile presentation.

tests/integration/test_backward_lens.py
janmenjayap: Addressed in adfc7c69. The output-projection hook edit now uses `tensor * 3.0`, making pre-edit versus post-edit capture observably different while retaining the reconstruction checks for both projection paths.

tests/unit/tools/test_backward_lens.py
janmenjayap: Addressed in adfc7c69. Removed the `Class.__name__` assertions. The smoke test now relies on importing and referencing every public Backward Lens symbol, so a missing export still fails without tautological checks.

tests/unit/tools/test_backward_lens.py
janmenjayap: Addressed in adfc7c69. Added an integration test that uses `torch.func.functional_call` with an isolated copied FF2 weight, applies `weight - 1e-4 * gradient`, and asserts a measured positive increase in the analyzed target logit. The live module parameter is not mutated.

jlarson4:
Great work @janmenjayap! There was an issue with export ordering from the rebase you did onto the most recent version of `dev`, I fixed that and am rerunning CI now, assuming it passes I will get this merged!

tests/integration/test_backward_lens.py
```diff
@@ -0,0 +1,706 @@
...
+    )
+    parameter_name = next(
+        name
+        for name, parameter in gpt2_bridge.original_model.named_parameters()
+        if parameter is weight
+    )
+    updated_weight = original_weight - 1e-4 * weight_gradient
+
+    with torch.no_grad():
+        baseline_output = torch.func.functional_call(
```
jlarson4: `torch.func.functional_call` fails to restore parameters through the bridge's `original_model`: after this call, layer 11's `c_proj` weight is left as the plain override tensor, which is what errors the three `public_*` fixtures later in this module — the current coverage-job failure (Compatibility jobs never run this tier, so only coverage shows it). Swap the value in place instead — `weight.copy_(updated_weight)` under `no_grad` with a `try/finally` restore via `weight.copy_(original_weight)` — same assertion, no reparametrization through the wrapped tree.

janmenjayap: Good catch; fixed now. I replaced the `functional_call` reparametrization with an in-place `weight.copy_(updated_weight)` under `no_grad`, using a `try/finally` block to restore the original weight. The same assertions remain, and there is no longer any override left in the wrapped `original_model` tree.

jlarson4: Thanks! I am also making a specific fix related to this issue on `dev`, there will now be a `temporarily_swap_parameter` utility that handles these kinds of cases. I appreciate the contribution and the quick fix, I wouldn't have found this specific bug if not for your hard work! Assuming CI passes I will get this merged and it'll be included in 3.9.0

PR #1727, opened 2026-08-30
Fix Cohere compatibility logit scaling

### LINKED ISSUE #1726: [Bug Report] Cohere compatibility mode double-applies logit_scale in forward
**Describe the bug**

For Cohere and Cohere2 bridges, weight processing folds `config.logit_scale` into the live
unembedding weight even though the wrapped Hugging Face forward still multiplies the resulting logits
by the same scale. The first `process_weights()` / compatibility-mode call therefore applies the scale
twice relative to the raw Bridge output.

With `logit_scale=0.5`, processing changes the output to exactly `raw * 0.5`. The call returns
normally, so this is a silent numerical correctness problem for logits and weight-space analyses.

`Cohere2ArchitectureAdapter` inherits the same preprocessing behavior from
`CohereArchitectureAdapter`.

**Code example**

This reproduction uses a randomly initialized tiny model from config and does not access the Hub:

```python
import torch
from transformers import CohereConfig, CohereForCausalLM

torch.manual_seed(0)
config = CohereConfig(
    vocab_size=32,
    hidden_size=16,
    intermediate_size=32,
    num_hidden_layers=1,
    num_attention_heads=4,
    num_key_value_heads=2,
    max_position_embeddings=32,
    logit_scale=0.5,
    attention_bias=False,
    use_qk_norm=False,
    tie_word_embeddings=True,
    bos_token_id=1,
    eos_token_id=2,
    pad_token_id=0,
)
hf_model = CohereForCausalLM(config).eval()
    hf_model,
    "CohereForCausalLM",
    hf_config=config,
    dtype=torch.float32,
    device="cpu",
)

tokens = torch.tensor([[1, 2, 3, 4]])
raw = bridge(tokens).detach()

bridge.enable_compatibility_mode(
    disable_warnings=True,
    fold_ln=False,
    center_writing_weights=False,
    center_unembed=False,
    fold_value_biases=False,
)
processed = bridge(tokens).detach()

print((processed - raw * 0.5).abs().max().item())
```

Current output on `dev` at `572946b7`:

```text
0.0
```

The cause is the interaction between:

- `CohereArchitectureAdapter.preprocess_weights()`, which multiplies `unembed.weight` and
  `unembed.bias` by `cfg.logit_scale`; and
- `CohereForCausalLM.forward()`, which still multiplies the logits produced from that live weight by
  its instance-level `logit_scale`.

**Expected behavior**

- The first compatibility-mode weight processing call should preserve Cohere's single
  `logit_scale` application in forward.
- The scale should remain folded into the unembedding weight for compatibility-mode weight-space
  analyses.
- Tied input embeddings must remain unscaled.
- `cfg.logit_scale` should retain the architecture metadata, and `logit_scale=1.0` should remain a
  no-op.

**Suggested scope**

Keep the fix focused on the Cohere/Cohere2 output-scale contract and add a network-free tiny-config
regression covering both direct `process_weights()` and the first compatibility-mode call. Existing
raw-HF parity, tied-embedding, and Jacobian-lens scale behavior should remain intact.

**System Info**

- Installed from source on the current `dev` branch (`572946b7`)
- Windows 11, CPU
- Python 3.12.10
- PyTorch 2.10.0+cpu
- Transformers 5.13.0

**Additional context**

Compatibility mode is documented as a one-shot transformation because generic centering operations
are not reversible or generally idempotent. A second call producing `raw * 0.25` helped isolate the
compounding fold, but this report does not propose making the whole compatibility-mode pipeline
repeatable.

Open and closed Issue/PR searches were run for `Cohere compatibility logit_scale` and the broader
`logit_scale` term. Closed #1325 and merged #1348 cover non-trivial-scale test parameterization, not
the first-call forward double application.

### Checklist

emerardd:
# Description

Fixes #1726

Cohere and Cohere2 weight processing folds `cfg.logit_scale` into the live unembedding weights, but the Hugging Face model forward still applied its model-level scale afterward. As a result, the first valid compatibility-mode conversion produced logits with the scale applied twice.

This change adds an architecture post-install lifecycle hook. The Cohere adapter uses it to neutralize the Hugging Face model-level `logit_scale` only after the scaled unembedding weights have been installed. The Bridge config retains the original scale as metadata, and tied input embeddings remain unscaled.

Network-free tiny-config regressions cover both direct `process_weights()` and the first `enable_compatibility_mode()` call.

## Type of change

## Validation

- `tests/integration/model_bridge/test_cohere_adapter.py`: 24 passed
- Cohere/Cohere2 adapter unit surface: 74 passed
- Weight-processing unit surface: 27 passed, 1 skipped
- Full non-slow unit surface: 5,435 passed; the sole nonzero result was a native-Windows GBK decode error in a source-roster test
- The exact roster test passed after enabling Python UTF-8 mode: 1 passed
- `mypy .`: Success, no issues found in 392 source files
- pycln, pinned isort, Black, and `git diff --check`: passed

# Checklist:

- [x] I have made corresponding changes to the documentation (no public documentation change is required; the internal lifecycle hook is documented in code)

### DIFF
--- a/transformer_lens/model_bridge/architecture_adapter.py
+++ b/transformer_lens/model_bridge/architecture_adapter.py
@@ -300,6 +300,13 @@ class ArchitectureAdapter:
         """
         return state_dict
 
+    def postprocess_weights(self, bridge: Any) -> None:
+        """Apply architecture-specific updates after processed weights are installed.
+
+        Args:
+            bridge: The TransformerBridge whose live source components received the weights.
+        """
+
     def get_component_mapping(self) -> ComponentMapping:
         """Get the full component mapping.
 
--- a/transformer_lens/model_bridge/supported_architectures/cohere.py
+++ b/transformer_lens/model_bridge/supported_architectures/cohere.py
@@ -94,6 +94,7 @@ class CohereArchitectureAdapter(ArchitectureAdapter):
         # Cohere-specific dynamic attribute accessed later in preprocess_weights.
         _ls = getattr(cfg, "logit_scale", None)
         self.cfg.logit_scale = float(_ls) if _ls is not None else 0.0625  # type: ignore[attr-defined]
+        self._logit_scale_fold_pending = False
 
         # --- RoPE theta (informational metadata) ---
         # CohereRotaryEmbedding reads config.rope_parameters["rope_theta"] directly;
@@ -154,17 +155,30 @@ class CohereArchitectureAdapter(ArchitectureAdapter):
     def preprocess_weights(self, state_dict: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         """Fold logit_scale into unembed weights before ProcessWeights runs.
 
-        bridge.py lines 726-732 clone unembed.weight before calling this, so
+        bridge.py clones unembed.weight before calling this, so
         scaling does not affect the tied embed.weight.
         logit_scale=1.0 is a no-op (skipped for efficiency).
         """
+        self._logit_scale_fold_pending = False
         scale: float = getattr(self.cfg, "logit_scale")  # always set by __init__
         if scale != 1.0:
             for key in ("unembed.weight", "unembed.bias"):
                 if key in state_dict:
                     orig_dtype = state_dict[key].dtype
                     state_dict[key] = (state_dict[key].float() * scale).to(orig_dtype)
+                    self._logit_scale_fold_pending = True
         return state_dict
+
+    def postprocess_weights(self, bridge: Any) -> None:
+        """Disable HF's outer scale after folding it into the live unembed weights."""
+        if not self._logit_scale_fold_pending:
+            return
+
+        model = getattr(bridge, "original_model", None)
+        if model is None or not hasattr(model, "logit_scale"):
+            raise RuntimeError("Cohere weight processing requires a model-level logit_scale")
+        model.logit_scale = 1.0
+        self._logit_scale_fold_pending = False
 
     def apply_output_logits_transform(self, logits: torch.Tensor) -> torch.Tensor:
         """Match Cohere's ``lm_head -> logit_scale -> optional softcap`` path."""
--- a/transformer_lens/model_bridge/supported_architectures/cohere.py
+++ b/transformer_lens/model_bridge/supported_architectures/cohere.py
@@ -175,9 +175,8 @@ class CohereArchitectureAdapter(ArchitectureAdapter):
             return
 
         model = getattr(bridge, "original_model", None)
-        if model is None or not hasattr(model, "logit_scale"):
-            raise RuntimeError("Cohere weight processing requires a model-level logit_scale")
-        model.logit_scale = 1.0
+        if model is not None and hasattr(model, "logit_scale"):
+            model.logit_scale = 1.0
         self._logit_scale_fold_pending = False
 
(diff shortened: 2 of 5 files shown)


transformer_lens/model_bridge/architecture_adapter.py
```diff
@@ -300,6 +300,13 @@ def preprocess_weights(self, state_dict: dict[str, torch.Tensor]) -> dict[str, t
         """
         return state_dict
 
+    def postprocess_weights(self, bridge: Any) -> None:
```
jlarson4, line 303: `supported_architectures/AGENTS.md` "When to override `preprocess_weights`" tells adapter authors to fold any config knob HF applies in forward, with Cohere's `logit_scale` as its worked example and a copy-paste skeleton. Following it produces exactly this bug, because it never says the forward-time factor also has to be taken out.

Please add `postprocess_weights` to that section: when the factor lives outside the module the bridge wraps, folding it is only half the change.

transformer_lens/model_bridge/supported_architectures/cohere.py
```diff
@@ -154,18 +155,31 @@ def __init__(self, cfg: Any) -> None:
...
+                    self._logit_scale_fold_pending = True
         return state_dict
 
+    def postprocess_weights(self, bridge: Any) -> None:
+        """Disable HF's outer scale after folding it into the live unembed weights."""
+        if not self._logit_scale_fold_pending:
+            return
+
+        model = getattr(bridge, "original_model", None)
+        if model is None or not hasattr(model, "logit_scale"):
```
jlarson4: If `original_model` has no `logit_scale`, its forward isn't applying one, so the fold alone is already correct and there is nothing to neutralize. A Cohere-architecture module in that shape processes cleanly on `dev` and hard-fails here. This occurs after the folded weights are installed and `_weights_processed` is set, so a caller that catches the error is left holding the mis-scaled bridge.

Return instead of raising when the attribute is absent.

jlarson4 (commented):
Thanks for documenting this issue and putting together the fix @emerardd! Just a note on the CI failure, that is a testing issue that is currently being resolved in #1730, and should be ready soon. A couple small comments on this fix, let me know if you have any questions

transformer_lens/model_bridge/architecture_adapter.py
emerardd: Addressed in `1544c4ce`. I expanded the `preprocess_weights` guidance in `supported_architectures/AGENTS.md` to cover the paired post-install lifecycle: when HF applies a factor outside the wrapped component, adapters now track whether the fold occurred and use `postprocess_weights()` to neutralize the outer runtime factor after installation. The guidance and skeleton also call out that a missing runtime attribute means there is no outer factor to neutralize.

transformer_lens/model_bridge/supported_architectures/cohere.py
emerardd: Addressed in `1544c4ce`. `postprocess_weights()` now neutralizes `model.logit_scale` only when the wrapped model exposes it, and always clears the pending fold state without raising when the attribute is absent. I added a network-free regression that removes the model-level attribute, runs `process_weights()`, and verifies that processing completes, the attribute remains absent, the unembed weight is folded by `0.5`, and `cfg.logit_scale` remains `0.5`. Focused validation passed: 3 end-to-end regressions, 74 Cohere/Cohere2 adapter tests, 27 weight-processing tests with 1 skip, formatting, and mypy over 392 source files.

PR #1741, opened 2026-09-03
feat(jacobian_lens): anchored J-space coordinate patching (offline primitive)

### LINKED ISSUE #1739: [Proposal] Add J-space coordinate patching over sparse decompositions
### Proposal 

Add J-space coordinate patching over sparse decompositions

cc @jlarson4 (Would appreciate your review and guidance on this. Thank you)

**Status:** Maintainer-approved. The implementation is complete locally and undergoing final validation before a PR.
**Suggested labels:** `enhancement`, `complexity-moderate`, `TransformerBridge`

### Motivation

This follows the Jacobian-Lens proposal (#1505), sparse-decomposition follow-up (#1539), and merged
decomposition implementation (#1596).

`JacobianLens.decompose(...)` answers which vocabulary-aligned J-space vectors are active in an
activation. `swap_hooks(...)` can exchange coordinates in a basis containing exactly two named
vectors, but it ignores the other co-active vectors found by decomposition.

Researchers currently cannot express: “replace active source concept `s` with target concept `t`
while retaining the other recovered coordinates and reconstruction residual.” Implementing that
delta provides a direct causal consumer of the decomposition introduced in #1596.

PR #1738 is adjacent but non-overlapping: it adds a documentation notebook that analyzes sparse
supports, selected-span variance and occupancy, solver behavior, and whole-span causal ablations.
It deliberately changes no library API and does not repeat coordinate swaps. A selected-span
ablation removes or projects an entire subspace; coordinate patching instead changes one anchored
source/target coordinate while preserving every other recovered coordinate and the decomposition
residual.

### Proposed Semantics

Given activation `x`, dictionary rows `D`, and sparse decomposition

`x = r + D[S]ᵀ c`,

where `S` is active support, `c >= 0`, and `r = x - reconstruction`:

1. Require `source_idx` to be in `S`; otherwise raise rather than pretend the source was active.
2. If `target_idx` is absent from active support, append its atom to the edit basis with coordinate
   zero. Report separately whether it was selected by pursuit but assigned a zero coordinate.
3. `swap`: exchange source and target coordinates.
4. `substitute`: set source to zero and target to the source coordinate, discarding any old target
   coordinate.
5. Interpolate with `c_alpha = c + alpha * (c_edit - c)`.
6. Return `x_patched = x + B @ (c_alpha - c)`. Use an explicit `alpha == 0` fast path returning an
   unchanged clone so the no-op contract is bit exact.

All non-edited coordinates and the anchored residual remain unchanged by construction. The target
append rule makes absent-target substitution well-defined.

### API sketch

Model-free core:

- `solve_coordinate_patch(x, dictionary, source_idx, target_idx, *, decomposition=None,
  k=DEFAULT_K, mode="substitute", alpha=1.0,
  algorithm="nonnegative_orthogonal_matching_pursuit")`
- returns `CoordinatePatch` with supports, coordinates before/after, reconstruction before/after,
  residual, delta, patched activation, selected/append metadata, source-target cosine, and basis
  diagnostics.
- when `decomposition` is supplied, reuse it after strict activation/dictionary compatibility
  validation instead of repeating the vocabulary-scale atom scan.

Bridge wrapper:

- `JacobianLens.coordinate_patch(...)` resolves a prompt/position or raw activation exactly as
  `decompose(...)`, resolves token inputs to dictionary-row indices, then invokes the model-free
  core.

Dynamic `coordinate_patch_hooks(...)` are a separate follow-up because each selected activation
would otherwise require a vocabulary-scale decomposition during the forward pass.

### Correctness contract

- `alpha=0` is an exact no-op.
- `alpha` must be finite, and source/target indices must be valid and distinct.
- Inputs are not mutated.
- The anchored residual is unchanged within float32 compute tolerance.
- Every non-edited coordinate remains bitwise unchanged in the coordinate vector.
- Missing target is appended with coordinate zero.
- Missing source raises.
- Identical source and target raises.
- Near-parallel source/target vectors follow the existing `swap_hooks` warning/error policy.
- The shared near-parallel policy lives in a dependency both intervention paths can import, avoiding
  circular imports and duplicated thresholds.
- Rank and condition number are computed on the column-normalized active-plus-target basis.
- Rank deficiency or high condition number warns but does not fail: anchored editing uses no inverse.
- On an exact two-atom positive decomposition, `swap` agrees with `swap_hooks` within tolerance.
- Output follows the decomposition module's float32 compute convention and preserves its device; it
  does not promise the activation's original storage dtype.

### Proposed repository changes

- New `transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py`.
- Additive methods in `transformer_lens/tools/analysis/jacobian_lens.py`.
- Shared private near-parallel diagnostics with no behavior change to `swap_hooks(...)`.
- Exports from `transformer_lens/tools/analysis/__init__.py`.
- New model-free unit tests and focused wrapper tests.
- One Bridge integration invariant test using a cached model/artifact.
- Short API documentation; no notebook in PR1.

### Alternatives considered

1. **Use `V⁺x` over the expanded basis.** Rejected for absent targets because appending one atom
   can redistribute every coordinate and violate the “preserve the recovered frame” contract.
2. **Use only `swap_hooks`.** It is useful but solves a different two-vector coordinate problem.
3. **Edit all-vocabulary coordinates.** Intractable and not the sparse frame identified by the
   paper or shipped decomposition.
4. **Patch the orthogonal `j_space_component`.** It lacks the nonnegative interpretable coordinates
   required to identify source and target slots.
5. **Ship the behavioral benchmark together.** Deferred because model capability and prompt design
   would obscure review of the algebraic primitive.
6. **Use the decomposition-guided span ablation from PR #1738.** That analysis tests whether the
   selected span matters relative to matched random spans, but it cannot replace one active concept
   with another while retaining the other sparse coordinates.

### Non-goals

- J-lens fitting, registry, loading, readout, occupancy, or sparse-solver semantic changes. PR1 may include the narrow CPU-before-float64 transfer ordering needed for existing decomposition code on MPS, as a separate prerequisite commit.
- Arbitrary multi-concept permutations.
- Dynamic `coordinate_patch_hooks(...)` in PR1.
- Corpus-scale swap-success evaluation.
- Notebook or generated result artifacts.
- HookedTransformer support in PR1.
- Claims that a successful intervention proves unique causal mediation.

### Validation

Unit tests use exact planted dictionaries with analytic expected patches. Integration tests assert
no-op, residual-preservation, finite-output, and restricted pairwise parity—not behavioral success
rates. They also verify that fresh and supplied precomputed decompositions produce the same patch.
Full formatting, typing, unit, integration, and PR test surfaces run before submission.

### Maintainer decisions requested

1. **PR1 scope:** offline model-free primitive plus wrapper, with dynamic hooks deferred?
   **Recommendation: yes.**
2. **Already-active target:** should `substitute` overwrite its old coordinate while `swap` exchanges
   values? **Recommendation: yes, while reporting the overwritten coordinate.**
3. **Conditioning:** should normalized-basis rank deficiency/high condition number warn but remain
   non-fatal? **Recommendation: yes, because the edit performs no inverse.**
4. **Reuse:** should the core accept an optional precomputed `JSpaceDecomposition` for repeated
   edits? **Recommendation: yes, with strict compatibility validation.**

Unless maintainers prefer otherwise, the public names are `solve_coordinate_patch(...)`,
`CoordinatePatch`, and `JacobianLens.coordinate_patch(...)`.

### Checklist

- [x] Existing code and issue lineage reviewed.
- [x] No competing coordinate-patching issue, PR, or source implementation found as of 2026-09-02;
      PR #1738 is a documentation-only decomposition analysis, not a coordinate-edit API.
- [x] Maintainer agreement on anchored-coordinate semantics.
- [x] Maintainer agreement on offline-only PR1 scope.
- [x] Maintainer agreement on target, conditioning, and decomposition-reuse policies.
- [ ] Tracking issue number added to the implementation plan.

janmenjayap:
### Summary

Adds an **offline, model-free coordinate-patching primitive** on top of the shipped J-space sparse
decomposition, plus a thin `JacobianLens` wrapper. Given `x = residual + reconstruction`, it edits
only named coordinates in the active sparse frame and holds `residual = x - reconstruction` and
every other coordinate fixed. `substitute` moves the source coordinate onto the target (zeroing the
source); `swap` exchanges them; an absent target is appended at zero; `alpha` interpolates, with
`alpha=0` an exact no-op.

The edit **inverts nothing** — coordinates come from the anchored decomposition chart, and
`patched = residual + B @ coordinates_after`. The two numerical diagnostics (basis conditioning and
source/target near-parallelism) therefore **warn but never raise**.

This is PR1 of the plan: the offline primitive. Dynamic forward hooks, a causal-swap benchmark, a
notebook, arbitrary multi-slot permutations, and HookedTransformer support are explicitly deferred
to a follow-up.

Fixes #1739 

### Motivation

Sparse decomposition identifies which J-space concepts are active in an activation; nothing yet
edits one of those coordinates while preserving the rest of the recovered frame and the residual.
Coordinate patching returns an activation suitable for intervention studies without silently
re-solving `B⁺x` (which would perturb every coordinate when a target atom is appended).

The plan called for three reviewable units; a fourth prerequisite fix rides along (see "Note on the
MPS commit").

1. **`fix(jacobian_lens): support sparse decomposition on MPS`**
   Stages the selected-span pseudoinverse and the NNLS solve on CPU before their unsupported float64
   conversion, keeping vector outputs on the MPS device. Required for the coordinate patch (and the
   existing decomposition) to run on Apple Silicon. Adds two `tests/mps` cases.

2. **`feat(jacobian_lens): add anchored coordinate-patching core`**
   New `transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py`: the `CoordinatePatch`
   frozen dataclass (report-only, all tensors detached — retains no model/lens/tokenizer/hook/full
   dictionary) and `solve_coordinate_patch(...)`. Includes the Q2 overwrite field, the Q3/Q3-ext
   real warnings, the Q4 precomputed-reuse path with strict validation, and postcondition witnesses.
   Model-free tests first.

3. **`feat(jacobian_lens): expose coordinate patching on the Bridge`**
   `JacobianLens.coordinate_patch(...)` resolves a raw activation or prompt+position exactly like
   `decompose(...)`, resolves token inputs to dictionary rows, then calls the core. **No forward
   hooks.** Also performs the shared near-parallel policy refactor: `_SWAP_WARN_COSINE` /
   `_SWAP_ERROR_COSINE` and a `_diagnose_intervention_pair` helper move into
   `jacobian_lens_decomposition.py`; `swap_hooks` re-points to them with **byte-identical** warn/raise
   behavior.

4. **`feat(jacobian_lens): export coordinate patching with docs and integration test`**
   Exports `CoordinatePatch` and `solve_coordinate_patch` from `transformer_lens.tools.analysis`,
   extends the package docstring, documents the API and warning policy in `jacobian_lens_fitting.md`,
   and adds one cached-model integration test asserting **algebraic invariants only** (no-op at
   `alpha=0`, residual preservation, finite output, unchanged non-edited coordinates) — not a
   token-flip.

### Reviewer decision ledger

| # | Decision | How it's satisfied |
|---|----------|--------------------|
| Q1 | Offline primitive + wrapper; dynamic hooks deferred | Core and wrapper land as separate, independently reviewable commits. |
| Q2 | `substitute` overwrites, `swap` exchanges, overwritten value reported | `overwritten_target_coordinate: float \| None` (`None` for absent target or `swap`); a value-level test asserts the **exact** discarded coordinate, not just field presence. |
| Q3 | Rank-deficient / high-condition basis warns, non-fatal | Real `warnings.warn` naming the **measured condition number**; warn-path test + over-warning control. |
| Q3-ext | Near-parallel source/target warns, non-fatal | Real `warnings.warn` naming the **measured cosine**; warn-path test. Patch warns (never raises) at the parallel extreme because it performs no inverse. |
| Q4 | Optional precomputed `JSpaceDecomposition` | Accepted with strict compatibility validation before reuse; a fresh and a validated-supplied decomposition produce identical patches. |

### Testing

- `tests/unit/tools/test_jacobian_lens_coordinate_patch.py` — model-free core (exact expected
  outputs, alpha semantics, error paths, malformed activation/dictionary, Q2/Q3/Q3-ext/Q4, no-mutation,
  determinism, both decomposition algorithms).
- `tests/unit/tools/test_jacobian_lens.py` — wrapper: raw vs prompt agreement, token resolution,
  restricted `swap_hooks` pseudoinverse oracle, unfitted-layer rejection, warning propagation,
  `swap_hooks` regression after the constant move.
- `tests/integration/test_jacobian_lens.py` — one published-GPT-2 algebraic-invariant test.
- `tests/mps/test_mps_basic.py` — decomposition and patch stay on the MPS device.

Local gates green: unit tier passes; `black`, `isort`, `mypy` clean on touched sources
(`tests/` are excluded from mypy per `pyproject.toml`).

### Out of scope (deferred to a follow-up PR)

Dynamic `coordinate_patch_hooks(...)`, the causal-swap benchmark and any notebook, arbitrary
multi-slot permutations, and HookedTransformer support.

### Note on the MPS commit

Commit 1 (`fix ... support sparse decomposition on MPS`) is a prerequisite, not part of the
reviewer's original three-unit scope. It is kept in this PR because the coordinate patch depends on
it on Apple Silicon and splitting it would introduce a merge-order dependency between two PRs. It is
isolated in its own commit and can be cherry-picked out into a standalone PR on request.

### Checklist

- [x] Base set to `dev`.
- [x] Full `make test-pr` run attached (unit + docstring + acceptance + integration).
- [x] Integration test run against a cached GPT-2 with `.env` sourced.
- [ ] Reviewer sign-off on keeping the MPS fix in-PR vs. splitting it.

### DIFF
--- a/docs/source/content/jacobian_lens_fitting.md
+++ b/docs/source/content/jacobian_lens_fitting.md
@@ -331,14 +331,14 @@ and source and target must be distinct.
 Both diagnostics warn but never raise, because an anchored reconstruction needs no inverse:
 
 - **Poor conditioning.** When the column-normalized active-plus-target basis is rank deficient or its
-  `basis_condition_number` exceeds the threshold, coordinate patching emits a `UserWarning` that
-  names the measured condition number; coordinate attribution is then non-unique, but the edit still
-  completes.
+  `basis_condition_number` exceeds `1 / sqrt(float32 eps)` (≈ 2896), coordinate patching emits a
+  `UserWarning` that names the measured condition number; coordinate attribution is then non-unique,
+  but the edit still completes.
 - **Near-parallel source/target.** When the absolute source–target cosine (reported signed in
-  `source_target_cosine`) exceeds the shared `_SWAP_WARN_COSINE` threshold, coordinate patching emits
-  a `UserWarning` that names the measured cosine, since a swap between near-parallel atoms is close to
-  a no-op. Unlike `swap_hooks`, which raises above `_SWAP_ERROR_COSINE`, the patch core only warns at
-  the parallel extreme.
+  `source_target_cosine`) exceeds `_SWAP_WARN_COSINE` (0.99), coordinate patching emits a
+  `UserWarning` that names the measured cosine, since a swap between near-parallel atoms is close to
+  a no-op. Unlike `swap_hooks`, which raises above `_SWAP_ERROR_COSINE` (0.999), the patch core only
+  warns at the parallel extreme.
 
 Read the measured condition number and cosine straight out of each warning message (and from the
 `basis_condition_number` and `source_target_cosine` fields on the returned `CoordinatePatch`) to
--- a/docs/source/content/jacobian_lens_fitting.md
+++ b/docs/source/content/jacobian_lens_fitting.md
@@ -273,8 +273,7 @@ So two vector outputs also need not coincide:
 
 - `reconstruction` -- the nonnegative combination over the active `support`.
 - `j_space_component` (the *J-space component*) -- the orthogonal projection of the activation onto
-  the span of `selected_support` -- with `non_j_space_component = x - j_space_component`, the
-  residual the interventions leave unchanged.
+  the span of `selected_support` -- with `non_j_space_component = x - j_space_component`.
 
 For the default exact NNLS re-solve the `reconstruction` equals the projection onto the *active*
 support (KKT stationarity), so it differs from `j_space_component` exactly when a selected vector
@@ -298,6 +297,57 @@ the paper; its projected step is accepted only when it does not increase the res
 algorithms share the same greedy selection *rule* but, because their coefficient residuals
 differ, may select different vectors at later steps and so return a different `support` and
 `reconstruction`.
+
+### Coordinate patching
+
+`JacobianLens.coordinate_patch` turns a decomposition into an anchored causal edit. Given
+`x = residual + reconstruction`, it changes only named coordinates in the active sparse frame and
+keeps `residual = x - reconstruction` plus every other coordinate fixed. This residual differs from
+`non_j_space_component` when `selected_support` contains zero-coordinate atoms.
+
+```python
+prompt = "The Eiffel Tower is in the city of"
+decomposition = lens.decompose(model, prompt, layer=6, position=-1, k=8)
+source_id = int(decomposition.support[0])
+patch = lens.coordinate_patch(
+    model,
+    prompt,
+    layer=6,
+    source_token=source_id,
+    target_token=" Paris",
+    position=-1,
+    decomposition=decomposition,
+    mode="substitute",
+)
+patched_activation = patch.patched
+```
+
+`substitute` sets the source coordinate to zero and the target to the source value, overwriting an
+existing target coordinate (the discarded value is reported in `overwritten_target_coordinate`).
+`swap` exchanges the two values. An absent target is appended at zero, and `alpha` interpolates
+between the original and edited coordinates (`alpha=0` is an exact no-op). The source must be active
+and source and target must be distinct.
+
+Both diagnostics warn but never raise, because an anchored reconstruction needs no inverse:
+
+- **Poor conditioning.** When the column-normalized active-plus-target basis is rank deficient or its
+  `basis_condition_number` exceeds the threshold, coordinate patching emits a `UserWarning` that
+  names the measured condition number; coordinate attribution is then non-unique, but the edit still
+  completes.
+- **Near-parallel source/target.** When the absolute source–target cosine (reported signed in
+  `source_target_cosine`) exceeds the shared `_SWAP_WARN_COSINE` threshold, coordinate patching emits
+  a `UserWarning` that names the measured cosine, since a swap between near-parallel atoms is close to
+  a no-op. Unlike `swap_hooks`, which raises above `_SWAP_ERROR_COSINE`, the patch core only warns at
+  the parallel extreme.
+
+Read the measured condition number and cosine straight out of each warning message (and from the
+`basis_condition_number` and `source_target_cosine` fields on the returned `CoordinatePatch`) to
+judge how much to trust the edit.
+
+Without `decomposition=`, coordinate patching runs the vocabulary-scale sparse decomposition first.
+Reuse a compatible result for repeated edits to avoid that scan. The method returns an offline
+activation and diagnostics, not forward hooks; applying patches dynamically would otherwise perform
+a vocabulary-scale solve inside the model run.
 
 ### Interpreting the numbers honestly
 
--- a/tests/unit/tools/test_jacobian_lens_coordinate_patch.py
+++ b/tests/unit/tools/test_jacobian_lens_coordinate_patch.py
@@ -416,3 +416,74 @@ def test_well_conditioned_control_emits_no_warning() -> None:
         result = solve_coordinate_patch(activation, dictionary, 0, 2, k=2)
 
     assert torch.isfinite(result.patched).all()
+
+
+def test_large_d_model_full_rank_basis_reports_finite_condition() -> None:
+    """C2: the rank tolerance must scale by the column count, not the d_model row count.
+
+    A full-rank but poorly conditioned edit basis stays finite at model scale. Scaling the
+    rank tolerance by ``max(shape)`` (the d_model row count) instead crosses the smallest
+    singular value once ``d_model >= ~2896``, reporting the basis as rank deficient with an
+    infinite condition number on Llama/Mistral/Gemma-scale dictionaries.
+    """
+    d_model = 3000
+
+    def _atom(*entries: tuple[int, float]) -> torch.Tensor:
+        atom = torch.zeros(d_model)
+        for index, value in entries:
+            atom[index] = value
+        return atom
+
+    dictionary = torch.stack(
+        [_atom((0, 1.0)), _atom((1, 1.0)), _atom((0, 1.0), (1, 1.0), (2, 1e-4)), _atom((3, 1.0))]
+    )
+    activation = dictionary[0] + dictionary[1] + dictionary[2]
+    selected_atoms = dictionary[[0, 1, 2]].T
+    component = selected_atoms @ torch.linalg.pinv(selected_atoms) @ activation
+    decomposition = JSpaceDecomposition(
+        support=torch.tensor([0, 1, 2]),
+        coordinates=torch.ones(3),
+        selected_support=torch.tensor([0, 1, 2]),
+        reconstruction=activation.clone(),
+        j_space_component=component,
+        non_j_space_component=activation - component,
+    )
+
+    with pytest.warns(UserWarning, match="condition number="):
+        result = solve_coordinate_patch(activation, dictionary, 0, 3, decomposition=decomposition)
+
+    assert result.basis_rank == 4
+    assert math.isfinite(result.basis_condition_number)
+
+
+def test_high_condition_decomposition_is_accepted() -> None:
+    """C8: the compatibility tolerance must scale with the selected span's condition number.
+
+    An exactly valid, hand-built decomposition is rejected by a fixed 32-eps tolerance when the
+    activation lies along the near-null direction of an ill-conditioned selected span: the
+    float32 pinv recompute of the projection then diverges by ~cond*eps. Scaling the tolerance
+    with the measured condition number keeps the genuine decomposition valid.
+    """
+    separation = 1e-3  # atom1/atom2 near-parallel; their difference direction is e2
+    atom0 = torch.tensor([1.0, 0.0, 0.0, 0.0])  # active support and edit source
+    atom1 = torch.tensor([0.0, 1.0, 0.0, 0.0])
+    atom2 = torch.tensor([0.0, 1.0, separation, 0.0])
+    atom2 = atom2 / atom2.norm()
+    target_atom = torch.tensor([0.0, 0.0, 0.0, 1.0])  # appended by the edit
+    dictionary = torch.stack([atom0, atom1, atom2, target_atom])
+
+    # The activation sits in the selected span but concentrates on e2, the ill-conditioned
+    # (near-null) direction of the atom1/atom2 pair, so recovering it needs the amplifying inverse.
+    activation = atom0 + torch.tensor([0.0, 0.0, 1.0, 0.0])
+    decomposition = JSpaceDecomposition(
+        support=torch.tensor([0]),
+        coordinates=torch.tensor([1.0]),  # NNLS fit of the activation over the active atom e0
+        selected_support=torch.tensor([0, 1, 2]),
+        reconstruction=atom0.clone(),
+        j_space_component=activation.clone(),  # activation lies entirely in the selected span
+        non_j_space_component=torch.zeros(4),
+    )
+
+    result = solve_coordinate_patch(activation, dictionary, 0, 3, decomposition=decomposition)
+
+    assert torch.isfinite(result.patched).all()
--- a/tests/unit/tools/test_jacobian_lens_coordinate_patch.py
+++ b/tests/unit/tools/test_jacobian_lens_coordinate_patch.py
@@ -41,9 +41,36 @@ def test_substitute_appends_absent_target_and_preserves_frame() -> None:
     torch.testing.assert_close(result.patched, result.residual + result.reconstruction_after)
     torch.testing.assert_close(result.delta, torch.tensor([-2.0, 0.0, 2.0, 0.0, 0.0]))
     assert result.target_was_appended is True
+    assert result.target_was_selected is False
     assert result.overwritten_target_coordinate is None
     assert result.nonedited_coordinate_max_delta == pytest.approx(0.0, abs=1e-6)
     assert result.residual_max_delta == pytest.approx(0.0, abs=1e-6)
+
+
+def test_target_was_selected_true_when_pursuit_drops_it_to_zero() -> None:
+    """Q6: a target selected by pursuit but assigned a zero coordinate is distinguishable from a
+    target pursuit never considered -- both leave ``target_was_appended`` True, but only the
+    former leaves ``target_was_selected`` True."""
+    activation, dictionary = _problem()
+    decomposition = get_sparse_decomposition(activation, dictionary, k=2)
+    # Atom 2 is orthogonal to both selected atoms and to the activation, so folding it into
+    # ``selected_support`` (as if pursuit had picked it and the NNLS re-solve zeroed it) doesn't
+    # perturb ``j_space_component`` or any other validated quantity.
+    selected_with_dropped_target = replace(
+        decomposition,
+        selected_support=torch.cat([decomposition.selected_support, torch.tensor([2])]),
+    )
+
+    result = solve_coordinate_patch(
+        activation,
+        dictionary,
+        source_idx=0,
+        target_idx=2,
+        decomposition=selected_with_dropped_target,
+    )
+
+    assert result.target_was_appended is True
+    assert result.target_was_selected is True
 
 
 def test_active_target_substitute_overwrites_while_swap_exchanges() -> None:
@@ -62,6 +89,8 @@ def test_active_target_substitute_overwrites_while_swap_exchanges() -> None:
     torch.testing.assert_close(swap.patched, torch.tensor([5.0, 2.0, 0.0]))
     assert substitute.target_was_appended is False
     assert swap.target_was_appended is False
+    assert substitute.target_was_selected is True
+    assert swap.target_was_selected is True
 
 
 def test_overwritten_target_coordinate_reports_exact_discarded_value() -> None:
@@ -83,6 +112,23 @@ def test_overwritten_target_coordinate_reports_exact_discarded_value() -> None:
     assert substitute.overwritten_target_coordinate == pytest.approx(7.25)
     assert swap.overwritten_target_coordinate is None
     assert absent.overwritten_target_coordinate is None
+
+
+def test_overwritten_target_coordinate_is_alpha_independent_hypothetical_value() -> None:
+    """Q10: overwritten_target_coordinate reports the coordinate the requested substitute would
+    discard at full strength, independent of alpha -- so at alpha=0, where patched equals the
+    input and nothing is actually discarded, the field is still the hypothetical value rather
+    than None."""
+    dictionary = torch.eye(3)
+    activation = torch.tensor([2.0, 7.25, 0.0])
+    decomposition = get_sparse_decomposition(activation, dictionary, k=2)
+
+    result = solve_coordinate_patch(
+        activation, dictionary, 0, 1, decomposition=decomposition, mode="substitute", alpha=0.0
+    )
+
+    assert torch.equal(result.patched, activation)
+    assert result.overwritten_target_coordinate == pytest.approx(7.25)
 
 
 def test_alpha_interpolates_and_zero_is_bit_exact() -> None:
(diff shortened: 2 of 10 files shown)


transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+        patched=patched.detach(),
+        nonedited_coordinate_max_delta=nonedited_max,
+        residual_max_delta=residual_max,
+        source_target_cosine=cosine,
+        basis_rank=rank,
+        basis_condition_number=condition,
+    )
+
+
+def _postcondition_witnesses(
```
jlarson4, line 494: Both witnesses are algebraic identities of the construction (`patched − recon_after ≡ residual` by definition; the non-edited delta is `alpha·0`), so a basis/coordinate misalignment, a dropped edit, or an ignored alpha all pass them. Recompute the check from the dictionary rows (`patched ≈ residual + dictionary[support_after].T @ coordinates_after`) so the witness has an independent component.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+
+def _basis_diagnostics(basis: torch.Tensor) -> tuple[int, float]:
+    """Rank and condition number of the column-normalized edit basis, warning when ill-conditioned.
+
+    Both warnings are non-fatal: the anchored edit performs no inverse of ``basis``, so a
+    rank-deficient or poorly conditioned basis only means coordinate attribution is non-unique.
+    """
+    normalized = basis / torch.linalg.vector_norm(basis, dim=0, keepdim=True)
+    diagnostic_basis = normalized.detach().cpu().double()
+    rank_rtol = torch.finfo(torch.float32).eps * max(diagnostic_basis.shape)
```
jlarson4: `rank_rtol` scales by `max(shape)` (the row count), so for `d_model ≥ ~2896` matrix_rank declares rank-deficiency before the warn threshold is reachable, the warning always says `condition number=inf` on Llama/Mistral/Gemma-scale bases and `basis_rank` under-reports genuinely full-rank ones. Scale `rank_rtol` by the column count instead, which keeps the finite branch live at every d_model.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+        decomposition.non_j_space_component,
+    )
+    if any(vector.ndim != 1 or vector.shape[0] != d_model for vector in vectors):
+        raise ValueError("precomputed decomposition vector outputs must have shape [d_model]")
+    local_vectors = [vector.float().to(dictionary.device) for vector in vectors]
+    if not all(bool(torch.isfinite(vector).all()) for vector in local_vectors):
+        raise ValueError("precomputed decomposition contains non-finite vector outputs")
+    support_atoms = dictionary[support_cpu].float()
+    expected_reconstruction = support_atoms.T @ coordinates
+    if not _close(local_vectors[0], expected_reconstruction):
```
jlarson4: `reconstruction`/`coordinates` are validated against each other and the dictionary, but nothing checks they fit the supplied activation. A self-consistently scaled pair passes and yields a silently wrong patch. Add an x-coupled check: the sibling `_validate_nnls_kkt` stationarity test, or `reconstruction ≈ j_space_component` when the supports coincide.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+
+    reconstruction_before = decomposition.reconstruction.float().to(dictionary.device).clone()
+    residual = target - reconstruction_before
+    if float(alpha) == 0.0:
+        reconstruction_after = reconstruction_before.clone()
+        delta = torch.zeros_like(target)
+        patched = target.clone()
+    else:
+        reconstruction_after = basis @ coordinates_after
+        delta = reconstruction_after - reconstruction_before
```
jlarson4: `delta = basis @ c_after − reconstruction_before` carries an alpha-independent recompute floor (~2e-6 at model scale), making `patched` discontinuous at alpha=0 and `delta` meaningless below alpha≈1e-4. `basis @ (c_after − c_before)` removes the floor and the discontinuity, retires the alpha=0 special case, and makes comment 1's residual witness non-vacuous.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+    if mode == "swap":
+        edit_coordinates[source_slot] = target_coordinate
+        overwritten_target_coordinate: Optional[float] = None
+    else:
+        edit_coordinates[source_slot] = 0.0
+        overwritten_target_coordinate = (
+            float(target_coordinate.item()) if target_was_active else None
+        )
+    edit_coordinates[target_slot] = source_coordinate
+    coordinates_after = coordinates_before + float(alpha) * (edit_coordinates - coordinates_before)
```
jlarson4, line 438: alpha=2.0 or −1.0 silently produces negative coordinates in a frame defined by `c ≥ 0`. Warn (or flag on the result) when `coordinates_after` exits the nonnegative orthant, or document extrapolation explicitly.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+        basis_condition_number: Condition number of that basis, or infinity when rank deficient.
+    """
+
+    support_before: torch.Tensor
+    support_after: torch.Tensor
+    coordinates_before: torch.Tensor
+    coordinates_after: torch.Tensor
+    source_slot: int
+    target_slot: int
+    target_was_appended: bool
```
jlarson4, line 93: The issue's append rule asks to "report separately whether it was selected by pursuit but assigned a zero coordinate"; `target_was_appended` alone can't express that. Add `target_was_selected` (or an enum) so the distinction survives into the result.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+        raise ValueError("precomputed decomposition contains non-finite vector outputs")
+    support_atoms = dictionary[support_cpu].float()
+    expected_reconstruction = support_atoms.T @ coordinates
+    if not _close(local_vectors[0], expected_reconstruction):
+        raise ValueError("precomputed decomposition reconstruction is incompatible with dictionary")
+    if not _close(local_vectors[1] + local_vectors[2], target):
+        raise ValueError("precomputed decomposition is incompatible with activation")
+    if selected_cpu.numel():
+        selected_atoms = dictionary[selected_cpu].float().T
+        if target.device.type == "mps":
```
jlarson4: This is the third copy of the device-detect → `.cpu()` → op → restore idiom (with `jacobian_lens_decomposition.py:476` and `projection_kernel.py:366`). Extract one helper, the way this PR already consolidated `_SWAP_WARN_COSINE`.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+def _validate_index(name: str, value: int, num_atoms: int) -> int:
+    if isinstance(value, bool) or not isinstance(value, Integral):
+        raise ValueError(f"{name} must be an integer atom index")
+    index = int(value)
+    if not 0 <= index < num_atoms:
+        raise ValueError(f"{name}={index} out of range for dictionary with {num_atoms} atoms")
+    return index
+
+
+def _close(actual: torch.Tensor, expected: torch.Tensor) -> bool:
```
jlarson4: The selected-span recompute error scales with the span's condition number, but `_close` uses a fixed 32-eps tolerance — an exactly-valid hand-built decomposition at cond≈2800 is rejected as "incompatible with dictionary". Lets scale the tolerance with the measured condition.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,434 @@
...
+    atoms = dictionary[indices].float()
+    if not bool(torch.isfinite(atoms).all()):
+        raise ValueError(f"{context} contains a non-finite dictionary atom")
+    if bool((torch.linalg.vector_norm(atoms, dim=1) == 0).any()):
+        raise ValueError(f"{context} contains a zero-norm dictionary atom")
+    return atoms
+
+
+def _validate_index(name: str, value: int, num_atoms: int) -> int:
+    if isinstance(value, bool) or not isinstance(value, Integral):
```
jlarson4, line 143: A boolean token input coerces to id 1 before the bool rejection can fire, yielding "source_idx=1 is not in the active support". Screen out bools in the wrapper's token resolution first.

tests/unit/tools/test_jacobian_lens_coordinate_patch.py
```diff
@@ -0,0 +1,387 @@
...
+        activation, dictionary, 0, 1, decomposition=decomposition, mode="swap"
+    )
+
+    torch.testing.assert_close(substitute.patched, torch.tensor([0.0, 2.0, 0.0]))
+    torch.testing.assert_close(swap.patched, torch.tensor([5.0, 2.0, 0.0]))
+    assert substitute.target_was_appended is False
+    assert swap.target_was_appended is False
+
+
+def test_overwritten_target_coordinate_reports_exact_discarded_value() -> None:
```
jlarson4, line 96: `overwritten_target_coordinate` is computed before alpha blends, so at alpha=0 it reports a value nothing discarded. Decide whether that is the intended reading (hypothetical full-strength report) or it should be None on a no-op, document it, and add the one-line test either way.

docs/source/content/jacobian_lens_fitting.md
```diff
@@ -299,6 +298,57 @@ algorithms share the same greedy selection *rule* but, because their coefficient
...
+and source and target must be distinct.
+
+Both diagnostics warn but never raise, because an anchored reconstruction needs no inverse:
+
+- **Poor conditioning.** When the column-normalized active-plus-target basis is rank deficient or its
+  `basis_condition_number` exceeds the threshold, coordinate patching emits a `UserWarning` that
+  names the measured condition number; coordinate attribution is then non-unique, but the edit still
+  completes.
+- **Near-parallel source/target.** When the absolute source–target cosine (reported signed in
+  `source_target_cosine`) exceeds the shared `_SWAP_WARN_COSINE` threshold, coordinate patching emits
```
jlarson4: The new section names `_SWAP_WARN_COSINE`/`_SWAP_ERROR_COSINE` (private identifiers) without their values, and "exceeds the threshold" never says ~2896. Give readers the numbers; drop or keep the identifiers as you prefer.

jlarson4 (commented):
Thanks for putting this together @janmenjayap, the core features are great! Take a look at my review edits below and let me know if you have any questions

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 69fc098c. The residual witness now recomputes the reconstruction independently from the dictionary rows (`basis @ coordinates_after`, where `basis = dictionary[support_after].T`) rather than reusing the `delta` that built `patched`. So `recovered_residual = patched - basis @ coordinates_after` no longer cancels as an algebraic identity of the construction; it now depends on `reconstruction_before` matching `basis @ coordinates_before` from an independent code path (the decomposition object), so a basis/coordinate misalignment surfaces here.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 5f735eb1. `rank_rtol` now scales by the column count (`diagnostic_basis.shape[1]`, the number of edit atoms) instead of `max(shape)` (the `d_model` row count), so the finite condition-number branch stays live at every `d_model` instead of forcing `condition=inf` past approximately 2896. Added `test_large_d_model_full_rank_basis_reports_finite_condition` at `d_model=3000` asserting a full-rank basis now reports a finite condition number.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in ac304a36. `_validate_decomposition` now reuses the solver's `_validate_nnls_kkt` stationarity check, coupling `coordinates` to the activation `target` itself rather than just to the dictionary, since the active coordinates are already validated `> 0`, stationarity is the exact condition for `coordinates` being the nonnegative fit of `target`. Tolerances mirror the solver's own float32 certification so genuinely rounded decompositions still pass. Added `test_rejects_scaled_but_self_consistent_decomposition`, which scales `(reconstruction, coordinates)` together (passes every self-referential check) and confirms it's now rejected with a `ValueError` mentioning "activation".

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 69fc098c. `delta` is now `basis @ (coordinates_after - coordinates_before)` instead of `basis @ coordinates_after - reconstruction_before`, so it's exactly zero at `alpha=0` and strictly proportional to alpha with no recompute floor. The `alpha == 0.0` special case is retired, and `reconstruction_after` is anchored as `reconstruction_before + delta` so the residual is preserved exactly. Added `test_small_alpha_delta_is_proportional_not_floored`, which asserts `delta` at `alpha=1e-4` is proportional to `delta` at `alpha=1` (within `rtol=1e-3`) and stays below `1e-3` in magnitude, so no floor survives.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in f7499033. Added `_warn_if_alpha_extrapolates`, which warns (non-fatal, same style as `_warn_if_near_parallel`/`_basis_diagnostics`) and sets a new `coordinates_after_nonnegative` field on the result whenever `coordinates_after` exits `c >= 0`. `alpha` in `[0, 1]` always stays in-orthant; `alpha` outside that range (for example, `2.0` or `-1.0`) is now explicitly flagged as an extrapolation. Covered by `test_alpha_within_unit_interval_stays_in_nonnegative_orthant` and `test_alpha_extrapolation_beyond_orthant_warns_and_flags`.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in f7499033. Added `target_was_selected`, computed from `decomposition.selected_support`, so `target_was_appended and not target_was_selected` means pursuit never considered the target, while `target_was_appended and target_was_selected` means pursuit selected it but assigned it a zero coordinate. Documented on the dataclass. Covered by `test_target_was_selected_true_when_pursuit_drops_it_to_zero`.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 5f735eb1. `_close` now takes an optional `condition_number` that scales the fixed 32-eps tolerance; `_validate_decomposition` measures the selected span's condition number once via a new `_condition_number` helper (MPS-safe, built on the C7 helper) and threads it through all three compatibility checks. Added `test_high_condition_decomposition_is_accepted`, a hand-built decomposition where the activation concentrates on the near-null direction of an ill-conditioned atom pair, previously rejected, now validates.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 0da6667e. Extracted `_linalg_on_cpu_if_mps` into `jacobian_lens_decomposition.py` and replaced all three call sites: the pinv block in `_validate_decomposition` (`jacobian_lens_coordinate_patch.py`), the pinv block in `get_sparse_decomposition` (`jacobian_lens_decomposition.py` itself), and `_singular_values` in `projection_kernel.py`. `jacobian_lens_decomposition.py` has no intra-analysis imports, so `projection_kernel.py` importing it introduces no cycle. Pure refactor: CPU path is bit-identical, MPS branch preserved, no test changes needed.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Fixed in 511e0515. `_to_token_ids` now raises `ValueError` on a `bool` token before the bare `int(token)` call, at the wrapper's token-resolution boundary; the core `_validate_index` already rejected bools, but only after the misleading id had already been produced. Added `test_to_token_ids_rejects_bool`, parametrized over `True` and `False`.

tests/unit/tools/test_jacobian_lens_coordinate_patch.py
janmenjayap: Landed in f7499033. Decided the reading: `overwritten_target_coordinate` is alpha-independent by design: it reports the coordinate a `substitute` targets for replacement at full strength, not the blended outcome, so it's documented as non-`None` even at `alpha=0` where `patched` equals the input and nothing is actually discarded. Added `test_overwritten_target_coordinate_is_alpha_independent_hypothetical_value` to lock in that reading and guard against a future change silently flipping it.

docs/source/content/jacobian_lens_fitting.md
janmenjayap: Fixed in 7079e70a. The docs now state the numbers inline: `basis_condition_number` warns above `1 / sqrt(float32 eps)` (approximately 2896), `_SWAP_WARN_COSINE` is `0.99`, and `_SWAP_ERROR_COSINE` (used by `swap_hooks`) is `0.999`. Kept the identifiers alongside the values and the existing field pointers (`source_target_cosine`, `basis_condition_number`).

jlarson4:
Looks good great @janmenjayap! Merging now. I am going to spend some time reviewing your proposal from #1742 this morning, I will tag you on that thread once I have my thoughts together

PR #1745, opened 2026-09-04
fix(head_detector): forward keyword arguments on the multi-prompt path, and stop mutating the ActivationCache

WatchTree-19:
### 1. Every keyword argument is silently dropped when `seq` is a list

`detect_head` recurses positionally on the multi-prompt path:

```python
batch_scores = [detect_head(model, seq, detection_pattern) for seq in seq]
```

so `error_measure`, `exclude_bos`, `exclude_current_token` and `heads` are all reset to their defaults. The call returns a number, on a different scale, with no warning.

Measured on `gpt2-small` with a 4-prompt list, comparing the list result against averaging the per-prompt results computed correctly:

| argument passed | max abs error vs correct | note |
|---|---|---|
| `error_measure="abs"` | 0.4217 | silently computes `"mul"` instead; sign flips on some heads |
| `heads={0: [0, 1]}` | 1.1465 | filter ignored entirely, all heads scored |
| `exclude_bos=True` | 0.0937 | |
| `exclude_current_token=True` | 0.1483 | |

`"abs"` and `"mul"` are not the same scale and are not monotonically related, so the sign flip is not cosmetic: heads rank differently.

This repo's own `demos/Head_Detector_Demo.ipynb` calls `detect_head` on a list of prompts with `error_measure="abs"`, so the published plots in that notebook are `"mul"` scores.

### 2. `compute_head_attention_similarity_score` masks the cache in place

```python
attention_pattern[:, 0, :] = 0          # exclude_bos
attention_pattern[:, range(n), range(n)] = 0   # exclude_current_token
```

`attention_pattern` is a view into the caller's `ActivationCache`, so those writes are permanent. Row sums of a cached attention pattern after one call with `exclude_bos=True`:

```
before: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
after:  [0.0, 0.1529, 0.3884, 0.7915, 0.8771, 0.9045, 0.9528]
```

Any later read of that cache, by this function or anything else, sees altered activations. Row 0 is now entirely zero. This also makes `detect_head` order-dependent: run it twice on the same cache and the second call gets different numbers.

## Fix

- Forward all keyword arguments on the recursive call.
- `.clone()` the pattern before masking.

## Tests

Seven regression tests added to `tests/integration/test_head_detector.py`: one per forwarded keyword argument, asserting the list path matches the manually-averaged per-prompt result; three asserting the cache is unmodified and its rows still sum to 1.

Full file passes (67 tests). `black` clean.

One pre-existing failure, `test_detect_head_with_invalid_head_name`, is unrelated: it takes a single string, never enters the list branch, and depends on jaxtyping/beartype runtime checking being active in the environment. It fails identically on unpatched `main` here.

### DIFF
--- a/transformer_lens/head_detector.py
+++ b/transformer_lens/head_detector.py
@@ -113,7 +113,18 @@ def detect_head(
     if isinstance(detection_pattern, str):
         assert detection_pattern in HEAD_NAMES, INVALID_HEAD_NAME_ERR % detection_pattern
         if isinstance(seq, list):
-            batch_scores = [detect_head(model, seq, detection_pattern) for seq in seq]
+            batch_scores = [
+                detect_head(
+                    model,
+                    batch_seq,
+                    detection_pattern,
+                    heads=heads,
+                    exclude_bos=exclude_bos,
+                    exclude_current_token=exclude_current_token,
+                    error_measure=error_measure,
+                )
+                for batch_seq in seq
+            ]
             return torch.stack(batch_scores).mean(0)
         detection_pattern = cast(
             torch.Tensor,
@@ -248,6 +259,12 @@ def compute_head_attention_similarity_score(
     # mul
 
     if error_measure == "mul":
+        # Clone before masking. attention_pattern is a view into the caller's
+        # ActivationCache, so masking in place permanently corrupts the cached
+        # pattern: its rows stop summing to 1 and every later use of that cache
+        # silently reads modified activations.
+        if exclude_bos or exclude_current_token:
+            attention_pattern = attention_pattern.clone()
         if exclude_bos:
             attention_pattern[:, 0] = 0
         if exclude_current_token:
--- a/transformer_lens/head_detector.py
+++ b/transformer_lens/head_detector.py
@@ -26,6 +26,12 @@ INVALID_HEAD_NAME_ERR = (
     f"detection_pattern must be a Tensor or one of head names: {HEAD_NAMES}; got %s"
 )
 
+CACHE_WITH_SEQ_LIST_ERR = (
+    "A single cache cannot be reused across multiple prompts, so `cache` is not\n"
+    "supported when `seq` is a list. Pass one prompt at a time, or omit `cache`\n"
+    "and let each prompt be run separately."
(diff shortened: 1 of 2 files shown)


transformer_lens/head_detector.py
```diff
@@ -113,7 +113,18 @@ def detect_head(
     if isinstance(detection_pattern, str):
         assert detection_pattern in HEAD_NAMES, INVALID_HEAD_NAME_ERR % detection_pattern
         if isinstance(seq, list):
-            batch_scores = [detect_head(model, seq, detection_pattern) for seq in seq]
+            batch_scores = [
+                detect_head(
```
jlarson4, line 128: `cache` is still silently dropped on the multi-prompt path. The same failure mode this PR fixes for the other four arguments and forwarding it as-is would be wrong since a single cache can't serve every prompt. Please make the drop explicit: raise (or warn) when `cache is not None` and `seq` is a list, or document that the list path ignores it.

jlarson4 (commented):
Thanks for putting this together @WatchTree-19! Just one comment, and would you please retarget the PR to `dev`?

transformer_lens/head_detector.py
WatchTree-19: @jlarson4 you are right, and it is the same failure as the four arguments this PR was already fixing. Thanks.

Forwarding `cache` is not an option — an `ActivationCache` holds one prompt's activations, so reusing it across a list would score every prompt against the first one's attention patterns. So the list path now rejects it:

```python
if isinstance(seq, list):
    if cache is not None:
        raise ValueError(CACHE_WITH_SEQ_LIST_ERR)
```

with the message pointing at passing one prompt at a time, or omitting `cache` and letting each prompt be run separately.

Three tests added: passing a cache with a list raises, passing a cache with a single prompt still matches the uncached result, and the list path without a cache is unchanged.

Happy to downgrade it to a warning instead if you would rather not break anyone who is currently passing a cache and silently getting it ignored — but a raise seemed right given the result they are getting today is wrong rather than merely unoptimised.

PR #1747, opened 2026-09-05
Add clean-coordinate clamping for JacobianLens swaps

### LINKED ISSUE #1746: `JacobianLens.swap_hooks` re-reads live coordinates, causing odd/even cancellation across layer bands
## Describe the bug

`JacobianLens.swap_hooks` re-reads lens coordinates from the live activation at every hooked layer:

```python
coords = selected.float() @ local_pinv.T
delta = alpha * ((coords[..., [1, 0]] - coords) @ local_basis.T)
```

A coordinate exchange is an involution. Across a band whose in-plane coordinates are approximately preserved, the second hook therefore undoes the first, the third redoes it, and so on. This makes the intervention alternate with band-width parity; at `alpha=2`, the antisymmetric coordinate is multiplied by `-3` per application and can diverge rapidly.

This differs from the paper's clamp control, which uses clean-pass coordinate values. The released hook instead exchanges whatever coordinates are present at each layer.

Across three fitted lenses, the odd-minus-even paired margin effect was:

- Qwen3.5-9B: +2.13 nat, 95% CI [1.26, 3.07]
- Qwen3.5-4B: +1.87 nat, 95% CI [1.09, 2.57]
- Qwen3-4B (dense attention, independently fitted): +3.99 nat, 95% CI [1.66, 6.49] (+3.54 after fp32 refitting)

On Qwen3.5-9B, band widths 1/2/3/4/13 gave flip rates `.19/.00/.19/.02/.10`, while a clean-coordinate clamp rose to `.41`. At width 13, `alpha=2` used about `3e10` times the intervention energy of `alpha=1`.

Odd/even parity across three models

Alpha divergence

## Code example

With any model and fitted lens, compare consecutive band widths:

```python
def margin(layers):
    hooks = lens.swap_hooks(model, source_token, target_token, layers=layers)
    with model.hooks(fwd_hooks=hooks):
        logits = model(tokens)[0, -1].float().log_softmax(-1)
    return float(logits[target_answer] - logits[source_answer])

print(margin([L]))
print(margin([L, L + 1]))      # approximately returns toward clean
print(margin([L, L + 1, L + 2]))
```

The full experiment and artifacts are public at https://github.com/koriyoshi2041/rlens-probe-swap.

## Suggested fix

Keep the current live-coordinate exchange for backward compatibility, but expose a clamp form that fixes each layer's target to the exchanged coordinates from a clean run:

```text
h <- h + V (c_target - V⁺h)
```

The existing method should also document that repeated live swaps may cancel across layers, and direct multi-layer users to the clamp form.

## System info

- TransformerLens main at `c03d51037e32ae5e34b3e23d63f63d0e6e16bd1` (the same implementation remains on current `main` and in v3.8.1)
- Source installation
- Linux, CUDA, Python 3.11
- Models: Qwen3.5-9B, Qwen3.5-4B, Qwen3-4B via `TransformerBridge`

## Checklist

- [x] I checked for similar open and closed issues and PRs in the repository.

koriyoshi2041:
# Description

Adds `JacobianLens.swap_clamp_hooks`, which fixes each hooked layer's source/target lens coordinates to the exchanged values from a clean-run cache. This implements the multi-layer clamp form without changing the existing live-coordinate semantics of `swap_hooks`.

Also documents that repeated live swaps are involutive and can cancel across a layer band.

Fixes #1746

## Type of change

### Screenshots

Not applicable; numerical evidence is attached to #1746.

# Validation

- `make format`
- `uv run pytest tests/unit/tools/test_jacobian_lens.py -q` (93 passed)
- `uv run mypy transformer_lens/tools/analysis/jacobian_lens.py`
- `git diff --check`

# Checklist

- [x] New and existing relevant unit tests pass locally with my changes


transformer_lens/tools/analysis/jacobian_lens.py
jlarson4: The `layers` entry still reads "Layers to intervene at (the paper clamps the swap across an intermediate-layer band)". Anyone reading Args rather than the summary gets the advice this paragraph was added to retract, and "clamps" now names a different method. Please drop the parenthetical from this line to improve accuracy.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -1375,6 +1381,102 @@ def transform(
             )
         return hooks
 
+    def swap_clamp_hooks(
```
jlarson4, line 1408: `demos/Jacobian_Lens_Demo.ipynb` line 352 says "The swap is clamped at every token position across an intermediate-layer band (layers 10–24 …)", but line 394 and the 96-condition sweep at line 715 both call `swap_hooks` over `range(10, 25)`, at alpha 1 and 2. The notebook's stored results are published as the clamp protocol while running the form this PR documents as wrong for bands. Can we update the demo to properly document `swap_clamp_hooks`?

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -1375,6 +1381,102 @@ def transform(
...
+        self,
+        model: Any,
+        source_token: TokenInput,
+        target_token: TokenInput,
+        layers: Sequence[int],
+        clean_cache: Mapping[str, torch.Tensor],
```
jlarson4: `ActivationCache` is what `run_with_cache` returns and what the docstring tells the caller to pass, but it is not a `Mapping`, so the documented call fails mypy. `transformer_lens` ships `py.typed`, so that error lands on users. Can you please annotate `clean_cache` as `ActivationCache`, as `direct_path_patching.py:162` does for the same parameter?

tests/unit/tools/test_jacobian_lens.py
```diff
@@ -1080,6 +1080,31 @@ def test_swap_leaves_orthogonal_complement_unchanged(
     assert orthogonal_part.abs().max().item() < 1e-4
 
 
+def test_swap_clamp_holds_each_layer_at_exchanged_clean_coordinates(
```
jlarson4, line 930: The clamp's coordinate exchange is covered, but deleting any of the identical-token guard, the near-parallel diagnostic, the clean-activation shape validation, the clean/live shape guard, or the `positions` slice of the clean cache leaves the suite green. A perturbation injected outside `span{v_s, v_t}` also passes, which `swap_hooks` catches at line 1066. Can we extend the new tests to cover those, mirroring the `swap_hooks` cases directly above?

jlarson4 (commented):
Thanks for documenting and putting together a fix for this @koriyoshi2041. The parity mechanism reproduces as described, and the clamp holds at every band width. Just a couple small comments below on some edge cases that were missed by the CI

transformer_lens/tools/analysis/jacobian_lens.py
koriyoshi2041: Removed that parenthetical so the layers argument no longer implies clamp semantics.

transformer_lens/tools/analysis/jacobian_lens.py
koriyoshi2041: Updated the primary demo to cache each clean prompt and call swap_clamp_hooks. The separate 96-condition alpha sweep is now explicitly labeled as the historical live-swap comparison, and I cleared the stale output from the changed clamp cell.

transformer_lens/tools/analysis/jacobian_lens.py
koriyoshi2041: Changed clean_cache to ActivationCache and added the public import. Targeted mypy now passes.

tests/unit/tools/test_jacobian_lens.py
koriyoshi2041: Added matched coverage for identical tokens, near-parallel diagnostics, malformed clean activations, clean/live batch mismatch, position slicing, and orthogonal-complement preservation. The full Jacobian Lens unit file passes 108 tests.

demos/Jacobian_Lens_Demo.ipynb
```diff
@@ -391,13 +380,22 @@
...
+    "    return [model.tokenizer.decode([t]) for t in top]\n",
+    "\n",
     "BAND = range(10, 25)\n",
-    "swap = lens.swap_hooks(model, \" France\", \" China\", layers=BAND, alpha=1.0)\n",
     "\n",
     "for prompt in [\"Most people in France speak\", \"The capital of France is\"]:\n",
     "    print(f\"{prompt!r}\")\n",
     "    print(\"  baseline :\", show_next_tokens(prompt))\n",
-    "    print(\"  swapped  :\", show_next_tokens(prompt, hooks=swap))"
+    "    print(\"  clamped :\", clamped_next_tokens(prompt, \" France\", \" China\", BAND))"
```
jlarson4: Removing `swap = lens.swap_hooks(...)` from this cell left the next cell's
`with model.hooks(fwd_hooks=swap):` (line 430) unbound, no cell assigns `swap` any more, so a
fresh run raises `NameError` there. That cell also still carries its stored output from the
deleted live-swap band, under prose that now describes the clamp. Please move cell 13 onto `swap_clamp_hooks` too, then re-execute the notebook so the stored
outputs match the code.

transformer_lens/tools/analysis/jacobian_lens.py
jlarson4: I'm sorry, I should've been clearer with my original comment. `run_with_cache(..., return_cache_object=False)` returns a plain dict, which this annotation now reject. Let's widen this type to `Union[ActivationCache, Mapping[str, torch.Tensor]]` in order to cover both cases.

demos/Jacobian_Lens_Demo.ipynb
koriyoshi2041: Moved the rank cell onto swap_clamp_hooks: cell 12 now defines a clamped_logits helper and cell 13 reuses it, so nothing references the old swap variable any more. I also reworded the section 4 intro so the live swap is no longer described as a clamp, and dropped the stale name from the cleanup cell. Then re-executed the whole notebook top to bottom with the pinned model and lens revisions, so every stored output comes from the committed code, and updated the section 3 and 4 prose to match this run's outputs.

transformer_lens/tools/analysis/jacobian_lens.py
koriyoshi2041: Widened it to Union[ActivationCache, Mapping[str, torch.Tensor]], and the docstring now says both the default ActivationCache and the plain dict from return_cache_object=False are accepted. Added test_swap_clamp_accepts_plain_cache_dict, which passes the raw dict straight through the clamp. mypy is clean on the module and on a typed caller passing each form, and the unit file passes.

jlarson4:
Thanks for closing out those comments @koriyoshi2041, looks great! merging now

PR #1729, opened 2026-08-31
Add leakage-safe k-sparse probing over activation tensors

lorenzozanee:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes 1728

TransformerLens exposed activations but had no maintained k-sparse probing primitive, so leakage-prone splitting and selection were rebuilt per experiment. This adds a leakage-safe, model-free probe that fixes a single stratified split before any learned statistic and reports held-out F1 with deterministic controls.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Screenshots

Please attach before and after screenshots of the change if applicable.

No visual change.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

Reviewers: @jlarson4 @c03d5103 @9a793522

PR #1749, opened 2026-09-05
feat(jacobian_lens): add dynamic J-space coordinate-patch hooks

### LINKED ISSUE #1746: `JacobianLens.swap_hooks` re-reads live coordinates, causing odd/even cancellation across layer bands

### LINKED ISSUE #1748: [Proposal] Dynamic hooks for J-space coordinate patching
---

### Proposal

Add `JacobianLens.coordinate_patch_hooks(...)`, a forward-hook variant of the offline
`coordinate_patch(...)` primitive shipped in #1741, so an anchored J-space coordinate edit can run
live inside `model.run_with_hooks(...)` / `model.generate(...)` instead of only on one
already-captured activation.

**Suggested labels:** `enhancement`, `complexity-moderate`, `TransformerBridge`
cc @jlarson4 (following the review lineage from #1739 / #1741)

---

### Motivation

This follows the Jacobian-Lens proposal (#1505), the sparse-decomposition follow-up (#1539), the
merged decomposition implementation (#1596), and the merged offline coordinate-patching primitive
(#1739 / PR #1741).

`JacobianLens.coordinate_patch(...)` edits one already-captured activation offline and returns a
report-only `CoordinatePatch`; it installs no forward hooks. #1739 deferred a hooked variant to a
separate follow-up for a concrete reason: each hooked position would otherwise require its own
vocabulary-scale sparse decomposition (`get_sparse_decomposition` scans the full lens dictionary)
fired live during the forward pass — "do not hide the cost behind a convenient API."

The gap that leaves: researchers cannot study how a patched coordinate propagates through the rest
of a forward pass, and cannot patch during `model.generate(...)` or any multi-token pass. They must
manually re-capture, patch, and re-inject activations outside the hook system. This issue tracks
closing that gap by wrapping the existing primitive as a forward hook, while making the
per-position cost explicit rather than silent.

---

### Proposed semantics

Given a forward-pass residual chunk `activations` of shape `[batch, num_positions, d_model]` at a
requested layer, for every `(batch_idx, position)` pair the hook body:

1. Slices that pair's activation `x`.
2. Reuses a validated `JSpaceDecomposition` from `decomposition_cache[(layer, batch_idx, position)]`
   if present; otherwise runs `get_sparse_decomposition(x, dictionary, k, ...)` once and stores it.
3. Calls `solve_coordinate_patch(x, dictionary, source_idx, target_idx, decomposition=..., mode=...,
   alpha=...)` — the exact PR1 primitive — to anchor-edit only the source/target coordinates while
   holding every other recovered coordinate and the residual fixed.
4. Writes that pair's `patched` activation back into the requested slice; untouched positions pass
   through unchanged.

A source concept active in one pair never affects another pair: each pair gets its own independent
decomposition and edit, matching the anchored, per-pair contract of the underlying primitive.

---

### API sketch

Model-free core (Commit 1, in `jacobian_lens_coordinate_patch.py`):

```python
def solve_coordinate_patch_positions(
    activations: torch.Tensor,                # [batch, num_positions, d_model]
    dictionary: torch.Tensor,
    position_labels: Sequence[int],           # real sequence position per column; keys the cache
    source_idx: int,
    target_idx: int,
    *,
    layer: int,
    decomposition_cache: Optional[MutableMapping[Tuple[int, int, int], JSpaceDecomposition]] = None,
    k: int = DEFAULT_K,
    mode: str = "substitute",
    alpha: float = 1.0,
    algorithm: str = "nonnegative_orthogonal_matching_pursuit",
) -> Tuple[torch.Tensor, Dict[Tuple[int, int, int], CoordinatePatch]]:
```

Loops `solve_coordinate_patch(...)` over every `(batch_idx, position)` pair — never once per
position shared across the batch — and returns `(patched, patches)` where `patches` maps
`(layer, batch_idx, position)` to that pair's full `CoordinatePatch`.

Bridge wrapper (Commit 2b, in `jacobian_lens.py`):

```python
def coordinate_patch_hooks(
    self,
    model: Any,
    source_token: TokenInput,
    target_token: TokenInput,
    layers: Sequence[int],
    *,
    positions: Sequence[int],                 # required — no "defaults to all"
    decomposition_cache: Optional[MutableMapping[Tuple[int, int, int], JSpaceDecomposition]] = None,
    k: int = DEFAULT_K,
    mode: str = "substitute",
    alpha: float = 1.0,
    algorithm: str = "nonnegative_orthogonal_matching_pursuit",
) -> List[Tuple[str, Any]]:
```

Resolves token inputs and layers exactly like `coordinate_patch(...)`, then builds one forward hook
per layer at `_resid_post_hook_name(layer)` = `blocks.{layer}.hook_out`, following the same eager
dictionary-build pattern `swap_hooks` already uses. Returns `[(hook_name, fn), ...]` for
`model.hooks(fwd_hooks=...)` / `run_with_hooks(...)`.

Two departures from `swap_hooks`, both deliberate:

1. **`positions` is required**, not defaulted to all. `swap_hooks` defaults to all positions because
   its transform is a cheap closed-form pseudoinverse over a fixed 2-vector basis; coordinate
   patching's transform is a full sparse solve per position, so a silent full-sequence default would
   be a cost trap the codebase elsewhere warns against explicitly.
2. **Optional caller-owned `decomposition_cache`** keyed `(layer, batch_idx, position)`. When a
   caller re-runs the same activation repeatedly (e.g. an interactive causal-tracing loop holding the
   prompt fixed while varying `alpha` or `mode`), a cache hit skips the vocabulary-scale scan and
   reuses the strict-compatibility path already validated in `solve_coordinate_patch`; a miss solves
   and stores. Purely a performance path — a cache hit and a fresh solve produce an identical patch.

---

### Correctness contract

- `alpha=0` is an exact no-op, position by position.
- Only the requested `(layer, position)` entries change; every other position in the forward pass is
  bit-identical to an un-hooked run.
- Oracle parity: the patched activation at a given position matches an offline `coordinate_patch(...)`
  call on that position's pre-hook activation.
- Each `(batch_idx, position)` pair is decomposed and edited independently — batch rows do not share
  a support.
- If `source_token` is inactive at **any** pair touched by a hook firing, the whole forward pass
  raises (fail-fast; see decision 2), rather than silently patching a subset.
- `positions` must be non-empty; `source_token` and `target_token` must resolve to distinct ids.
- Inputs are not mutated; the output preserves the decomposition module's float32 compute convention
  and device (it does not promise the activation's original storage dtype).
- The conditioning and near-parallel warnings from `solve_coordinate_patch` still fire from inside
  the hook, per pair, exactly as they would offline — never swallowed or re-wrapped.

---

### Design decisions (pinned; maintainer sign-off requested on 2 and 3)

1. **Batch handling — settled by construction.** Loop `(batch_idx, position)` pairs explicitly and
   solve once per pair, never once per position shared across the batch: two batch rows can have
   different active supports. Required correctness, not a style choice.
2. **Fail-fast on any inactive source (needs sign-off).** If `source_idx` is inactive at any one
   `(batch, position)` pair touched by a hook firing, `solve_coordinate_patch`'s existing
   `ValueError: "... is not in the decomposition's active support"` propagates out of the whole
   forward pass. There is no `try/except` around the per-pair loop, so nothing is a partial write.
   The alternative — skip that pair with a warning and copy the row through unedited — is a two-line
   change if preferred. Fail-fast is recommended: a silently partial intervention is a worse failure
   mode than a loud error. But it changes ergonomics for batched prompts where not every prompt has
   the source concept active, so it warrants explicit sign-off.
3. **`decomposition_cache` is caller-owned (needs sign-off).** A plain dict (any `MutableMapping`) the
   caller passes in and can inspect, not a `JacobianLens` attribute — matching `CoordinatePatch`'s
   report-only, no-retained-state design.
4. **Warning timing.** One `UserWarning` per call at hook-construction time, naming
   `len(layers) x len(positions)` — not a per-position warning (that would spam), and not an attempt
   to report the exact number of live (non-cache-hit) solves, which is only known at forward time.
   This is the plan's own call, flagged for review.

---

### Relationship to #1746 / #1747

Many thanks to the authors of #1746 and #1747 for the careful write-up there — it's a useful
reference for this section. #1746 reports that `swap_hooks` re-reads *live* lens coordinates at
every hooked layer, which makes a multi-layer swap band involutive (odd/even cancellation); #1747
proposes a separate clean-coordinate clamp for J-lens swaps to address that. `coordinate_patch_hooks`
shares the structural property of re-decomposing the live activation per layer, so we'd expect a
multi-layer install to compound across the band in a similar way — but since a `substitute` patch is
not an involution, it should not exhibit the exact cancellation #1746 describes. We'd welcome a
clean-coordinate clamp variant of coordinate patching as a follow-up if the multi-layer concern turns
out to be material in practice; we've left it out of scope for this issue. To be clear, this work
does not fix or depend on #1746 / #1747 — we just want to flag that both areas touch
`jacobian_lens.py`, the `__init__.py` exports, and `jacobian_lens_fitting.md`, so whichever lands
second will need a small mechanical rebase

---

### Proposed repository changes

- New model-free core `solve_coordinate_patch_positions(...)` in
  `transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py`.
- Additive `JacobianLens.coordinate_patch_hooks(...)` in
  `transformer_lens/tools/analysis/jacobian_lens.py` (no change to `swap_hooks` / `coordinate_patch`).
- Export `solve_coordinate_patch_positions` from `transformer_lens/tools/analysis/__init__.py`.
- Extract the shared `_ToyBridge` test fixture into `tests/unit/tools/conftest.py` (mechanical, no
  behavior change) so a new hook-test file can reuse it.
- New model-free unit tests, a dedicated wrapper test file
  (`tests/unit/tools/test_jacobian_lens_coordinate_patch_hooks.py`, not folded into the already-large
  `test_jacobian_lens.py`), and one cached-model integration invariant test.
- Short docs in `jacobian_lens_fitting.md`: cost model, required `positions`, cache semantics,
  fail-fast behavior.

---

### Non-goals

- Arbitrary multi-slot permutations (already out of scope for the primitive).
- Any behavioral / causal-swap benchmark (separate follow-up, PR3).
- HookedTransformer support (separate follow-up, PR4).
- A clean-coordinate clamp variant of coordinate patching (candidate future PR; see above).
- Claims that a successful intervention proves unique causal mediation.

---

### Validation

Unit tests use exact planted dictionaries with analytic expected patches; integration tests assert
algebraic invariants only (`alpha=0` no-op end-to-end, unedited positions bit-identical through a
real `run_with_hooks` pass) — not behavioral success rates or token-flip claims, per PR1's own policy
against behavior-dependent assertions. Cache-hit and fresh-solve paths are checked to produce
identical patches. Full formatting, typing, unit, and integration surfaces run before submission.

---

### Checklist

- [x] Confirm no competing open issue/PR for dynamic coordinate-patch hooks
- [ ] Maintainer sign-off on design decisions 2 (fail-fast vs. skip-with-warning) and 3
      (caller-owned `decomposition_cache`).

---

janmenjayap:
---

### Summary

Adds `JacobianLens.coordinate_patch_hooks(...)`, a forward-hook variant of the offline
`coordinate_patch(...)` primitive shipped in #1741, so an anchored J-space coordinate edit can run
live inside `model.run_with_hooks(...)` / `model.generate(...)` instead of only on one
already-captured activation.

Underneath, a new model-free core function, `solve_coordinate_patch_positions(...)`, loops
`solve_coordinate_patch(...)` over every `(batch_idx, position)` pair in a `[batch, num_positions,
d_model]` chunk, with an optional caller-owned decomposition cache keyed `(layer, batch_idx,
position)`. `coordinate_patch_hooks(...)` is a thin wrapper that builds one forward hook per layer —
following the same `_resid_post_hook_name` / eager dictionary-build pattern as `swap_hooks` — whose
hook body slices the requested positions, calls the core loop, and writes the result back.

This is PR2 of the plan tracked by the companion issue. Arbitrary multi-slot permutations, a
causal-swap benchmark, and HookedTransformer support remain explicitly deferred (PR3/PR4).

Fixes #1748 

### Motivation

`coordinate_patch(...)` edits one pre-captured activation offline and returns a report-only
`CoordinatePatch`. Studying how a patched coordinate propagates through the rest of a forward pass —
or patching during `generate(...)` — previously required manually re-capturing, patching, and
re-injecting activations outside the hook system. #1739 deferred the hooked variant because each
hooked position would otherwise fire its own vocabulary-scale sparse decomposition live during the
forward pass. This PR ships that variant while making the per-position cost explicit (a required
`positions` argument and a construction-time warning) rather than hiding it behind a convenient API.

1. **`feat(jacobian_lens_coordinate_patch): add per-position patch loop core`**
   New model-free `solve_coordinate_patch_positions(...)` in `jacobian_lens_coordinate_patch.py`:
   loops `solve_coordinate_patch(...)` over every `(batch_idx, position)` pair — never once per
   position shared across the batch, since two batch rows can have different active supports.
   Supports an optional caller-owned `decomposition_cache` keyed `(layer, batch_idx, position)`; a
   cache hit skips `get_sparse_decomposition(...)`, a miss solves once and populates it. No
   `try/except` around the per-pair loop — an inactive source at any one pair propagates
   `solve_coordinate_patch`'s existing `ValueError` out of the whole call, so nothing is a partial
   write. Model-free tests first, mirroring PR1's own core-before-wrapper split.

2. **`test(tools): extract shared J-lens toy-bridge fixtures into conftest`**
   Mechanical move, no behavior change. Extracts the `_ToyBridge` fake `TransformerBridge` (and its
   supporting constants/classes/fixture) that `test_jacobian_lens.py` already defined into
   `tests/unit/tools/conftest.py`, so the new hook-test file in commit 3 can reuse it instead of
   duplicating ~90 lines that would otherwise risk drifting between two copies. `test_jacobian_lens.py`
   imports from the conftest; its pass/fail set is byte-identical to before the move.

3. **`feat(jacobian_lens): expose coordinate_patch_hooks on the Bridge`**
   `JacobianLens.coordinate_patch_hooks(model, source_token, target_token, layers, *, positions,
   decomposition_cache=None, ...)` resolves token inputs and layers exactly like `coordinate_patch`,
   then builds one forward hook per layer via `solve_coordinate_patch_positions`. `positions` is
   required — no full-sequence default, since a silent default would trigger a vocabulary-scale
   sparse solve at every position. Emits one `UserWarning` per call, at hook-construction time,
   naming `len(layers) x len(positions)` — not a per-position warning, and not an attempt to report
   the exact number of live (non-cache-hit) solves, which is only known at forward time. New
   dedicated test file (`test_jacobian_lens_coordinate_patch_hooks.py`), not folded into the
   already-1600+-line `test_jacobian_lens.py`.

4. **`feat(jacobian_lens): export and document coordinate_patch_hooks`**
   Exports `solve_coordinate_patch_positions` from `transformer_lens.tools.analysis`, extends the
   package docstring, replaces the now-stale "dynamic patching would require a vocabulary-scale
   solve" claim in `jacobian_lens_fitting.md` with a new `#### Dynamic coordinate-patch hooks`
   subsection, and adds one cached-model GPT-2 integration test asserting **algebraic invariants
   only** (`alpha=0` is an exact no-op; untouched positions are bit-identical through a real
   `run_with_hooks` pass) — not a token-flip claim.

### Design decisions (per the tracking issue)

| # | Decision | Status |
|---|----------|--------|
| 1 | Batch handling: loop `(batch_idx, position)` pairs explicitly, never once per position shared across the batch | Required correctness, not a style choice — settled by construction |
| 2 | Fail-fast (not skip-with-warning) when a source is inactive at any one pair in a batch | **Needs maintainer sign-off** — the alternative (`try/except ValueError` + `warnings.warn` + copy the row through unedited) is a two-line hunk if preferred |
| 3 | `decomposition_cache` is caller-owned (`Optional[MutableMapping[Tuple[int, int, int], JSpaceDecomposition]]`), not a `JacobianLens` attribute | Needs maintainer sign-off, lower risk than `#2` |
| 4 | Warning fires once per `coordinate_patch_hooks(...)` call, at construction time, naming `layers x positions` counts (not the exact number of live solves) | This plan's own call, not yet reviewed — flagged explicitly here |

Please weigh in on `#2` and `#3` in particular — they're the most likely to get pushback and are cheap
to change if so.

### Testing

- `tests/unit/tools/test_jacobian_lens_coordinate_patch.py` — model-free core: parity with
  `solve_coordinate_patch`, batch independence, fail-fast on any inactive source, cache miss/hit
  behavior, cache-hit vs. fresh-solve identical output, mismatched `position_labels` rejection.
- `tests/unit/tools/test_jacobian_lens.py` — unchanged behavior after the conftest extraction
  (byte-identical pass/fail set).
- `tests/unit/tools/test_jacobian_lens_coordinate_patch_hooks.py` — wrapper: hook-list shape mirrors
  `swap_hooks`, once-per-call warning naming layer/position counts, unfitted-layer rejection,
  empty-`positions` rejection, identical-token rejection, only-requested-positions-change, oracle
  parity with offline `coordinate_patch`, cache hits skip resolve across repeated hook firings,
  core errors and core warnings propagate uncaught through the hook.
- `tests/integration/test_jacobian_lens.py` — one cached GPT-2 test: `alpha=0` no-op and
  bit-identical untouched positions through a real `run_with_hooks` pass.

Local gates green:
- `uv run mypy .` — `Success: no issues found in 397 source files`.
- `uv run pytest tests/unit/tools/test_jacobian_lens_coordinate_patch.py tests/unit/tools/test_jacobian_lens.py tests/unit/tools/test_jacobian_lens_coordinate_patch_hooks.py` — `157 passed`.
- `source .env && uv run pytest tests/integration/test_jacobian_lens.py -k coordinate_patch_hooks` — `1 passed`.

### Related work

- **#1739 (closed) / #1741 (merged)** — the offline coordinate-patching proposal and the PR that
  implemented the `coordinate_patch(...)` primitive this PR wraps. Strict dependency; no overlap.
- **#1746 / #1747 (open)** — a bug report that `swap_hooks` re-reads *live* lens coordinates at every
  hooked layer, making a multi-layer swap band involutive (odd/even cancellation), and a proposal to
  add clean-coordinate clamping for J-lens swaps. **No functional conflict** with this PR — different
  method, different semantics — but two things reviewers should note: (1) both areas touch
  `jacobian_lens.py`, the `__init__.py` exports, and `jacobian_lens_fitting.md`, so whichever lands
  second needs a mechanical rebase; and (2) `coordinate_patch_hooks` shares the same structural
  property — it re-decomposes the *live* activation at each hooked layer, so a multi-layer install
  compounds across the band. Unlike a swap, a `substitute` patch is not an involution, so it does not
  exhibit the exact odd/even cancellation #1746 describes; a clean-coordinate clamp variant of
  coordinate patching is a reasonable future PR if the same multi-layer concern proves material. This
  PR does not attempt to fix or depend on #1746 / #1747.

### Out of scope (deferred to a follow-up PR)

Arbitrary multi-slot permutations, any behavioral / causal-swap benchmark (PR3), and
HookedTransformer support (PR4) remain deferred to follow-up PRs.

### Checklist

- [ ] Full `make test-pr` run attached (unit + docstring + acceptance + integration).
- [ ] `uv run build-docs` run clean (new `#### Dynamic coordinate-patch hooks` section renders).
- [ ] Reviewer sign-off on design decisions `#2` (fail-fast vs. skip-with-warning) and `#3`
      (caller-owned `decomposition_cache` shape).


.cursor/rules/transformerlens.mdc
jlarson4, line 1: Please restore this file

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -1375,6 +1386,119 @@ def transform(
...
+        this installs a forward hook that solves :func:`solve_coordinate_patch` independently for
+        every ``(batch_idx, position)`` pair at each requested layer -- a vocabulary-scale sparse
+        decomposition per pair, per hook firing, unless ``decomposition_cache`` supplies one
+        already validated for that ``(layer, batch_idx, position)`` key.
+
+        Args:
+            model: The model the hooks will run on.
+            source_token: Active source concept, as a single-token string or token id.
+            target_token: Distinct target concept, as a single-token string or token id.
+            layers: Layers to intervene at.
```
jlarson4, line 1417: On GPT-2 with the published lens, only a single layer at a single position runs. `positions=[-1, -2]`, `positions=list(range(seq_len))`, and the bands `[6, 7]`, `[6, 9]`, `[6, 10]` all raise `source_idx=… is not in the decomposition's active support` mid-forward; `[6, 8]` happens to survive. For a band, the earlier hook is what causes it,  the source is in the clean support at layer 7 and the layer-6 patch removes it, which is exactly what `substitute` and `swap` do to the source coordinate. So the real precondition is that the source must be in the top-`k` active support of every position at every layer *after* the earlier hooks in the band have edited it, which is far stronger than the `Raises:` note, and no test covers a plural install end to end.

Let's state that precondition in the docstring, and add a test that installs on two fitted layers and runs a forward pass that also pins the per-layer `layer=layer` / `dictionary=dictionary` binding on line 1471, which nothing currently does.

docs/source/content/jacobian_lens_fitting.md
```diff
@@ -350,9 +350,53 @@ Read the measured condition number and cosine straight out of each warning messa
...
-a vocabulary-scale solve inside the model run.
+Reuse a compatible result for repeated edits to avoid that scan. `coordinate_patch` returns an
+offline activation and diagnostics, not forward hooks; see "Dynamic coordinate-patch hooks" below
+for the live variant, which performs the same scan per `(batch, position)` pair on every forward
+pass unless a cache hit avoids it.
+
+### Dynamic coordinate-patch hooks
+
+`JacobianLens.coordinate_patch_hooks` installs the same anchored edit as a forward hook, so it can
+run inside `model.run_with_hooks(...)` or `model.generate(...)` instead of on one pre-captured
```
jlarson4: On GPT-2 with the published lens, `generate(max_new_tokens=3)` with `layers=[6], positions=[-1]` raises `source_idx=3240 is not in the decomposition's active support`, with and without a `decomposition_cache`. Each decode step re-decomposes a freshly generated token's residual, and a new token generally does not carry the source concept in its top-`k` support, so fail-fast aborts the generation.

Drop `model.generate(...)` from this sentence, and from the same claim in the PR description's Motivation.

docs/source/content/jacobian_lens_fitting.md
```diff
@@ -350,9 +350,53 @@ Read the measured condition number and cosine straight out of each warning messa
...
+```
+
+Two departures from `coordinate_patch`, both deliberate:
+
+- **`positions` is required.** There is no full-sequence default: each hooked position performs
+  its own vocabulary-scale sparse decomposition (unless a cache hit avoids it), and a silent
+  full-sequence default would trigger that scan at every position without the caller asking for
+  it.
+- **`decomposition_cache` (optional, caller-owned).** A plain `dict` (or any `MutableMapping`)
+  keyed `(layer, batch_idx, position)`. Pass the same dict across repeated `model.hooks(...)` calls
```
jlarson4, line 383: `position` in the key is the chunk-local index. With `use_past_kv_cache=True` the prefill sees `[1, seq, d]` and every decode step sees `[1, 1, d]`, so step 2 collides with step 1's key and dies with `precomputed decomposition coordinates do not fit the activation (NNLS result violates stationarity on positive coordinates)`, which does not point at the cause. The same applies to reusing one dict across prompts of different lengths.

Please mention in this bullet that the cache is valid only across passes with identical chunking.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -491,6 +491,99 @@ def solve_coordinate_patch(
...
+                dictionary,
+                source_idx,
+                target_idx,
+                decomposition=decomposition,
+                k=k,
+                mode=mode,
+                alpha=alpha,
+                algorithm=algorithm,
+            )
+            patched[batch_idx, column] = patch.patched
```
jlarson4: `CoordinatePatch.patched` is detached, so scattering it into the residual stream cuts the patched position's gradient path: gradient arriving there is exactly 0 against 7.5e5 for an unhooked pass, and upstream parameter gradients come out silently understated (3322 to 2700 on a 9-token prompt, exactly 0 for a single-token one). `swap_hooks` on the same model and position keeps them. Attribution patching or `backward_lens` over these hooks would report wrong numbers with no error.

Please write `patched[batch_idx, column] = x + patch.delta` instead. `patch.patched` is `x.float() + delta`, so it is bitwise identical for the float32 activations the hook supplies, and it gives a straight-through gradient rather than none.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
```diff
@@ -491,6 +491,99 @@ def solve_coordinate_patch(
...
+    if len(position_labels) != num_positions:
+        raise ValueError(
+            f"position_labels has {len(position_labels)} entries, expected {num_positions} "
+            "to match activations.shape[1]"
+        )
+    patched = activations.new_empty(activations.shape)
+    patches: Dict[Tuple[int, int, int], CoordinatePatch] = {}
+    for batch_idx in range(batch_size):
+        for column, position in enumerate(position_labels):
+            key = (layer, batch_idx, position)
```
jlarson4, line 563: Every cache test uses `[1, 1, d_model]` activations. Replacing this key with `(layer, 0, position)` passes the whole unit tier, so nothing pins the batch component of the documented key shape.

Please add a core test that patches a batch of two with a shared `decomposition_cache` and asserts both `(layer, 0, position)` and `(layer, 1, position)` are populated with different decompositions.

tests/mps/test_mps_basic.py
```diff
@@ -28,6 +28,7 @@
     SubspaceBasis,
     projection_kernel,
 )
+from transformer_lens.tools.analysis.sparse_probing import fit_sparse_probe
```
jlarson4: `transformer_lens.tools.analysis.sparse_probing` exists neither on this branch nor on `dev`, so `pytest tests/mps` fails at collection with `ModuleNotFoundError` instead of running. The MPS job only runs on PRs based on `main`, so this lands on `dev` unnoticed and surfaces on the next release PR. Please move this important and the related test to 1729, if it is not already in the code you recommended they cherry pick

docs/source/index.md
```diff
@@ -58,6 +58,7 @@ content/hook_system
 content/compatibility_mode
 content/ssm_interpretability
 content/projection_kernel
+content/sparse_probing
```
jlarson4: Remove, not related to this PR

jlarson4 (commented):
Thanks for putting this together @janmenjayap! Review comments below:

.cursor/rules/transformerlens.mdc
janmenjayap: Thanks for catching this. Fixed in ea135c04. Restored the file verbatim from `dev`. Its deletion
was an unrelated leak from another branch, not an intentional change in this PR.

transformer_lens/tools/analysis/jacobian_lens.py
janmenjayap: Good catch, thank you. Fixed in b304f732. The `Raises:` note on `coordinate_patch_hooks` now
states the real precondition: the source must be in the top-`k` active support of every patched
`(batch_idx, position)` pair *at the moment its hook fires*, not merely active on a clean forward
pass. For a band of layers, an earlier hook's `substitute`/`swap` edits the residual that a later
layer re-decomposes and can remove the source coordinate from that later layer's support even
though it was active pre-hook, so stacking layers/positions makes the precondition progressively
harder to satisfy, exactly as observed on GPT-2 (`[6,7]`, `[6,9]`, `[6,10]` raise; `[6,8]`
survives).

Also added `test_coordinate_patch_hooks_plural_install_binds_each_layer_to_its_own_dictionary`:
the first end-to-end test that installs on two fitted layers and runs one real forward pass
through `model.hooks(...)`. It uses deliberately different per-layer dictionaries (so a mis-bound
closure produces a numerically wrong edit rather than one hidden by scale-covariance) and pins the
`layer=layer, dictionary=dictionary` default-argument binding at `jacobian_lens.py:1471`. It also
searches for a source that satisfies the band precondition rather than assuming one exists, and
asserts both `(0, 0, position)` and `(1, 0, position)` land in the shared `decomposition_cache`.

docs/source/content/jacobian_lens_fitting.md
janmenjayap: Thanks, you're right to flag that. Fixed in 8074f2ea. Removed `model.generate(...)` from the
dynamic coordinate-patch hooks section. Confirmed on GPT-2 with the published lens that
`generate(max_new_tokens=3)` with `layers=[6]`, `positions=[-1]` raises `source_idx=3240 is not in
the decomposition's active support` with or without a `decomposition_cache`, since each decode
step re-decomposes a freshly generated token whose top-k support generally doesn't carry the
source concept. `model.run_with_hooks(...)` stays.
**Still open:** the same claim in the PR description's Motivation is a GitHub-side edit, not a
tracked file. I will update the PR body directly.

docs/source/content/jacobian_lens_fitting.md
janmenjayap: Appreciate the thorough read. Fixed in 8074f2ea. The `decomposition_cache`
bullet now states the cache is valid **only across passes with identical chunking**: the
`position` in the `(layer, batch_idx, position)` key is the chunk-local index, so with
`use_past_kv_cache=True` the prefill's `[1, seq, d]` and each decode step's `[1, 1, d]` collide on
position `0`, and reusing one dict across prompts of different lengths hits the same problem. Both
cases surface as an NNLS stationarity error that doesn't point back at the cache; this is now
called out explicitly so it's diagnosable from the docs instead of from the traceback.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Nice catch, thank you. Fixed in a903536e. `solve_coordinate_patch_positions` now scatters
`x + patch.delta` instead of the detached `patch.patched`. `patch.delta` is a detached constant
w.r.t. `x`, and `patch.patched == x.float() + delta`, so the forward value is bitwise identical for
the float32 activations the hook supplies, while `x`'s graph stays attached: an identity gradient
at the patched position instead of the silent zero this had been producing (confirmed: exactly 0
vs. ~7.5e5 unhooked, and 3322 → 2700 on a 9-token prompt). `swap_hooks` was never affected since it
doesn't route through this detached path. Added
`test_solve_coordinate_patch_positions_preserves_gradient_to_patched_position`, which asserts a
finite nonzero gradient at the patched position and bitwise-identical forward output before/after.

transformer_lens/tools/analysis/jacobian_lens_coordinate_patch.py
janmenjayap: Thanks for pointing this out. Fixed in 7d4c8433. Added
`test_solve_coordinate_patch_positions_populates_batch_component_of_cache_key`: two batch rows with
different active supports share one `decomposition_cache` at the same `(layer, position)`; the test
asserts both `(layer, 0, position)` and `(layer, 1, position)` populate, with different supports,
so a collapse to `(layer, 0, position)` (which every other cache test's `[1, 1, d_model]` fixtures
would have let through) is now caught.

tests/mps/test_mps_basic.py
janmenjayap: Thanks for catching this. Fixed in ea135c04. Removed the `sparse_probing` import and the test it
supported: `transformer_lens.tools.analysis.sparse_probing` doesn't exist on this branch or on
`dev`, so it was failing `pytest tests/mps` at collection with `ModuleNotFoundError`. Confirmed
this was an accidental pickup from another branch. I will confirm the corresponding test is
carried in **#1729** before that PR merges rather than dropping the coverage.

docs/source/index.md
janmenjayap: Good catch, thanks. Fixed in ea135c04. Removed the toctree line referencing the nonexistent `content/sparse_probing` page.

jlarson4:
Looks good, great work @janmenjayap Merging now

Issue #1742, opened 2026-09-04
[Proposal] Attribution Patching + EAP/EAP-IG: linearized activation patching and faithful edge-level circuit discovery


janmenjayap:
---

### Proposal 

Add a self-contained `tools/analysis/` module that estimates the causal effect of **every node and edge** in a model's computational graph in a **single 2-forward-plus-1-backward pass**, instead of one forward pass per intervention. The module implements three layered methods on the same substrate:

1. **Attribution patching** — a first-order Taylor estimate of activation patching: `effect ≈ (clean_act − corrupt_act) · grad_of_metric`, computed for all components at once from one gradient cache.
2. **EAP (Edge Attribution Patching)** — score every **edge** in the residual-stream computational graph (node = head / MLP / input at a position; edge = residual write→read), not just nodes, yielding a ranked edge list.
3. **EAP-IG** — replace the single saturating gradient with an **integrated-gradients path** between corrupt and clean activations, giving markedly more **faithful** circuits (the ablate-everything-outside-the-circuit metric).
4. **Faithfulness evaluation** — given a candidate edge set, mean-ablate every edge *outside* it and report how much of the task metric (e.g. IOI logit-diff) is recovered.
5. A **demo** on `gpt2-small` IOI reproducing the known circuit from EAP/EAP-IG scores alone.

This is the linearized-patching substrate that **ACDC** (Automatic Circuit Discovery — a separate, not-yet-filed sibling proposal for a patch-and-prune circuit-discovery tool that would consume this module's ranked edge list as its pre-filter) and attribution-graph work build on — the single highest-leverage addition to TL's circuit-discovery toolkit.

**Status:** exploratory proposal, no code changes.
**Suggested labels:** `enhancement`, `tooling`, `TransformerBridge`, `complexity: moderate-to-high`

---

### Motivation

**What TL has today.** TL already exposes every primitive this method needs but assembles none of them into an all-at-once attributor. `run_with_cache` caches activations and can retain their gradients for a manual backward; `get_act_patch_direct_path` (in `tools/analysis/direct_path_patching.py`) does a **forward-mode linear perturbation** of a **single** A→B edge (no gradients); `patching.py:generic_activation_patch` does exact (expensive) node patching; `direct_logit_attribution` maps a component to the logit; the hook graph enumerates the nodes. What is missing is the tool that reads **one** gradient cache and produces **every** node/edge score in a single sweep.

[...]

PR #1750, opened 2026-09-06
feat(attribution_patching): node attribution-patching substrate on TransformerBridge

janmenjayap:
---

### Summary

Adds `attribution_patching`, a new analysis tool under `transformer_lens/tools/analysis/`, together with its reusable substrate and public API. This first PR ships **node-level (first-order Taylor) attribution only**. The API surface (`granularity`, `ig_steps`) is designed here so it stays stable across the planned series — edge scoring (EAP), the integrated-gradient path (EAP-IG), ablate-outside faithfulness, oracle parity, and the demo land in follow-on PRs

Attribution patching estimates the causal effect of every residual-stream component on a task metric by reading a single gradient cache instead of running one forward pass per intervention. For a clean/corrupt prompt pair it runs a clean forward (for `a_clean`), a corrupt forward with retained gradients plus a manual `metric.backward()` (for `a_corrupt` and `g = d(metric)/d(a)`), and scores each node with the first-order estimate `effect(node) = (a_clean − a_corrupt) · g`. Per-node scores are averaged across a batch of clean/corrupt pairs before ranking.

Everything targets the `TransformerBridge` API exclusively; there is no new reference to `HookedTransformer`.

---

### What's included

- **Names-filtered gradient-cache substrate.** A helper runs a Bridge forward with a names filter, retains gradients only on the requested hook families, and invokes the user metric with a manual `metric.backward()` — necessary because `run_with_cache(..., incl_bwd=True)` only backprops the model's scalar output, not a custom logit-diff metric. Retaining gradients everywhere roughly doubles cache memory, so callers filter to the hook families their analysis reads.
- **Typed computational-graph node model.** A `Node` dataclass keyed by `(layer, position, head, kind)` with node families for embed / attn-head-out / mlp-out, enumerated from the Bridge hook graph. A missing required hook point raises rather than being silently skipped.
- **Config + result API.** `EdgeAttributionConfig` exposes `granularity` (`"node"` | `"edge"`) and `ig_steps` (int) only — there is deliberately no `method` field, which collapses the overlapping axes flagged in review; `ig_steps=1` is plain attribution/EAP. `AttributionResult` carries `node_scores` and `top_nodes(k=...)`. `granularity="edge"` and `ig_steps>1` raise`NotImplementedError` pointing at the follow-on PRs, so downstream code can pin against a stable surface now.
- **`attribution_patch` entry point** assembling two forwards and one manual backward, with token-length parity validation (mismatched clean/corrupt tokenization raises) and dataset averaging across a batch of pairs.
- **Pinned sign/direction convention** (denoising form): the gradient is taken on the corrupt run and the estimate points toward the clean activation, so a positive score means patching that node from corrupt toward clean moves the metric in the positive direction. The docstring notes a follow-on PR will map this convention onto the pinned oracle rather than assuming the two agree.
- **Linear-model reconstruction identity test.** A tiny `TransformerBridge` (chosen over a model-free stub so the real hook graph and cache plumbing are exercised end-to-end) is configured to be linear — identity MLP activation, a metric linear in the logits, frozen attention patterns, and LayerNorm scale treated as constant — so the first-order identity is exact. The test asserts summed node attributions equal `m(clean) − m(corrupt)` under tight `atol`, and asserts the documented sign/direction. Exact equality is a linear-network property and does not hold for a stock nonlinear config; the test documents this constraint.

---

### Files

- `transformer_lens/tools/analysis/attribution_patching.py` (new)
- `transformer_lens/tools/analysis/__init__.py` (exports `attribution_patch`, `EdgeAttributionConfig`, `AttributionResult`, `Node`; `__all__` updated)
- `tests/unit/tools/test_attribution_patching.py` (new)

---

### Scope / follow-ups

Node granularity only in this PR. The following land in follow-on PRs:

- Faithfulness default (corrupt-patch / ablate-outside)
- Edge scoring (EAP)
- Integrated gradients (EAP-IG, `ig_steps>1`)
- Oracle parity
- The uniquely-named demo notebook

Addresses part of #1742 — this PR covers node-granularity attribution patching; edge scoring (EAP) and integrated gradients (EAP-IG), which #1742 also asks for, are tracked as follow-on work above.

---

### Checklist

- [x] `make unit-test` green
- [x] `uv run mypy .` clean
- [x] `make format` applied
- [x] Linear-model reconstruction identity holds under exact equality
- [x] Token-length parity and missing-hook paths raise
- [x] API is `granularity` + `ig_steps` only; `edge` / `ig_steps>1` raise `NotImplementedError`
- [x] No new reference to `HookedTransformer`; all model interaction via `TransformerBridge`

---

### DIFF
--- a/tests/unit/tools/test_attribution_patching.py
+++ b/tests/unit/tools/test_attribution_patching.py
@@ -19,6 +19,8 @@ import torch.nn as nn
 from transformer_lens.hook_points import HookPoint
 from transformer_lens.model_bridge import TransformerBridge
 from transformer_lens.tools.analysis.attribution_patching import (
+    AttributionResult,
+    EdgeAttributionConfig,
     GradientCache,
     Node,
     cache_activation_and_gradient,
@@ -275,3 +277,48 @@ def test_enumerate_nodes_raises_on_missing_hook() -> None:
 
     with pytest.raises(ValueError, match="blocks.1.hook_mlp_out"):
         enumerate_nodes(_cfg_stub(), cache)
+
+
+# ---------------------------------------------------------------------------
+# Commit 3 — config + result API
+# ---------------------------------------------------------------------------
+
+
+def test_config_defaults_are_the_supported_node_sweep() -> None:
+    config = EdgeAttributionConfig()
+    assert config.granularity == "node"
+    assert config.ig_steps == 1
+
+
+def test_config_rejects_invalid_values() -> None:
+    with pytest.raises(ValueError, match="ig_steps"):
+        EdgeAttributionConfig(ig_steps=0)
+
+
+def test_config_unsupported_paths_raise_not_implemented() -> None:
+    with pytest.raises(NotImplementedError, match="PR2"):
+        EdgeAttributionConfig(granularity="edge")
+    with pytest.raises(NotImplementedError, match="PR3"):
+        EdgeAttributionConfig(ig_steps=5)
+
+
+def test_top_nodes_ranks_by_effect_magnitude() -> None:
+    small = Node(kind="embed", position=0)
+    big_negative = Node(kind="mlp_out", layer=0, position=1)
+    medium = Node(kind="attn_head_out", layer=1, head=0, position=2)
+    result = AttributionResult(
+        node_scores={small: 0.1, big_negative: -5.0, medium: 2.0},
+    )
+
+    ranked = result.top_nodes(k=2)
+    assert [node for node, _ in ranked] == [big_negative, medium]
+
+    # k beyond the node count returns every node, still magnitude-ordered.
+    assert [node for node, _ in result.top_nodes(k=10)] == [big_negative, medium, small]
+
+
+def test_top_edges_not_implemented_until_pr2() -> None:
+    result = AttributionResult(node_scores={})
+    assert result.edge_scores == {}
+    with pytest.raises(NotImplementedError, match="PR2"):
+        result.top_edges()
--- a/tests/unit/tools/test_attribution_patching.py
+++ b/tests/unit/tools/test_attribution_patching.py
@@ -256,7 +256,7 @@ def test_node_hook_name_and_key_validation() -> None:
     )
     assert Node(kind="mlp_out", layer=0, position=1).hook_name == "blocks.0.hook_mlp_out"
 
-    # The typed key rejects malformed nodes (Risk 1: explicit graph).
+    # The typed key rejects malformed nodes rather than building a wrong graph.
     with pytest.raises(ValueError):
         Node(kind="embed", position=0, layer=0)
     with pytest.raises(ValueError):
@@ -314,9 +314,9 @@ def test_config_rejects_invalid_values() -> None:
 
 
 def test_config_unsupported_paths_raise_not_implemented() -> None:
-    with pytest.raises(NotImplementedError, match="PR2"):
+    with pytest.raises(NotImplementedError, match="edge"):
         EdgeAttributionConfig(granularity="edge")
-    with pytest.raises(NotImplementedError, match="PR3"):
+    with pytest.raises(NotImplementedError, match="integrated gradient"):
         EdgeAttributionConfig(ig_steps=5)
 
 
@@ -335,10 +335,10 @@ def test_top_nodes_ranks_by_effect_magnitude() -> None:
     assert [node for node, _ in result.top_nodes(k=10)] == [big_negative, medium, small]
 
 
-def test_top_edges_not_implemented_until_pr2() -> None:
+def test_top_edges_not_implemented() -> None:
     result = AttributionResult(node_scores={})
     assert result.edge_scores == {}
-    with pytest.raises(NotImplementedError, match="PR2"):
+    with pytest.raises(NotImplementedError, match="edge"):
         result.top_edges()
 
 
@@ -516,9 +516,9 @@ def test_attribution_patch_raises_on_batch_size_mismatch() -> None:
 # Commit 5 — linear-model reconstruction identity
 # ---------------------------------------------------------------------------
 #
-# ``_NodeGraphToyBridge`` is fully linear by construction — the plan's
-# "neutralize the remaining nonlinearities on the residual->metric path"
-# conditions hold without extra freezing: ``ln_final`` is ``nn.Identity``, the
+# ``_NodeGraphToyBridge`` is fully linear by construction — the remaining
+# nonlinearities on the residual->metric path are neutralized without extra
+# freezing: ``ln_final`` is ``nn.Identity``, the
 # attention projection is a plain linear map with no softmax pattern, the MLP has
 # no activation, and ``_metric_fn`` is linear in the logits. The first-order
 # Taylor estimate each node score uses is therefore *exact*, so the reconstruction
--- a/tests/unit/tools/test_attribution_patching.py
+++ b/tests/unit/tools/test_attribution_patching.py
@@ -1,0 +1,189 @@
+"""Model-free unit tests for the attribution-patching substrate.
+
+These tests target the ``TransformerBridge`` API exclusively (TransformerLens v4
+deprecates ``HookedTransformer``). They use a tiny, deliberately *linear*
+``TransformerBridge`` subclass so gradients have a closed form and, later, the
+first-order attribution identity holds exactly.
+"""
+
+from __future__ import annotations
+
+from contextlib import contextmanager
+from types import SimpleNamespace
+from typing import Any, Callable, Iterator
+
+import torch
+import torch.nn as nn
+
+from transformer_lens.hook_points import HookPoint
+from transformer_lens.model_bridge import TransformerBridge
+from transformer_lens.tools.analysis.attribution_patching import (
+    cache_activation_and_gradient,
+)
+
+D_MODEL = 4
+D_VOCAB = 6
+N_LAYERS = 2
+SEQ_LEN = 3
+
+
+class _LinearBlock(nn.Module):
+    """Position-wise residual-linear block with a single output hook point."""
+
+    def __init__(self, d_model: int, layer: int, dtype: torch.dtype) -> None:
+        super().__init__()
+        self.linear = nn.Linear(d_model, d_model, bias=False, dtype=dtype)
+        nn.init.normal_(self.linear.weight, std=0.2)
+        self.hook_out = HookPoint()
+        self.hook_out.name = f"blocks.{layer}.hook_out"
+
+    def forward(self, residual: torch.Tensor) -> torch.Tensor:
+        return self.hook_out(residual + self.linear(residual))
+
+
+class _LinearToyBridge(TransformerBridge):
+    """Tiny fully-linear ``TransformerBridge`` with Bridge-native hook points.
+
+    The production constructor needs a Hugging Face model and an architecture
+    adapter; unit tests only need the hook graph, a forward pass, and the
+    ``hooks()`` context, so this subclass initializes ``nn.Module`` directly while
+    keeping the concrete ``TransformerBridge`` isinstance contract. Every layer is
+    linear and ``ln_final`` is the identity, so the residual->metric map is linear.
+    """
+
+    def __init__(self, *, dtype: torch.dtype = torch.float32) -> None:
+        nn.Module.__init__(self)
+        # TransformerBridge.__setattr__ registers HookPoints here; create it before
+        # any HookPoint attribute is assigned.
+        self._hook_registry: dict[str, HookPoint] = {}
+        self.context_level = 0
+        torch.manual_seed(0)
+        self.cfg = SimpleNamespace(
+            n_layers=N_LAYERS,
+            d_model=D_MODEL,
+            d_vocab=D_VOCAB,
+            d_vocab_out=D_VOCAB,
+            model_name="linear-toy-bridge",
+            dtype=dtype,
+            device="cpu",
+        )
+        self.compatibility_mode = False
+        self._weights_processed = False
+        self.embed = nn.Embedding(D_VOCAB, D_MODEL, dtype=dtype)
+        nn.init.normal_(self.embed.weight, std=0.2)
+        self.hook_embed = HookPoint()
+        self.hook_embed.name = "hook_embed"
+        self.blocks = nn.ModuleList(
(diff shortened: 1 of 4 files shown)


jlarson4 (commented):
Thanks for building this out @janmenjayap! Review comments below, let me know if you have any questions

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -0,0 +1,438 @@
...
+        raise ValueError("names_filter matched no hook points")
+
+    live: dict[str, torch.Tensor] = {}
+    activations: dict[str, torch.Tensor] = {}
+
+    def make_hook(name: str) -> Callable[..., None]:
+        def hook(tensor: torch.Tensor, *, hook: Any) -> None:
+            del hook
+            if isinstance(tensor, torch.Tensor):
+                tensor.retain_grad()
```
jlarson4: `TransformerBridge` applies a hook conversion before calling a forward hook, so at `blocks.*.attn.hook_z` the tensor handed to this hook is a reshaped view that the model's own forward does not consume. `retain_grad()` on it is inert and `.grad` stays `None`, so `attribution_patch` raises on the first attention-head node for every real bridge, which I verified on pythia-14m and gpt2. 84 of pythia-14m's 345 hook points carry a conversion, including all of `hook_z`, `hook_q/k/v`, and `hook_attn_out`, so the per-head families the edge sweep will need are affected too. Compatibility mode does not change it.

Let's capture gradients with `bwd_hooks` instead. Backward hooks go through the same conversion, so a backward hook at `blocks.0.attn.hook_z` delivers `[batch, seq, n_heads, d_head]`, matching the cached activation. They key by canonical name, so you will need the alias-to-canonical map from `transformer_lens/model_bridge/bridge.py:86`.

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -0,0 +1,438 @@
...
+                live[name] = tensor
+                activations[name] = tensor.detach().clone()
+            return None
+
+        return hook
+
+    fwd_hooks = [(name, make_hook(name)) for name in names]
+
+    model.zero_grad(set_to_none=True)
```
jlarson4: `metric.backward()` at line 321 accumulates into every parameter's `.grad`, and this `zero_grad` discards whatever the caller had there first, gradients a user accumulated on their model before calling `attribution_patch` are lost. Each call also allocates a model-sized gradient buffer the tool never reads, roughly 500MB on gpt2-small.

Please take the gradients with `torch.autograd.grad` and drop the `zero_grad` call, the way `backward_lens.py` does.

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -0,0 +1,438 @@
...
+            f"length {clean.shape[1]} and corrupt length {corrupt.shape[1]}. "
+            "Attribution patching aligns activations position-by-position."
+        )
+
+    node_hook_names = _required_hook_names(int(model.cfg.n_layers))
+    batch = int(clean.shape[0])
+    totals: dict[Node, float] = {}
+
+    for index in range(batch):
+        clean_cache = cache_activation_and_gradient(
```
jlarson4, line 506: The clean forward goes through `cache_activation_and_gradient`, which always calls `metric.backward()`, but `_node_effects` reads only `clean_cache.activations`. A two-pair sweep runs four backward passes where the documented two forwards + one backward needs two.

Please adjust it to run the clean pass activation-only.

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -0,0 +1,438 @@
...
+        return names_filter
+    wanted = set(names_filter)
+    return lambda name: name in wanted
+
+
+def cache_activation_and_gradient(
+    model: Any,
+    tokens: torch.Tensor,
+    metric_fn: MetricFn,
+    names_filter: NamesFilter = None,
```
jlarson4, line 268: Line 282 documents this default as caching every hook point, but a bridge's `hook_dict` contains gated hook points (`blocks.0.hook_mlp_in`, `blocks.0.attn.hook_result`, the split-QKV inputs) that `add_hook` rejects unless the matching `set_use_*` flag is on. The default raises `Cannot add hook blocks.0.hook_mlp_in because use_hook_mlp_in is False`.

Please default `names_filter` to the node graph's hook set, or say in the docstring that a filter is required on the Bridge.

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -0,0 +1,438 @@
...
+
+    granularity: Granularity = "node"
+    ig_steps: int = 1
+
+    def __post_init__(self) -> None:
+        if self.ig_steps < 1:
+            raise ValueError(f"ig_steps must be >= 1, got {self.ig_steps}")
+        if self.granularity == "edge":
+            raise NotImplementedError(
+                "granularity='edge' (EAP edge scoring) lands in PR2; this substrate "
```
jlarson4: This message and the one at line 160 tell a user their call "lands in PR2" / "lands in PR3", but users have no way to know where theses references are meant to point. The same applies to `top_edges`'s raise at line 197 and to the "Risk 1" / "proposal step 6" references, which name sections of #1742. In general, please avoid adding references like this to any of your work, I understand what you're referring to, but other maintainers may not and neither will other contributors.

Describe the missing capability instead ("edge scoring is not implemented yet", etc) and drop the PR / Risk / step labels from lines 18, 75, 137, 177, 195, 228, 376, 382, and 396 here and from lines 243, 301, 303, 325, and 478 of the test file.

tests/unit/tools/test_attribution_patching.py
```diff
@@ -0,0 +1,576 @@
...
+        self.linear = nn.Linear(d_model, d_model, bias=False, dtype=dtype)
+        nn.init.normal_(self.linear.weight, std=0.2)
+        self.hook_out = HookPoint()
+        self.hook_out.name = f"blocks.{layer}.hook_out"
+
+    def forward(self, residual: torch.Tensor) -> torch.Tensor:
+        return self.hook_out(residual + self.linear(residual))
+
+
+class _LinearToyBridge(TransformerBridge):
```
jlarson4, line 53: `_LinearToyBridge` inherits `TransformerBridge` but overrides `hook_dict`, `hooks()`, and `check_hooks_to_add`, which are three behaviors `cache_activation_and_gradient` actually depends on. Its hook points carry no conversion and its `check_hooks_to_add` is a no-op, so both bridge failures above pass this suite green.

Can we add an integration test that runs `attribution_patch` against a real small bridge and asserts finite scores for all three node families?

tests/unit/tools/test_attribution_patching.py
```diff
@@ -0,0 +1,576 @@
...
+    with torch.no_grad():
+        m_corrupt = float(metric(model(corrupt)))
+
+    covered: set[str] = set()
+    for node, score in result.top_nodes(k=len(result.node_scores)):
+        delta_m = _patch_node_toward_clean(model, corrupt, node, clean_cache.activations, metric)
+        delta_m -= m_corrupt
+        assert delta_m == pytest.approx(score, abs=1e-6)
+        if abs(score) > 1e-6:
+            assert (delta_m > 0) == (score > 0)  # denoising sign convention
```
jlarson4, line 615: A linear model has an input-independent Jacobian, so the clean and corrupt caches hold identical gradients. Pointing `_node_effects` at `clean_cache.gradients` leaves all 16 tests passing, so the "gradients from the corrupt run" half of the denoising convention this test claims to pin is uncovered.

Please cover the convention on a model with a nonlinearity, where the two gradients differ.

transformer_lens/tools/analysis/attribution_patching.py
janmenjayap: Fixed in e8f6571c. The forward hooks now store the detached activation clone plus a live reference to the on-path tensor, and I register a backward hook at each cached point via `model.hooks(fwd_hooks=..., bwd_hooks=...)`. The backward hook goes through the same conversion, so at `blocks.0.attn.hook_z` it receives the gradient in the converted `[batch, seq, n_heads, d_head]` shape that matches the cached activation. Each bwd hook closes over its own alias `name` and writes into `gradients[name]`, so the gradient lands under the key `_node_effects` reads with no remap; the most-upstream cached point (whose own bwd hook never fires) is backfilled from the `torch.autograd.grad` return. The new integration test in fc98be49 boots a real GPT-2 Bridge and asserts finite scores for the `attn_head_out` family, the family this bug broke, and reproduces red on the pre-fix code.

janmenjayap: Thanks for catching this. It was a real bug and your diagnosis is right: at a converted point the tensor the forward hook sees is a reshaped view the model's forward never consumes, so `retain_grad()` does nothing and `.grad` stays `None`.

Fixed in 95e1423c. The forward hooks now store the detached activation clone plus a live reference to the on-path tensor, and I register a backward hook at each cached point via `model.hooks(fwd_hooks=..., bwd_hooks=...)`. The backward hook goes through the same conversion, so at `blocks.0.attn.hook_z` it receives the gradient in the converted `[batch, seq, n_heads, d_head]` shape that matches the cached activation. Each bwd hook closes over its own alias `name` and writes into `gradients[name]`, so the gradient lands under the key `_node_effects` reads with no remap; the most-upstream cached point (whose own bwd hook never fires) is backfilled from the `torch.autograd.grad` return. The new integration test in 36c4808 boots a real GPT-2 Bridge and asserts finite scores for the `attn_head_out` family, the family this bug broke, and reproduces red on the pre-fix code.

transformer_lens/tools/analysis/attribution_patching.py
janmenjayap: Thanks, agreed, and fixed in the same commit (95e1423c). I dropped `model.zero_grad(set_to_none=True)` and replaced `metric.backward()` with `torch.autograd.grad(metric, inputs=<live on-path tensors>, allow_unused=True, retain_graph=False)`, mirroring `backward_lens.py`. The gradients I keep come from the backward hooks; the `autograd.grad` return is discarded except for the upstream-point backfill noted above. No parameter `.grad` buffer is touched now, so no caller-accumulated grads are lost and the ~500 MB model-sized buffer is never allocated.

tests/unit/tools/test_attribution_patching.py
janmenjayap: Good point: with an input-independent Jacobian the clean and corrupt caches hold identical gradients, so nothing pins which run the score reads from. Covered in d1ad6ae. I added a `_NonlinearNodeGraphToyBridge` whose MLP carries a GELU, so `d(metric)/d(activation)` is input-dependent and the two runs' gradients genuinely differ. The new `test_nonlinear_node_scores_read_the_corrupt_run_gradient` first asserts the two caches' gradients differ at >=1 node (guarding the fixture against silently going linear), then checks the score's sign against the measured delta-metric under a corrupt-to-clean patch, so it fails if `_node_effects` reads the clean gradients instead of the corrupt ones.

tests/unit/tools/test_attribution_patching.py
janmenjayap: Agreed, and thank you for the sharp catch. `_LinearToyBridge` overrides `hook_dict`, `hooks()`, and `check_hooks_to_add`, so its hook points carry no conversion and its gate check is a no-op, which is exactly why both real-Bridge failures pass the model-free suite green. Added `tests/integration/test_attribution_patching.py` in 36c4808. It boots a real GPT-2 Bridge (module-scoped fixture, following `test_backward_lens.py`), runs `attribution_patch` with the default config and no explicit `names_filter`, and asserts the result scores exactly the node graph, every score is finite, and all three node families are present (`{embed, attn_head_out, mlp_out}`). I confirmed it reproduces the reported crash on the pre-95e1423c code and passes after the fixes.

transformer_lens/tools/analysis/attribution_patching.py
janmenjayap: Understood, and thank you, you are right that these are opaque to anyone outside the proposal; I will keep them out going forward. Cleaned up in 3039679. The `NotImplementedError` and raise messages now describe the missing capability directly ("granularity='edge' (EAP edge scoring) is not implemented yet", "ig_steps>1 (EAP-IG integrated gradients) is not implemented yet", "edge scoring is not implemented yet; run a node-granularity sweep and use top_nodes()"). The docstrings and the two "Risk 1" comments were reworded to keep the invariant description without the label. I updated the test assertions that matched on the old strings (`match="edge"` / `match="integrated gradient"`) and renamed `test_top_edges_not_implemented_until_pr2` to `test_top_edges_not_implemented`. `grep -nE 'PR[0-9]|Risk [0-9]|proposal step|lands in'` over the module and the test file now returns nothing.

transformer_lens/tools/analysis/attribution_patching.py
janmenjayap: Thanks for flagging this: the default `None` was documented as "every hook point", but a real `hook_dict` exposes gated points (`hook_mlp_in`, `attn.hook_result`, the split-QKV inputs) that `add_hook` rejects unless the `set_use_*` flag is on. Fixed in b266796: when `names_filter is None`, the helper falls back to the node graph's hook set via `_required_hook_names(model.cfg.n_layers)` rather than every hook point. The docstring now states that `None` means the node-granularity set (`hook_embed`, per-layer `attn.hook_z`, `hook_mlp_out`) and that an explicit filter is required to reach anything outside it. Covered end-to-end by the integration test in 36c4808, which relies on the default rather than an explicit filter, plus a unit assertion that `None` on the toy bridge caches exactly the node hook set.

transformer_lens/tools/analysis/attribution_patching.py
janmenjayap: Good catch, thank you. Fixed in 586628e. `cache_activation_and_gradient` now takes `compute_gradient: bool = True`; with `compute_gradient=False` it registers forward hooks only, skips the `bwd_hooks` and the `autograd.grad` call, and returns a `GradientCache` whose gradients are all `None`. `attribution_patch` now builds the clean cache with `compute_gradient=False`, so a pair runs the documented two forwards plus one backward instead of two backwards. A new unit assertion confirms a `compute_gradient=False` cache has all-`None` gradients and populated activations.

jlarson4:
Thanks for the quick resolution on all of these items @janmenjayap! Looks great, merging now

PR #1754, opened 2026-09-07
Clarify quantized dtype ownership; keep the whole-model skip (#1743)

### LINKED ISSUE #1743: [Bug Report] maybe_cast_floating_params skips dtype normalization for an entire quantized model, not just quantizer-owned tensors
**Describe the bug**

#1716 fixed a real bug (#1713): casting every floating parameter to the requested dtype during `boot_transformers` was corrupting FP8 scale tensors on quantized checkpoints (e.g. MXFP4's `float8_e8m0fnu` scales getting silently upcast to bf16, breaking the weight/scale pairing). The fix was implemented at two levels: a narrow per-parameter guard in `cast_floating_params_to_dtype` (skip any floating tensor with `itemsize < 2`, i.e. FP8/narrow floats), and a call-site guard in the new `maybe_cast_floating_params` that skips the cast for the *entire model* whenever `quantization_method(model.config)` returns non-None.

**What's directly proven (function-level, verified by running the actual merged code, not inferred from reading the diff):** `maybe_cast_floating_params` -- the exact function `boot()` calls -- skips *all* dtype casting for a model the moment any `quantization_config` is present, regardless of any individual parameter's actual dtype. The narrower per-parameter guard already in `cast_floating_params_to_dtype` is sufficient on its own to reproduce #1713's fix (protects the FP8 scale) while still correctly casting an ordinary floating parameter on the same model -- see the repro below.

**What's inferred, not directly checked against a live checkpoint (real-world impact):** quantization methods (bitsandbytes, GPTQ, AWQ, MXFP4, ...) typically only replace specific layers (usually `nn.Linear`); embeddings, layernorms, and often `lm_head` commonly stay in their original, non-quantized dtype. If that holds for a given checkpoint, `boot_transformers(repo_id, dtype=torch.bfloat16, ...)` would silently leave those non-quantized parameters in whatever dtype they loaded in rather than the caller's requested dtype, with no error or warning. This part is standard quantization practice and consistent with `cast_floating_params_to_dtype` existing in `boot()` since #1459 (replacing an even older raw per-parameter cast loop, i.e. it has been the relied-upon mechanism for honoring `dtype=`, not a redundant safety net) -- but I don't have a live quantized checkpoint on hand to confirm end-to-end against `boot_transformers` itself.

**Code example**
```python
import torch
import torch.nn as nn
from transformer_lens.utilities.multi_gpu import cast_floating_params_to_dtype, maybe_cast_floating_params

class TinyModel(nn.Module):
    def __init__(self):
        super().__init__()
        # A layernorm-like param that is NOT quantizer-owned -- realistic for
        # any bnb/GPTQ/AWQ/MXFP4 checkpoint, which quantize Linear layers only.
        self.layernorm_weight = nn.Parameter(torch.ones(8, dtype=torch.float32))
        self.fp8_scale = nn.Parameter(torch.ones(4, dtype=torch.float8_e8m0fnu))  # #1713's exact dtype

class FakeConfig:
    class quantization_config:
        quant_method = "bitsandbytes"

model = TinyModel()
model.config = FakeConfig()

# The narrow per-parameter guard alone already protects the FP8 scale AND
# correctly casts the non-quantized param:
cast_floating_params_to_dtype(model, torch.bfloat16)
print(model.layernorm_weight.dtype)  # bfloat16 (correct)
print(model.fp8_scale.dtype)         # float8_e8m0fnu, untouched (correct)

# The actual boot() call path uses maybe_cast_floating_params instead, which
# skips the WHOLE model:
model2 = TinyModel()
model2.config = FakeConfig()
maybe_cast_floating_params(model2, torch.bfloat16)
print(model2.layernorm_weight.dtype)  # float32 -- the dtype=bfloat16 request was silently ignored
```

The existing test suite currently encodes this as intended behavior: `tests/unit/utilities/test_multi_gpu_unit.py::TestMaybeCastFloatingParams::test_skips_quantized_model` asserts a plain fp32 `nn.Linear.weight` (not an FP8 scale, not a packed quantized tensor) stays uncast solely because *some* `quantization_config` is present anywhere on the model -- that's the exact behavior this report is about.

**System Info**

- Installed from source, `dev-4.x`
- OS: macOS (arm64), CPU-only (mechanism is device-independent, reproduces on a plain `nn.Module`)

Remove (or substantially narrow) the call-site guard in `maybe_cast_floating_params`; the per-parameter `itemsize < 2` guard already in `cast_floating_params_to_dtype` is sufficient to protect quantizer-owned narrow-float scale tensors (proven above) while still correctly normalizing every other floating parameter to the caller's requested dtype, matching the behavior for unquantized models. Flagging the mechanism with confidence; flagging the real-world severity as a likely-but-unconfirmed consequence, since I couldn't test against an actual quantized checkpoint here.

Acceptance:
- [ ] A quantized model's non-quantizer-owned floating parameters (layernorms, embeddings, unquantized `lm_head`, ...) are cast to the caller's requested `dtype` during `boot_transformers`, same as an unquantized model
- [ ] #1713's exact repro (FP8/narrow-float scale tensors) remains protected
- [ ] `test_skips_quantized_model` is updated to reflect the corrected, narrower scope (or replaced with a test using a realistic mixed-parameter module, matching `test_cast_skips_fp8_scales_even_if_called`'s pattern)

### Checklist

Canonik:
## Description

Fixes #1743, by correcting its premise rather than its proposed fix.

#1743 is right about the mechanism: `maybe_cast_floating_params` skips dtype normalization for an entire model the moment a `quantization_config` is present. Its proposed fix is to drop that call-site guard and rely on the `itemsize < 2` check inside `cast_floating_params_to_dtype`. That would reintroduce #1713.

`itemsize < 2` is not an ownership test. transformers' finegrained-FP8 picks `weight_scale_inv`'s storage dtype from the checkpoint:

```python
sf_dtype = _get_ue8m0_dtype() if scale_fmt == "ue8m0" else torch.float32
```

Same parameter, same quantizer, same role, two widths, and **float32 is the default**. `activation_scale` is float32 under either format, and fbgemm-fp8 stores `weight_scale` and its expert scales as float32 `nn.Parameter`s. Running the merged `cast_floating_params_to_dtype` against a real `FP8Linear` rewrites the float32 `weight_scale_inv` to bf16, which is #1713's corruption in a scale format the guard cannot see. It also corrupts `activation_scale` on `ue8m0` checkpoints, so even #1713's own format is not fully protected by dtype alone.

| `scale_fmt` | quantizer-owned params rewritten by the cast |
| --- | --- |
| `"ue8m0"` | `activation_scale` (float32) |
| `"float"` (default) | `weight_scale_inv`, `activation_scale` (both float32) |

Nothing is lost by skipping. `from_pretrained` settles the load dtype before this
helper runs, and on a quantized checkpoint that is the quantizer's *effective* dtype
rather than necessarily the requested one: nine quantizers override it in
`HfQuantizer.update_dtype` (AWQ downgrades bfloat16 to float16 whenever CUDA or XPU is
available, whatever the placement; fbgemm-FP8 and FP-Quant force bfloat16). Re-casting
here would overwrite those deliberate choices along with quantizer-owned storage.

On bitsandbytes, which overrides nothing, every ordinary floating parameter arrives at
the requested dtype already: 4-bit pre-quantized and 8-bit on-the-fly both show 0
mismatches, including with a deliberately conflicting `config.dtype=float32`. On the
4-bit checkpoint that is 100 distinct floating tensors against 48 packed `Params4bit`
(a naive module walk reports 101, because GPT-2 ties `wte.weight` and `lm_head.weight`).

This also matches the line transformers draws itself: `PreTrainedModel.to(dtype=...)` raises for bitsandbytes and GPTQ, and `.half()` / `.float()` raise for any quantized model ("the model has already been casted to the correct dtype"), all gated on the same whole-model `is_quantized` flag. And the gate releases in step with HF: `HfQuantizer.postprocess_model` calls `remove_quantization_config` on a dequantized load, deleting `config.quantization_config`, so `quantization_method` returns None and normalization resumes. That is the one way the guard could have been genuinely too broad, and it isn't.

**No runtime behavior change.** The contribution is corrected rationale plus regression tests for the actual ownership invariant. The previous docstring claimed the `itemsize < 2` branch skips "quantizer-owned scale parameters", which is false and is what made removing the call-site guard look safe.

## Type of change

## Tests

- `test_itemsize_guard_does_not_identify_quantizer_owned_scales[ue8m0|float]` uses a real `FP8Linear`. The expected set is derived from the cast's own predicate (floating, `itemsize >= 2`, not already at target) rather than dtype width, so it stays correct if transformers moves `weight` to wide packed-integer storage like GPTQ's int32.
- `test_preserves_quantizer_owned_scales_of_any_width[ue8m0|float]`: both scale widths survive an active quantizer. The float32 side fails outright if the cast is ever re-enabled behind only the one-byte guard.
- `test_skips_quantized_model`: restored, with the evidence for why.
- `test_casts_once_hf_releases_quantizer_ownership`: casting resumes once ownership is released.
- `test_hf_still_clears_quantization_config_when_dequantizing`: pins the upstream behavior this design depends on, so a transformers change fails loudly instead of silently stranding dequantized checkpoints.

Anti-tautology check: applying #1743's proposed fix fails 4 of these tests.

## Checklist

`make unit-test`: 5581 passed. The 8 failures are pre-existing on `dev` and reproduce unchanged at `b652cfa8` in a clean worktree. 6 are CPU/CUDA device-mixing failures that pass under `CUDA_VISIBLE_DEVICES=""` (CI runs CPU-only per `tests/QUARANTINES.md`), and 2 are `test_attention.py` bitsandbytes stub tests that CI skips via `skipif(not is_bitsandbytes_available())`. `uv run mypy .` clean, `black` / `isort` / `pycln` clean.

### DIFF
--- a/tests/unit/utilities/test_multi_gpu_unit.py
+++ b/tests/unit/utilities/test_multi_gpu_unit.py
@@ -300,9 +300,20 @@ class TestMaybeCastFloatingParams:
         assert model.weight.dtype == torch.bfloat16
 
     def test_skips_quantized_model(self):
-        """Quantized models should NOT have their params cast."""
-        from types import SimpleNamespace
+        """An active quantizer means HF owns the storage dtypes, so nothing is cast.
 
+        The skip is whole-model deliberately. ``from_pretrained`` is responsible for
+        applying the requested dtype to ordinary floating parameters before this helper
+        runs, and TransformerLens must not second-guess quantizer-owned storage
+        afterward. Any parameter still off the requested dtype here may be
+        quantizer-owned, and dtype alone cannot distinguish ownership safely (see
+        ``test_itemsize_guard_does_not_identify_quantizer_owned_scales``). transformers
+        draws the same line itself: ``.to(dtype=...)`` raises for bitsandbytes and
+        GPTQ models, ``.half()`` / ``.float()`` for any quantized model.
+
+        See: https://github.com/TransformerLensOrg/TransformerLens/issues/1713
+        See: https://github.com/TransformerLensOrg/TransformerLens/issues/1743
+        """
         from transformer_lens.utilities.multi_gpu import maybe_cast_floating_params
 
         model = nn.Linear(4, 4)
@@ -311,6 +322,28 @@ class TestMaybeCastFloatingParams:
 
         maybe_cast_floating_params(model, torch.bfloat16)
         assert model.weight.dtype == torch.float32  # NOT cast
+
+    @pytest.mark.parametrize("scale_fmt", ["ue8m0", "float"])
+    def test_preserves_quantizer_owned_scales_of_any_width(self, scale_fmt):
+        """Both widths of ``weight_scale_inv`` survive, which the dtype guard cannot do.
+
+        The ordinary parameter is set up as the loader would have delivered it, at the
+        requested dtype, so the correct outcome for the whole model is that nothing
+        moves. Under ``scale_fmt="float"`` the scale is float32, so this fails outright
+        if the cast is ever re-enabled behind only the one-byte-float guard.
+        """
+        from transformer_lens.utilities.multi_gpu import maybe_cast_floating_params
+
+        model = nn.Module()
+        model.quantized = _fp8_linear(scale_fmt)
+        model.ln_weight = nn.Parameter(torch.ones(4, dtype=torch.bfloat16))
+        model.config = SimpleNamespace(quantization_config=SimpleNamespace(quant_method="fp8"))
+
+        before = {name: param.dtype for name, param in model.named_parameters()}
+        maybe_cast_floating_params(model, torch.bfloat16)
+
+        assert {name: param.dtype for name, param in model.named_parameters()} == before
+        assert model.quantized.weight_scale_inv.dtype == before["quantized.weight_scale_inv"]
 
     def test_skips_model_without_config(self):
         """Models without a config attribute should be cast (no quantization)."""
--- a/tests/unit/utilities/test_multi_gpu_unit.py
+++ b/tests/unit/utilities/test_multi_gpu_unit.py
@@ -135,6 +135,31 @@ class TestGetDeviceForBlockIndex:
         assert result.type == "cpu"
 
 
+def _fp8_linear(scale_fmt: str) -> nn.Module:
+    """A real transformers finegrained-FP8 ``Linear``, or skip if the integration moved."""
+    integration = pytest.importorskip(
+        "transformers.integrations.finegrained_fp8",
+        reason="requires transformers' finegrained-FP8 integration",
+    )
+    fp8_linear = getattr(integration, "FP8Linear", None)
+    if fp8_linear is None:
+        pytest.skip("transformers.integrations.finegrained_fp8.FP8Linear is unavailable")
+    return fp8_linear(
+        in_features=128,
+        out_features=128,
+        block_size=(128, 128),
+        activation_scheme="static",
+        scale_fmt=scale_fmt,
+        has_bias=True,
+    )
+
+
+# Quantizer-owned storage on FP8Linear: the packed weight and its scales. ``bias`` is
+# excluded on purpose -- HF keeps biases in the model's compute dtype, so a bias is an
+# ordinary parameter that the quantizer does not own.
+_FP8_OWNED_PARAMS = ("weight", "weight_scale_inv", "activation_scale")
+
+
 class TestCastFloatingParamsToDtype:
     """Regression tests for cast_floating_params_to_dtype.
 
@@ -211,6 +236,46 @@ class TestCastFloatingParamsToDtype:
         assert model.standard_weight.dtype == torch.bfloat16
         assert model.fp8_scale.dtype == torch.float8_e4m3fn
         assert model.packed_weight.dtype == torch.int8
+
+    @pytest.mark.parametrize("scale_fmt", ["ue8m0", "float"])
+    def test_itemsize_guard_does_not_identify_quantizer_owned_scales(self, scale_fmt):
+        """``itemsize < 2`` is a narrow-float guard, not an ownership test.
+
+        Transformers picks ``weight_scale_inv``'s storage dtype from the checkpoint's
+        ``scale_fmt``: a one-byte UE8M0 float for "ue8m0", float32 for "float" (the
+        default). Both spell the same quantizer-owned scale, and ``activation_scale``
+        is float32 under either format. So a one-byte test protects some quantizer-owned
+        scales and silently rewrites others, which is why the cast has to be gated on
+        the model having no active quantizer rather than on per-parameter dtype.
+
+        See: https://github.com/TransformerLensOrg/TransformerLens/issues/1743
+        """
+        target = torch.bfloat16
+        module = _fp8_linear(scale_fmt)
+        before = {name: param.dtype for name, param in module.named_parameters()}
+        cast_floating_params_to_dtype(module, target)
+        after = {name: param.dtype for name, param in module.named_parameters()}
+
+        owned = {name: before[name] for name in _FP8_OWNED_PARAMS if name in before}
+        assert owned, "FP8Linear exposed none of its quantizer-owned parameters"
+        # Mirror the cast's whole predicate, not dtype width alone: it rewrites a
+        # parameter only when that parameter is floating, wider than one byte and not
+        # already at the target. (Its fourth condition, meta, cannot apply to this
+        # materialized fixture.) Deriving the set this way stays correct if
+        # transformers moves `weight` to a wide packed-integer storage like GPTQ's
+        # int32, which the cast is right to skip.
+        eligible = {
+            name
+            for name, dtype in owned.items()
+            if dtype.is_floating_point and dtype.itemsize >= 2 and dtype != target
+        }
+        assert eligible, "expected a quantizer-owned float wider than one byte"
+        rewritten = {name for name, dtype in owned.items() if after[name] != dtype}
+        assert rewritten == eligible, (
+            "the cast should rewrite exactly those quantizer-owned parameters eligible "
+            f"under its dtype-only predicate: rewritten={sorted(rewritten)} "
+            f"eligible={sorted(eligible)}"
+        )
 
 
 class TestMaybeCastFloatingParams:
--- a/tests/unit/utilities/test_multi_gpu_unit.py
+++ b/tests/unit/utilities/test_multi_gpu_unit.py
@@ -345,6 +345,48 @@ class TestMaybeCastFloatingParams:
         assert {name: param.dtype for name, param in model.named_parameters()} == before
         assert model.quantized.weight_scale_inv.dtype == before["quantized.weight_scale_inv"]
 
+    def test_casts_once_hf_releases_quantizer_ownership(self):
+        """The guard is a hand-off, not a permanent opt-out.
+
+        ``HfQuantizer.postprocess_model`` calls ``remove_quantization_config`` when a
+        checkpoint is loaded with ``dequantize=True``, which deletes
+        ``config.quantization_config``. ``quantization_method`` then returns None and
+        normalization resumes. That is what stops the whole-model guard from stranding a
+        genuinely dequantized checkpoint in its load dtype.
+        """
+        from transformer_lens.utilities.multi_gpu import maybe_cast_floating_params
+
+        model = nn.Linear(4, 4)
+        model.weight = nn.Parameter(torch.zeros(4, 4, dtype=torch.float32))
+        model.config = SimpleNamespace(quantization_config=SimpleNamespace(quant_method="mxfp4"))
+
+        maybe_cast_floating_params(model, torch.bfloat16)
+        assert model.weight.dtype == torch.float32
+
+        del model.config.quantization_config  # what remove_quantization_config does
+        maybe_cast_floating_params(model, torch.bfloat16)
+        assert model.weight.dtype == torch.bfloat16
+
+    def test_hf_still_clears_quantization_config_when_dequantizing(self):
+        """Pins the upstream behaviour the whole-model guard is calibrated against.
+
+        If transformers ever stops deleting ``quantization_config`` on a dequantized
+        load, ``quantization_method`` would keep reporting a method for a model whose
+        storage the quantizer no longer owns, and the guard really would be too broad.
+        Fail here rather than silently stranding those checkpoints.
+        """
+        base = pytest.importorskip("transformers.quantizers.base")
+
+        model = nn.Linear(4, 4)
+        model.config = SimpleNamespace(quantization_config=SimpleNamespace(quant_method="mxfp4"))
+        model.is_quantized = True
+        # Called unbound: remove_quantization_config only touches `model`, and building
+        # a real quantizer would need a live quantization config per method.
+        base.HfQuantizer.remove_quantization_config(None, model)
+
+        assert not hasattr(model.config, "quantization_config")
+        assert model.is_quantized is False
+
     def test_skips_model_without_config(self):
         """Models without a config attribute should be cast (no quantization)."""
         from transformer_lens.utilities.multi_gpu import maybe_cast_floating_params
--- a/tests/unit/utilities/test_multi_gpu_unit.py
+++ b/tests/unit/utilities/test_multi_gpu_unit.py
@@ -144,6 +144,10 @@ def _fp8_linear(scale_fmt: str) -> nn.Module:
     fp8_linear = getattr(integration, "FP8Linear", None)
     if fp8_linear is None:
         pytest.skip("transformers.integrations.finegrained_fp8.FP8Linear is unavailable")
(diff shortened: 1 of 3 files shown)


transformer_lens/utilities/multi_gpu.py
```diff
@@ -273,11 +281,27 @@ def cast_floating_params_to_dtype(model: nn.Module, dtype: torch.dtype) -> None:
 def maybe_cast_floating_params(model: nn.Module, dtype: torch.dtype) -> None:
     """Cast floating params to dtype, skipping models with active quantization.
 
-    When a model has an active quantization_config, the quantizer owns specific
-    dtypes (e.g., FP8 scales) that must not be overwritten. This helper wraps
-    the cast with that check.
+    The skip is whole-model on purpose, and it is a division of responsibility rather
+    than a loss. ``from_pretrained`` is the component that applies the requested dtype
```
jlarson4: Three quantizers deliberately override the requested dtype in `update_dtype`, which `from_pretrained` calls before any parameter is materialised: AWQ turns `bfloat16` into `float16` on CUDA/XPU, and fbgemm-FP8 and FP-Quant turn anything that is not `bfloat16` into `bfloat16`. On those checkpoints the layernorms and embeddings end up at the quantizer's dtype, not the caller's, and they are not quantizer-owned. A reader who tests this docstring's claim finds a counterexample and reopens #1743. The bitsandbytes measurement in the parenthetical can't surface it, because bitsandbytes is one of the quantizers with no `update_dtype` override.

Please update to mention that the skip defers to HF's *effective* load dtype, which the quantizer may have overridden on purpose, rather than asserting the requested dtype was applied. Mirror the same wording in the comment at `transformer_lens/model_bridge/sources/transformers.py:851-855`.

transformer_lens/utilities/multi_gpu.py
```diff
@@ -273,11 +281,27 @@ def cast_floating_params_to_dtype(model: nn.Module, dtype: torch.dtype) -> None:
...
+    to ordinary floating parameters, and it has already run by this point; whatever is
+    still not at the requested dtype afterward may be quantizer-owned storage, and
+    dtype alone cannot distinguish ownership safely. (bitsandbytes 4-bit and 8-bit
```
jlarson4: "bitsandbytes 4-bit and 8-bit checkpoints were checked… Not every quantization backend was reachable to test" describes only one test environment at one moment. It duplicates the PR description and can go become incorrect over time. At 23 lines this is now the longest docstring in the file by a wide margin. Can we trim this down? Cut the measurement parenthetical (lines 288-292) and the `.half()` / `.float()` paragraph (lines 294-301). Keep the ownership rule and the two issue links.

tests/unit/utilities/test_multi_gpu_unit.py
```diff
@@ -247,6 +323,70 @@ def test_skips_quantized_model(self):
...
+
+        model = nn.Module()
+        model.quantized = _fp8_linear(scale_fmt)
+        model.ln_weight = nn.Parameter(torch.ones(4, dtype=torch.bfloat16))
+        model.config = SimpleNamespace(quantization_config=SimpleNamespace(quant_method="fp8"))
+
+        before = {name: param.dtype for name, param in model.named_parameters()}
+        maybe_cast_floating_params(model, torch.bfloat16)
+
+        assert {name: param.dtype for name, param in model.named_parameters()} == before
```
jlarson4, line 358: `FP8Linear(..., has_bias=True)` gives a float32 `bias`, and the whole-dict assertion pins it as uncast. Line 157-159 calls a bias an ordinary parameter the quantizer does not own, and the docstring says the ordinary parameter is set up at the requested dtype, but only `ln_weight` is. A future narrowing that preserves every quantizer-owned scale still turns this test red, on `quantized.bias` alone, so it fails for a reason its name disclaims. `test_skips_quantized_model` already covers ordinary parameters.

Please put the bias at the target dtype in the fixture near line 339 (`model.quantized.bias = nn.Parameter(model.quantized.bias.to(torch.bfloat16))`). The test will still fail on `activation_scale` and `weight_scale_inv` if the whole-model guard is removed.

tests/unit/utilities/test_multi_gpu_unit.py
```diff
@@ -247,6 +323,70 @@ def test_skips_quantized_model(self):
...
+        Fail here rather than silently stranding those checkpoints.
+        """
+        base = pytest.importorskip("transformers.quantizers.base")
+
+        model = nn.Linear(4, 4)
+        model.config = SimpleNamespace(quantization_config=SimpleNamespace(quant_method="mxfp4"))
+        model.is_quantized = True
+        # Called unbound: remove_quantization_config only touches `model`, and building
+        # a real quantizer would need a live quantization config per method.
+        base.HfQuantizer.remove_quantization_config(None, model)
```
jlarson4: The docstring says the test fires if transformers stops clearing `quantization_config` on a dequantized load, but calling `remove_quantization_config` directly only pins that method's body. Delete the call from `HfQuantizer.postprocess_model` and the test still passes.

Lets drive `HfQuantizer.postprocess_model` instead, with a stub carrying `pre_quantized=True` and a `quantization_config` with `dequantize=True`, then assert `quantization_config` is gone.

tests/unit/utilities/test_multi_gpu_unit.py
```diff
@@ -135,6 +135,31 @@ def test_cpu_device_is_returned_unchanged(self):
         assert result.type == "cpu"
 
 
+def _fp8_linear(scale_fmt: str) -> nn.Module:
+    """A real transformers finegrained-FP8 ``Linear``, or skip if the integration moved."""
+    integration = pytest.importorskip(
```
jlarson4, line 140: `pyproject.toml` declares `torch>=2.6`, but `scale_fmt="ue8m0"` reaches `_get_ue8m0_dtype()`, which raises `RuntimeError` when `torch.float8_e8m0fnu` is missing, which will happen on torch 2.6 (this was changed in 2.7). Both `ue8m0` parameterisations error rather than skip, despite the helper's docstring promising a skip. Please add `if not hasattr(torch, "float8_e8m0fnu"): pytest.skip(...)` before constructing the `ue8m0` variant.

jlarson4 (commented):
Thanks for thinking this through thoroughly @Canonik, rather than taking the issue at face value. I agree with the ownership argument and the anti-tautology check reproduced exactly. Comments on the rationale in `maybe_cast_floating_params` + the fixture in the scale-preservation test + the upstream-pinning test.

transformer_lens/utilities/multi_gpu.py
Canonik: Confirmed, and this is the one that matters. Nine quantizers override `update_dtype` in transformers 5.13 and bitsandbytes 4-bit/8-bit are not among them, so the measurement in that parenthetical was structurally incapable of catching this. `modeling_utils.py:879` calls `hf_quantizer.update_dtype(dtype)` during dtype resolution, before any parameter is materialised.

The three you named check out: AWQ turns `bfloat16` into `float16` when CUDA or XPU is available, fbgemm-FP8 and FP-Quant turn anything that is not `bfloat16` into `bfloat16`. So on an AWQ checkpoint loaded with `dtype=torch.bfloat16` the layernorms and embeddings land at float16, they are not quantizer-owned, and my docstring asserted they were at the requested dtype.

Reworded to say the skip defers to the load dtype `from_pretrained` settled, which on a quantized checkpoint is the quantizer's effective dtype rather than necessarily the requested one, and mirrored in the `transformers.py` call-site comment.

This actually strengthens the case for the guard: if a quantizer overrode the dtype deliberately, post-casting to the caller's dtype would be overriding that decision, which is the same failure class as #1713.

While rewording I found the same defect one function up: `cast_floating_params_to_dtype` claimed `activation_scale` is float32 "either way", but it only exists under `activation_scheme="static"` and the default is `"dynamic"`, where it is `None`. Replaced with the unconditional version (finegrained-FP8 keeps `weight_scale_inv` float32 unless the checkpoint asks for ue8m0, fbgemm-FP8 keeps its scales float32 outright).

transformer_lens/utilities/multi_gpu.py
Canonik: Agreed. The measurement was environment-and-moment specific and duplicated the PR description, which is the right home for it since that does not have to stay true indefinitely. Cut it, along with the `.to()` / `.half()` / `.float()` enumeration.

Docstring is 23 lines to 16, so no longer the longest in the file (`get_device_for_block_index` is 19).

Two notes on what I kept. Your other comment asks for the effective-versus-requested dtype distinction here, so that paragraph got rewritten rather than just shortened. And I kept one sentence on the dequantize release, because `test_hf_still_clears_quantization_config_when_dequantizing` pins it and it is the reason the whole-model guard is not a permanent opt-out. Happy to drop that too if you would rather it lived only in the test.

tests/unit/utilities/test_multi_gpu_unit.py
Canonik: Agreed, and it contradicted my own comment on `_FP8_OWNED_PARAMS`, which excludes `bias` precisely because HF keeps biases in the compute dtype. `FP8Linear(has_bias=True)` allocates `bias` at float32, so `before` pinned it as uncast and the docstring's "the ordinary parameter is set up at the requested dtype" was only true of `ln_weight`.

Took your fixture change and fixed the docstring to say every ordinary parameter. Verified the failure set with the guard removed:

- `scale_fmt="ue8m0"`: `quantized.activation_scale`
- `scale_fmt="float"`: `quantized.weight_scale_inv`, `quantized.activation_scale`

`bias` is no longer in it, so the test now fails only on quantizer-owned storage, which is what its name claims.

tests/unit/utilities/test_multi_gpu_unit.py
Canonik: You are right. Calling `remove_quantization_config` directly only pinned that method's body, and the branch this design actually depends on is in `postprocess_model`, so the docstring's claim was false.

Switched to your suggestion: an `HfQuantizer` subclass with `pre_quantized=True` and a `quantization_config` carrying `dequantize=True`, implementing the two abstract members (`is_serializable`, `is_trainable`) and the two `_process_model_*` hooks as no-ops, then asserting `quantization_config` is gone and `is_quantized` is False after `postprocess_model(model)`.

Verified it bites, by patching the branch out of the installed transformers:

- with the dequantize branch removed from `postprocess_model`, the new test FAILS
- the old direct-call version still passes under the same mutation

So the test now pins the wiring rather than the helper.

tests/unit/utilities/test_multi_gpu_unit.py
Canonik: Confirmed. `pyproject.toml:21` declares `torch>=2.6`, and `_get_ue8m0_dtype()` raises `RuntimeError` rather than falling back when `torch.float8_e8m0fnu` is absent, which is torch < 2.7. Both `ue8m0` parameterisations errored on 2.6 while the helper docstring promised a skip.

Added the guard before constructing the `ue8m0` variant, with the same `torch < 2.7` reason string `test_skips_one_byte_floats_fp8_scales` already uses in this file.

Verified by deleting `torch.float8_e8m0fnu` with `monkeypatch.delattr` after importing the module: `_fp8_linear("ue8m0")` now raises `Skipped` carrying that reason instead of `RuntimeError`, and the `"float"` variant still constructs with a float32 `weight_scale_inv`.

tests/unit/utilities/test_multi_gpu_unit.py
Canonik: Correction to my reply above. I repeated the `_FP8_OWNED_PARAMS` rationale ("HF keeps biases in the compute dtype") as if it were true, and it is not.

`FbgemmFp8Linear.__init__` hardcodes the bias at float32 (`integrations/fbgemm_fp8.py:115`) while `FbgemmFp8HfQuantizer.update_dtype` forces the model to bfloat16, so an fbgemm-FP8 model genuinely carries float32 biases in a bfloat16 model. Same overclaiming pattern you flagged in the other thread, one function away.

The accurate reason to exclude `bias` is ownership, not dtype: `param_needs_quantization` returns False for `tensor_name == "bias"` in both `quantizer_finegrained_fp8.py` and `quantizer_fbgemm_fp8.py`, so the quantizer never claims it. Rewrote the comment on that basis, and dropped the "as the loader would have delivered it" phrasing from the test docstring since that was an unverified claim about loader behaviour too. The fixture change itself is unaffected.

transformer_lens/utilities/multi_gpu.py
Canonik: One follow-up on precision here. The docstring as I first reworded it said AWQ downgrades "on CUDA/XPU", which reads as conditioned on placement. `quantizer_awq.py:58` keys off `torch.cuda.is_available() or torch.xpu.is_available()` and never consults `device_map`, so an AWQ checkpoint loaded with `device_map={"": "cpu"}` on a GPU box still comes back float16. Changed to "whenever CUDA or XPU is available, whatever the placement" and mirrored it in the call-site comment.

jlarson4:
Thanks for closing all of those comment requests @Canonik! Looks great, no additional thoughts, merging now

PR #1757, opened 2026-09-08
fix(bridge): decode every batch row in generate_stream

### LINKED ISSUE #1756: [Bug Report] generate_stream(return_type="str") decodes only the first sequence of a batch
**Describe the bug**

`TransformerBridge.generate_stream(..., return_type="str")` decodes only row 0 of the batch:

```python
def _maybe_decode(tokens: torch.Tensor) -> Union[torch.Tensor, str]:
    if return_type == "str":
        assert self.tokenizer is not None
        return self.tokenizer.decode(tokens[0], skip_special_tokens=True)
    return tokens
```

So a batched stream yields a single string — the first sequence's text — and every other row of the batch is silently unreachable. The generation loop itself is fine (all rows are generated, and `return_type="tokens"` yields the full `[batch, seq_len]` tensor); only the decode step drops them.

This diverges from `generate()`, which decodes each row and unwraps a one-row batch:

```python
decoded_texts = [
    self.tokenizer.decode(tokens, skip_special_tokens=True) for tokens in output_tokens
]
result = decoded_texts[0] if len(decoded_texts) == 1 else decoded_texts
```

Single-string / single-row streaming is unaffected, which is why the existing coverage (`tests/acceptance/model_bridge/test_generate_stream.py::test_stream_returns_strings`) passes.

**Code example**

```python

bridge = TransformerBridge.boot_transformers("distilgpt2", device="cpu")
prompts = ["The capital of France is", "My favourite colour is"]

print(bridge.generate(prompts, max_new_tokens=4, do_sample=False, verbose=False))
for chunk in bridge.generate_stream(prompts, max_new_tokens=4, do_sample=False, verbose=False):
    print(repr(chunk))
```

Actual:

```
['The capital of France is the capital of the', 'My favourite colour is the red. I']   # generate()
'The capital of France is the capital of the'                                          # generate_stream()
```

Expected — one string per batch row, matching `generate()`:

```
['The capital of France is the capital of the', 'My favourite colour is the red. I']
```

**System Info**

* Installed from source (`uv sync`), branch at `dev` (b652cfa8)
* Windows 11; nothing platform-specific — this is the plain CPU path
* Python 3.12

**Additional context**

Fix: decode every row and keep the `generate()` unwrap convention (bare string for a one-row batch, `list[str]` otherwise), widening the yield type to `Generator[Union[torch.Tensor, str, List[str]], None, None]`.

`HookedTransformer.generate_stream` does not share this bug — it never decodes at all and always yields tensors, even for `return_type="str"`. That is a separate inconsistency; this issue is Bridge-only, so no mirroring change is required.

I have a fix with unit + integration coverage and will open a PR against `dev`.

### Checklist

emerardd:
# Description

`TransformerBridge.generate_stream(..., return_type="str")` decoded `tokens[0]` only, so a batched stream yielded the first sequence's text and the rest of the batch was unreachable. Generation itself was fine — all rows are produced, and `return_type="tokens"` yields the full `[batch, seq_len]` tensor — only the decode step dropped them.

This decodes every row and keeps `generate()`'s unwrap convention: a bare string for a one-row batch, one string per row otherwise. The yield type widens to `Generator[Union[torch.Tensor, str, List[str]], None, None]`.

```python
prompts = ["The capital of France is", "My favourite colour is"]
bridge.generate(prompts, max_new_tokens=4, do_sample=False, verbose=False)
# ['The capital of France is the capital of the', 'My favourite colour is the red. I']
```

| `generate_stream(prompts, ...)` | Before | After |
| --- | --- | --- |
| yield | `'The capital of France is the capital of the'` | `['The capital of France is the capital of the', 'My favourite colour is the red. I']` |

Single-string and single-row streaming are unchanged, which is why `tests/acceptance/model_bridge/test_generate_stream.py::test_stream_returns_strings` passed throughout.

`HookedTransformer.generate_stream` does not share the bug — it never decodes and always yields tensors, even for `return_type="str"`. That is a separate inconsistency, so no mirroring change ([AGENTS.md §2](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/AGENTS.md#2-two-systems-live-in-this-repo)) is required here.

Fixes #1756

## Type of change

# Checklist:

Tests added — both fail on `dev` and pass with the fix:

- `tests/unit/model_bridge/test_generate_stream_batch_decode.py` — stubbed token stream, no model load; pins the batched `list[str]` yield and the single-row bare-string unwrap.
- `tests/integration/model_bridge/test_generate_stream_batched.py` — distilgpt2 end-to-end; each yield carries one string per prompt and the first yield echoes each row's own prompt.

### DIFF
--- a/tests/integration/model_bridge/test_generate_stream_batched.py
+++ b/tests/integration/model_bridge/test_generate_stream_batched.py
@@ -1,0 +1,48 @@
+"""Batched ``generate_stream(return_type="str")`` decodes every row of the batch.
+
+The streaming decoder used to decode ``tokens[0]`` only, so a batched call
+yielded the first sequence's text for the whole batch and the other rows were
+never observable. See ``generate()``, which decodes each row and unwraps a
+single-row batch — ``generate_stream`` now matches that contract.
+"""
+
+PROMPTS = ["The capital of France is", "My favourite colour is"]
+
+
+def test_batched_stream_yields_one_string_per_row(distilgpt2_bridge):
+    chunks = list(
+        distilgpt2_bridge.generate_stream(
+            PROMPTS,
+            max_new_tokens=4,
+            max_tokens_per_yield=2,
+            do_sample=False,
+            verbose=False,
+            return_type="str",
+        )
+    )
+
+    assert chunks, "expected at least one yield"
+    for chunk in chunks:
+        assert isinstance(chunk, list)
+        assert len(chunk) == len(PROMPTS)
+        assert all(isinstance(text, str) for text in chunk)
+
+    # The first yield carries the input tokens, so each row must echo its own prompt.
+    for prompt, streamed in zip(PROMPTS, chunks[0]):
+        assert prompt in streamed
+
+
+def test_single_prompt_stream_stays_a_bare_string(distilgpt2_bridge):
+    """A one-element batch keeps the scalar contract shared with generate()."""
+    chunks = list(
+        distilgpt2_bridge.generate_stream(
+            [PROMPTS[0]],
+            max_new_tokens=2,
+            do_sample=False,
+            verbose=False,
+            return_type="str",
+        )
+    )
+
+    assert chunks
+    assert all(isinstance(chunk, str) for chunk in chunks)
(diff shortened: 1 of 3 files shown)


jlarson4:
Great work on this @emerardd! Assuming Full Code Coverage passes, I will merge

PR #1761, opened 2026-09-09
Fix batched hf_generate padding and attention masks

### LINKED ISSUE #1760: [Bug Report] hf_generate right-pads batched strings and drops the attention mask
**Describe the bug**

`TransformerBridge.hf_generate()` silently produces incorrect continuations for shorter rows in a
variable-length `list[str]` batch.

The list-input branch tokenizes with `padding=True` using the tokenizer's current padding side. For
GPT-2-family tokenizers this is normally right padding. Although the tokenizer returns an
`attention_mask`, `hf_generate()` extracts only `input_ids` and calls
`original_model.generate(input_ids, **generation_kwargs)`, so Hugging Face sees the pad columns as
real context.

This is difficult to notice because the pad token is commonly the EOS token and
`skip_special_tokens=True` hides the padding in the returned text. The call succeeds and returns a
plausible but different continuation from generating the same prompt alone.

**Code example**

On current `dev` (`7edc5ddf`), using `distilgpt2`, greedy generation, and two deliberately unequal
prompts:

```python
prompts = ["The capital of France is", "Hi"]

batched = bridge.hf_generate(
    prompts,
    max_new_tokens=8,
    do_sample=False,
    return_type="input",
)
solo = [
    bridge.hf_generate(
        prompt,
        max_new_tokens=8,
        do_sample=False,
        return_type="input",
    )
    for prompt in prompts
]
```

Observed output for the short row:

```text
batched: 'HiThe following is a list of the most'
solo:    'Hi everyone,\n\n\n\n\n\n'
```

`TransformerBridge.generate(prompts, ...)` agrees with the solo control for the short row because
that path forces left padding and preserves the relevant mask semantics.

The current equal-length batch test uses `['Hello', 'World']`, so no padding is emitted and it cannot
detect this failure.

**Impact**

- Every row shorter than the longest prompt can be continued from padding rather than its last real
  token.
- Batched `hf_generate()` can disagree with row-by-row `hf_generate()` under greedy decoding.
- Stateful/SSM `generate(use_past_kv_cache=False)` delegates to `hf_generate()`, so its batched string
  fallback reaches the same path.

**Suggested scope**

For `list[str]` input, force left padding without permanently mutating the shared tokenizer, retain
the tokenizer's `attention_mask`, and pass it to `original_model.generate()`. Preserve tensor-input
behavior and HF `ModelOutput` flags.

A larger alternative would route through `to_tokens()`, but that also changes BOS and truncation
semantics. I plan to keep this fix focused unless maintainers prefer that broader unification.

**System Info**

- Windows 11 / PowerShell 7
- Python 3.12.10
- Reproduced on `upstream/dev` commit `7edc5ddfccc24619109e3c98db29a870d99977ec`

**Additional context**

Issue #1690 / PR #1701 fixed related batched padding behavior in the normal Bridge tokenization and
generation paths, but did not change `hf_generate()`. This report is limited to the HF-delegation
path.

### Checklist

emerardd:
# Description

Fixes #1760.

`TransformerBridge.hf_generate()` now left-pads batched string prompts and forwards the tokenizer's `attention_mask` to the underlying Hugging Face `generate()` call. The tokenizer's original padding side is restored even if tokenization fails, and an explicitly supplied `attention_mask` still takes precedence.

This prevents shorter prompts from treating right-padding tokens as generation context. Tensor inputs and scalar string inputs retain their existing behavior.

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation (no documentation change is required; this restores the documented batched-generation behavior)
- [ ] New and existing unit tests pass locally with my changes (the focused unit and integration surfaces pass; the full suite was not rerun)

## Validation

- `uv run --no-cache --no-sync pytest tests/unit/model_bridge/test_hf_generate_batch_padding.py -q`: 4 passed
- `uv run --no-cache --no-sync pytest tests/integration/test_generation_compatibility.py::TestTransformerBridgeHFGenerate::test_hf_generate_batch_generation tests/integration/test_generation_compatibility.py::TestTransformerBridgeHFGenerate::test_hf_generate_variable_length_batch_matches_solo_generation -q`: 2 passed
- pinned pycln, isort, and Black on all changed Python files
- `uv run --no-cache --no-sync mypy .`: success across 396 source files

### DIFF
--- a/tests/integration/test_generation_compatibility.py
+++ b/tests/integration/test_generation_compatibility.py
@@ -376,7 +376,10 @@ class TestTransformerBridgeHFGenerate:
 
     def test_hf_generate_variable_length_batch_matches_solo_generation(self, gpt2_bridge):
         """Test padded batch rows retain the same context as solo generation."""
-        prompts = ["Hi", "The capital of France is"]
+        prompts = [
+            "A",
+            "Mechanistic interpretability studies how neural networks represent and transform information internally.",
+        ]
 
         batched = gpt2_bridge.hf_generate(
             prompts,
--- a/tests/integration/test_generation_compatibility.py
+++ b/tests/integration/test_generation_compatibility.py
@@ -374,6 +374,28 @@ class TestTransformerBridgeHFGenerate:
         assert isinstance(result, list), "Batch input should return list"
         assert len(result) == len(prompts), "Output list should match input length"
 
+    def test_hf_generate_variable_length_batch_matches_solo_generation(self, gpt2_bridge):
+        """Test padded batch rows retain the same context as solo generation."""
+        prompts = ["Hi", "The capital of France is"]
+
+        batched = gpt2_bridge.hf_generate(
+            prompts,
+            max_new_tokens=2,
+            do_sample=False,
+            output_logits=True,
+        )
+        solo = [
+            gpt2_bridge.hf_generate(
+                prompt,
+                max_new_tokens=2,
+                do_sample=False,
+                output_logits=True,
+            )
+            for prompt in prompts
+        ]
+
+        assert batched == solo
+
 
 class TestGenerationBackwardCompatibility:
     """Tests to ensure backward compatibility with existing generation usage."""
--- a/tests/unit/model_bridge/test_hf_generate_batch_padding.py
+++ b/tests/unit/model_bridge/test_hf_generate_batch_padding.py
@@ -1,0 +1,104 @@
+"""Batched ``hf_generate`` calls preserve each prompt's generation context."""
+
+from types import MethodType, SimpleNamespace
+from typing import Any
+
+import torch
+
+from transformer_lens.model_bridge.bridge import TransformerBridge
+
+
+class _BatchEncoding(dict[str, torch.Tensor]):
+    def to(self, device: torch.device) -> "_BatchEncoding":
+        return _BatchEncoding({key: value.to(device) for key, value in self.items()})
+
+
+class _Tokenizer:
+    eos_token_id = 0
+    padding_side = "right"
+
+    def __init__(self) -> None:
+        self.padding_sides: list[str] = []
+
+    def __call__(self, inputs: str | list[str], **kwargs: Any) -> _BatchEncoding:
+        self.padding_sides.append(self.padding_side)
+        if isinstance(inputs, str):
+            return _BatchEncoding(
+                {
+                    "input_ids": torch.tensor([[11, 12]]),
+                    "attention_mask": torch.tensor([[1, 1]]),
+                }
+            )
+        return _BatchEncoding(
+            {
+                "input_ids": torch.tensor([[0, 11], [21, 22]]),
+                "attention_mask": torch.tensor([[0, 1], [1, 1]]),
+            }
+        )
+
+    def decode(self, tokens: torch.Tensor, **kwargs: Any) -> str:
+        return " ".join(str(token) for token in tokens.tolist())
+
+
+class _Model(torch.nn.Module):
+    def __init__(self) -> None:
+        super().__init__()
+        self.generation_kwargs: dict[str, Any] = {}
+
+    def generate(self, input_ids: torch.Tensor, **kwargs: Any) -> torch.Tensor:
+        self.generation_kwargs = kwargs
+        return input_ids
+
+
+def _make_bridge() -> tuple[TransformerBridge, _Tokenizer, _Model]:
+    bridge = object.__new__(TransformerBridge)
+    torch.nn.Module.__init__(bridge)
+    bridge.cfg = SimpleNamespace(device=torch.device("cpu"))
+    tokenizer = _Tokenizer()
+    bridge.tokenizer = tokenizer
+    model = _Model()
+    bridge.original_model = model
+    bridge._ensure_generation_supported = MethodType(lambda self, api: None, bridge)
+    return bridge, tokenizer, model
+
+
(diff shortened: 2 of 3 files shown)


transformer_lens/model_bridge/bridge.py
```diff
@@ -4677,10 +4678,16 @@ def hf_generate(
             input_ids = inputs["input_ids"]
             input_type = "str"
         elif isinstance(input, list):
-            inputs = self.tokenizer(input, return_tensors="pt", padding=True, truncation=False).to(
-                self.cfg.device
-            )
+            original_padding_side = self.tokenizer.padding_side
+            self.tokenizer.padding_side = "left"
```
jlarson4: This forces left padding for every `list[str]` input, including encoder-decoder models. BART, mBART, Pegasus and Marian assign encoder positions by column index and ignore the attention mask, so left padding shifts every real token and the short rows still disagree with solo generation. `generate()` deliberately keeps the tokenizer's own side for encoder-decoder list input (`bridge.py:3838-3848`).

Let's adjust this to force left padding only when `self.original_model.config.is_encoder_decoder` is false, and forward the tokenizer's mask on both paths.

transformer_lens/model_bridge/bridge.py
```diff
@@ -4690,6 +4697,8 @@ def hf_generate(
 
         # Build generation_kwargs from explicit args and kwargs
         generation_kwargs = dict(generation_kwargs) if generation_kwargs is not None else {}
+        if input_attention_mask is not None:
+            generation_kwargs.setdefault("attention_mask", input_attention_mask)
```
jlarson4: `setdefault` lets a mask passed through `**generation_kwargs` win over the one the tokenizer just built. A `list[str]` caller cannot know the padded layout, and this change flips it from right to left, so a shape-correct mask built for the old layout silently masks real tokens and unmasks pads. For list input use the tokenizer's mask unconditionally, and drop `test_hf_generate_preserves_explicit_attention_mask`.

tests/integration/test_generation_compatibility.py
```diff
@@ -374,6 +374,28 @@ def test_hf_generate_batch_generation(self, gpt2_bridge):
         assert isinstance(result, list), "Batch input should return list"
         assert len(result) == len(prompts), "Output list should match input length"
 
+    def test_hf_generate_variable_length_batch_matches_solo_generation(self, gpt2_bridge):
+        """Test padded batch rows retain the same context as solo generation."""
+        prompts = ["Hi", "The capital of France is"]
```
jlarson4: With these two prompts the forwarded mask alone is enough to make batched match solo, so the test passes even without the left-padding change and guards only half the fix. A larger length gap does diverge. Lets widen the gap, try `"A"` against a ~16-token prompt.

tests/unit/model_bridge/test_hf_generate_batch_padding.py
```diff
@@ -0,0 +1,104 @@
...
+    bridge._ensure_generation_supported = MethodType(lambda self, api: None, bridge)
+    return bridge, tokenizer, model
+
+
+def test_hf_generate_list_uses_left_padding_and_restores_tokenizer() -> None:
+    bridge, tokenizer, _ = _make_bridge()
+
+    bridge.hf_generate(["short", "a longer prompt"], return_type="tokens")
+
+    assert tokenizer.padding_sides == ["left"]
```
jlarson4: `padding_sides == ["left"]` records that `tokenizer.padding_side` was reassigned. An implementation that passes `padding_side="left"` as a tokenizer call kwarg pads correctly, never touches the shared tokenizer, and fails this assertion. Let's have `_Tokenizer.__call__` pad according to the side it is given, and assert the `input_ids` reaching `generate` are left-aligned.

jlarson4 (commented):
Good catch on this bug @emerardd, thanks for documenting and putting together this fix. Just a couple small comments, altogether your solution is solid.

tests/integration/test_generation_compatibility.py
emerardd: Updated in 9cc6a5dc. The prompts are now 2 versus 16 GPT-2 tokens. I also checked the negative control: right padding with the correct mask gives 'A, a' while solo generation gives 'A new study', so the test now fails without the left-padding half of the fix.

transformer_lens/model_bridge/bridge.py
emerardd: Updated in 9cc6a5dc. The tokenizer-produced mask now replaces any attention_mask supplied through generation_kwargs for list[str] input. I removed the old preserves-explicit-mask contract and replaced it with coverage asserting that the tokenizer mask wins.

tests/unit/model_bridge/test_hf_generate_batch_padding.py
emerardd: Updated in 9cc6a5dc. The fake tokenizer now honors the per-call padding_side, and the fake model records the input_ids it receives. The test asserts the actual left-padded layout reaching generate and verifies that the tokenizer's default padding_side remains right.

transformer_lens/model_bridge/bridge.py
emerardd: Updated in 9cc6a5dc. hf_generate now checks original_model.config.is_encoder_decoder: encoder-decoder list inputs keep the tokenizer's native padding side, while decoder-only lists request left padding per tokenizer call. The attention mask produced by that same tokenization call is forwarded in both cases.

jlarson4:
Looks great, thank you for putting this together @emerardd

PR #1763, opened 2026-09-09
Require eval mode for JacobianLens fitting

### LINKED ISSUE #1762: [Bug] JacobianLens.fit silently mixes dropout realizations in train mode
### Describe the bug

`JacobianLens.fit()` can silently assemble Jacobian rows from different dropout realizations when the model, or one of its relevant submodules, is in training mode.

The method expands one prompt across the batch dimension and maps output dimension `i` to batch row `i`. With active dropout, those rows receive different masks. Later backward chunks can also use new realizations. The returned matrix is therefore not the Jacobian of any single model execution, even though `fit()` succeeds with the expected shape.

This also violates two documented expectations: fitting is described as deterministic for fixed prompts, and `dim_batch` is presented as a performance/memory tradeoff rather than a semantic parameter.

### Reproduction

Using the existing test `_ToyBridge`, an identity lens, and `Dropout(p=0.5)` in later blocks, the same weights, prompt, and seed produced:

```text
train_weights_equal=True
train_dim_batch_max_diff=9.480900764465332
train_dim_batch_rel_diff=1.5122860670089722
train_repeat_max_diff=4.875718593597412
eval_dim_batch_max_diff=0.0
```

In training mode, `dim_batch=1` and `dim_batch=6` differ by order-one amounts, and consecutive fits differ as well. In evaluation mode, the same comparison agrees exactly. This is not floating-point drift.

The trigger is realistic during fine-tuning, training-time lens monitoring, after `model.train()`, or when an individual stochastic submodule remains in training mode. Ordinary inference loads often start in evaluation mode, which makes the bug easy to miss.

### Expected behavior

`JacobianLens.fit()` should fail fast before fitting if the model or any of its submodules is in training mode, with an actionable message instructing the caller to run `model.eval()`.

The method should not silently switch modes internally: preserving mixed per-submodule training states across success and exception paths would add hidden state mutation to an analysis API.

### Additional context

The affected implementation is in `transformer_lens/tools/analysis/jacobian_lens.py`. `_frozen_parameters(model)` changes only `requires_grad`; it does not change training mode.

Required regression coverage should include:

- a root model in training mode;
- a root model in eval mode with one nested submodule still in training mode;
- unchanged eval-mode fitting behavior;
- an actionable `model.eval()` error message.

Validated against `upstream/dev` at `7edc5ddfccc24619109e3c98db29a870d99977ec` on 2026-09-09. Searches across open and closed Issues and PRs for `JacobianLens dropout train mode`, `JacobianLens dim_batch`, and `Jacobian lens eval mode` found no matching report.

### System Info

```text
OS: Windows
Python: 3.12
TransformerLens: upstream/dev at 7edc5ddfccc24619109e3c98db29a870d99977ec
```

### Checklist

- [x] I have provided a minimal reproduction or clear numerical evidence
- [x] I have included the expected behavior and a focused fix direction

emerardd:
# Description

Fixes #1762.

`JacobianLens.fit()` documents deterministic fitting, but an active training-mode module can apply different dropout realizations to the replicated batch rows used for different Jacobian output dimensions. That silently assembles rows from different stochastic executions, so the result is not the Jacobian of any single forward pass and can change substantially with `dim_batch`.

This change makes fitting fail fast unless the Bridge and every submodule are in evaluation mode. The check covers both the Bridge module tree and its deliberately unregistered `original_model`, reports an actionable `model.eval()` message, and does not mutate the caller's training state. The public docstring now records the requirement.

The regression tests cover a training root model, a training nested Bridge module under an eval root, and a training nested module inside the hidden original model. They also verify that every mixed training state remains unchanged after the error.

## Type of change

# Checklist:

- [x] New and existing unit tests pass locally with my changes (affected unit file: 104 passed; full repository suite was not rerun)

## Validation

- `uv run --no-cache --no-sync pytest tests/unit/tools/test_jacobian_lens.py -q`: 104 passed
- focused training-state regression surface after final assertions: 3 passed
- pinned pycln, isort, and Black on both changed Python files
- `uv run --no-cache --no-sync mypy .`: success across 396 source files
- `git diff --check`

### DIFF
--- a/tests/unit/tools/test_jacobian_lens.py
+++ b/tests/unit/tools/test_jacobian_lens.py
@@ -124,6 +124,7 @@ class _ToyBridge(TransformerBridge):
         self.blocks = nn.ModuleList(blocks)
         self.ln_final = nn.Identity()
         self.unembed = nn.Linear(D_MODEL, D_VOCAB, bias=False, dtype=dtype)
+        self.eval()
 
     @property
     def W_U(self) -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


jlarson4:
Looks good @emerardd, great work!

Issue #1755, opened 2026-09-08
[Proposal] Relevance Lens (R-lens): a RelP/ LRP-based transport-matrix estimator for Jacobian Lens (J-lens) fit, readout, and intervention


janmenjayap:
---

### Proposal

Add **Relevance Lens (R-lens)** as an alternative estimator for the shipped Jacobian Lens. R-lens keeps J-lens's fitting loop and downstream readout unchanged, but computes each per-layer `d_model × d_model` transport matrix with a **RelP/LRP-modified backward pass** instead of the ordinary Jacobian vector–Jacobian product (VJP). Only backward semantics change; the forward pass is bit-identical to the model's native forward.

#### Scope decision: Bridge-only

HookedTransformer is deprecated as of TransformerLens 3.0 (emits `DeprecationWarning`, `TransformerBridge` is the default surface, `transformers>=5.9.0` required, removal scheduled for the next major version). **R-lens targets `TransformerBridge` exclusively** and adds no HookedTransformer path.

#### Proposed API

Names are adjustable to maintainer preference.

```python
from transformer_lens.model_bridge import TransformerBridge
from transformer_lens.tools.analysis import RelevanceLens

model = TransformerBridge.boot_transformers(model_id, device="cuda")
lens = RelevanceLens.fit(model, prompts, corpus="pile-10k:fixed-manifest")

result = lens.readout(model, "The Eiffel Tower is in the city of")
hooks = lens.ablation_hooks(model, " Paris", layers=[6, 7])
```

`RelevanceLens` reuses J-lens's matrix-based `readout`, vocabulary vectors/dictionary, sparse decomposition, and steering/ablation/swap interventions unchanged. It differs from `JacobianLens` only in how the transport matrices are estimated and in the provenance recorded on the artifact.

---

### Motivation

The source post reports that ordinary gradients accumulate noise at early layers, and that a small set of LRP rules produce earlier, cleaner, and more causally load-bearing readouts. Those findings motivate the estimator; they are **not** proposed as small-model CI assertions (the post reports no advantage on its smallest models).

---

#### The three rules (first dense slice)

Applied exactly as in the source post; everything else stays on ordinary autograd.

1. **LN-rule (residual-stream norms only).** Preserve each residual LayerNorm/RMSNorm forward value while treating its denominator as a constant in the backward pass. Does **not** touch q/k norms or other attention-internal normalization.
[...]

PR #1764, opened 2026-09-10
refactor(jacobian_lens): extract estimator-independent fit driver

janmenjayap:
---

### Summary

Extracts the estimator-independent part of `JacobianLens.fit` / `_jacobian_for_prompt` into asingle shared driver, `_fit_transport_matrices`, that accepts a **swappable backward provider**.The ordinary J-lens path passes an ordinary-VJP wrapper (`_ordinary_vjp` over`torch.autograd.grad`), so the fitted transport matrices are **byte-identical** to before.

This is a pure refactor. It creates the seam a relevance-modified backward pass later plugs into,and ships **no** new public API and **no** relevance-lens surface.

Part of #1755 (R-lens). This is PR1 of the 3-PR plan in that issue; PR2 lands the shared LN/Identity/Half custom-VJP backend, PR3 adds `RelevanceLens.fit`.

---

### Motivation

R-lens reuses J-lens's entire fitting loop unchanged (frozen-parameter lifecycle, one-hot cotangent batching, valid-position selection, source-position averaging, prompt accumulation) and differs only in the backward step, which becomes a RelP/LRP-modified VJP instead of the ordinary Jacobian VJP. Isolating the estimator-independent mechanics behind a narrow backward-provider seam lets that later work plug in without touching or risking the ordinary J-lens numerics. Landing it first, on its own, keeps the risky refactor separate from the new estimator and gives a pinned regression baseline the rest of the plan must preserve.

---

1. **`test(jacobian_lens): pin ordinary-estimator fit regression fixture`** Captures a deterministic small-model `JacobianLens.fit` as a golden reference before the refactor. A tiny locally-built GPT-2 with RNG-independent arithmetic weights makes the fitted transport matrices reproducible across environments. Also asserts repeat-fit determinism and joint == merged consistency.

2. **`refactor(jacobian_lens): extract estimator-independent fit driver`** Adds a `BackwardProvider` seam and an `_ordinary_vjp` wrapper so the fit backward step goes through a swappable callable instead of a direct `torch.autograd.grad`. Routes `_jacobian_for_prompt` through the provider; the ordinary path passes `_ordinary_vjp` so the numerics are identical. Adds the `_fit_transport_matrices` driver owning the frozen-parameter lifecycle, prompt accumulation, and source-position averaging.

3. **`refactor(jacobian_lens): route JacobianLens.fit through shared driver`** Rewrites `fit` to delegate to `_fit_transport_matrices` with the ordinary provider and removes the duplicated drive-loop body (frozen-parameter lifecycle, per-prompt accumulation, empty-fit guard).

4. **`test(jacobian_lens): driver accepts alternate backward provider`** Drives `_fit_transport_matrices` with an alternate provider that scales the ordinary VJP by an exact power of two and asserts every transport matrix is byte-identical to `SCALE * baseline`. Confirms the ordinary provider reproduces the pinned golden matrices and that the public `JacobianLens.fit` path is untouched. Also asserts the model forward runs exactly once per prompt, the invariant PR2's per-forward fail-closed hook check depends on.

5. **`chore(jacobian_lens): format, mypy`** Annotates the shared fit kwargs dicts as `dict[str, Any]` so `**` unpacking type-checks under mypy; imports `typing.Any` in the integration test. Format and mypy clean on touched files.

---

### Testing

- `tests/integration/test_jacobian_lens.py` — 18 passed (includes the new pinned regression fixture and the backward-provider seam test).
- `tests/integration/test_backward_lens.py` — 32 passed (defensive net).
- `tests/unit/model_bridge/generalized_components/test_normalization_hook_semantics.py` — 11 passed (defensive net).
- `uv run mypy` on touched sources — clean; `black` / `isort` — clean.

The `backward_lens` and hook-semantics suites are a defensive regression net, not a functional gate this change touches no `NormalizationBridge` code, backward hooks, or the autograd forward dispatch, and running them makes the "pure refactor, no behavior change" claim verifiable rather than asserted.

---

### Out of scope (later PRs on #1755)

The LN/Identity/Half custom-VJP backend, native-forward preservation, the raw-`GatedMLPBridge` recompute path (PR2), and `RelevanceLens.fit` with provenance/merge/registry guards and docs (PR3). No relevance-rule code lands here.

---

### API

No public API added or changed.

---

### Checklist

- [x] Base set to `dev`; working branch `feat/r-lens-drive-loop`.
- [x] Ordinary-estimator regression fixture captured before the refactor and green after it.
- [x] `JacobianLens.fit` numerics unchanged; backward-provider seam exercised by a test with no
  R-lens code yet.
- [x] Driver invokes model forward exactly once per prompt.
- [x] `backward_lens` and hook-semantics suites green and untouched.
- [x] `mypy` and formatting clean on touched files.
- [ ] Full `make test-pr` run (unit + docstring + acceptance + integration) attached before submit.
- [ ] Integration test run against a cached GPT-2 with `.env` sourced.

---

### DIFF
--- a/tests/integration/test_jacobian_lens.py
+++ b/tests/integration/test_jacobian_lens.py
@@ -615,3 +615,125 @@ def test_decompose_gemma_activation_is_valid():
     assert torch.allclose(
         result.j_space_component + result.non_j_space_component, activation, atol=1e-2
     )
+
+
+# --------------------------------------------------------------------------- #
+# Ordinary-estimator fit regression fixture                                   #
+# --------------------------------------------------------------------------- #
+#
+# This pins the exact numerics of the current ``JacobianLens.fit`` drive loop
+# before the estimator-independent driver is extracted. The
+# refactor must reproduce these golden transport matrices, so a change to
+# cotangent batching, valid-position selection, source-position averaging,
+# prompt accumulation, or the graph root would surface here as a failure.
+#
+# The fixture avoids brittleness two ways: it builds a tiny GPT-2 locally (no
+# model download) and overwrites every parameter with an RNG-version-independent
+# arithmetic ramp, so the fitted matrices are byte-identical across machines and
+# torch versions. In-run determinism is asserted with ``torch.equal`` (a repeat
+# fit) and the golden comparison uses a tight tolerance that only absorbs
+# cross-BLAS float rounding on the 4-wide matmuls.
+
+REGRESSION_CORPUS = "jlens-drive-loop-regression-v1"
+REGRESSION_PROMPTS = [
+    "the quick brown fox jumps over the lazy dog again",
+    "colorless green ideas sleep furiously beneath the ancient stone bridge",
+]
+REGRESSION_FIT_KWARGS = dict(
+    source_layers=[0, 1],
+    dim_batch=3,  # < d_model=4 so the ragged final chunk is exercised
+    max_seq_len=16,
+    skip_first_positions=1,
+    show_progress=False,
+)
+# Golden transport matrices captured from the pre-refactor JacobianLens.fit.
+REGRESSION_GOLDEN_JACOBIANS = {
+    0: [
+        [1.0194597244262695, -0.10382787883281708, 0.01114815566688776, 0.07321998476982117],
+        [0.017510414123535156, 0.7570222020149231, 0.17482545971870422, 0.05064191669225693],
+        [-0.06917503476142883, -0.09777483344078064, 1.198944091796875, -0.03199436515569687],
+        [0.05239897966384888, 0.06332147866487503, -0.1471048891544342, 1.0313844680786133],
+    ],
+    1: [
+        [1.022949457168579, -0.05948571860790253, -0.08073973655700684, 0.11727607250213623],
+        [-0.011972094886004925, 0.8772072792053223, 0.12830956280231476, 0.006455251481384039],
+        [-0.04260174185037613, 0.0968940407037735, 1.0271172523498535, -0.08140958845615387],
+        [0.01806488260626793, -0.12579867243766785, -0.007413430605083704, 1.1151472330093384],
+    ],
+}
+
+
+def _build_tiny_deterministic_gpt2():
+    """A 4-wide, 3-layer GPT-2 bridge with fixed arithmetic weights and a real tokenizer.
+
+    Weights are RNG-independent so the fit is reproducible across environments; the
+    GPT-2 tokenizer (small, cached) lets ``JacobianLens.fit`` take string prompts.
+    """
+    from transformers import AutoTokenizer, GPT2Config, GPT2LMHeadModel
+
+    from transformer_lens.model_bridge.sources import build_bridge_from_module
+
+    tokenizer = AutoTokenizer.from_pretrained("gpt2")
+    config = GPT2Config(
+        vocab_size=tokenizer.vocab_size,
+        n_positions=64,
+        n_embd=4,
+        n_layer=3,
+        n_head=2,
+        resid_pdrop=0.0,
+        embd_pdrop=0.0,
+        attn_pdrop=0.0,
+    )
+    hf_model = GPT2LMHeadModel(config).eval()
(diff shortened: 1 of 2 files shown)


transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -1890,15 +1909,96 @@ def capture_fn(
...
+    iterator = tqdm(prompts, desc="fitting J-lens", disable=not show_progress)
+    with _frozen_parameters(model):
+        for prompt in iterator:
+            tokens = model.to_tokens(prompt)[:, :max_seq_len]
+            seq_len = tokens.shape[1]
+            if seq_len <= skip_first_positions + 1:
+                warnings.warn(
+                    f"skipping prompt with only {seq_len} tokens "
+                    f"(need > {skip_first_positions + 1})",
+                    stacklevel=2,
```
jlarson4: This warning is a frame deeper than it was in `fit`, so `stacklevel=2` now resolves to `jacobian_lens.py` rather than the user's `fit` call. The `fit_dtype` warning at line 1487 did not move, so a single `fit` can emit two warnings pointing at different places. Let's update this to `stacklevel=3`

tests/integration/test_jacobian_lens.py
```diff
@@ -615,3 +617,196 @@ def test_decompose_gemma_activation_is_valid():
...
+    def scaled_provider(target, sources, cotangent, retain_graph):
+        nonlocal calls
+        calls += 1
+        grads = _ordinary_vjp(target, sources, cotangent, retain_graph)
+        return tuple(SCALE * grad for grad in grads)
+
+    scaled, n_scaled = _fit_transport_matrices(
+        model, REGRESSION_PROMPTS, backward_provider=scaled_provider, **REGRESSION_FIT_KWARGS
+    )
+    assert calls > 0  # the driver actually took its backward step through the seam
```
jlarson4, line 820: `calls` counts backward-provider invocations, not forwards. A second `model(replicated, return_type=None)` inside `_jacobian_for_prompt` passes every test in this file. Can we count bridge forwards across the `_fit_transport_matrices` call and assert one per prompt?

jlarson4 (commented):
Great work on this @janmenjayap, just a couple small consistency comments

transformer_lens/tools/analysis/jacobian_lens.py
janmenjayap: Thanks for the careful read, and you are right. Now that `fit` delegates to `_fit_transport_matrices`, the `warnings.warn` site sits one frame deeper, so `stacklevel=2` pointed at `jacobian_lens.py` instead of the caller's `fit(...)` line. I have bumped it to `stacklevel=3` so the warning attributes back to the user's call site.

To keep this from drifting again if the driver ever picks up another wrapper frame, I also added a regression test (`test_fit_attributes_short_prompt_warning_to_caller_frame`). It runs `fit` in a bare subprocess (this suite's jaxtyping/beartype wrapping would otherwise add its own frames and mask the difference) and asserts the caught warning's `filename` resolves to the caller. Addressed in commit 5abd16c7

tests/integration/test_jacobian_lens.py
janmenjayap: Good catch. You are right that `calls` only tracks `backward_provider` invocations, which grow with `dim_batch` chunking rather than with forwards, so `assert calls > 0` could not tell one forward per prompt from several.

I added an independent counter: a forward pre-hook on the bridge that increments on each top-level `model(...)` call across the `_fit_transport_matrices` run, then `assert forward_calls == len(REGRESSION_PROMPTS)`, that is, exactly one bridge forward per contributing prompt (no prompt is skipped in this case, so every prompt contributes). I kept the existing `assert calls > 0` alongside it as a weaker seam check rather than replacing it. Addressed in commit 9d6bea7d.

janmenjayap:
Appreciate the review and the consistency catches. Both are addressed on the branch, format and mypy are clean, and the two affected integration tests pass.

jlarson4:
Thanks for your quick response on these changes requests @janmenjayap! Everything is covered, assuming this passes CI I will get it merged shortly to ensure it is included in 3.9 and 4.0

PR #1766, opened 2026-09-10
Fix JacobianLens dictionary cache invalidation

### LINKED ISSUE #1765: [Bug] JacobianLens reuses a stale full-vocabulary dictionary after unembedding weights change
## Bug description

`JacobianLens.lens_vector_dictionary()` caches its full-vocabulary result only by `(layer, device)`, even though the result also depends on the supplied model's current `W_U`.

After `W_U` changes, the method silently returns the old dictionary. This affects `decompose()`, `coordinate_patch()`, and `occupancy()`, and can assign a residual direction to the wrong vocabulary token while still reporting zero reconstruction error.

## Reproduction

On current `dev` (`90ac6f0b`), using the existing network-free `_ToyBridge` and an identity Jacobian lens:

```text
baseline_max_diff=0.000000000
cache_object_reused=True
direct_changed_ratio=2.000000000
cached_changed_max=0.000000000
stale_dictionary_max_diff=0.405797064
after_clear_max_diff=0.000000000
```

The sequence was:

1. Populate `lens_vector_dictionary()`.
2. Mutate the same model's unembedding weight in place by multiplying it by 2.
3. Compare the cached dictionary with `lens_vectors()` over the whole vocabulary.

`lens_vectors()` immediately reflects the new weights, while `lens_vector_dictionary()` returns the previous object unchanged. Calling `clear_device_cache()` repairs the result.

A row-swap variant is more concerning: after swapping unembedding rows 0 and 1, decomposition of the current token-0 direction can select the old token label with zero reconstruction error.

## Root cause

The cache currently uses:

```python
cached = self._dictionary_cache.get((layer, device))
```

The key contains no model, parameter/storage identity, or tensor version. `validate_model()` checks compatible metadata and dimensions, but it cannot detect weight mutation. This includes supported workflows such as temporary parameter swaps, optimizer steps, `copy_()`, whole-parameter replacement, and reuse with another compatible model instance.

## Suggested fix

Associate each cached dictionary with a fingerprint of the `W_U` used to build it. The fingerprint should distinguish model/storage identity and include the tensor version, shape, dtype, and device. Model identity alone is insufficient because an optimizer step mutates the same parameter object.

The implementation should retain cache hits for an unchanged model while rebuilding after:

- in-place `W_U.copy_()` or optimizer updates;
- whole-parameter replacement;
- a temporary parameter swap, both inside and after the context;
- use with a second compatible model instance.

## Duplicate check

Searched open and closed Issues and PRs for `JacobianLens dictionary cache W_U`, `lens_vector_dictionary cache`, and related stale-cache/weight-mutation terms on 2026-09-10; no match found.

emerardd:
# Description

`JacobianLens.lens_vector_dictionary()` previously cached a full-vocabulary dictionary only by layer and device. Reusing the lens after an optimizer step, an in-place or temporary unembedding edit, a parameter replacement, or with another compatible model could therefore return a stale dictionary and silently attach residual directions to the wrong token labels.

This change retains one detached `W_U` snapshot per device and compares it with the current unembedding using on-device `torch.equal`. Unchanged weights retain the existing object-level dictionary cache hit; a content change invalidates every cached layer on that device. This covers `.data` writes that bypass PyTorch's version counter and parameter replacements after allocator address reuse, without copying weights to the host or hashing them on every cache lookup.

The regression coverage verifies cache hits for unchanged weights and invalidation after `copy_()`, an optimizer step, `.data` mutation without a version increment, whole-parameter replacement after the old parameter has been released, temporary parameter swapping in both directions, and reuse with a second compatible model. It also verifies that one snapshot is shared across layers, a change invalidates every layer on the affected device, and an unembedding row swap cannot make decomposition return the old token label with zero residual.

Fixes #1765

## Type of change

### Screenshots

Not applicable.

# Validation

- `uv run --no-cache --no-sync pytest tests/unit/tools/test_jacobian_lens.py tests/unit/tools/test_jacobian_lens_coordinate_patch_hooks.py -q` — 122 passed
- `uv run --no-cache --no-sync mypy .` — success across 397 source files
- pycln, isort, Black, and `git diff --check` — passed for the affected surface

The complete `make test-pr` surface was not run locally.

# Checklist:

- [ ] New and existing unit tests pass locally with my changes — only the affected unit-test surface was run

### DIFF
--- a/transformer_lens/tools/analysis/jacobian_lens.py
+++ b/transformer_lens/tools/analysis/jacobian_lens.py
@@ -64,6 +64,7 @@ import warnings
 from dataclasses import dataclass
 from importlib.metadata import version
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from weakref import ref
 
 import torch
 from jaxtyping import Float, Int
@@ -213,6 +214,29 @@ class JacobianLensReadout:
         return out
 
 
+@dataclass(frozen=True)
+class _UnembeddingFingerprint:
+    model_ref: Any
+    data_ptr: int
+    version: int
+    shape: Tuple[int, ...]
+    stride: Tuple[int, ...]
+    dtype: torch.dtype
+    device: torch.device
+
+
+def _unembedding_fingerprint(model: Any, unembed: torch.Tensor) -> _UnembeddingFingerprint:
+    return _UnembeddingFingerprint(
+        model_ref=ref(model),
+        data_ptr=unembed.data_ptr(),
+        version=unembed._version,
+        shape=tuple(unembed.shape),
+        stride=tuple(unembed.stride()),
+        dtype=unembed.dtype,
+        device=unembed.device,
+    )
+
+
 class JacobianLens:
     """A fitted Jacobian lens: one transport matrix per source layer.
 
@@ -255,7 +279,9 @@ class JacobianLens:
         self.d_model = int(d_model)
         self.metadata: Dict[str, Any] = dict(metadata or {})
         self._device_jacobians: Dict[Tuple[int, torch.device], torch.Tensor] = {}
-        self._dictionary_cache: Dict[Tuple[int, torch.device], torch.Tensor] = {}
+        self._dictionary_cache: Dict[
+            Tuple[int, torch.device], Tuple[_UnembeddingFingerprint, torch.Tensor]
+        ] = {}
 
     @property
     def source_layers(self) -> List[int]:
@@ -832,9 +858,9 @@ class JacobianLens:
         """Full-vocabulary J-lens dictionary at ``layer``: ``[d_vocab, d_model]``.
 
         Row ``t`` is the J-lens vector ``v_t = J[layer]^T W_U[:, t]`` -- this is
-        :meth:`lens_vectors` over the entire vocabulary. The result is cached per
-        (layer, device) so a sparse decomposition can reuse it; :meth:`clear_device_cache`
-        releases it.
+        :meth:`lens_vectors` over the entire vocabulary. The result is cached while the
+        model's unembedding is unchanged so a sparse decomposition can reuse it;
+        :meth:`clear_device_cache` releases it.
 
         The dictionary is vocabulary-sized and cached on the model's device
         (``d_vocab * d_model`` fp32 values, on the order of gigabytes for a large
@@ -849,13 +875,16 @@ class JacobianLens:
         """
         self.validate_model(model)
         layer = _normalize_layer(layer, model.cfg.n_layers)
-        device = torch.device(model.W_U.device)
-        cached = self._dictionary_cache.get((layer, device))
-        if cached is None:
+        unembed = model.W_U
+        device = torch.device(unembed.device)
+        fingerprint = _unembedding_fingerprint(model, unembed)
+        entry = self._dictionary_cache.get((layer, device))
+        if entry is None or entry[0] != fingerprint:
             matrix = self._matrix_on(layer, device)  # [d_model, d_model]
-            cached = (matrix.T @ model.W_U.float()).T  # [d_vocab, d_model]
-            self._dictionary_cache[(layer, device)] = cached
-        return cached
+            dictionary = (matrix.T @ unembed.float()).T  # [d_vocab, d_model]
+            entry = (fingerprint, dictionary)
+            self._dictionary_cache[(layer, device)] = entry
+        return entry[1]
 
     @torch.no_grad()
     def decompose(
--- a/transformer_lens/tools/analysis/jacobian_lens.py
+++ b/transformer_lens/tools/analysis/jacobian_lens.py
@@ -60,7 +60,6 @@ Example::
     print(result.top_tokens(model.tokenizer, k=5)[8][-1])  # layer 8, final position
 """
 
-import hashlib
 import warnings
 from dataclasses import dataclass
 from importlib.metadata import version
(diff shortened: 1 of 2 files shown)


transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -213,6 +214,29 @@ def top_tokens(self, tokenizer: Any, k: int = 5) -> Dict[int, List[List[str]]]:
...
+    stride: Tuple[int, ...]
+    dtype: torch.dtype
+    device: torch.device
+
+
+def _unembedding_fingerprint(model: Any, unembed: torch.Tensor) -> _UnembeddingFingerprint:
+    return _UnembeddingFingerprint(
+        model_ref=ref(model),
+        data_ptr=unembed.data_ptr(),
+        version=unembed._version,
```
jlarson4: `.data` deliberately does not share a tensor's version counter, and a `.data` write also leaves `data_ptr`, `shape`, `stride`, `dtype`, `device` and the model object identical. Every fingerprint field matches and the stale dictionary is served. #1765's headline symptom survives here: swap unembedding rows with `weight.data[[0, 1]] = weight.data[[1, 0]]` and `decompose` labels the current token-0 direction as token 1 with a zero residual. `weight.data.copy_()` and `weight.data.fill_()` are an established pattern in this repo — `test_linear_bridge.py` uses both.

Please base the fingerprint on the unembedding's contents rather than on its identity metadata; a cheap digest costs far less than the `d_model² × d_vocab` matmul it guards.

transformer_lens/tools/analysis/jacobian_lens.py
```diff
@@ -213,6 +214,29 @@ def top_tokens(self, tokenizer: Any, k: int = 5) -> Dict[int, List[List[str]]]:
...
+    shape: Tuple[int, ...]
+    stride: Tuple[int, ...]
+    dtype: torch.dtype
+    device: torch.device
+
+
+def _unembedding_fingerprint(model: Any, unembed: torch.Tensor) -> _UnembeddingFingerprint:
+    return _UnembeddingFingerprint(
+        model_ref=ref(model),
+        data_ptr=unembed.data_ptr(),
```
jlarson4: `data_ptr()` is an address, not an identity. Release the unembedding parameter and allocate its replacement, and the allocator hands back the same address; a replacement built by `torch.empty(...).fill_()` or loaded from a checkpoint carries the same small `_version` as the parameter it replaced. Both fields collide, the fingerprint matches, and `decompose`, `coordinate_patch` and `occupancy` all keep serving the pre-replacement dictionary. Whole-parameter replacement is one of the scenarios this change is meant to cover; `test_lens_vector_dictionary_invalidates_after_parameter_replacement` misses it because `2 * weight.detach()` is allocated while the old parameter is still alive, so the two can never share an address.

Can we add a replacement case that releases the old parameter before allocating its replacement and make the fingerprint survive it?

jlarson4 (commented):
@emerardd Great work on this, thanks for bringing this issue to our attention. Just a couple small comments on the `_unembedding_fingerprint` function

transformer_lens/tools/analysis/jacobian_lens.py
emerardd: Fixed in c84e4a72. The replacement regression now removes the old registered parameter, runs gc.collect(), asserts that a weak reference is dead, and only then allocates the replacement. The production fingerprint no longer uses data_ptr, so allocator address reuse cannot produce a stale cache hit; if the logical contents match, the cached dictionary remains valid.

transformer_lens/tools/analysis/jacobian_lens.py
emerardd: Fixed in c84e4a72. The fingerprint now hashes the logical W_U contents with BLAKE2b (plus shape, dtype, and device), so .data writes no longer depend on _version. I added the row-swap regression and assert the version remains unchanged; decomposition selects the current token label with zero residual.

jlarson4: A small follow up on this: 

Would it be possible to detect the change on-device instead of hashing host bytes? Something like `torch.equal` against a retained copy of `W_U`.

As it's built here, `.contiguous()` always materializes a full copy, `.cpu()` adds a device->host transfer and a sync on a GPU model, and then `blake2b` hashes the whole buffer single-threaded, against a matmul that is multi-threaded BLAS. At GPT-2 scale the guard costs roughly eight times the rebuild it protects, so the cache is now slower than having no cache at all, and it is still a net loss at Llama-3-8B's `d_model`.

emerardd: Addressed in 5d76b4c6. I replaced the host-side BLAKE2b fingerprint with one detached W_U snapshot per device and compare it with the current unembedding using on-device torch.equal. When the contents change, all dictionary entries on that device are invalidated and the single shared snapshot is refreshed, avoiding both the device-to-host transfer and per-layer snapshot duplication. clear_device_cache() now releases the snapshots as well. I added coverage that verifies the snapshot is shared across layers and that a mutation invalidates every cached layer on the affected device.

koriyoshi2041 (approved):
Validated the latest on-device snapshot implementation at `5d76b4c6`: the two JacobianLens test modules pass 122/122 locally, including `.data` mutation, parameter replacement, cross-model reuse, and cross-layer invalidation; diff hygiene is clean. The shared per-device snapshot also removes the earlier host-transfer/hash cost while preserving content-based invalidation.

transformer_lens/tools/analysis/jacobian_lens.py
jlarson4: Awesome, thank you @emerardd, looks great, merging now

PR #1768, opened 2026-09-10
feat(svd_circuits): per-head QK/OV SVD with degeneracy guard

### LINKED ISSUE #1767: [Proposal] SVD Circuits: singular-vector decomposition of a head's QK/ OV into causally-validated subfunctions
---

### Proposal

Component-level circuit analysis (`direct_logit_attribution`, path patching, head detectors) treats an attention head as an indivisible unit. But a single canonical head — e.g. a GPT-2-small IOI name-mover — empirically superposes **multiple subfunctions on distinct low-rank directions** of its QK and OV maps. This proposes `svd_circuits`, a self-contained tool in `transformer_lens/tools/analysis/` that takes the **SVD of a head's QK (`W_Q W_Kᵀ`) and OV (`W_V W_O`) matrices**, exposes each **orthogonal singular direction** as an interpretable sub-component, and — crucially — **causally gates** every claimed subfunction by patching activations along that individual singular direction. This is a finer-than-component decomposition that resolves within-head superposition component-level tools miss.

Capabilities:

1. **Per-head QK/OV SVD** via `FactoredMatrix` — get singular values/vectors of `W_Q W_Kᵀ` and `W_V W_O` without materializing the `d_model × d_model` matrix.
2. **Direction → vocab readout** — project each OV singular direction into the unembedding basis (reusing `SVDInterpreter`) and each singular direction into logit space via `direct_logit_attribution`, so a direction gets a human-readable "what tokens does this subfunction move" signature.
3. **Activation projection / attribution** — project cached per-head activations onto the singular directions to measure how much each subfunction fires on a given prompt (per token/position).
4. **Causal validation (patch-along-a-direction)** — reconstruct the head output using only a chosen singular subspace (or ablate a single direction) via `generic_activation_patch`, and report the behavior change (e.g. IOI logit-diff), so a subfunction claim is accepted only if its direction is causally load-bearing.
5. **Degeneracy guard** — detect near-equal singular values (rotation/degeneracy ambiguity) and refuse to attribute individual directions inside a degenerate block, reporting the block as a subspace instead.

**Status:**PR1 (weight-space core: `decompose_head`, `HeadSVD`, degeneracy guard) implemented and verified on branch `feat/svd-circuits-core` — unit tests, `mypy`, and `make unit-test` all green (#1768 ). PR2/PR3 not started.
**Suggested labels:** `enhancement`, `tooling`, `TransformerBridge`, `complexity: moderate`

---

### Motivation

The existing TL baseline for "what does this head do" is `direct_logit_attribution` (`transformer_lens/tools/analysis/direct_logit_attribution.py`) and `head_detector` — both operate at head granularity. The IOI work established name-mover / S-inhibition / duplicate-token heads as the atomic units of the circuit. The claim in *Beyond Components* is that this granularity is too coarse: an IOI name-mover head is not one function but several, packed onto near-orthogonal low-rank directions of its QK/OV weights, and only an SVD-based decomposition surfaces them. This tool lets a TL user go **inside** a canonical head and attribute/patch its subfunctions — the natural next rung below component-level circuit analysis.

**Why no maintained TL implementation exists.** TL already ships the two hardest primitives — `FactoredMatrix` (low-rank OV/QK SVD/eigendecomposition without materializing the full matrix) and `SVDInterpreter` (projects OV/`w_in`/`w_out` singular vectors into vocab) — but neither closes the loop from "singular direction" to "causally-validated subfunction." `SVDInterpreter` is a static weight-inspection utility (its own docstring warns SVD directions are not reliably interpretable); it never touches activations or patching. The missing piece is the activation-projection + causal-patch layer, which is exactly what makes the decomposition trustworthy.

**Implementation status elsewhere.** The authors' research repo exists — [`github.com/Exploration-Lab/Beyond-Components`](https://github.com/Exploration-Lab/Beyond-Components) (~7 commits; IOI / GP / GT examples) — but it is not maintained or TL-integrated. This would be the first **TL-native** implementation. This is **not** SAE/dictionary-learning territory (no learned features, no overcomplete basis) — it is a closed-form weight decomposition, so it sits inside TL rather than at the SAELens boundary.

---

### Pitch

Add `svd_circuits` as a self-contained analysis tool under `transformer_lens/tools/analysis/`. It builds on `FactoredMatrix` for the weight-space SVD, `SVDInterpreter` for vocab readout, `ActivationCache`/`direct_logit_attribution` for activation projection and logit signatures, and `generic_activation_patch` for the mandatory causal gate.

#### Proposed API

Names adjustable to maintainer preference.

```python
from transformer_lens.tools.analysis import svd_circuits

# TransformerBridge is the supported TL 3.x path; HookedTransformer is deprecated (removed next major).
model = TransformerBridge.boot_transformers("gpt2-small")
model.enable_compatibility_mode()  # folds final LN into W_U — required for the OV vocab/logit readout

# 1. Decompose a head's QK and OV into singular directions (FactoredMatrix under the hood).
decomp = svd_circuits.decompose_head(model, layer=9, head=9, which=("QK", "OV"))
decomp.OV.S            # singular values, sorted desc
decomp.OV.rank_report  # (idx, sigma, sigma/sigma_max, is_degenerate) per direction

# 2. Readout: each OV singular direction -> top tokens in the unembedding basis.
decomp.OV.vocab_readout(k=10)     # wraps SVDInterpreter.get_singular_vectors
decomp.OV.logit_signature(model, prompt=ioi_prompt)  # via direct_logit_attribution

# 3. Attribution: how much does each direction fire on this prompt?
proj = svd_circuits.project_activations(model, decomp, prompt=ioi_prompt)  # [pos, direction]

# 4. Causal gate: keep only directions {0,3}, measure IOI logit-diff change.
result = svd_circuits.patch_along_directions(
    model, decomp.OV, keep=[0, 3], prompt=ioi_prompt,
    metric=ioi_logit_diff,      # any callable(logits) -> scalar
)
result.delta_metric   # behavior change attributable to that subspace
result.gated          # True only if |delta| exceeds the causal threshold
```

#### Design (algorithm)

For a head `(layer ℓ, head h)`:

1. **Build factored maps.** `OV = FactoredMatrix(W_V[h], W_O[h])` (shape `d_model × d_model`, rank ≤ `d_head`); `QK = FactoredMatrix(W_Q[h], W_K[h].T)`. Both stay factored — never materialize `d_model²`.
2. **SVD.** `U, S, V = M.svd()` (`M == U @ S.diag() @ V.transpose(-2, -1)`; `U` and `V` are both `[d_model, rank]`, with the *i*-th singular direction in **column** `i` — `U[:, i]`, `V[:, i]`). Right singular vectors `V[:, i]` are the **input directions** (which residual-stream directions the subfunction reads); left singular vectors `U[:, i]` are the **output directions**. Access them via the `.V` property — `.Vh` is a **deprecated alias that returns the same tensor as `.V`** (it emits a `DeprecationWarning`), not its Hermitian transpose; do not introduce a new call site that relies on it.
3. **Degeneracy guard.** Flag any run of singular values with `σᵢ/σᵢ₊₁ − 1 < ε` (default `ε = 1e-2`) as a degenerate block: individual directions inside it are rotation-ambiguous, so attribute the block as a subspace, not per-direction.
4. **Vocab readout (OV).** For each output direction `U[:, i]`, project through the (LN-folded) unembedding — reuse `SVDInterpreter.get_singular_vectors("OV", ℓ, head_index=h)` — to get its top-token signature.
5. **Activation projection.** Cache the head's per-position input (residual stream into the head; for OV attribution use `stack_head_results` / the head's value stream). Project onto `V[:, i]` to get a `[pos]` firing coefficient per direction `i`.
6. **Logit signature.** Feed the rank-1 reconstruction `σᵢ · U[:, i] V[:, i]ᵀ` of the head output through `direct_logit_attribution` to get each direction's signed logit effect on the task metric.
7. **Causal patch-along-direction.** Using `generic_activation_patch`, replace the head's output (at `blocks.ℓ.attn.hook_z` / `hook_result`) with its projection onto the chosen singular subspace `span(U[:, keep])` (or zero out a single direction for ablation), rerun, and record `Δmetric`. A subfunction is **accepted only if** its direction/subspace produces a causal `Δmetric` above threshold — the decomposition alone never suffices.

Reuse map:

| Sub-step | TL primitive | `file:symbol` |
|---|---|---|
| Factored QK/OV, low-rank SVD/eigen | `FactoredMatrix` (`.svd()`, `.U/.S/.V`, `.eigenvalues`) | `transformer_lens/FactoredMatrix.py:FactoredMatrix` |
| OV direction → vocab readout | `SVDInterpreter` | `transformer_lens/SVDInterpreter.py:SVDInterpreter.get_singular_vectors` |
| Per-head activations for projection | `ActivationCache` | `transformer_lens/ActivationCache.py:stack_head_results` (:957) |
| Direction → logit effect | `direct_logit_attribution` | `transformer_lens/tools/analysis/direct_logit_attribution.py:direct_logit_attribution` |
| Bridge LN-folding guard (reuse pattern) | `_validate_bridge_compatibility` | `transformer_lens/tools/analysis/direct_logit_attribution.py:_validate_bridge_compatibility` |
| Characterize resulting subfunction | head detector | `transformer_lens/head_detector.py` |
| Causal patch-along-direction | `generic_activation_patch` + setter | `transformer_lens/patching.py:generic_activation_patch` |
| Weights / unembed / final LN | `W_Q/W_K/W_V/W_O`, `model.W_U` | Bridge `tl_parameters()` / legacy `transformer_lens/HookedTransformer.py` |

#### Main correctness risk (named): interpreting individual singular directions under rotation/degeneracy ambiguity

SVD directions are unique only when singular values are distinct; equal (or near-equal) singular values leave the corresponding singular subspace defined only up to an arbitrary rotation, so any "direction 3 = the surname subfunction" claim inside a degenerate block is meaningless. Compounding this, "the SVD direction is interpretable" is **not** guaranteed even for well-separated directions (`SVDInterpreter`'s own docstring flags numerical instability and cross-device inconsistency). **Mitigation, not a warning:** (a) the degeneracy guard (step 3) *raises/refuses* per-direction attribution inside a near-degenerate block and downgrades it to a subspace; (b) **every** claimed subfunction must pass the causal patch-along-direction gate (step 7) — a direction that is not causally load-bearing is never reported as a subfunction, regardless of how clean its vocab readout looks. The tool's contract is "causal evidence gates interpretive claims," not "SVD is interpretable."

1. **Analytic SVD oracle.** On a tiny synthetic head with hand-set `W_V/W_O` (and `W_Q/W_K`), the tool's `S`, `U`, `V` match `torch.linalg.svd` of the explicitly materialized matrix to `atol=1e-5` (sign/degeneracy handled).
2. **Reconstruction fidelity.** Summing all rank-1 direction reconstructions reproduces the full head output (via the cache) to `atol=1e-4`; keeping top-`k` directions monotonically increases recovered `Δmetric`.
3. **Degeneracy guard fires.** A synthetic head with two exactly-equal singular values → the guard flags the block and per-direction attribution raises; subspace attribution still works.
4. **Causal gate discriminates.** Ablating a high-`σ`, causally-relevant direction changes the IOI logit-diff materially; ablating a null-space / random direction does not (baseline reported alongside).
5. **Paper sanity check (slow).** On `gpt2-small` head L9H9 (name-mover), the tool surfaces ≥2 causally-gated OV subfunctions whose vocab readouts qualitatively match the paper's name-mover split; reported as a qualitative sanity check, not a pinned numeric threshold.

#### Scope of the vertical slice

Per-head QK/OV SVD + direction→vocab readout + direction→logit signature + **one** causal patch-along-direction validation, on `gpt2-small` IOI. Defer full multi-head circuit assembly and automated subfunction labeling. Ships as **3 sequential PRs**, not one — base off `dev`:

1. `feat(svd_circuits): per-head QK/OV SVD with degeneracy guard` — `decompose_head`, `HeadSVD`, degeneracy guard. Pure weight-space, no public export yet.
2. `feat(svd_circuits): singular-direction readout, projection, and causal patch gate` — vocab/logit readout, `project_activations`, `patch_along_directions`, public exports, Bridge compatibility-mode validation, `gpt2-small` IOI integration test.
3. `docs(svd_circuits): add demo notebook, slow name-mover parity, and tool docs` — slow oracle-parity test, `SVD_Circuits_Demo.ipynb`, docs section.

Files:

- New: `transformer_lens/tools/analysis/svd_circuits.py`
- Export: `transformer_lens/tools/analysis/__init__.py`
- Unit: `tests/unit/tools/test_svd_circuits.py` (synthetic head; checks 1–4, no HF download)
- Integration: `tests/integration/test_svd_circuits.py` (`gpt2-small`, IOI patch-along-direction end-to-end)
- Oracle-parity (slow): `tests/integration/test_svd_circuits_oracle_parity.py` (check 5, `@pytest.mark.slow`)
- Demo: `demos/SVD_Circuits_Demo.ipynb` (nbval)

#### Follow-up work

Deferred to a separate tiered issue, filed after PR3 merges: multi-head circuit assembly, automated subfunction labeling, QK-side pattern attribution, more models.

#### Model coverage

- **CI reference / oracle:** tiny synthetic head (no download) for unit correctness; `gpt2-small` for the integration + slow paper sanity check (CI-cacheable, canonical IOI).
- **Demo model:** `gpt2-small`.
- **Honesty caveat:** the paper's headline subfunction taxonomy is scale- and model-dependent; `gpt2-small` reproduces the *mechanism* (a name-mover splits into causally-distinct directions) but not necessarily the exact subfunction inventory of larger models. Baselines (random-direction ablation) are reported next to every headline `Δmetric`.

---

### Alternatives

- **Stay at head granularity (`direct_logit_attribution` / head detectors).** Rejected: cannot resolve within-head superposition — the whole point.
- **Use `SVDInterpreter` alone.** Rejected: static weight readout with no activation projection and no causal gate; its own docstring disclaims interpretability of the directions.
- **Sparse autoencoders on head outputs (SAELens).** Rejected here: needs training, an overcomplete learned basis, and lives outside TL; `svd_circuits` is closed-form, training-free, and reuses TL primitives.
- **Full path-patching over a rank-1 direction grid.** Rejected as the first slice: combinatorially expensive and premature before the single-direction gate is validated; folded into Tier 2 of the follow-up issue.

---

### Correctness oracle

**Brute-force / analytic reference** (no external frozen impl exists). Primary oracle: `torch.linalg.svd` on an explicitly materialized synthetic head, compared to the `FactoredMatrix`-based path within threshold (`atol=1e-5`), plus a reconstruction-fidelity check. Secondary (slow, `@pytest.mark.slow`): the `gpt2-small` name-mover paper sanity check, treated as qualitative because the paper ships no numeric reference to pin against.

---

### Additional context

- **Paper:** Areeb Ahmad, Abhinav Joshi, Ashutosh Modi (IIT Kanpur), "Beyond Components: Singular Vector-Based Interpretability of Transformer Circuits," arXiv **2511.20273**, NeurIPS 2025 poster (poster 119702, OpenReview `7UbXEQNny7`).
- **External implementation status:** authors' research repo [`Exploration-Lab/Beyond-Components`](https://github.com/Exploration-Lab/Beyond-Components) exists (research-grade, ~7 commits, 4 stars, no `transformer_lens` dependency); this would be the first **TL-native** implementation. Paper nuance: it augments weight matrices with biases folded in and applies SVD uniformly to attention and MLP — this slice covers attention QK/OV (OV logit-facing) only.
- **Artifact sources:** none required — the decomposition is computed from model weights at call time; no downloaded lens/SAE artifacts, so no registry entry is needed for the vertical slice.
- **Algorithm note for reviewers:** OV singular directions are attributed through the unembedding (output side); QK singular directions describe which residual directions the head compares (attention-pattern side) and are attributed via the pattern, not the unembedding — the vertical slice focuses on OV for logit-facing claims and includes QK SVD as a readout only. The causal gate is what separates this from a pretty-picture SVD tool; keep it mandatory.
- **Substrate history:** SVD convention fix #341 (closed) → #1300 (merged) (`torch.linalg.svd`; `.Vh` deprecated in favour of `.V`); #328 (closed, 2023, "SVD tests fail on GPU") is direct evidence for the cross-device instability the degeneracy guard + causal gate backstop.
- **Provisional-citation flags:** arXiv id `2511.20273` and NeurIPS-2025 venue are confirmed (no open citation flags).

---

### Checklist

- [x] Checked no similar issue / tool exists (`SVDInterpreter` is static weight readout only; no causal singular-direction tool in `tools/analysis/`).
- [x] Scope limited to a vertical slice (single head, OV logit-facing, one causal gate); follow-up tiered separately.
- [x] Correctness oracle identified (analytic `torch.linalg.svd` + reconstruction; slow qualitative paper check).
- [x] Targets `TransformerBridge` (the supported TL 3.x path) via shared `ActivationCache`; also runs on the deprecated `HookedTransformer` compat layer.
- [ ] Claim-in-comments before starting a tier item.

---

janmenjayap:
---

### Description

Adds `svd_circuits.decompose_head`, the first of three PRs implementing per-head singular-vector decomposition of a head's QK (`W_Q W_K^T`) and OV (`W_V W_O`) maps (tracking issue: #1767 ).

This PR is **weight-space only**: it reads `W_Q`/`W_K`/`W_V`/`W_O` and needs no forward pass, no activation cache, and no compatibility mode.

- `decompose_head(model, layer, head, which=("QK", "OV"))` — factored SVD via `FactoredMatrix`, never materializing the `d_model x d_model` product.
- `HeadSVD` / `HeadDecomposition` dataclasses with a per-direction `rank_report`.
- Degeneracy guard: near-equal singular values are grouped into a block; `require_isolated` raises `DegenerateDirectionError` for per-direction attribution inside a block instead of silently returning a rotation-dependent direction.
- Reads right singular vectors from `FactoredMatrix.V`; the deprecated `.Vh` alias is never touched (enforced by a test that turns its `DeprecationWarning` into an error).

**By design, this PR does not export a public API.** `svd_circuits` is intentionally absent from `transformer_lens/tools/analysis/__init__.py` `__all__` — the causal patch-along-direction gate (PR2) is what turns a
readout into a validated claim, so the public surface should never ship without it attached.

Unit-tested model-free: synthetic weight tensors only, no `from_pretrained`, no HF download. SVD parity against `torch.linalg.svd` within `atol=1e-5`; reconstruction within `atol=1e-4`.

Follow-ups (not in this PR): OV vocab/logit readout, `project_activations`, the mandatory `patch_along_directions` causal gate, and public exports (PR2); demo notebook, slow paper-sanity check, and docs (PR3).

Part of #1767 

---

### Type of change

---

### Checklist:

- [ ] I have made corresponding changes to the documentation (Will be done in PR3)

---

### DIFF
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -5,6 +5,7 @@ degeneracy guard directly, so no model is loaded and no pretrained weights are d
 """
 
 import warnings
+from types import SimpleNamespace
 
 import pytest
 import torch
@@ -15,9 +16,45 @@ from transformer_lens.tools.analysis.svd_circuits import (
     RankReportRow,
     _degeneracy_blocks,
     _factored_head_svd,
+    decompose_head,
 )
 
 D_MODEL, D_HEAD = 12, 4
+
+
+class _StubAttn:
+    def __init__(self, W_Q, W_K, W_V, W_O):
+        self.W_Q, self.W_K, self.W_V, self.W_O = W_Q, W_K, W_V, W_O
+
+
+class _StubBlock:
+    def __init__(self, attn):
+        self.attn = attn
+
+
+class _StubModel:
+    """Model-free stand-in exposing cfg.n_layers/n_heads and blocks[i].attn.W_*.
+
+    W_K/W_V carry ``n_kv_heads`` rows (the grouped-query layout); W_Q/W_O carry
+    ``n_heads`` rows, matching the per-block shapes ``decompose_head`` reads.
+    """
+
+    def __init__(
+        self,
+        n_heads,
+        n_kv_heads,
+        n_layers=1,
+        d_model=D_MODEL,
+        d_head=D_HEAD,
+        seed=0,
+    ):
+        g = torch.Generator().manual_seed(seed)
+        W_Q = torch.randn(n_heads, d_model, d_head, generator=g)
+        W_K = torch.randn(n_kv_heads, d_model, d_head, generator=g)
+        W_V = torch.randn(n_kv_heads, d_model, d_head, generator=g)
+        W_O = torch.randn(n_heads, d_head, d_model, generator=g)
+        self.cfg = SimpleNamespace(n_layers=n_layers, n_heads=n_heads)
+        self.blocks = [_StubBlock(_StubAttn(W_Q, W_K, W_V, W_O)) for _ in range(n_layers)]
 
 
 # --------------------------------------------------------------------------- #
@@ -211,3 +248,79 @@ def test_degeneracy_blocks_groups_equal_run_directly():
 def test_degeneracy_blocks_on_well_separated_are_all_singletons():
     blocks = _degeneracy_blocks(torch.tensor([8.0, 4.0, 2.0, 1.0]), eps=1e-2)
     assert blocks == [[0], [1], [2], [3]]
+
+
+# --------------------------------------------------------------------------- #
+# decompose_head wiring and input guards (model-free, via _StubModel)
+# --------------------------------------------------------------------------- #
+def test_decompose_head_mha_reads_requested_head():
+    """MHA stub (n_heads == n_kv_heads): each head's OV decomposes its own W_V/W_O."""
+    model = _StubModel(n_heads=4, n_kv_heads=4)
+    attn = model.blocks[0].attn
+    for h in range(4):
+        result = decompose_head(model, layer=0, head=h, which=("OV",))
+        S_ref = torch.linalg.svd(attn.W_V[h] @ attn.W_O[h]).S
+        assert torch.allclose(result.OV.S, S_ref[:D_HEAD], atol=1e-4)
+
+
+def test_decompose_head_gqa_maps_query_to_kv_head():
+    """GQA stub (4 query heads, 2 kv heads): every head decomposes without IndexError,
+    and heads 0,1 recover kv head 0's W_K/W_V while heads 2,3 recover kv head 1's."""
+    model = _StubModel(n_heads=4, n_kv_heads=2)
+    attn = model.blocks[0].attn
+    n_heads = model.cfg.n_heads
+    n_kv_heads = attn.W_K.shape[0]
+    for h in range(n_heads):
+        result = decompose_head(model, layer=0, head=h, which=("QK", "OV"))
+        kv_head = h // (n_heads // n_kv_heads)
+        ov_ref = torch.linalg.svd(attn.W_V[kv_head] @ attn.W_O[h]).S
+        qk_ref = torch.linalg.svd(attn.W_Q[h] @ attn.W_K[kv_head].transpose(-1, -2)).S
+        assert torch.allclose(result.OV.S, ov_ref[:D_HEAD], atol=1e-4)
+        assert torch.allclose(result.QK.S, qk_ref[:D_HEAD], atol=1e-4)
+
+
+def test_decompose_head_qk_transpose_wiring():
+    """QK factors W_Q_h and W_K_h.T; res.QK.S matches svd(W_Q_h @ W_K_h.T)."""
+    model = _StubModel(n_heads=4, n_kv_heads=4)
+    attn = model.blocks[0].attn
+    result = decompose_head(model, layer=0, head=2, which=("QK",))
+    S_ref = torch.linalg.svd(attn.W_Q[2] @ attn.W_K[2].transpose(-1, -2)).S
+    assert torch.allclose(result.QK.S, S_ref[:D_HEAD], atol=1e-4)
+
+
+def test_decompose_head_which_filter():
+    """which=("OV",) returns only OV; which=("QK",) returns only QK."""
+    model = _StubModel(n_heads=2, n_kv_heads=2)
+    ov_only = decompose_head(model, layer=0, head=0, which=("OV",))
+    assert ov_only.OV is not None and ov_only.QK is None
+    qk_only = decompose_head(model, layer=0, head=0, which=("QK",))
+    assert qk_only.QK is not None and qk_only.OV is None
+
+
+def test_decompose_head_rejects_bad_layer():
+    model = _StubModel(n_heads=2, n_kv_heads=2, n_layers=1)
+    with pytest.raises(ValueError):
+        decompose_head(model, layer=1, head=0)
+
+
+def test_decompose_head_rejects_bad_head():
+    model = _StubModel(n_heads=2, n_kv_heads=2)
+    with pytest.raises(ValueError):
+        decompose_head(model, layer=0, head=2)
+
+
+def test_decompose_head_rejects_empty_which():
+    model = _StubModel(n_heads=2, n_kv_heads=2)
+    with pytest.raises(ValueError):
+        decompose_head(model, layer=0, head=0, which=())
+
+
+def test_decompose_head_rejects_unknown_which():
+    """An unknown which entry is rejected; the jaxtyping/beartype import hook enforced
+    by this suite's pytest config raises its own violation before the manual ValueError
+    guard runs, so the check is on the exception class name rather than ValueError."""
+    model = _StubModel(n_heads=2, n_kv_heads=2)
+    with pytest.raises(Exception) as exc_info:
+        decompose_head(model, layer=0, head=0, which=("QK", "XX"))
+    exc_name = type(exc_info.value).__name__
+    assert "TypeCheckError" in exc_name or "Beartype" in exc_name
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -214,11 +214,39 @@ def test_near_but_not_equal_respects_eps():
 
 
 def test_null_run_grouped():
-    """A run of near-zero singular values forms one degenerate null block."""
+    """A spectrum the relative-gap rule cannot group (8e-3 is 40x above 2e-4) is grouped by
+    an explicit null_rtol wide enough to catch both as numerically null."""
+    res = _factored_head_svd(
+        *_factored_with_spectrum([1.0, 8e-3, 2e-4]),
+        which="OV",
+        layer=0,
+        head=0,
+        eps=1e-2,
+        null_rtol=1e-2,
+    )
+    assert res.degenerate_blocks() == [[1, 2]]
+    with pytest.raises(DegenerateDirectionError):
+        res.require_isolated(1)
+    with pytest.raises(DegenerateDirectionError):
+        res.require_isolated(2)
+
+
+def test_null_default_does_not_overgroup():
+    """With the default null_rtol, 8e-3 is not treated as null, so it is not over-grouped
+    with the tail direction the way the eps=1e-2 near-equal threshold used to."""
+    res = _factored_head_svd(
+        *_factored_with_spectrum([1.0, 8e-3, 2e-4]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    assert not res.is_degenerate(1)
+    assert not res.is_degenerate(2)
+
+
+def test_null_run_groups_true_null_tail():
+    """The default null_rtol still catches a genuinely near-zero tail as a null block."""
     res = _factored_head_svd(
         *_factored_with_spectrum([5.0, 1.0, 1e-9, 1e-9]), which="OV", layer=0, head=0, eps=1e-2
     )
-    assert res.is_degenerate(2) and res.is_degenerate(3)
+    assert res.degenerate_blocks() == [[2, 3]]
     with pytest.raises(DegenerateDirectionError):
         res.require_isolated(2)
     with pytest.raises(DegenerateDirectionError):
@@ -265,12 +293,12 @@ def test_degeneracy_blocks_partition_all_indices():
 # --------------------------------------------------------------------------- #
 def test_degeneracy_blocks_groups_equal_run_directly():
     """_degeneracy_blocks groups a repeated value and isolates the well-separated neighbours."""
-    blocks = _degeneracy_blocks(torch.tensor([5.0, 3.0, 3.0, 1.0]), eps=1e-2)
+    blocks = _degeneracy_blocks(torch.tensor([5.0, 3.0, 3.0, 1.0]), eps=1e-2, null_rtol=1e-2)
     assert blocks == [[0], [1, 2], [3]]
 
 
 def test_degeneracy_blocks_on_well_separated_are_all_singletons():
-    blocks = _degeneracy_blocks(torch.tensor([8.0, 4.0, 2.0, 1.0]), eps=1e-2)
+    blocks = _degeneracy_blocks(torch.tensor([8.0, 4.0, 2.0, 1.0]), eps=1e-2, null_rtol=1e-2)
     assert blocks == [[0], [1], [2], [3]]
 
 
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -36,7 +36,8 @@ class _StubModel:
     """Model-free stand-in exposing cfg.n_layers/n_heads and blocks[i].attn.W_*.
 
     W_K/W_V carry ``n_kv_heads`` rows (the grouped-query layout); W_Q/W_O carry
-    ``n_heads`` rows, matching the per-block shapes ``decompose_head`` reads.
+    ``n_heads`` rows, matching the per-block shapes ``decompose_head`` reads. Each
+    layer draws its own weights so a read from the wrong layer is visible.
     """
 
     def __init__(
@@ -47,14 +48,27 @@ class _StubModel:
         d_model=D_MODEL,
         d_head=D_HEAD,
         seed=0,
+        dtype=torch.float32,
+        requires_grad=False,
     ):
         g = torch.Generator().manual_seed(seed)
-        W_Q = torch.randn(n_heads, d_model, d_head, generator=g)
-        W_K = torch.randn(n_kv_heads, d_model, d_head, generator=g)
-        W_V = torch.randn(n_kv_heads, d_model, d_head, generator=g)
-        W_O = torch.randn(n_heads, d_head, d_model, generator=g)
         self.cfg = SimpleNamespace(n_layers=n_layers, n_heads=n_heads)
-        self.blocks = [_StubBlock(_StubAttn(W_Q, W_K, W_V, W_O)) for _ in range(n_layers)]
+        self.blocks = []
+        for _ in range(n_layers):
+            shapes = [
+                (n_heads, d_model, d_head),
+                (n_kv_heads, d_model, d_head),
+                (n_kv_heads, d_model, d_head),
+                (n_heads, d_head, d_model),
+            ]
+            weights = [torch.randn(*shape, generator=g).to(dtype) for shape in shapes]
+            if requires_grad:
+                weights = [torch.nn.Parameter(w) for w in weights]
+            self.blocks.append(_StubBlock(_StubAttn(*weights)))
+
+
+def _reconstruct(res):
+    return res.U @ torch.diag(res.S) @ res.V.transpose(-2, -1)
 
 
 # --------------------------------------------------------------------------- #
@@ -100,8 +114,7 @@ def test_reconstruction_matches_materialized_matrix():
     """U @ diag(S) @ V.T rebuilds W_V @ W_O, sidestepping the U/V column-sign ambiguity."""
     W_V, W_O = _random_ov()
     res = _factored_head_svd(W_V, W_O, which="OV", layer=0, head=0, eps=1e-2)
-    reconstruction = res.U @ torch.diag(res.S) @ res.V.transpose(-2, -1)
-    assert torch.allclose(reconstruction, W_V @ W_O, atol=1e-4)
+    assert torch.allclose(_reconstruct(res), W_V @ W_O, atol=1e-4)
 
 
 def test_singular_values_sorted_descending():
@@ -191,6 +204,23 @@ def test_degenerate_error_message_points_to_block():
     assert "subspace" in message.lower()
 
 
+def test_refusal_message_names_the_rule_that_formed_the_block():
+    """A gap-formed block is refused for its eps gap; a null block is refused as null."""
+    gap = _factored_head_svd(
+        *_factored_with_spectrum([5.0, 3.0, 3.0, 1.0]), which="QK", layer=2, head=3, eps=1e-2
+    )
+    with pytest.raises(
+        DegenerateDirectionError, match=r"QK SVD of head L2H3.*eps=0\.01"
+    ) as gap_exc:
+        gap.require_isolated(2)
+    assert "null" not in str(gap_exc.value)
+    null = _factored_head_svd(
+        *_factored_with_spectrum([5.0, 2e-9, 1e-9]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(DegenerateDirectionError, match="numerically null"):
+        null.require_isolated(1)
+
+
 def test_well_separated_spectrum_flags_nothing():
     """A well-separated spectrum (8, 4, 2, 1) has no degenerate block and no flagged row."""
     res = _factored_head_svd(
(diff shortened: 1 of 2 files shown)


transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+    Float[torch.Tensor, "d_head d_model"],
+]:
+    """Return ``(W_Q_h, W_K_h, W_V_h, W_O_h)`` for one head as float tensors.
+
+    The upcast to ``float`` promotes fp16/bf16 weights before the SVD, where
+    reduced precision is a known source of instability.
+    """
+    return (
+        model.W_Q[layer, head].float(),
+        model.W_K[layer, head].float(),
```
jlarson4: On `TransformerBridge`, `W_K` and `W_V` have shape `[n_layers, n_kv_heads, d_model, d_head]` – the grouped key-value layout, not the query layout. `head` is validated against `cfg.n_heads`, so on any grouped-query model `decompose_head` raises `IndexError` for every query head at or past `n_kv_heads`, and returns the wrong key-value head for the rest. On `Qwen/Qwen2.5-0.5B` (14 query heads, 2 key-value heads), 12 of 14 heads raise and 1 of the remaining 2 is silently wrong. 

Can we map the query head to its key-value head before indexing — `bridge._expand_kv_heads` (`transformer_lens/model_bridge/bridge.py:1650`) is what the bridge's own `QK`/`OV` accessors use for this.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+    Float[torch.Tensor, "d_model d_head"],
+    Float[torch.Tensor, "d_head d_model"],
+]:
+    """Return ``(W_Q_h, W_K_h, W_V_h, W_O_h)`` for one head as float tensors.
+
+    The upcast to ``float`` promotes fp16/bf16 weights before the SVD, where
+    reduced precision is a known source of instability.
+    """
+    return (
+        model.W_Q[layer, head].float(),
```
jlarson4: `model.W_Q`/`W_K`/`W_V`/`W_O` are uncached properties that `torch.stack` across every layer, so these four lines build four full-model tensors and then slice a single head out of each. At gpt2-small that is 113 MB allocated to reach 0.79 MB of weights, 48x the `d_model x d_model` product the module's docstring is built around avoiding. At 80 layers and `d_model` 8192 it is about 43 GB.

Lets read the one layer's weights instead: `model.blocks[layer].attn.W_Q` and friends, as `SVDInterpreter._get_OV_matrix` and `projection_kernel._read_role_matrix` both do.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+        head=head,
+        U=U,
+        S=S,
+        V=V,
+        rank_report=rank_report,
+        eps=eps,
+    )
+
+
+def decompose_head(
```
jlarson4, line 340: Nothing in `tests/unit/tools/test_svd_circuits.py` calls `decompose_head`. Can we add tests that call `decompose_head` against a stub object exposing `cfg` and the four weight properties? That should cover the QK transpose wiring, the `which` filter, and the four `ValueError` guards with no model load.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+        return []
+    values = [float(x) for x in S.tolist()]
+    top = max(values[0], _SIGMA_FLOOR)
+    blocks: List[List[int]] = []
+    current = [0]
+    for i in range(1, n):
+        prev = max(values[i - 1], _SIGMA_FLOOR)
+        relative_gap = 1.0 - (values[i] / prev)
+        null_run = (values[i] / top) < eps and (values[i - 1] / top) < eps
+        if relative_gap < eps or null_run:
```
jlarson4: A direction joins the current block on its gap to the previous direction alone, and that relation chains, so nothing bounds how far a block's last member sits from its first. A spectrum decaying 0.9% per step puts all 64 directions in one block spanning 1.77x, and `require_isolated(0)` then refuses the top direction as "near-equal singular values" while handing back the entire map as the subspace to attribute instead. On gpt2-small L4H7 one OV block covers 29 of 64 directions across a 13.3% spread. Lets require each member to be within `eps` of the block's largest value as well, so a reported block matches the near-equality the error message claims.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+    if n == 0:
+        return []
+    values = [float(x) for x in S.tolist()]
+    top = max(values[0], _SIGMA_FLOOR)
+    blocks: List[List[int]] = []
+    current = [0]
+    for i in range(1, n):
+        prev = max(values[i - 1], _SIGMA_FLOOR)
+        relative_gap = 1.0 - (values[i] / prev)
+        null_run = (values[i] / top) < eps and (values[i - 1] / top) < eps
```
jlarson4: `null_run` groups two neighbors whenever both fall below `eps x sigma_max`, using the relative-gap threshold as a ratio-to-top cutoff. With the 1e-2 default, `[1, 8e-3, 2e-4]` becomes one block: two values 40x apart reported as near-equal and refused for per-direction attribution. Can we give the null-space cutoff its own parameter rather than reusing `eps`?

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -0,0 +1,343 @@
...
+rotation, so the result carries a per-direction degeneracy report. Callers can
+use it to attribute an ambiguous block as a subspace instead of trusting a
+single, rotation-dependent direction.
+
+Example::
+
+    from transformer_lens import HookedTransformer
+    from transformer_lens.tools.analysis.svd_circuits import decompose_head
+
+    model = HookedTransformer.from_pretrained("gpt2", device="cpu")
```
jlarson4: Please rewrite this example using `TransformerBridge`. `HookedTransformer` is on the verge of being officially deprecated as part of 4.x releasing.

If there was an additional handling needed to support HookedTransformer, please strip it out, and remove any references to `HookedTransformer`.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -0,0 +1,213 @@
...
+        *_factored_with_spectrum(spectrum), which="OV", layer=0, head=0, eps=1e-2
+    )
+    assert grouped.is_degenerate(0) and grouped.is_degenerate(1)
+    separated = _factored_head_svd(
+        *_factored_with_spectrum(spectrum), which="OV", layer=0, head=0, eps=1e-3
+    )
+    assert not separated.is_degenerate(0) and not separated.is_degenerate(1)
+
+
+def test_null_run_grouped():
```
jlarson4, line 246: The trailing `1e-9, 1e-9` pair has a relative gap of exactly zero, so it groups through the ordinary `relative_gap < eps` rule. Deleting the `null_run` clause from `_degeneracy_blocks` leaves this test passing. Separately, `degenerate_blocks()` can be made to return `[]` unconditionally with the whole file still green: `test_well_separated_spectrum_flags_nothing` only pins the empty case, and the partition assertion at line 191 holds for any duplicate-free output.

Lets use a spectrum only `null_run` can group, such as `[1, 8e-3, 2e-4]`, and assert the actual grouping somewhere — `[5, 3, 3, 1]` should give `degenerate_blocks() == [[1, 2]]`.

jlarson4 (commented):
Hi @janmenjayap! Thanks for building out this tool! it is very cool. Just a couple notes on areas where we might be able to improve our efficiency or close a couple testing gaps.

Additionally, I noticed this includes references to `HookedTransformer`. That class will be official removed soon, any new features do not require supporting HookedTransformer going forward.

------------

Sorry for the delay on this, I wrote up all my comments this morning and forgot to post them. Let me know if you have any questions.

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: You are right, and thank you for the concrete Qwen numbers. `_head_weights` was indexing `W_K`/`W_V` with the query head, but on grouped-query attention those tensors carry one row per key-value head, so `decompose_head` raised on any query head at or past `n_kv_heads` and returned the wrong key-value head for the rest.

Fixed in commit `8018eefe` (`fix(svd_circuits): read one layer's weights and map query heads to key-value heads`): the query head is now mapped to its key-value head with `head // (n_heads // n_kv_heads)` before indexing, mirroring `bridge._expand_kv_heads`. This is a no-op for multi-head attention, where the head counts already match. Added a regression test that exercises every query head under a grouped-query stub (`n_heads=4`, `n_kv_heads=2`) and asserts the recovered `W_V`/`W_K` singular values match the intended key-value head.

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: Agreed, thank you for pointing at the actual bytes. The same commit `8018eefe` now reads `model.blocks[layer].attn.W_Q` and friends directly instead of the full-model `W_Q`/`W_K`/`W_V`/`W_O` stacks, so only one layer is materialized, matching the pattern `SVDInterpreter._get_OV_matrix` and `projection_kernel._read_role_matrix` already use.

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: Added, also in commit `8018eefe`. There is a lightweight stub model exposing `cfg` and `blocks[layer].attn.{W_Q,W_K,W_V,W_O}`, with no `from_pretrained` and no download. Tests cover multi-head head selection, the grouped-query mapping above, the QK transpose wiring, the `which` filter, and all four `ValueError` guards (bad layer, bad head, empty `which`, unknown `which`).

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: Agreed, `[1, 8e-3, 2e-4]` should not be grouped as near-equal. Fixed in commit `9aadfbb6` (`fix(svd_circuits): base the null-space cutoff on effective rank`): `decompose_head` now takes a dedicated `null_rtol` parameter, threaded through `_factored_head_svd` into `_degeneracy_blocks`, so the null-run grouping no longer reuses `eps`. The default resolves to `d_model * torch.finfo(S.dtype).eps`, matching `torch.linalg.matrix_rank`'s default relative tolerance, so it lines up with the effective-rank framing in the source paper (Ahmad, Joshi, Modi, arXiv:2511.20273) rather than a hand-tuned epsilon. The resolved value is stored on `HeadSVD.null_rtol` for reproducibility.

tests/unit/tools/test_svd_circuits.py
janmenjayap: Both are addressed:

- The `null_run` deletion gap: `test_null_run_grouped` (commit `9aadfbb6`) now uses `[1, 8e-3, 2e-4]`, a spectrum only `null_run` can group, called with an explicit `null_rtol`, so deleting the `null_run` clause now fails it. Also added `test_null_default_does_not_overgroup` and `test_null_run_groups_true_null_tail` to pin the default cutoff's behavior in both directions.
- The `degenerate_blocks() == []` gap: commit `bbe1d3cb` adds a test pinning the actual grouping for `[5, 3, 3, 1]` to `[[1, 2]]`, rather than only asserting non-degeneracy.

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: Fixed in commit `da84acf1` (`docs(svd_circuits): rewrite the example on TransformerBridge and drop HookedTransformer`): the module docstring example now boots a `TransformerBridge` via `TransformerBridge.boot_transformers`, the "works with both HookedTransformer and TransformerBridge" sentence is gone, and the remaining `HookedTransformer` mention in the `Args:` block is replaced with a description of the per-block bridge accessors `decompose_head` reads. There was no `HookedTransformer`-only code path to strip, so this commit is documentation only. `grep -rn HookedTransformer transformer_lens/tools/analysis/svd_circuits.py` now returns nothing.

transformer_lens/tools/analysis/svd_circuits.py
janmenjayap: Agreed, and thank you for the gpt2-small L4H7 example, that made the failure mode very clear. Fixed in commit `bbe1d3cb` (`fix(svd_circuits): bound degenerate blocks by distance to the block anchor`): a direction now joins the open block only when it is within `eps` of both the previous direction and the block's anchor (its first, largest member), so a reported block cannot spread further than the near-equality the error message claims. Added a regression test with a spectrum decaying about 0.9 percent per step (every pairwise gap under `eps`), confirming it no longer collapses into one block and the top direction stays isolated under `require_isolated(0)`.

jlarson4:
@janmenjayap After reviewing your changes, I realized I provided some incorrect feedback in my latest review. 

I requested you cap how wide a degenerate block can get, but that was bad advice. Under the capped rule, a long run of near-equal singular values gets cut at arbitrary points. So a direction can be reported as isolated while its neighbor is within `eps`, which is ambiguous by the module's own definition.

I went ahead and cleaned up that error and pushed the fix to your branch. Assuming this passes CI, I am going to merge it so that we can guarantee inclusion in 3.9.0. Thanks you for putting this together, and apologies for my mistake.

PR #1773, opened 2026-09-12
Fix IOIDataset BOS handling

### LINKED ISSUE #1772: [Bug Report] IOIDataset double-prepends BOS and ignores prepend_bos=False
**Describe the bug**

`IOIDataset.__getitem__()` calls `tokenizer.encode()` with its default special-token behavior and then manually prepends `bos_token_id` when `prepend_bos=True`.

TransformerLens tokenizers configured with `add_bos_token=True` therefore produce two leading BOS tokens on the default path. With `prepend_bos=False`, `encode()` still adds one BOS, so the flag does not disable BOS insertion.

The same unqualified `encode()` calls are used for the `IO` and `S` labels. Their shared automatic BOS shifts the first differing label-token index, and together with the prompt's second BOS causes the default `ioi_eval()` path to read logits one position later than intended.

**Code example**

This network-free reproduction uses a tokenizer stub with the same `add_special_tokens` contract:

```python
from unittest.mock import MagicMock

from transformer_lens.evals import IOIDataset

tokenizer = MagicMock()
tokenizer.bos_token_id = 0

def encode(text, add_special_tokens=True):
    payload = [11, 12, 21] if len(text) > 7 else [21 if "Alice" in text else 22]
    return ([0] if add_special_tokens else []) + payload

tokenizer.encode.side_effect = encode

for prepend_bos in (True, False):
    dataset = IOIDataset(
        tokenizer,
        templates=["[A] gave [B] to [A]"],
        names=["Alice", "Bob"],
        nouns={},
        num_samples=1,
        prepend_bos=prepend_bos,
        seed=0,
    )
    item = dataset[0]
    print(prepend_bos, item["prompt"].tolist(), item["IO"].tolist(), item["S"].tolist())
```

On current `dev`:

```text
True  [0, 0, 11, 12, 21] [0, 22] [0, 21]
False [0, 11, 12, 21]    [0, 22] [0, 21]
```

Expected:

- `prepend_bos=True`: exactly one leading BOS in `prompt`.
- `prepend_bos=False`: no leading BOS in `prompt`.
- `IO` and `S`: no automatically inserted special tokens, so their first differing content token is used directly.

A focused fix would encode all three fields with `add_special_tokens=False`, then let `prepend_bos` be the only code path that adds a prompt BOS.

**System Info**

- Base: `dev` at `e2581b5f3f3f62ae1fa8c770cb95adb187843fa4`
- OS: Windows NT 10.0.26200.0
- Python: 3.12.10

**Additional context**

The existing unit tokenizer stub always returns a fixed list and does not model automatic special-token insertion, so it cannot detect this contract violation. Regression coverage should exercise both values of `prepend_bos`, assert that label encodings omit special tokens, and verify the logit position used by `ioi_eval()`.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/TransformerLensOrg/TransformerLens/issues) in the repo

emerardd:
# Description

`IOIDataset` previously called `tokenizer.encode()` with automatic special-token insertion enabled and then manually prepended a BOS. TransformerLens tokenizers configured to add BOS therefore produced two leading BOS tokens by default, while `prepend_bos=False` still left one BOS in place. Automatic BOS tokens in the IO/S label encodings also shifted the first differing-token index used by `ioi_eval()`, causing the default evaluation path to read logits one position too late.

This change encodes prompts and labels with `add_special_tokens=False`, leaving the existing `prepend_bos` branch as the sole source of a prompt BOS. The parameter documentation now states that contract.

Regression tests model a tokenizer that automatically inserts BOS, cover both values of `prepend_bos`, verify that IO/S contain only content tokens, and use position-sensitive logits to confirm that `ioi_eval()` reads the prediction immediately before the first answer token.

Fixes #1772

## Type of change

### Screenshots

Not applicable.

# Validation

- `uv run --no-cache --no-sync pytest tests/unit/test_evals_ioi.py -q` — 7 passed
- `uv run --no-cache --no-sync mypy .` — success across 398 source files
- pycln, isort, Black, and `git diff --check` — passed for the affected surface

The complete unit-test suite was not run locally.

# Checklist:

- [ ] New and existing unit tests pass locally with my changes — only the affected unit-test file was run

### DIFF
--- a/tests/unit/test_evals_ioi.py
+++ b/tests/unit/test_evals_ioi.py
@@ -1,18 +1,34 @@
-"""Tests for IOIDataset in transformer_lens/evals.py.
-
-Regression test for https://github.com/TransformerLensOrg/TransformerLens/issues/515
-"""
+"""Tests for IOIDataset and ioi_eval in transformer_lens/evals.py."""
 
 from unittest.mock import MagicMock
 
-from transformer_lens.evals import IOIDataset
+import torch
+
+from transformer_lens.evals import IOIDataset, ioi_eval
 
 
 def _make_tokenizer():
     """Minimal mock tokenizer sufficient for IOIDataset."""
     tok = MagicMock()
-    tok.encode.side_effect = lambda text: [1, 2, 3]
+    tok.encode.side_effect = lambda text, add_special_tokens=True: [1, 2, 3]
     tok.bos_token_id = 0
+    return tok
+
+
+def _make_automatic_bos_tokenizer():
+    """Tokenizer stub whose default encode path inserts a BOS token."""
+    tok = MagicMock()
+    tok.bos_token_id = 0
+
+    def encode(text, add_special_tokens=True):
+        if text.startswith(" "):
+            payload = [30 if text.strip() == "Alice" else 31]
+        else:
+            target = 30 if text.split()[-1] == "Alice" else 31
+            payload = [10, 20, target]
+        return ([tok.bos_token_id] if add_special_tokens else []) + payload
+
+    tok.encode.side_effect = encode
     return tok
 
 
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (approved):
The fix matches the tokenizer contract: prompt and label encodings now exclude automatic special tokens, so  is the only prompt BOS path and  compares the first answer token. I reran the focused unit file at this exact head (7 passed) and ; both passed.

koriyoshi2041:
Correction to my approval wording: `prepend_bos` is the only prompt BOS path, and `ioi_eval` compares the first answer token. The focused unit file passed 7/7 at `54a3727d`; `git diff --check` also passed.

jlarson4:
Looks good and covers the reported issues, thanks @emerardd

PR #1774, opened 2026-09-12
feat(sparse_probing): add leakage-safe k-sparse probing over activation tensors

### LINKED ISSUE #1728: [Proposal] Add leakage-safe k-sparse probing over activation tensors
### Proposal 

Add a dependency-free, model-free analysis module,
`transformer_lens/tools/analysis/sparse_probing.py`, for fitting binary k-sparse probes to a
supplied activation matrix `X` and label vector `y`.

The first PR contains only:

1. train/test splitting before all learned statistics;
2. train-only mean-difference feature selection;
3. deterministic binary logistic fitting on the selected coordinates;
4. held-out accuracy, precision, recall, and F1;
5. k-sweeps using one fixed split;
6. shuffled-label and random-coordinate controls.

Activation extraction, model wrappers, plotting, notebooks, multiclass probing, and optimal MIP
selection are follow-ups.

---

### Motivation

A central mechanistic-interpretability question is how strongly a labeled feature is concentrated in
individual activation coordinates. Gurnee et al. study this with k-sparse linear probes: select at
most `k` neurons, retrain a classifier on those neurons, and evaluate held-out performance as `k`
changes.

TransformerLens already exposes activations through `run_with_cache`, but has no maintained probing
primitive. Users must repeatedly rebuild splitting, preprocessing, selection, fitting, and control
logic, where leakage is easy to introduce.

The model-free core is also generic machine-learning infrastructure. Whether TransformerLens should
own and maintain a logistic solver, rather than document composition with sklearn, is the primary
repository-fit question for maintainers.

The output supports hypotheses about coordinate concentration. It does **not** establish that the
model uses the decoded feature, that a selected neuron is monosemantic, or that a gradual k-curve is
by itself evidence of superposition.

---

#### Proposed API

Names and optimizer details remain subject to maintainer approval.

```python
from transformer_lens.tools.analysis import (
    fit_sparse_probe,
    sweep_sparse_probe,
)

probe = fit_sparse_probe(
    X,
    y,
    k=4,
    test_fraction=0.3,
    positive_label=1,
    preprocess="none",
    l2_strength=1e-2,
    seed=0,
)

sweep = sweep_sparse_probe(
    X,
    y,
    ks=[1, 2, 4, 8, 16],
    test_fraction=0.3,
    positive_label=1,
    preprocess="none",
    l2_strength=1e-2,
    n_random_subsets=20,
    n_label_shuffles=20,
    seed=0,
)
```

`X` has shape `[example, feature]`; `y` is a Boolean or integer label tensor containing exactly two
values, one equal to `positive_label`. Callers obtain `X` through any appropriate `run_with_cache`
workflow and choose their own position aggregation policy.

#### Design (algorithm)

For `X ∈ R^{n×d}` and binary labels `y`:

1. Validate finite floating `X`, binary Boolean/integer `y`, `1 <= k <= d`,
   `0 < test_fraction < 1`, positive L2 strength, and at least two examples from each class.
2. Create one deterministic stratified split before computing any learned statistic. For class `c`,
   use `n_test_c = clamp(ceil(test_fraction * n_c), 1, n_c - 1)` and report realized counts.
3. On `X_train` only, compute the paper's raw mean-difference score
   `mean(X_train[y=positive]) - mean(X_train[y=negative])`; rank by absolute value with
   deterministic tie-breaking.
4. Select the top `k` coordinates. `preprocess="none"` matches the reference default. Optional
   `"standardize"` uses selected-column train mean and population standard deviation, maps a
   zero standard deviation to scale one, and applies the same transform to test data. Because L2 is
   scale-sensitive, this mode intentionally changes the fitted objective and is recorded in results.
5. Fit logits `z = Xw + b` by minimizing mean class-balanced binary cross-entropy plus
   `l2_strength * ||w||² / 2`; use weights `n / (2 n_c)` and do not regularize `b`.
6. Optimize selected matrices on CPU in float64 with Torch LBFGS and strong-Wolfe line search.
   Require finite output and a configurable final gradient-infinity-norm threshold; otherwise raise.
7. Predict positive at `z >= 0`. Report held-out confusion counts, accuracy, precision, recall, and
   F1, returning zero for precision/F1 when their denominators are zero. F1 is primary.
8. For a sweep, reuse the exact split across every strictly increasing unique `k`. Random-coordinate
   controls sample supports without replacement and fit the same classifier. Label-shuffle controls
   permute training labels, repeat selection/fitting, and evaluate against untouched test labels.

Control repeat counts are nonnegative; zero disables that control. The implementation uses a local
`torch.Generator` and must not mutate global RNG state. It adds no scikit-learn, Gurobi, plotting, or
dataset dependency.

Every k-sweep uses one fixed preprocessing mode and L2 strength. Curve shape is conditional on those
choices; PR1 performs no hyperparameter search and must not present one setting as canonical.

#### Main correctness risk (named)

**Leakage and the probe-fit-vs-model-use conflation.** The single thing most likely to be subtly wrong is **information leaking from test into the probe** — the most common trap is **column standardization or unit-selection computed on the full dataset** (not the train split), which silently inflates held-out accuracy. The close-second is **interpretation**: a high-accuracy probe means the feature is *linearly decodable*, **not** that the model *uses* it — over-claiming here is the field's canonical error.

Mitigations:

- Split first; compute selection and optional normalization on training data only. The function
  cannot detect leakage already introduced into caller-provided `X`, which documentation must state.
- Use deterministic stratification and balanced class weights rather than treating accuracy as safe
  under imbalance.
- Return precision, recall, and F1 next to accuracy.
- Return raw control distributions rather than assigning automatic "significance" labels.
- Document that controls detect some probe-pipeline failures, not causal model use.

1. **Selector oracle.** Independently compute raw train-only class-mean differences and assert exact
   selected indices, signed scores, and deterministic tie-breaking.
2. **Planted sparse feature.** On seeded synthetic data with one known predictive coordinate, `k=1`
   selects it and held-out F1 exceeds a fixed threshold.
3. **Distributed decodability.** On a seeded feature spread across several coordinates, larger `k`
   improves held-out F1 by a fixed margin. The test makes no superposition claim.
4. **Leakage guard.** In standardize mode, hand-computed train-only means/stds and transformed test
   values match metadata; changing only held-out values cannot alter selected coordinates.
5. **Controls.** Label-shuffle and random-coordinate distributions are deterministic for a seed and
   remain below the planted-feature result by fixed margins.
6. **Optimizer.** Objective and gradient match independent formulas; a tiny analytic dataset reaches
   the configured gradient threshold, while forced non-convergence raises.
7. **Validation.** Empty classes, non-binary labels, non-finite values, invalid `k`, infeasible
   splits, duplicate `k` values, and constant selected columns follow explicit tested policies.

#### Scope of the first PR (vertical slice)

Exact paths:

- `transformer_lens/tools/analysis/sparse_probing.py`
- `tests/unit/tools/test_sparse_probing.py`
- exports from `transformer_lens/tools/analysis/__init__.py`
- `docs/source/content/sparse_probing.md`, including composition with `run_with_cache`
- `docs/source/index.md`, adding the guide to the Resources toctree

Public results should include class labels/counts, selected indices/scores, coefficients/intercept,
preprocessing metadata, split indices, held-out metrics, final objective/gradient/iterations, and raw
control distributions. They must not expose a `.plot()` method or assign a representation label.

#### Follow-up work

Deferred: activation extraction, position aggregation, model integration, plotting, notebooks,
multiclass labels, SAE-latent probing, MIP selection, feature batteries, and causal validation.

---

### Alternatives

- **Document sklearn composition instead:** ecosystem-standard and lower maintenance, but leaves
  leakage-sensitive orchestration and result contracts to each user. This may be preferable if
  maintainers do not want TransformerLens to own a classifier optimizer.
- **Depend on the paper repository:** introduces research-code and Gurobi coupling.
- **Ship MIP selection now:** adds a solver dependency and changes the project from a small PR.
- **Bundle model extraction:** forces unresolved prompt/position aggregation policy into the core.

---

### Correctness oracle

The selector oracle is the independently computed train-only raw class-mean difference. The
mean-difference heuristic is **not** required to approximate the exhaustive best subset; no such
guarantee exists. Logistic fitting is checked through analytic planted data, loss/convergence
metadata, and deterministic repetition.

---

### Additional context

- **Paper:** Wes Gurnee et al., ["Finding Neurons in a Haystack: Case Studies with Sparse
  Probing"](https://arxiv.org/abs/2305.01610), TMLR 2023
  ([OpenReview record](https://openreview.net/forum?id=JYs1R9IMJr)).
- **Reference code:** [`wesg52/sparse-probing-paper` pinned at `a610e102`](https://github.com/wesg52/sparse-probing-paper/tree/a610e102c6e25a6ef9cc16c3a2abb736aa90849b):
  [mean-difference selector](https://github.com/wesg52/sparse-probing-paper/blob/a610e102c6e25a6ef9cc16c3a2abb736aa90849b/experiments/probes.py#L168-L195),
  [split and optional scaling](https://github.com/wesg52/sparse-probing-paper/blob/a610e102c6e25a6ef9cc16c3a2abb736aa90849b/experiments/inner_loops.py#L26-L37),
  [heuristic k-sweep](https://github.com/wesg52/sparse-probing-paper/blob/a610e102c6e25a6ef9cc16c3a2abb736aa90849b/experiments/inner_loops.py#L321-L364),
  [experiment defaults](https://github.com/wesg52/sparse-probing-paper/blob/a610e102c6e25a6ef9cc16c3a2abb736aa90849b/config.py#L5-L30), and
  [classification metrics](https://github.com/wesg52/sparse-probing-paper/blob/a610e102c6e25a6ef9cc16c3a2abb736aa90849b/experiments/metrics.py#L14-L35).
- **Paper alignment:** rank coordinates by raw absolute class-mean difference, retrain a balanced
  logistic probe on the selected support, and treat held-out F1 as the primary metric.
- **Verified reference defaults:** the heuristic sweep first creates a random train/test split
  (`test_set_frac=0.3` by default), optionally scales using `X_train` standard deviations, and ranks
  on `X_train`; normalization is disabled by default and, when enabled, divides by clamped standard
  deviation times ten without centering. Its sklearn SAGA fit uses balanced weights, L2, and
  `max_iter=200` (other regularization/tolerance settings use sklearn defaults).
- **Intentional divergences:** stratify, offer explicit centered train-only standardization, use a
  defined Torch loss/L2 convention, add deterministic controls and convergence metadata, and state
  claim boundaries.
- **TransformerLens touchpoints:** current
  [`tools.analysis` package](https://github.com/TransformerLensOrg/TransformerLens/tree/dev/transformer_lens/tools/analysis),
  [`HookedRootModule.run_with_cache`](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/transformer_lens/HookedRootModule.py#L381),
  [`TransformerBridge.run_with_cache`](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/transformer_lens/model_bridge/bridge.py#L2707),
  [`ActivationCache`](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/transformer_lens/ActivationCache.py#L55), and
  [dependency manifest](https://github.com/TransformerLensOrg/TransformerLens/blob/dev/pyproject.toml).
- **Artifact requirements:** none; PR1 tests use only synthetic tensors.

---

### Maintainer decisions requested

1. **Repository fit and scope:** should TransformerLens own a model-free sparse-probe solver at all?
   If yes, should PR1 remain model-free with extraction/plotting deferred? **Recommendation: proceed
   only if maintainers explicitly accept optimizer ownership; then keep PR1 model-free.**
2. **Optimizer:** is the explicitly defined CPU-float64 Torch LBFGS objective acceptable instead of
   adding sklearn? **Recommendation: yes, with final-gradient metadata and failure.**
3. **Controls:** should random-coordinate and shuffled-training-label distributions be first-class
   sweep outputs? **Recommendation: yes, disabled only when repetition count is zero.**
4. **Preprocessing:** should `"none"` match the paper-code default, with optional train-only
   `"standardize"` explicitly documented as changing the L2 objective? **Recommendation: yes.**
5. **Device/dtype:** should score reductions run in at least float32 on the input device, while only
   the selected `[example, k]` matrices move to CPU float64 for deterministic LBFGS?
   **Recommendation: yes; return CPU analysis results and document the transfer explicitly.**

---

### Checklist

- [x] No competing TransformerLens or SAELens probing issue, PR, or implementation found as of
  2026-08-30.
- [x] Paper heuristic and primary metric checked against arXiv:2305.01610.
- [x] Reference defaults and leakage order checked against commit `a610e102`.
- [x] Torch 2.10 CPU-float64 LBFGS prototype converged deterministically across FP16/BF16/FP32/FP64
  inputs; this does not replace tests on the repository's supported Torch/Python matrix.
- [x] PR1 narrowed to model-free fitting, sweeping, controls, tests, exports, and docs.
- [ ] Maintainer decisions recorded before implementation.
- [ ] Follow-up extraction/plotting/model work tracked separately if requested.

---

**Suggested labels:** `enhancement`, `tooling`

---

janmenjayap:
---

### Description

Adds `transformer_lens/tools/analysis/sparse_probing.py`, a dependency-free, model-free module for fitting binary k sparse probes to a supplied activation matrix `X` and label vector `y`, following the method in Gurnee et al., ["Finding Neurons in a Haystack: Case Studies with Sparse Probing"](https://arxiv.org/abs/2305.01610) (TMLR 2023).

TransformerLens already exposes activations via `run_with_cache`, but has no maintained probing primitive: users currently rebuild splitting, feature selection, fitting, and control logic themselves, and it is easy to leak test information into that process. This PR adds a leakage-safe, dependency-free core for that workflow. Activation extraction, model wrappers, plotting, notebooks, multiclass probing, and optimal (MIP) selection are deliberately deferred to follow-up work.

Fixes #1728

---

### What's included

- `fit_sparse_probe(X, y, k, ...)` — deterministic stratified train/test split computed before any learned statistic, train-only raw mean-difference feature selection, and a CPU-float64 Torch LBFGS balanced-logistic fit with explicit objective/gradient-norm convergence diagnostics (raises rather than silently returning an unconverged fit).
- `sweep_sparse_probe(X, y, ks, ...)` — reuses one fixed split across a k-grid and reports raw random-coordinate and shuffled-training-label control distributions alongside each `k`, without assigning an automatic "significance" label.
- `SparseProbeResult` / `SparseProbeSweep` result dataclasses reporting held-out accuracy, precision, recall, and F1 (F1 primary), selected indices/scores, coefficients/intercept, preprocessing metadata, split indices/class counts, and control distributions.
- A regression test asserting that MPS-resident features are moved to CPU before the float64 fit (float64 is unsupported on MPS).
- A `jaxtyping>=0.3` compatibility fix: the rebase onto `origin/dev` pulled in #1732, which now raises `jaxtyping.TypeCheckError` instead of letting `BeartypeCallHintParamViolation` propagate, and reclassifies `float8_e4m3fn` as a valid `Float` dtype at the annotation level. Updated `test_runtime_typecheck_rejects_invalid_tensor_contracts` to use `tests/typecheck_errors.TYPECHECK_ERRORS`, and moved the float8 case to this module's own `ValueError` dtype-rejection table, where it's now actually caught.
- Exports from `transformer_lens.tools.analysis`, plus `docs/source/content/sparse_probing.md` (contracts, claim boundaries, and a `run_with_cache` composition example) linked from `docs/source/index.md`.

---

### Design decisions

Issue #1728 asked maintainers to weigh in on five open questions before implementation. This PR takes the issue's own recommended position on each:

1. **Repository fit / model-free scope** — kept model-free; extraction and plotting are out of scope for this PR.
2. **Optimizer** — uses an explicit CPU-float64 Torch LBFGS objective (no new `sklearn` dependency), and reports final gradient/objective metadata, raising on non-convergence.
3. **Controls** — random-coordinate and shuffled-training-label distributions are first-class sweep outputs, disabled only when their repeat count is zero.
4. **Preprocessing** — defaults to `"none"` (matching the paper's reference code); optional train-only `"standardize"` is documented as intentionally changing the L2 objective.
5. **Device/dtype** — score reductions run in at least float32 on the input device; only the selected `[example, k]` matrices move to CPU float64 for the deterministic LBFGS fit.

---

### Non-goals (tracked as follow-ups)

Activation extraction and position-reduction helpers, model wrappers/downloads, multiclass/one-vs-rest probing, plotting and notebooks, an optimal/MIP selector, feature batteries, SAE-latent composition, and causal validation are all explicitly out of scope for this PR.

---

### Claim boundaries

A high held-out F1 means the labeled feature is *linearly decodable* from the supplied activations — it does not by itself establish that the model *uses* that feature, that a selected coordinate is monosemantic, or that a smooth k-sweep curve is evidence of superposition. The docs and result contracts call this out explicitly, and the module exposes no `.plot()` method or automatic "significant" label.

---

### Test plan

- `uv run pytest tests/unit/tools/test_sparse_probing.py tests/mps/test_mps_basic.py -q` — 68 passed.
- `make test-pr` (unit + docstring + acceptance + integration): unit 5,827 passed, docstring 18 passed, acceptance 209 passed, integration 1,459 passed / 1 failed. The one failure, `test_granite_eager_scan_device_correctness[mps]`, is an unrelated Granite MoE Hybrid eager/fused-scan divergence — reproduced identically on a clean `origin/dev` checkout in an isolated worktree, confirming it's pre-existing on trunk and not introduced by this branch.
- `make format` and `uv run mypy .` clean.
- `docs/source/index.md` includes the new guide exactly once; docs build succeeds.

---

### Type of change

---

### Checklist:

---

### DIFF
--- a/docs/source/content/sparse_probing.md
+++ b/docs/source/content/sparse_probing.md
@@ -1,0 +1,153 @@
+# Sparse Probing
+
+Sparse probing measures how strongly binary-label information is concentrated in a small
+set of activation coordinates. TransformerLens provides a model-independent fit and sweep API;
+callers decide how to collect and aggregate activations.
+
+Probe performance establishes held-out decodability under the chosen split, preprocessing,
+regularization, and sparsity. It does not establish causal model use, neuron monosemanticity,
+or superposition.
+
+## Fit one probe
+
+```python
+import torch
+
+from transformer_lens.tools.analysis import fit_sparse_probe
+
+features = torch.randn(200, 768)
+labels = torch.arange(200) % 2
+result = fit_sparse_probe(
+    features,
+    labels,
+    k=8,
+    positive_label=1,
+    test_fraction=0.3,
+    preprocess="none",
+    class_weight="balanced",
+    l2_strength=1e-2,
+    seed=0,
+)
+
+print(result.selected_features)
+print(result.metrics.f1)
+```
+
+`features` must have shape `[example, feature]` and dtype float16, bfloat16, float32, or float64.
+Labels must be a one-dimensional Boolean or integer tensor containing exactly two values, one equal
+to `positive_label`. Each class needs at least two examples.
+
+## Selection and split contract
+
+The function creates a deterministic stratified split before computing any learned statistic.
+For each class $c$, its realized test count is
+
+$$
+n_{\mathrm{test},c} =
+\operatorname{clamp}(\lceil f n_c \rceil, 1, n_c - 1),
+$$
+
+where $f$ is `test_fraction`. Realized train and test counts are returned because small classes
+can differ materially from the requested aggregate fraction.
+
+On the training split only, feature $j$ receives the signed score
+
+$$
+s_j = \mathbb{E}[X_j \mid y=\mathrm{positive}]
+      - \mathbb{E}[X_j \mid y=\mathrm{negative}].
+$$
+
+The selected support contains the $k$ largest $|s_j|$. Equal scores are resolved by increasing
+feature index. `preprocess="none"` fits the selected raw coordinates. With
+`preprocess="standardize"`, selected columns are centered and scaled using training statistics;
+zero-variance columns receive scale one. The same transform is then applied to held-out values.
+
+Because L2 regularization is scale-sensitive, preprocessing can change the fitted probe and
+the resulting k-curve. A sweep therefore fixes preprocessing and L2 strength across every k.
+
+## Logistic objective
+
+For selected training features $X$, labels $y \in \{0,1\}$, coefficients $w$, and intercept $b$,
+the optimizer minimizes
+
+$$
+\frac{1}{n}\sum_i \alpha_{y_i}
+\operatorname{BCEWithLogits}(X_i w + b, y_i)
++ \frac{\lambda}{2}\lVert w\rVert_2^2,
+\qquad
+\alpha_c = \frac{n}{2n_c}.
+$$
+
+The displayed weights apply to the default `class_weight="balanced"`; pass `None` to use
+$\alpha_c=1$. The intercept is not regularized. Positive predictions have nonnegative logits.
+Accuracy, precision, recall, F1, and all four confusion counts are returned; precision or F1 is
+zero when its denominator is zero. F1 is the primary sparse-probing metric.
+
+Feature-score reductions use float64 for float64 inputs and float32 otherwise. Selected matrices
+move to CPU float64 for deterministic LBFGS fitting. All result tensors are detached CPU tensors.
+The fit raises when output is non-finite or the final objective-gradient infinity norm exceeds
+`gradient_tolerance`, which must lie in `(0, 1]`.
+Results retain the requested `k`, `max_iter`, and `gradient_tolerance` alongside the realized
+objective, gradient norm, iteration count, and convergence flag.
+
+## Sweep and controls
+
+```python
+from transformer_lens.tools.analysis import sweep_sparse_probe
+
+sweep = sweep_sparse_probe(
+    features,
+    labels,
+    ks=[1, 2, 4, 8, 16],
+    n_random_subsets=20,
+    n_label_shuffles=20,
+    seed=0,
+)
+
+for k, probe, random_control in zip(
+    sweep.ks,
+    sweep.results,
+    sweep.random_coordinate_controls,
+    strict=True,
+):
+    print(k, probe.metrics.f1, random_control.f1.median())
+```
+
+Every k uses the same split, preprocessing mode, and L2 strength. `ks` must be strictly
+increasing and unique.
+
+Random-coordinate controls sample k distinct coordinates and fit the same classifier.
+Label-shuffle controls permute training labels, repeat selection and fitting, and evaluate against
+the untouched held-out labels. The API returns raw control supports and metric distributions; it
+does not convert them into p-values or representation labels. A repeat count of zero disables that
+control.
+
+Controls can be expensive: the sweep performs one main fit plus both requested control counts for
+every k. Start with small grids and repeat counts.
+
+## Composing with cached activations
+
+Use `run_with_cache` to construct the feature matrix separately so token, position, batching, and
+aggregation choices remain explicit:
+
+```python
+tokens = model.to_tokens(prompts)
+_, cache = model.run_with_cache(tokens, names_filter=[hook_name])
+features = cache[hook_name][:, -1, :]
+result = fit_sparse_probe(features, labels, k=8)
+```
+
+The example selects the final sequence position, which is not appropriate for every dataset.
+Choose the hook and position policy before interpreting selected coordinates. The API cannot detect
+leakage already introduced into caller-provided `features`.
+
+## Reference
+
+The raw mean-difference selector and sparse-probing framing follow Wes Gurnee et al.,
+“Finding Neurons in a Haystack: Case Studies with Sparse Probing,”
+[TMLR 2023](https://openreview.net/forum?id=JYs1R9IMJr),
+with [reference code](https://github.com/wesg52/sparse-probing-paper).
+
+TransformerLens intentionally adds stratification, stable tie-breaking, explicit objective and
+convergence diagnostics, and deterministic controls. Its optional centered standardization and
+Torch LBFGS solver are not exact reproductions of the reference implementation.
--- a/tests/unit/tools/test_sparse_probing.py
+++ b/tests/unit/tools/test_sparse_probing.py
@@ -4,8 +4,8 @@ from dataclasses import FrozenInstanceError
 
 import pytest
 import torch
-from beartype.roar import BeartypeCallHintParamViolation
 
+from tests.typecheck_errors import TYPECHECK_ERRORS
 from transformer_lens.tools.analysis import (
     fit_sparse_probe as exported_fit_sparse_probe,
 )
@@ -288,6 +288,12 @@ def test_binary_metrics_zero_division_policy():
     [
         (torch.ones(0, 2), torch.empty(0, dtype=torch.int64), {}, "non-empty"),
         (torch.tensor([[1.0], [float("nan")]]), torch.tensor([0, 1]), {}, "finite"),
+        (
+            torch.ones(4, 2, dtype=torch.float8_e4m3fn),
+            torch.tensor([0, 1, 0, 1]),
+            {},
+            "supported dtype",
+        ),
         (torch.ones(4, 2), torch.zeros(4, dtype=torch.int64), {}, "exactly two"),
         (torch.ones(6, 2), torch.tensor([0, 1, 2, 0, 1, 2]), {}, "exactly two"),
         (torch.ones(4, 2), torch.tensor([0, 1, 0, 1]), {"positive_label": 2}, "positive_label"),
@@ -325,14 +331,13 @@ def test_rejects_invalid_inputs(features, labels, kwargs, message):
         ([[1.0], [2.0], [3.0], [4.0]], torch.tensor([0, 1, 0, 1])),
         (torch.ones(4), torch.tensor([0, 1, 0, 1])),
         (torch.ones(4, 2, dtype=torch.int64), torch.tensor([0, 1, 0, 1])),
-        (torch.ones(4, 2, dtype=torch.float8_e4m3fn), torch.tensor([0, 1, 0, 1])),
         (torch.ones(4, 2), torch.tensor([0.0, 1.0, 0.0, 1.0])),
         (torch.ones(4, 2), torch.tensor([[0, 1], [0, 1]])),
         (torch.ones(4, 2), torch.tensor([0, 1, 0])),
     ],
 )
 def test_runtime_typecheck_rejects_invalid_tensor_contracts(features, labels):
-    with pytest.raises(BeartypeCallHintParamViolation):
+    with pytest.raises(TYPECHECK_ERRORS):
         fit_sparse_probe(features, labels, k=1)
 
 
--- a/tests/unit/tools/test_sparse_probing.py
+++ b/tests/unit/tools/test_sparse_probing.py
@@ -1,0 +1,320 @@
+"""Unit tests for leakage-safe model-free sparse probing."""
+
+from dataclasses import FrozenInstanceError
+
+import pytest
+import torch
+from beartype.roar import BeartypeCallHintParamViolation
+
+from transformer_lens.tools.analysis.sparse_probing import (
+    SparseProbeResult,
+    _binary_metrics,
+    fit_sparse_probe,
+)
+
+
+def _planted_data(
+    *, n_examples: int = 400, n_features: int = 16, seed: int = 0
+) -> tuple[torch.Tensor, torch.Tensor]:
+    generator = torch.Generator().manual_seed(seed)
+    labels = torch.arange(n_examples) % 2
+    features = torch.randn(n_examples, n_features, generator=generator)
+    features[:, 3] += 2.5 * (2 * labels - 1)
+    permutation = torch.randperm(n_examples, generator=generator)
+    return features[permutation], labels[permutation]
+
+
+def _balanced_objective_gradient(
+    features: torch.Tensor,
+    labels: torch.Tensor,
+    coefficients: torch.Tensor,
+    intercept: torch.Tensor,
+    l2_strength: float,
+) -> torch.Tensor:
+    labels = labels.double()
+    count = labels.numel()
+    positive_count = labels.sum()
+    weights = torch.where(
+        labels == 1,
+        count / (2 * positive_count),
+        count / (2 * (count - positive_count)),
+    )
+    residual = weights * (torch.sigmoid(features @ coefficients + intercept) - labels) / count
+    return torch.cat(
+        (features.T @ residual + l2_strength * coefficients, residual.sum().reshape(1))
+    )
+
+
+def _newton_reference(
+    features: torch.Tensor,
+    labels: torch.Tensor,
+    l2_strength: float,
+) -> torch.Tensor:
+    features = features.double()
+    labels = labels.double()
+    design = torch.cat((features, torch.ones(features.shape[0], 1, dtype=torch.float64)), dim=1)
+    count = labels.numel()
+    positive_count = labels.sum()
+    weights = torch.where(
+        labels == 1,
+        count / (2 * positive_count),
+        count / (2 * (count - positive_count)),
+    )
+    penalty = torch.diag(
+        torch.tensor([l2_strength] * features.shape[1] + [0.0], dtype=torch.float64)
+    )
+    parameters = torch.zeros(design.shape[1], dtype=torch.float64)
+    for _ in range(100):
+        probability = torch.sigmoid(design @ parameters)
+        residual = weights * (probability - labels) / count
+        gradient = design.T @ residual + penalty @ parameters
+        curvature = weights * probability * (1 - probability) / count
+        hessian = design.T @ (curvature[:, None] * design) + penalty
+        parameters -= torch.linalg.solve(hessian, gradient)
+        if float(gradient.abs().max()) < 1e-12:
+            break
+    return parameters
+
+
+def test_fit_recovers_exact_train_only_mean_difference_and_planted_feature():
+    features, labels = _planted_data()
+
+    result = fit_sparse_probe(features, labels, k=1, positive_label=1, seed=17)
+
+    train_features = features[result.train_indices]
+    train_labels = labels[result.train_indices]
+    expected_scores = train_features[train_labels == 1].mean(0) - train_features[
+        train_labels == 0
+    ].mean(0)
+    expected_selected = torch.argsort(expected_scores.abs(), descending=True, stable=True)[:1]
+    assert isinstance(result, SparseProbeResult)
+    assert result.selected_features.tolist() == [3]
+    assert torch.equal(result.selected_features, expected_selected)
+    assert torch.allclose(result.feature_scores, expected_scores.double(), atol=1e-6)
+    assert result.metrics.f1 > 0.98
+    assert result.metrics.accuracy > 0.98
+    assert result.k == 1
+    assert result.max_iter == 200
+    assert result.gradient_tolerance == 1e-7
+
+
+def test_positive_label_controls_score_sign_and_class_metadata():
+    features, labels = _planted_data()
+    signed_labels = 1 - 2 * labels
+
+    result = fit_sparse_probe(features, signed_labels, k=1, positive_label=-1, seed=11)
+
+    assert result.positive_label == -1
+    assert result.negative_label == 1
+    assert result.feature_scores[3] > 0
+
+
+def test_unweighted_classification_policy_is_explicit():
+    features, labels = _planted_data(n_examples=100, n_features=6)
+
+    result = fit_sparse_probe(features, labels, k=2, class_weight=None, seed=2)
+
+    assert result.class_weight is None
+
+
+def test_standardization_is_train_only_and_heldout_values_do_not_change_selection():
+    features, labels = _planted_data(n_examples=200, n_features=8)
+    first = fit_sparse_probe(features, labels, k=2, preprocess="standardize", seed=9)
+    changed = features.clone()
+    changed[first.test_indices] += 10_000 * torch.randn_like(changed[first.test_indices])
+
+    second = fit_sparse_probe(changed, labels, k=2, preprocess="standardize", seed=9)
+
+    selected_train = features[first.train_indices][:, first.selected_features].double()
+    expected_mean = selected_train.mean(0)
+    expected_scale = selected_train.std(0, correction=0)
+    assert torch.equal(first.selected_features, second.selected_features)
+    assert torch.equal(first.feature_scores, second.feature_scores)
+    assert torch.allclose(first.preprocess_mean, expected_mean)
+    assert torch.allclose(first.preprocess_scale, expected_scale)
+    assert torch.equal(first.preprocess_mean, second.preprocess_mean)
+    assert torch.equal(first.preprocess_scale, second.preprocess_scale)
+    assert torch.equal(first.coefficients, second.coefficients)
+    assert torch.equal(first.intercept, second.intercept)
+    assert first.objective == second.objective
+
+
+def test_none_preprocessing_has_identity_metadata_and_constant_tie_order():
+    features = torch.zeros(20, 5)
+    labels = torch.arange(20) % 2
+
+    result = fit_sparse_probe(features, labels, k=3, preprocess="none", seed=1)
+
+    assert result.selected_features.tolist() == [0, 1, 2]
+    assert torch.equal(result.preprocess_mean, torch.zeros(3, dtype=torch.float64))
+    assert torch.equal(result.preprocess_scale, torch.ones(3, dtype=torch.float64))
+    assert result.constant_features.tolist() == [True, True, True]
+
+
+def test_lbfgs_matches_independent_newton_solution_and_gradient():
+    features, labels = _planted_data(n_examples=120, n_features=4, seed=4)
+    l2_strength = 0.02
+    result = fit_sparse_probe(
+        features,
+        labels,
+        k=4,
+        l2_strength=l2_strength,
+        gradient_tolerance=1e-7,
+        seed=3,
+    )
+    train_features = features[result.train_indices][:, result.selected_features].double()
+    train_labels = labels[result.train_indices]
+    expected = _newton_reference(train_features, train_labels, l2_strength)
+    gradient = _balanced_objective_gradient(
+        train_features,
+        train_labels,
+        result.coefficients,
+        result.intercept,
+        l2_strength,
+    )
+
+    assert torch.allclose(result.coefficients, expected[:-1], atol=2e-6, rtol=2e-6)
+    assert result.intercept.item() == pytest.approx(expected[-1].item(), abs=2e-6)
+    assert float(gradient.abs().max()) == pytest.approx(result.gradient_inf_norm, abs=1e-12)
+    assert result.gradient_inf_norm <= 1e-7
+
+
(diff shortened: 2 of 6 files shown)


transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,741 @@
...
+        raise RuntimeError("sparse probe optimizer failed") from error
+    detached = parameters.detach()
+    objective = float(
+        _objective(features, labels, detached, sample_weights, l2_strength).detach().item()
+    )
+    gradient = _objective_gradient(features, labels, detached, sample_weights, l2_strength)
+    gradient_inf_norm = float(gradient.abs().max().item())
+    if not math.isfinite(objective) or not math.isfinite(gradient_inf_norm):
+        raise RuntimeError("sparse probe optimizer produced non-finite output")
+    if gradient_inf_norm > gradient_tolerance:
```
jlarson4, line 381: The objective is mean-reduced and scale-invariant, but its gradient grows with `|X|`, so the absolute `1e-7` default rejects fits that match a float64 Newton solve to a few ulp: on n=1000, d=768 activations with per-coordinate std 60, every default `fit_sparse_probe` raises and `sweep_sparse_probe` dies on its first such fit. Because line 359 derives `tolerance_change` from `gradient_tolerance**2`, relaxing the tolerance also stops LBFGS earlier and `max_iter` changes nothing. Please make the acceptance test scale-relative (for example `gradient_inf_norm <= gradient_tolerance * max(1.0, initial_gradient_inf_norm)`), set `tolerance_change=0.0`, and add a guard test that fits the defaults on a matrix with a coordinate at std in the hundreds.

transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,741 @@
...
+        train_positive_count=int(train_labels.sum().item()),
+        train_negative_count=int((~train_labels).sum().item()),
+        test_positive_count=int(test_labels.sum().item()),
+        test_negative_count=int((~test_labels).sum().item()),
+        preprocess=validated.preprocess,
+        class_weight=validated.class_weight,
+        l2_strength=validated.l2_strength,
+        test_fraction=validated.test_fraction,
+        seed=validated.seed,
+        k=validated.k,
```
jlarson4, line 471: `sweep_sparse_probe` validates once with `k=ks[-1]` (line 668), so this line stamps that value on every result: for `ks=[1, 2, 4]` all three read `k=4` while their supports are sized 1, 2, 4, and any `(r.k, r.metrics.f1)` curve comes out flat in k. Please pass the loop's `k` into `_fit_result` explicitly and assert `tuple(r.k for r in sweep.results) == sweep.ks` in the sweep test.

tests/unit/tools/test_sparse_probing.py
```diff
@@ -0,0 +1,472 @@
...
+
+    result = fit_sparse_probe(features, labels.to(dtype), k=2, seed=0)
+
+    assert result.positive_label == 1
+    assert result.negative_label == 0
+
+
+def test_stratification_preserves_each_class_and_reports_realized_counts():
+    generator = torch.Generator().manual_seed(0)
+    features = torch.randn(100, 5, generator=generator)
```
jlarson4, line 228: This range check is the only direct assertion on a returned metric, and every other metric check compares two paths mutated the same way: scoring the training rows instead of the held-out rows, permuting the labels the control arms are scored against, ignoring `class_weight`, holding the split seed constant, swapping precision and recall, and moving the decision threshold to 0.5 each leave all 53 tests green. Please add a test that recomputes the confusion counts from `features[result.test_indices]`, `labels[result.test_indices]` and the returned coefficients and intercept and asserts equality with `result.metrics`, plus a shuffle-arm check that the scored labels are `labels[test_indices]`.

docs/source/content/sparse_probing.md
```diff
@@ -0,0 +1,153 @@
...
+$\alpha_c=1$. The intercept is not regularized. Positive predictions have nonnegative logits.
+Accuracy, precision, recall, F1, and all four confusion counts are returned; precision or F1 is
+zero when its denominator is zero. F1 is the primary sparse-probing metric.
+
+Feature-score reductions use float64 for float64 inputs and float32 otherwise. Selected matrices
+move to CPU float64 for deterministic LBFGS fitting. All result tensors are detached CPU tensors.
+The fit raises when output is non-finite or the final objective-gradient infinity norm exceeds
+`gradient_tolerance`, which must lie in `(0, 1]`.
+Results retain the requested `k`, `max_iter`, and `gradient_tolerance` alongside the realized
+objective, gradient norm, iteration count, and convergence flag.
```
jlarson4, line 91: `SparseProbeResult` has `objective`, `gradient_inf_norm`, `iterations` and `function_evaluations`, and a fit that misses the threshold raises rather than returning a flag, so `result.converged` is an `AttributeError`. Please replace "and convergence flag" with "and function-evaluation count" and say that a fit missing the threshold raises.

docs/source/content/sparse_probing.md
```diff
@@ -0,0 +1,153 @@
...
+```python
+tokens = model.to_tokens(prompts)
+_, cache = model.run_with_cache(tokens, names_filter=[hook_name])
+features = cache[hook_name][:, -1, :]
+result = fit_sparse_probe(features, labels, k=8)
+```
+
+The example selects the final sequence position, which is not appropriate for every dataset.
+Choose the hook and position policy before interpreting selected coordinates. The API cannot detect
+leakage already introduced into caller-provided `features`.
```
jlarson4, line 142: The row-level split does not know that several positions of one prompt share a label, so the common way of building `features` puts positions of the same prompt on both sides: on label-independent grouped data it scores 0.78 held-out accuracy where a group-held-out split scores 0.50. Please add one sentence here saying rows that share a source prompt must not straddle the split.

jlarson4 (commented):
Thanks for putting this together @janmenjayap. A few comments below:

transformer_lens/tools/analysis/sparse_probing.py
janmenjayap: You are right, and thank you for tracing this all the way to the `tolerance_change` coupling. The absolute `1e-7` bound rejected fits that already matched a float64 Newton solve, and because `tolerance_change` was derived from `gradient_tolerance ** 2`, loosening the tolerance only stopped LBFGS earlier while `max_iter` did nothing.

Fixed in commit deb0047f:

- Set `tolerance_change=0.0` so the stopping rule no longer depends on the acceptance rule.
- Captured the objective-gradient infinity norm at the starting parameters (`initial_gradient_inf_norm`) before the first `optimizer.step`.
- Replaced the acceptance test with the scale-relative form you suggested: `gradient_inf_norm <= gradient_tolerance * max(1.0, initial_gradient_inf_norm)`.
- Updated the raised message so the reported threshold is the scale-relative bound actually applied.

The same commit adds the guard test `test_default_tolerance_accepts_large_scale_activations`, which fits the defaults on a matrix with a coordinate at std in the hundreds and asserts the fit returns rather than raising. The docstring in `docs/source/content/sparse_probing.md` was reworded to describe the scale-relative bound.

transformer_lens/tools/analysis/sparse_probing.py
janmenjayap: Confirmed. Since the sweep validates once with `k=ks[-1]`, every result read back the largest requested k while the supports were correctly sized, so any `(r.k, r.metrics.f1)` curve came out flat in k.

Fixed in commit a03b090b:

- Added an explicit `k: int` parameter to `_fit_result` and set `k=k` on the returned `SparseProbeResult` instead of `k=validated.k`.
- The sweep loop now passes its loop variable `k`, and the single-fit `fit_sparse_probe` path passes `validated.k`, which is the correct requested k there.

The sweep test now asserts `tuple(result.k for result in sweep.results) == sweep.ks`, exactly as you proposed.

tests/unit/tools/test_sparse_probing.py
janmenjayap: This was a good catch on the weakness of the existing checks. The only direct assertion was a range check, and the mutation-symmetric comparisons would have missed scoring the training rows, permuting the control labels, ignoring `class_weight`, a constant split seed, a precision and recall swap, or a 0.5 decision threshold.

Addressed in commit a22ad223 with two independent tests:

- `test_metrics_match_an_independent_heldout_recompute_at_the_logit_zero_threshold` recomputes the confusion counts from `features[result.test_indices]`, `labels[result.test_indices]`, and the returned coefficients and intercept, then asserts equality with `result.metrics`. This pins the metrics to the held-out rows and the true logit-zero decision rule and catches a precision and recall swap.
- `test_label_shuffle_control_is_scored_against_the_true_heldout_labels` asserts the shuffle arm is scored against `labels[test_indices]`, so shuffling affects only the training labels used for support selection and fitting, not the scoring labels.

docs/source/content/sparse_probing.md
janmenjayap: Correct. `SparseProbeResult` exposes `objective`, `gradient_inf_norm`, `iterations`, and `function_evaluations`, and a fit that misses the threshold raises rather than returning a flag.

Fixed in commit 20809847: the field list now reads "and function-evaluation count" in place of "and convergence flag", and the sentence now states that a fit missing the acceptance threshold raises instead of returning.

docs/source/content/sparse_probing.md
janmenjayap: Thank you for measuring this. The row-level split is label-independent and does not know that several positions of one prompt share a label, so the usual feature-building approach can place positions of the same prompt on both sides, which inflated held-out accuracy from 0.50 to 0.78 on your grouped example.

Fixed in commit 20809847: the position-policy paragraph now states that rows sharing a source prompt, for example multiple positions of the same sequence, must be kept on one side of the split, and notes that the label-independent row split can otherwise place them on both sides and inflate held-out accuracy on grouped data.

janmenjayap:
Thanks again for the careful read. Happy to adjust any of the wording or test shapes if you would prefer a different form.

transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,749 @@
...
+        raise RuntimeError("sparse probe optimizer failed") from error
+    detached = parameters.detach()
+    objective = float(
+        _objective(features, labels, detached, sample_weights, l2_strength).detach().item()
+    )
+    gradient = _objective_gradient(features, labels, detached, sample_weights, l2_strength)
+    gradient_inf_norm = float(gradient.abs().max().item())
+    if not math.isfinite(objective) or not math.isfinite(gradient_inf_norm):
+        raise RuntimeError("sparse probe optimizer produced non-finite output")
+    acceptance_threshold = gradient_tolerance * max(1.0, initial_gradient_inf_norm)
```
jlarson4: This is something I suggested in my initial review and I definitely need to adjust my recommendation here. Apologies for leading you in the wrong direction.

The gradient at the zero starting point measures label–feature correlation rather than feature magnitude, so the probe support deserves a larger threshold while the random-subset and label-shuffle controls should be smaller, and the smallest are pinned back to the bare absolute tolerance by the `max(1.0, ...)` floor. At activation scale a sweep still dies on a converged control fit, discarding every completed fit with it, even though every single `fit_sparse_probe` at the same scale now passes. 

Can you please key the bound to the magnitude of the selected training columns instead, for example `gradient_tolerance * max(1.0, k * max|X_train|)` while leaving LBFGS's own `tolerance_grad` as it is, and add a sweep with both control arms at activation scale to the guard test?

transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,749 @@
...
+    l2_strength: float,
+    max_iter: int,
+    gradient_tolerance: float,
+) -> _FitOutcome:
+    labels = labels.to(dtype=torch.float64)
+    sample_weights = _sample_weights(labels, class_weight)
+    parameters = torch.zeros(features.shape[1] + 1, dtype=torch.float64, requires_grad=True)
+    optimizer = torch.optim.LBFGS(
+        [parameters],
+        max_iter=max_iter,
```
jlarson4, line 360: LBFGS is still told to stop at the absolute `gradient_tolerance`, which it cannot reach at scale, so it runs until `max_iter` or the evaluation budget binds, and the budget is never set here, so it is torch's undocumented `max_iter * 5 // 4`. The relative bound then accepts whatever the solver reached: at per-coordinate std in the thousands roughly half the accepted fits are such truncations, some of them far above the optimum in objective with coefficients well off the true ones even though their held-out metrics look fine, and a caller can only tell by comparing `function_evaluations` against a cap that appears nowhere in the signature or the guide. Please set `max_eval` explicitly and report the stop reason on the result, so a fit that ended on `max_iter` or the evaluation budget is distinguishable from one that met the tolerance.

transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,749 @@
...
+    validated_preprocess = cast(PreprocessMode, preprocess)
+    if class_weight not in ("balanced", None):
+        raise ValueError(f"class_weight must be 'balanced' or None, got {class_weight!r}")
+    validated_class_weight = cast(ClassWeightMode, class_weight)
+    validated_l2 = _finite_positive_real(l2_strength, "l2_strength")
+    if isinstance(seed, bool) or not isinstance(seed, int) or not 0 <= seed < 2**63:
+        raise ValueError(f"seed must be an integer in [0, 2**63), got {seed!r}")
+    validated_max_iter = _positive_integer(max_iter, "max_iter")
+    validated_tolerance = _finite_positive_real(gradient_tolerance, "gradient_tolerance")
+    if validated_tolerance > 1:
```
jlarson4: At the top of the documented `(0, 1]` range the acceptance threshold is at least the initial gradient, which no fit can miss, and LBFGS's own `tolerance_grad` of 1.0 stops the solve before it takes a step whenever the initial gradient is at or below 1, so the call returns coefficients that are all zero and an objective of exactly ln 2, the untrained probe, as a successful fit. Please reject `gradient_tolerance` at or above 1 here, no value there can express a convergence requirement.

tests/unit/tools/test_sparse_probing.py
```diff
@@ -0,0 +1,601 @@
...
+    assert result.feature_scores[3] > 0
+
+
+def test_unweighted_classification_policy_is_explicit():
+    features, labels = _planted_data(n_examples=100, n_features=6)
+
+    result = fit_sparse_probe(features, labels, k=2, class_weight=None, seed=2)
+    sweep = sweep_sparse_probe(features, labels, ks=[1], class_weight=None, seed=2)
+
+    assert result.class_weight is None
```
jlarson4: This checks that the metadata round-trips, not that the weighting was applied, and the fixture is balanced, so balanced weights are all 1.0 and neither mode can be distinguished: forcing the weights to ones, and separately ignoring `class_weight=None` so the fit is always balanced, both leave every test green, while on an imbalanced fixture the two modes give visibly different accuracies and coefficients. Please fit an imbalanced fixture in both modes and assert the coefficient vectors differ, each against a Newton reference that takes the same sample weights.

tests/unit/tools/test_sparse_probing.py
```diff
@@ -0,0 +1,601 @@
...
+        l2_strength,
+    )
+
+    assert torch.allclose(result.coefficients, expected[:-1], atol=2e-6, rtol=2e-6)
+    assert result.intercept.item() == pytest.approx(expected[-1].item(), abs=2e-6)
+    assert float(gradient.abs().max()) == pytest.approx(result.gradient_inf_norm, abs=1e-12)
+    assert result.gradient_inf_norm <= 1e-7
+
+
+def test_metrics_match_an_independent_heldout_recompute_at_the_logit_zero_threshold():
```
jlarson4: This recompute runs under the default `preprocess="none"`, where `preprocess_mean` and `preprocess_scale` are identity, so standardizing the held-out rows with their own statistics instead of the training ones leaves every test green while changing the held-out metrics, which is train-test leakage in the module whose headline contract is leakage safety. Please run this recompute under `preprocess="standardize"` as well, applying the returned `preprocess_mean` and `preprocess_scale` to the raw held-out rows, so held-out rows standardized with their own statistics fail it.

tests/unit/tools/test_sparse_probing.py
```diff
@@ -0,0 +1,601 @@
...
+        fit_sparse_probe(
+            features,
+            labels,
+            k=3,
+            max_iter=1,
+            gradient_tolerance=1e-12,
+        )
+
+
+def test_default_tolerance_accepts_large_scale_activations():
```
jlarson4: This test's fixture converges to an achieved gradient roughly ten times under the plain absolute `1e-7`, so the scale-relative bound is never exercised: reverting line 385 to the bare `gradient_tolerance` leaves every test green. Can we adjust to use a fixture whose accepted fit lands between the absolute and the relative thresholds? A per-coordinate scale in the thousands should give this without exhausting the evaluation budget. We will want to assert on the achieved gradient against the applied bound, so the test fails when the bound goes back to absolute.

transformer_lens/tools/analysis/sparse_probing.py
```diff
@@ -0,0 +1,749 @@
...
+        labels: Boolean or integer binary labels shaped ``[example]``.
+        k: Number of coordinates selected by absolute train class-mean difference.
+        test_fraction: Requested held-out fraction within each class.
+        positive_label: Label defining the positive class and score sign.
+        preprocess: ``"none"`` or train-only ``"standardize"``.
+        class_weight: ``"balanced"`` or ``None`` for unweighted BCE.
+        l2_strength: Positive coefficient penalty in the logistic objective.
+        seed: Local CPU-generator seed used only for the stratified split.
+        max_iter: Maximum LBFGS iterations.
+        gradient_tolerance: Required final objective-gradient infinity norm.
```
jlarson4: Both public docstrings (here and line 652) say `gradient_tolerance` is the "required final objective-gradient infinity norm", but since the fix the accepted quantity is that tolerance times `max(1, initial gradient infinity norm)`, so the API reference now describes a bound the code does not apply while the guide describes the one it does. Please reword both to the scale-relative bound, in whatever form you settle on from my comment on line 385.

jlarson4 (commented):
I appreciate you turning these around so quickly! The detailed replies are also great, having your thought process speeds up my reviewing process tremendously. The k metadata fix and the two new behavioural tests look great. I have a couple new comments below, this should hopefully be the last batch for this feature

jlarson4:
Looks good, merging!

PR #1775, opened 2026-09-12
feat(svd_circuits): singular-direction readout, projection, and causal patch gate

### LINKED ISSUE #1767: [Proposal] SVD Circuits: singular-vector decomposition of a head's QK/ OV into causally-validated subfunctions

janmenjayap:
---

### Description

Adds the causal-validation layer on top of PR1's per-head QK/OV SVD (#1768): a direction can now be read out to vocab/logit space, projected against a real forward pass, and, critically, causally patched before any subfunction claim is accepted. Second of three PRs implementing per-head singular-vector decomposition of a head's QK (`W_Q W_K^T`) and OV (`W_V W_O`) maps (tracking issue: #1767).

- `vocab_readout(model, head_svd, k=10)` — projects the top-k OV output directions (`HeadSVD.V`) through the unembedding, gated on `TransformerBridge` compatibility mode.
- `logit_signature(model, head_svd, direction, tokens)` — signed logit effect of one OV direction's rank-1 reconstruction; requires the direction to pass `require_isolated` first, since a rotation-ambiguous or null direction's signature is not attributable to it alone.
- `project_activations(model, head_svd, prompt)` — per-position firing coefficients of a head's actual `hook_result` output against its `V` basis; summing the coefficients against `V` reconstructs the real output to numerical precision, since both read the same basis the head actually writes in.
- `patch_along_directions(model, head_svd, prompt, metric, keep=..., ablate=...)` — the mandatory causal gate. Reconstructs or ablates a head's output onto a chosen span of `V` directions via a direct `hook_result` hook, and reports `delta_metric` against an equally-sized random-subspace baseline (`gated` is true only when the requested subspace beats that baseline, not merely when the metric moves at all). Degenerate blocks reported by `HeadSVD.degenerate_blocks()` must be kept or dropped whole; a caller cannot route around `require_isolated` by hand-picking part of a rotation-ambiguous block through `keep`/`ablate`.
- Public exports: `decompose_head`, `project_activations`, `patch_along_directions`, `vocab_readout`, `logit_signature`, and their supporting types now ship from `transformer_lens.tools.analysis` — the first time this module is part of the public API, by design, so a readout was never importable without the causal gate attached to it.
- `tests/integration/test_svd_circuits.py` (new): exercises `vocab_readout` and `patch_along_directions` end to end against GPT-2 small's layer 9 head 9 (name-mover head) on an IOI-style prompt.

**Correctness fix folded in first:** `HeadSVD`'s docstring had the OV output direction backwards — `U`'s columns are the value-computation input space, `V`'s are the residual-stream output space this head writes into and the one to project through `W_U`. The swap never raised a shape error because both spaces are `d_model`-dimensional, so every downstream readout, projection, or patch would have silently used the wrong basis. Fixed first, with a regression test that cross-checks against the already-shipped `SVDInterpreter` (which projects the OV map the other way already), so the assertion cannot pass under either labeling.

Two implementation notes for anyone diffing this against the original proposal:
- `vocab_readout` does not wrap `SVDInterpreter.get_singular_vectors`. That method indexes `W_V`/`W_K` with a raw query-head index, which is wrong on grouped-query models (the same bug PR1 already fixed for this module's own weight reads). `vocab_readout` reuses PR1's already-kv-mapped `HeadSVD.V` and projects it through `W_U` directly instead of introducing a second, buggy SVD path.
- `patch_along_directions` calls `model.run_with_hooks` directly rather than `generic_activation_patch`. That helper is built for a clean-vs-corrupted two-run sweep over an index grid; this tool is a single-prompt reconstruct/ablate on a live projection, which does not fit the sweep shape.

Unit-tested with a tiny no-download `TransformerBridge` (real forward passes, no Hub access) for projection/patch behavior, plus model-free tests for the pure weight-space math. Integration test downloads `gpt2-small`.

PR3 (docs, demo notebook, slow multi-head oracle-parity check) follows this one.

Part of #1767

---

### Type of change

---

### Checklist:

- [ ] I have made corresponding changes to the documentation (will be done in PR3)

---

### DIFF
--- a/tests/integration/test_svd_circuits.py
+++ b/tests/integration/test_svd_circuits.py
@@ -1,0 +1,66 @@
+"""Integration test: OV vocab readout and the causal patch gate on a real GPT-2 head.
+
+The only file in the ``svd_circuits`` suite that downloads a pretrained model. Drives the
+whole read-then-patch path against layer 9 head 9 (the paper's canonical name-mover head) on
+a minimal IOI-style prompt, to check that the readout and the causal gate agree on a real
+Bridge rather than only on the unit suite's synthetic and tiny-model fixtures.
+"""
+
+from __future__ import annotations
+
+import math
+
+import pytest
+import torch
+
+from transformer_lens.model_bridge import TransformerBridge
+from transformer_lens.tools.analysis.svd_circuits import (
+    decompose_head,
+    patch_along_directions,
+    vocab_readout,
+)
+
+CLEAN_PROMPT = "When Mary and John went to the store, John gave a drink to"
+LAYER, HEAD = 9, 9
+
+
+@pytest.fixture(scope="module")
+def gpt2_bridge():
+    model = TransformerBridge.boot_transformers("gpt2", device="cpu", dtype=torch.float32)
+    model.enable_compatibility_mode()
+    return model
+
+
+def _logit_diff_metric(model):
+    mary_token = model.to_single_token(" Mary")
+    john_token = model.to_single_token(" John")
+
+    def metric(logits: torch.Tensor) -> float:
+        return float(logits[0, -1, mary_token] - logits[0, -1, john_token])
+
+    return metric
+
+
+def test_svd_circuits_readout_and_patch_gate_on_name_mover_head(gpt2_bridge) -> None:
+    decomposition = decompose_head(gpt2_bridge, layer=LAYER, head=HEAD, which=("OV",))
+    ov = decomposition.OV
+    assert ov is not None
+
+    readout = vocab_readout(gpt2_bridge, ov, k=10)
+    assert readout.shape == (gpt2_bridge.cfg.d_vocab, 10)
+    assert torch.isfinite(readout).all()
+
+    # Pick the first non-degenerate direction rather than assuming index 0 is isolated.
+    top_direction = next(row.idx for row in ov.rank_report if not row.is_degenerate)
+
+    prompt = gpt2_bridge.to_tokens(CLEAN_PROMPT)
+    metric = _logit_diff_metric(gpt2_bridge)
+
+    result = patch_along_directions(gpt2_bridge, ov, prompt, metric, keep=[top_direction])
+
+    assert result.retained == [top_direction]
+    assert math.isfinite(result.original_metric)
+    assert math.isfinite(result.patched_metric)
+    assert math.isfinite(result.delta_metric)
+    assert math.isfinite(result.baseline_delta_metric)
+    assert isinstance(result.gated, bool)
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -16,10 +16,13 @@ import torch
 from transformer_lens.tools.analysis.svd_circuits import (
     DegenerateDirectionError,
     HeadSVD,
+    LogitSignature,
     RankReportRow,
     _degeneracy_blocks,
     _factored_head_svd,
     decompose_head,
+    logit_signature,
+    vocab_readout,
 )
 
 D_MODEL, D_HEAD = 12, 4
@@ -529,3 +532,59 @@ def test_ov_output_direction_matches_svd_interpreter(tiny_bridge):
     projected_via_U = ov.U[:, 0] @ W_U
     assert not torch.allclose(projected_via_U, reference[:, 0, 0], atol=1e-2)
     assert not torch.allclose(projected_via_U, -reference[:, 0, 0], atol=1e-2)
+
+
+# --------------------------------------------------------------------------- #
+# vocab_readout / logit_signature (OV output-direction readout)
+# --------------------------------------------------------------------------- #
+def test_vocab_readout_shape_and_which_guard():
+    """vocab_readout returns [d_vocab, k] for OV, and rejects QK input and out-of-range k."""
+    ov = _factored_head_svd(*_random_ov(), which="OV", layer=0, head=0, eps=1e-2)
+    vocab_size = 20
+    model = SimpleNamespace(W_U=torch.randn(D_MODEL, vocab_size))
+
+    result = vocab_readout(model, ov, k=3)
+    assert result.shape == (vocab_size, 3)
+
+    qk = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="QK", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(ValueError, match="OV"):
+        vocab_readout(model, qk, k=1)
+
+    with pytest.raises(ValueError, match="k must be"):
+        vocab_readout(model, ov, k=D_HEAD + 1)
+
+
+def test_vocab_readout_raises_without_compatibility_mode(tiny_bridge):
+    """A TransformerBridge without compatibility mode enabled refuses to project through W_U."""
+    decomposition = decompose_head(tiny_bridge, layer=0, head=0, which=("OV",))
+    with pytest.raises(ValueError, match="enable_compatibility_mode"):
+        vocab_readout(tiny_bridge, decomposition.OV)
+
+
+def test_logit_signature_matches_manual_projection():
+    """logit_signature's reconstruction matches a hand-computed S[i] * V[:, i] projection."""
+    ov = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    vocab_size = 16
+    W_U = torch.randn(D_MODEL, vocab_size)
+    model = SimpleNamespace(W_U=W_U)
+    token_ids = torch.tensor([1, 5, 9])
+
+    result = logit_signature(model, ov, direction=0, tokens=token_ids)
+    assert isinstance(result, LogitSignature)
+    expected = (ov.S[0] * ov.V[:, 0]) @ W_U[:, token_ids]
+    assert torch.allclose(result.values, expected, atol=1e-5)
+    assert result.direction == 0
+
+
+def test_logit_signature_raises_on_degenerate_direction():
+    """A rotation-ambiguous direction's 'signature' is not attributable, so this must raise."""
+    ov = _factored_head_svd(
+        *_factored_with_spectrum([5.0, 3.0, 3.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    model = SimpleNamespace(W_U=torch.randn(D_MODEL, 8))
+    with pytest.raises(DegenerateDirectionError):
+        logit_signature(model, ov, direction=1, tokens=torch.tensor([0]))
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -1,11 +1,14 @@
 """Unit tests for per-head QK/OV singular-vector decomposition.
 
-Model-free: they build synthetic weight tensors and exercise the factored SVD and the
-degeneracy guard directly, so no model is loaded and no pretrained weights are downloaded.
+Mostly model-free: most tests build synthetic weight tensors and exercise the factored SVD
+and the degeneracy guard directly, so no model is loaded and no pretrained weights are
+downloaded. A few tests instantiate a tiny, randomly-initialized TransformerBridge (CPU-only,
+no Hub access) where a real per-head decomposition is needed for a cross-check.
 """
 
 import warnings
 from types import SimpleNamespace
+from unittest.mock import MagicMock
 
 import pytest
 import torch
@@ -470,3 +473,59 @@ def test_decompose_head_rejects_bare_string_which():
     model = _StubModel(n_heads=2, n_kv_heads=2)
     with pytest.raises(ValueError, match="sequence"):
         decompose_head(model, layer=0, head=0, which="QK")
+
+
+# --------------------------------------------------------------------------- #
+# OV output-direction convention (cross-check against SVDInterpreter)
+# --------------------------------------------------------------------------- #
+@pytest.fixture(scope="module")
+def tiny_bridge():
+    """Tiny GPT-2 bridge: MHA (n_heads == n_kv_heads), CPU-only, no Hub access."""
+    from transformers import GPT2Config, GPT2LMHeadModel
+
+    from transformer_lens.config.transformer_bridge_config import (
+        TransformerBridgeConfig,
+    )
+    from transformer_lens.model_bridge import TransformerBridge
+    from transformer_lens.model_bridge.supported_architectures.gpt2 import (
+        GPT2ArchitectureAdapter,
+    )
+
+    torch.manual_seed(0)
+    hf_model = GPT2LMHeadModel(
+        GPT2Config(n_embd=32, n_layer=1, n_head=2, vocab_size=64, n_positions=32)
+    ).eval()
+    cfg = TransformerBridgeConfig(
+        d_model=32,
+        d_head=16,
+        n_heads=2,
+        n_layers=1,
+        n_ctx=32,
+        d_vocab=64,
+        architecture="GPT2LMHeadModel",
+    )
+    return TransformerBridge(hf_model, GPT2ArchitectureAdapter(cfg), tokenizer=MagicMock())
+
+
+def test_ov_output_direction_matches_svd_interpreter(tiny_bridge):
+    """OV's write/vocab-readout direction is ``.V``, not ``.U`` - checked against the
+    already-shipped ``SVDInterpreter``, not just internal self-consistency, so the
+    assertion cannot pass under either U/V labeling by construction."""
+    from transformer_lens.SVDInterpreter import SVDInterpreter
+
+    layer, head = 0, 0
+    decomposition = decompose_head(tiny_bridge, layer, head, which=("OV",))
+    ov = decomposition.OV
+    W_U = tiny_bridge.W_U
+
+    interpreter = SVDInterpreter(tiny_bridge)
+    reference = interpreter.get_singular_vectors("OV", layer, head_index=head, num_vectors=1)
+
+    projected_via_V = ov.V[:, 0] @ W_U
+    assert torch.allclose(projected_via_V, reference[:, 0, 0], atol=1e-4) or torch.allclose(
+        projected_via_V, -reference[:, 0, 0], atol=1e-4
+    )
+
+    projected_via_U = ov.U[:, 0] @ W_U
+    assert not torch.allclose(projected_via_U, reference[:, 0, 0], atol=1e-2)
+    assert not torch.allclose(projected_via_U, -reference[:, 0, 0], atol=1e-2)
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -6,6 +6,7 @@ downloaded. A few tests instantiate a tiny, randomly-initialized TransformerBrid
 no Hub access) where a real per-head decomposition is needed for a cross-check.
 """
 
+import math
 import warnings
 from types import SimpleNamespace
 from unittest.mock import MagicMock
@@ -18,11 +19,13 @@ from transformer_lens.tools.analysis.svd_circuits import (
     DegenerateDirectionError,
     HeadSVD,
     LogitSignature,
+    PatchResult,
     RankReportRow,
     _degeneracy_blocks,
     _factored_head_svd,
     decompose_head,
     logit_signature,
+    patch_along_directions,
     project_activations,
     vocab_readout,
 )
@@ -641,6 +644,146 @@ def test_project_activations_restores_use_attn_result(tiny_bridge):
         for initial in (False, True):
             tiny_bridge.set_use_attn_result(initial)
             project_activations(tiny_bridge, decomposition.OV, torch.tensor([[5, 63, 7, 9]]))
+            assert tiny_bridge.cfg.use_attn_result == initial
+    finally:
+        tiny_bridge.set_use_attn_result(original)
+
+
+# --------------------------------------------------------------------------- #
+# patch_along_directions (mandatory causal gate)
+# --------------------------------------------------------------------------- #
+class _PatchStubModel:
+    """Model-free stand-in for patch_along_directions' run_with_hooks/metric protocol.
+
+    Exposes a single fixed per-head "hook_result" activation that a hook can intercept
+    exactly as ``run_with_hooks`` would present it (the hook receives the activation and
+    a hook object, and returns the replacement), then reduces it to a scalar the same
+    way for the unmodified, patched, and baseline calls, so the block guard and gating
+    arithmetic can be tested without a real forward pass.
+    """
+
+    def __init__(self, d_model, n_heads, pos=3, seed=0):
+        g = torch.Generator().manual_seed(seed)
+        self.cfg = SimpleNamespace(use_attn_result=False)
+        self._result = torch.randn(1, pos, n_heads, d_model, generator=g)
+        self._readout = torch.randn(d_model, generator=g)
+
+    def set_use_attn_result(self, value):
+        self.cfg.use_attn_result = value
+
+    def _logits(self, result):
+        return result.sum(dim=2) @ self._readout  # [batch, pos]
+
+    def __call__(self, prompt):
+        return self._logits(self._result)
+
+    def run_with_hooks(self, prompt, fwd_hooks):
+        activation = self._result
+        for name, hook_fn in fwd_hooks:
+            activation = hook_fn(activation, hook=SimpleNamespace(name=name))
+        return self._logits(activation)
+
+
+def test_patch_along_directions_which_guard():
+    """QK has no write direction to reconstruct onto, so a QK HeadSVD is refused."""
+    qk = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="QK", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(ValueError, match="OV"):
+        patch_along_directions(SimpleNamespace(), qk, "prompt", lambda logits: 0.0, keep=[0])
+
+
+def test_patch_along_directions_requires_keep_xor_ablate():
+    """Passing both keep and ablate, or neither, is refused before any model access."""
+    ov = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    metric = lambda logits: 0.0
+    with pytest.raises(ValueError, match="exactly one"):
+        patch_along_directions(SimpleNamespace(), ov, "prompt", metric, keep=[0], ablate=[1])
+    with pytest.raises(ValueError, match="exactly one"):
+        patch_along_directions(SimpleNamespace(), ov, "prompt", metric)
+
+
+def test_patch_along_directions_rejects_partial_degenerate_block():
+    """keep must take a degenerate block whole or not at all; a partial slice is refused,
+    and the whole block plus an isolated direction is accepted."""
+    ov = _factored_head_svd(
+        *_factored_with_spectrum([5.0, 3.0, 3.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    metric = lambda logits: 0.0
+    with pytest.raises(DegenerateDirectionError):
+        patch_along_directions(SimpleNamespace(), ov, "prompt", metric, keep=[0, 1])
+
+    stub = _PatchStubModel(d_model=D_MODEL, n_heads=1)
+    result = patch_along_directions(
+        stub, ov, "prompt", lambda logits: float(logits.sum()), keep=[0, 1, 2]
+    )
+    assert isinstance(result, PatchResult)
+    assert result.retained == [0, 1, 2]
+
+
+def test_patch_along_directions_reports_delta_and_restores_use_attn_result():
+    """delta_metric is patched minus original metric, gated follows the documented
+    comparison against baseline_delta_metric, and use_attn_result is restored after."""
+    ov = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    stub = _PatchStubModel(d_model=D_MODEL, n_heads=1)
+    metric = lambda logits: float(logits.sum())
+    for initial in (False, True):
+        stub.set_use_attn_result(initial)
+        result = patch_along_directions(
+            stub, ov, "prompt", metric, keep=[0], rng=torch.Generator().manual_seed(0)
+        )
+        assert result.delta_metric == pytest.approx(result.patched_metric - result.original_metric)
+        assert math.isfinite(result.baseline_delta_metric)
+        expected_gated = abs(result.delta_metric) > abs(result.baseline_delta_metric)
+        assert result.gated == expected_gated
+        assert stub.cfg.use_attn_result == initial
+
+
+def test_patch_along_directions_discriminates_causal_direction(tiny_bridge):
+    """Ablating the head's strongest OV direction should move a metric aligned with that
+    direction materially more than ablating its weakest, least load-bearing direction."""
+    layer, head = 0, 0
+    decomposition = decompose_head(tiny_bridge, layer, head, which=("OV",))
+    ov = decomposition.OV
+    strong = ov.rank_report[0].idx
+    weak = ov.rank_report[-1].idx
+    assert not ov.is_degenerate(strong)
+    assert not ov.is_degenerate(weak)
+
+    target_token = int((ov.V[:, strong] @ tiny_bridge.W_U).argmax())
+    prompt = torch.tensor([[5, 63, 7, 9]])
+
+    def metric(logits):
+        return logits[0, -1, target_token].item()
+
+    strong_result = patch_along_directions(
+        tiny_bridge, ov, prompt, metric, ablate=[strong], rng=torch.Generator().manual_seed(0)
+    )
+    weak_result = patch_along_directions(
+        tiny_bridge, ov, prompt, metric, ablate=[weak], rng=torch.Generator().manual_seed(0)
+    )
+    assert isinstance(strong_result, PatchResult)
+    assert math.isfinite(strong_result.baseline_delta_metric)
+    assert math.isfinite(weak_result.baseline_delta_metric)
+    assert abs(strong_result.delta_metric) > abs(weak_result.delta_metric)
+
+
+def test_patch_along_directions_restores_use_attn_result(tiny_bridge):
+    """use_attn_result is restored to its prior value regardless of what it started as."""
+    decomposition = decompose_head(tiny_bridge, 0, 0, which=("OV",))
+    ov = decomposition.OV
+    metric = lambda logits: float(logits.sum())
+    original = tiny_bridge.cfg.use_attn_result
+    try:
+        for initial in (False, True):
+            tiny_bridge.set_use_attn_result(initial)
+            patch_along_directions(
+                tiny_bridge, ov, torch.tensor([[5, 63, 7, 9]]), metric, keep=[ov.rank_report[0].idx]
+            )
             assert tiny_bridge.cfg.use_attn_result == initial
     finally:
         tiny_bridge.set_use_attn_result(original)
--- a/tests/unit/tools/test_svd_circuits.py
+++ b/tests/unit/tools/test_svd_circuits.py
@@ -14,6 +14,7 @@ import pytest
 import torch
 
 from transformer_lens.tools.analysis.svd_circuits import (
+    ActivationProjection,
     DegenerateDirectionError,
     HeadSVD,
     LogitSignature,
@@ -22,6 +23,7 @@ from transformer_lens.tools.analysis.svd_circuits import (
     _factored_head_svd,
     decompose_head,
     logit_signature,
+    project_activations,
     vocab_readout,
 )
 
@@ -507,7 +509,11 @@ def tiny_bridge():
         d_vocab=64,
         architecture="GPT2LMHeadModel",
     )
-    return TransformerBridge(hf_model, GPT2ArchitectureAdapter(cfg), tokenizer=MagicMock())
+    tokenizer = MagicMock()
+    # A real tokenizer isn't available (no-download fixture), but to_str_tokens's return
+    # value is jaxtyped-checked as List[str], so batch_decode must return actual strings.
+    tokenizer.batch_decode = lambda tokens_list, **kwargs: [str(ids[0]) for ids in tokens_list]
+    return TransformerBridge(hf_model, GPT2ArchitectureAdapter(cfg), tokenizer=tokenizer)
 
 
 def test_ov_output_direction_matches_svd_interpreter(tiny_bridge):
@@ -588,3 +594,53 @@ def test_logit_signature_raises_on_degenerate_direction():
     model = SimpleNamespace(W_U=torch.randn(D_MODEL, 8))
     with pytest.raises(DegenerateDirectionError):
         logit_signature(model, ov, direction=1, tokens=torch.tensor([0]))
+
+
+# --------------------------------------------------------------------------- #
+# project_activations (per-position firing coefficients in the OV output basis)
+# --------------------------------------------------------------------------- #
+def test_project_activations_which_guard():
+    """QK has no write direction to project onto, so a QK HeadSVD is refused."""
+    qk = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="QK", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(ValueError, match="OV"):
+        project_activations(SimpleNamespace(), qk, "hello world")
+
+
+def test_project_activations_reconstructs_head_output(tiny_bridge):
+    """Coefficients recovered against V reconstruct the actual cached hook_result slice."""
+    layer, head = 0, 0
+    decomposition = decompose_head(tiny_bridge, layer, head, which=("OV",))
+    # A raw token tensor, not a string: tiny_bridge's tokenizer is a MagicMock and cannot
+    # tokenize text, but a token-id tensor bypasses that path entirely.
+    prompt = torch.tensor([[5, 63, 7, 9]])
+
+    projection = project_activations(tiny_bridge, decomposition.OV, prompt)
+    assert isinstance(projection, ActivationProjection)
+    assert projection.head_svd is decomposition.OV
+
+    previous = tiny_bridge.cfg.use_attn_result
+    tiny_bridge.set_use_attn_result(True)
+    try:
+        _, cache = tiny_bridge.run_with_cache(prompt)
+    finally:
+        tiny_bridge.set_use_attn_result(previous)
+    expected = cache[("result", layer, "attn")][0, :, head, :]
+
+    reconstructed = projection.coefficients @ decomposition.OV.V.transpose(-2, -1)
+    assert torch.allclose(reconstructed, expected, atol=1e-4)
+    assert projection.str_tokens == tiny_bridge.to_str_tokens(prompt)
+
+
+def test_project_activations_restores_use_attn_result(tiny_bridge):
+    """use_attn_result is restored to its prior value regardless of what it started as."""
+    decomposition = decompose_head(tiny_bridge, 0, 0, which=("OV",))
+    original = tiny_bridge.cfg.use_attn_result
+    try:
+        for initial in (False, True):
+            tiny_bridge.set_use_attn_result(initial)
+            project_activations(tiny_bridge, decomposition.OV, torch.tensor([[5, 63, 7, 9]]))
+            assert tiny_bridge.cfg.use_attn_result == initial
+    finally:
+        tiny_bridge.set_use_attn_result(original)
--- a/transformer_lens/tools/analysis/svd_circuits.py
+++ b/transformer_lens/tools/analysis/svd_circuits.py
@@ -52,7 +52,7 @@ Example::
 """
 
 from dataclasses import dataclass
-from typing import List, Literal, Optional, Sequence, Tuple
+from typing import List, Literal, Optional, Sequence, Tuple, Union
 
 import torch
 from jaxtyping import Float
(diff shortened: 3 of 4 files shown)


transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
             W_V_h, W_O_h, which="OV", layer=layer, head=head, eps=eps, null_rtol=null_rtol
         )
     return HeadDecomposition(layer=layer, head=head, QK=qk, OV=ov)
+
+
+def _validate_bridge_compatibility(model) -> None:
+    """Reject a ``TransformerBridge`` whose ``W_U`` would give a silently wrong projection.
+
+    ``HookedTransformer`` always has the final LayerNorm folded into ``W_U``, so this
```
jlarson4, line 428: `HookedTransformer` is to be deprecated, please remove any references to it, and any additional code that supports only `HookedTransformer`.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+    # Lazy import - keeps the module importable without the bridge as a hard dependency.
+    from transformer_lens.model_bridge import TransformerBridge
+
+    if not isinstance(model, TransformerBridge):
+        return
+    if not getattr(model, "compatibility_mode", False):
+        raise ValueError(
+            "Projecting an OV direction through W_U on a TransformerBridge requires "
+            "compatibility mode, so that LayerNorm weights are folded into W_U. Call "
+            "`model.enable_compatibility_mode()` after loading the bridge, then retry."
```
jlarson4, line 443: `enable_compatibility_mode()` folds ln1 into `W_V` and centres `W_O` and `W_U`, so a `HeadSVD` decomposed before the call describes a different OV map (on gpt2 L9H9 the top singular value goes 23.3 → 8.4 and direction 1's top-10 tokens share nothing with a fresh decomposition), yet the guard inspects only the model, so retrying with the stale object returns silently, and `project_activations` and `patch_along_directions` have no guard at all. Record the compatibility-mode state on `HeadSVD` at decomposition time, refuse a mismatch in all four consumers, and point this message at re-running `decompose_head`.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+            ``(0, rank]``, or if ``model`` is a ``TransformerBridge`` without
+            compatibility mode enabled.
+    """
+    if head_svd.which != "OV":
+        raise ValueError(f"vocab_readout requires an OV HeadSVD, got which={head_svd.which!r}")
+    rank = head_svd.V.shape[1]
+    if not 0 < k <= rank:
+        raise ValueError(f"k must be in (0, {rank}], got {k!r}")
+    _validate_bridge_compatibility(model)
+    return model.W_U.T @ head_svd.V[:, :k].float()
```
jlarson4, line 481: `V` is cast to fp32 while `W_U` keeps the model dtype and matmul does not promote, so `vocab_readout` and `logit_signature` (line 531) raise `expected m1 and m2 to have the same dtype` on every bf16, fp16 and fp64 model, while the projection and patch paths cast to the activation's dtype and work at all four. Cast both operands to a common dtype in both functions and add a dtype-parametrised test.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+            (batch-size-1) prompt.
+    """
+    if head_svd.which != "OV":
+        raise ValueError(
+            f"project_activations requires an OV HeadSVD, got which={head_svd.which!r}"
+        )
+    previous = getattr(model.cfg, "use_attn_result", False)
+    model.set_use_attn_result(True)
+    try:
+        _, cache = model.run_with_cache(prompt)
```
jlarson4: `run_with_cache(prompt)` retains every hook point to read one, about 1 GB against 19 MB on a 512-token gpt2-small prompt, and the batch guard at line 592 only runs after that forward. Pass `names_filter` for the one `hook_result` name and check a tensor prompt's batch dimension before the forward, keeping the post-cache check that catches a list of strings.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+    finally:
+        model.set_use_attn_result(previous)
+    result = cache[("result", head_svd.layer, "attn")][..., head_svd.head, :]
+    if result.shape[0] != 1:
+        raise ValueError(
+            f"project_activations requires a single prompt, got batch={result.shape[0]}"
+        )
+    result = result.squeeze(0).to(head_svd.V.dtype)
+    coefficients = result @ head_svd.V
+    str_tokens = model.to_str_tokens(prompt)
```
jlarson4: `str_tokens` is computed unconditionally after the forward pass and the projection, so a bridge built without a tokenizer, the from-scratch case the unit fixture has to fake with a `MagicMock`, dies here with `AttributeError: 'NoneType' object has no attribute 'batch_decode'` and cannot use `project_activations` at all. Compute `str_tokens` only when the model has a tokenizer and document the field as empty otherwise.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+
+    Exactly one of ``keep``/``ablate`` must be given; ``ablate``'s complement over the
+    map's full rank becomes the retained set. Raises :class:`DegenerateDirectionError`
+    if the result would split a degenerate block (see :func:`_validate_retained_blocks`).
+    """
+    if (keep is None) == (ablate is None):
+        raise ValueError("patch_along_directions requires exactly one of keep or ablate")
+    rank = head_svd.V.shape[1]
+    if keep is not None:
+        retained = sorted(set(keep))
```
jlarson4, line 636: `keep`, `ablate` and `logit_signature`'s `direction` (line 528) go straight into `V[:, ...]` and the `range(rank)` complement, so `ablate=[-1]` or `ablate=[999]` subtracts nothing and reports delta ≈ 0, gated False for an intervention that never ran (gpt2 L9H9: −9.5e-07 against −3.2e-02 for the last isolated direction), `keep=[-1]` passes the degenerate-block guard that `keep=[rank-1]` trips, `keep=[999]` raises a raw `IndexError`, and `keep=[]` ties the baseline by construction. Validate every index against `[0, rank)` the way `decompose_head` checks `layer` and `head`, coerce to `int`, and refuse an empty retained set unless an explicit `threshold` is given.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+    """Build a ``hook_result`` hook that reconstructs one head's output onto ``span(projector)``.
+
+    Leaves every other head's slice of the ``[batch, pos, head_index, d_model]`` tensor
+    untouched. Clones before mutating so the hook never writes into the activation the
+    forward pass itself is still using.
+    """
+
+    def hook_fn(activation: torch.Tensor, hook) -> torch.Tensor:
+        activation = activation.clone()
+        activation[:, :, head, :] = activation[:, :, head, :] @ projector.to(activation.dtype)
```
jlarson4, line 655: `baseline_projector` comes from `torch.randn(..., generator=rng)` with no device, so it is a CPU tensor while `kept_projector` sits on the model's device, and since this line reconciles dtype only the third forward pass raises `Tensor for argument #2 'mat2' is on CPU, but expected it to be on GPU` on any MPS or CUDA model (the draw has to stay on CPU: QR is unimplemented on MPS and a CUDA generator cannot feed a CPU `randn`). Move the finished projector to `activation.device` here alongside the dtype cast and add a device-skipif test that the call completes.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+def patch_along_directions(
+    model,
+    head_svd: HeadSVD,
+    prompt: Union[str, torch.Tensor],
+    metric: Callable[[torch.Tensor], float],
+    *,
+    keep: Optional[Sequence[int]] = None,
+    ablate: Optional[Sequence[int]] = None,
+    threshold: Optional[float] = None,
+    rng: Optional[torch.Generator] = None,
```
jlarson4, line 699: `gated` is thresholded on a single baseline draw from the global RNG, so identical calls disagree: twelve back-to-back `ablate=[top]` calls on the unit suite's tiny bridge gave twelve `baseline_delta_metric` values between 0.003 and 0.022 and `gated` True on two of them with `delta_metric` fixed, where `estimate_occupancy` in this package seeds by default and averages its control draws. Default `rng` to a seeded generator and average the baseline over several draws, documenting `baseline_delta_metric` as that average.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+            f"patch_along_directions requires an OV HeadSVD, got which={head_svd.which!r}"
+        )
+    retained = _resolve_retained(head_svd, keep, ablate)
+
+    V = head_svd.V
+    kept_projector = V[:, retained] @ V[:, retained].transpose(-2, -1)
+
+    d_model = V.shape[0]
+    width = len(retained)
+    random_input = torch.randn(d_model, width, generator=rng, dtype=V.dtype)
```
jlarson4, line 744: `randn(d_model, width)` draws the control subspace from the full residual stream, but the head's `hook_result` lies entirely inside `span(V)` (64 of 768 dimensions on gpt2), so the baseline retains about `width/d_model` of the head's output instead of `width/rank`: at width 1 it keeps under 0.1% against 3.9% for the requested direction and its delta equals the full-ablation delta, at width 63 it still discards over 90% of the head with a delta swinging from −0.06 to +0.32 across seeds, and on gpt2 L9H9 `ablate=[top]` gates False on every seed for that reason alone. Draw the baseline inside `span(V)`, for example `V @ Q[:, :width]` with `Q` a random orthogonal `rank × rank` matrix, and reword the `gated` and function docstrings to match.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+            fwd_hooks=[(hook_name, _make_subspace_hook(head_svd.head, baseline_projector))],
+        )
+        baseline_metric = float(metric(baseline_logits))
+    finally:
+        model.set_use_attn_result(previous)
+
+    delta_metric = patched_metric - original_metric
+    baseline_delta_metric = baseline_metric - original_metric
+    gate_threshold = abs(baseline_delta_metric) if threshold is None else threshold
+    gated = abs(delta_metric) > gate_threshold
```
jlarson4, line 768: After `_resolve_retained`, `keep=S` and `ablate=complement(S)` are the same experiment, so this single "moved more than the baseline" test can only answer the ablate question and answers the keep question with the sign reversed: keeping all 64 directions (exact reconstruction, delta −1e-06) gates False, `keep=[weakest isolated]` gates True exactly like `keep=[top]`, and `keep=[]` gates True under any explicit threshold, so the integration test's `keep=[top_direction]` call is not the check the PR describes. Make the success condition depend on the mode, `abs(delta) < threshold` for `keep` (the reconstruction preserves the metric better than the same-width baseline) and `abs(delta) > threshold` for `ablate`, and document both meanings on `gated`.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -406,3 +420,359 @@ def decompose_head(
...
+    V = head_svd.V
+    kept_projector = V[:, retained] @ V[:, retained].transpose(-2, -1)
+
+    d_model = V.shape[0]
+    width = len(retained)
+    random_input = torch.randn(d_model, width, generator=rng, dtype=V.dtype)
+    random_basis, _ = torch.linalg.qr(random_input)
+    baseline_projector = random_basis @ random_basis.transpose(-2, -1)
+
+    hook_name = f"blocks.{head_svd.layer}.attn.hook_result"
```
jlarson4: Nothing checks `head_svd.layer`, `head` or `d_model` against the model it is used with: `project_activations` (line 591) fails with a raw `KeyError` or `IndexError`, and here an out-of-range layer builds a hook name that matches nothing, `TransformerBridge.run_with_hooks` drops unknown names silently, and the call returns delta 0, baseline 0, gated False with no error. Validate the decomposition against `model.cfg` before the forward in both functions and fail if the hook never fired.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +481,309 @@ def test_decompose_head_rejects_bare_string_which():
...
+        architecture="GPT2LMHeadModel",
+    )
+    tokenizer = MagicMock()
+    # A real tokenizer isn't available (no-download fixture), but to_str_tokens's return
+    # value is jaxtyped-checked as List[str], so batch_decode must return actual strings.
+    tokenizer.batch_decode = lambda tokens_list, **kwargs: [str(ids[0]) for ids in tokens_list]
+    return TransformerBridge(hf_model, GPT2ArchitectureAdapter(cfg), tokenizer=tokenizer)
+
+
+def test_ov_output_direction_matches_svd_interpreter(tiny_bridge):
```
jlarson4, line 522: This test projects `ov.V[:, 0] @ W_U` by hand and `vocab_readout`'s own test checks shape only, so the U-for-V swap this PR's docstring fix exists to prevent passes every test when made in `vocab_readout` (line 481), as does returning zeros, and the same swap in the patch projector (line 739) passes too because it only collapses the strong/weak ratio in `test_patch_along_directions_discriminates_causal_direction` from 8.7× to 1.07× against a bare `>`. Compare `vocab_readout(model, ov, k=1)[:, 0]` against the `SVDInterpreter` reference here and give the discriminates test a 5× margin or a `keep=all` no-op assertion.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +481,309 @@ def test_decompose_head_rejects_bare_string_which():
...
+    qk = _factored_head_svd(
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="QK", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(ValueError, match="OV"):
+        project_activations(SimpleNamespace(), qk, "hello world")
+
+
+def test_project_activations_reconstructs_head_output(tiny_bridge):
+    """Coefficients recovered against V reconstruct the actual cached hook_result slice."""
+    layer, head = 0, 0
```
jlarson4, line 616: Every projection and patch test decomposes head 0 (the stub has one head), so hardcoding the head index in `project_activations` (line 591) or the patch hook (line 655), or writing the projector to every head, passes the whole suite, and the hook's "leaves every other head untouched" promise is asserted nowhere. Run the reconstruction and patch tests on head 1 of the tiny bridge as well and assert the sibling head's `hook_result` slice is unchanged.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +481,309 @@ def test_decompose_head_rejects_bare_string_which():
...
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="OV", layer=0, head=0, eps=1e-2
+    )
+    metric = lambda logits: 0.0
+    with pytest.raises(ValueError, match="exactly one"):
+        patch_along_directions(SimpleNamespace(), ov, "prompt", metric, keep=[0], ablate=[1])
+    with pytest.raises(ValueError, match="exactly one"):
+        patch_along_directions(SimpleNamespace(), ov, "prompt", metric)
+
+
+def test_patch_along_directions_rejects_partial_degenerate_block():
```
jlarson4, line 708: Both calls here go through `keep` and no `ablate` call anywhere asserts `retained`, so guarding `_validate_retained_blocks` with `if keep is not None` passes the suite, as does treating `ablate` as `keep`, since the discriminates test only orders |delta| for strong against weak. Add `ablate=[1]` on the `[5, 3, 3, 1]` spectrum expecting `DegenerateDirectionError` and assert `retained` on an `ablate` call.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +481,309 @@ def test_decompose_head_rejects_bare_string_which():
...
+    stub = _PatchStubModel(d_model=D_MODEL, n_heads=1)
+    metric = lambda logits: float(logits.sum())
+    for initial in (False, True):
+        stub.set_use_attn_result(initial)
+        result = patch_along_directions(
+            stub, ov, "prompt", metric, keep=[0], rng=torch.Generator().manual_seed(0)
+        )
+        assert result.delta_metric == pytest.approx(result.patched_metric - result.original_metric)
+        assert math.isfinite(result.baseline_delta_metric)
+        expected_gated = abs(result.delta_metric) > abs(result.baseline_delta_metric)
```
jlarson4, line 741: `expected_gated` is recomputed from the same result's fields, so it holds for any self-consistent implementation: `gated = True` hardcoded, the baseline comparison dropped, `threshold` ignored, `rng` ignored and the QR removed all pass, and no test passes `threshold=` or compares two seeded runs. Assert outcomes with a seeded generator instead, `ablate=[weakest isolated]` gating False, a `threshold` above `abs(delta_metric)` gating False, and the same seed reproducing `baseline_delta_metric`, updated to whatever the gate's semantics become.

tests/integration/test_svd_circuits.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Pick the first non-degenerate direction rather than assuming index 0 is isolated.
+    top_direction = next(row.idx for row in ov.rank_report if not row.is_degenerate)
+
+    prompt = gpt2_bridge.to_tokens(CLEAN_PROMPT)
+    metric = _logit_diff_metric(gpt2_bridge)
+
+    result = patch_along_directions(gpt2_bridge, ov, prompt, metric, keep=[top_direction])
+
+    assert result.retained == [top_direction]
```
jlarson4, line 61: Every assertion is shape, finiteness, `isinstance`, or `retained` echoing the `keep=` argument, so the only real-model test passes with the hook never installed, the gate hardcoded, `vocab_readout` returning zeros, or the wrong basis projected. Assert that `readout[:, i]` matches `SVDInterpreter(model).get_singular_vectors("OV", 9, head_index=9)` up to sign, that `keep=range(rank)` leaves the logit diff within 1e-4, and that ablating every direction moves it by more than 1e-2.

transformer_lens/tools/analysis/svd_circuits.py
jlarson4, line 588: `run_with_cache(prompt)` retains every hook point to read one, about 1 GB against 19 MB on a 512-token gpt2-small prompt, and the batch guard at line 592 only runs after that forward. Pass `names_filter` for the one `hook_result` name and check a tensor prompt's batch dimension before the forward, keeping the post-cache check that catches a list of strings.

jlarson4 (commented):
Thanks for the U/V correction and for cross-checking it against SVDInterpreter rather than the module's own arithmetic, that is the correct test point. Let me know if you have any questions on the below comments!

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+            baseline_logits = model.run_with_hooks(
+                prompt,
+                fwd_hooks=[(hook_name, _make_subspace_hook(head_svd.head, baseline_projector))],
+            )
+            baseline_deltas.append(float(metric(baseline_logits)) - original_metric)
+    finally:
+        model.set_use_attn_result(previous)
+
+    delta_metric = patched_metric - original_metric
+    baseline_delta_metric = sum(baseline_deltas) / len(baseline_deltas)
```
jlarson4: Averaging the control draws was my ask last round, but this averages them signed and then takes the absolute value, which is never larger than the mean of their magnitudes and is far smaller wherever the controls mix sign. At those widths the threshold understates the typical control effect and shrinks further as `n_baseline` grows, so ablate passes too easily and keep too rarely; on gpt2 L9H9 the default `ablate=[0]` gates True on a delta several times smaller than an arbitrary same-width subspace produces. Threshold on the mean of the per-draw magnitudes, and say whether `baseline_delta_metric` stays signed or becomes the magnitude, since the `abs()` below and the docstring at line 775 depend on that.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+            f"patch_along_directions requires an OV HeadSVD, got which={head_svd.which!r}"
+        )
+    if n_baseline < 1:
+        raise ValueError(f"n_baseline must be at least 1, got {n_baseline}")
+    _validate_decomposition_matches_model(model, head_svd)
+    retained = _resolve_retained(head_svd, keep, ablate)
+    # _resolve_retained has confirmed exactly one of keep/ablate is set, so the mode the
+    # caller expressed is unambiguous and selects the gate's success condition (see below).
+    mode = "keep" if keep is not None else "ablate"
+    if not retained and threshold is None:
```
jlarson4: `width == rank` has the tie this guard refuses for the empty set: the control projector is then `V Vᵀ`, the kept projector, so both runs apply the same projection and `gated` resolves on floating-point residue. It is reachable, since a full-rank keep passes the block guard, and it is the call the integration test makes at line 82. Extend this condition to a retained set that spans the full rank, give that line-82 call a `threshold`, and reword the message, which says the set is empty.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+            null_rtol=null_rtol,
+            compatibility_mode=compatibility_mode,
         )
     return HeadDecomposition(layer=layer, head=head, QK=qk, OV=ov)
+
+
+def _validate_bridge_compatibility(model) -> None:
+    """Reject a ``TransformerBridge`` whose ``W_U`` would give a silently wrong projection.
+
+    Projecting an OV direction through ``W_U`` requires the final LayerNorm folded
```
jlarson4, line 459: The replacement sentence is also untrue: the bridge sets `compatibility_mode = True` before it processes weights and skips the processing entirely under `no_processing`, so the flag can be True with `W_U` unfolded and this guard passes. Check the recorded folding state, `_weights_processed` and the adapter's `_fold_ln_requested`, which the sibling guards already read, instead of the flag; while here, name `direct_logit_attribution` as the one tool this mirrors and fix the staleness guard's docstring below, which says this function checks the folding.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+
+    Args:
+        model: A ``TransformerBridge``.
+        head_svd: An OV :class:`HeadSVD` from :func:`decompose_head`.
+        prompt: A single prompt (not a batch): a string or a ``[1, pos]`` token tensor.
+
+    Returns:
+        An :class:`ActivationProjection` with the per-position coefficients.
+
+    Raises:
```
jlarson4, line 657: `set_use_attn_result(True)` raises `NotImplementedError` on adapters without per-head results, and this function and `patch_along_directions` (line 852) both call it, but neither `Raises` block says so. Add it to both.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+) -> PatchResult:
+    """Causally validate a claimed OV subfunction by reconstructing the head's output onto it.
+
+    Requires ``head_svd.which == "OV"``: this reconstructs the write/output basis
+    ``V``, and QK has no such vector (see the module docstring). Runs the prompt with
+    ``use_attn_result`` enabled once unmodified, once with the head's ``hook_result``
+    slice reconstructed onto ``span(head_svd.V[:, retained])``, and once per random
+    control subspace. Each control subspace is drawn *inside the head's own OV span*
+    ``span(V)`` (not from the full residual stream, where a width-``w`` random subspace
+    would keep only ``w/d_model`` of a head output that lives entirely in ``w/rank`` of
```
jlarson4: The head output lives in `span(V)`, which is `rank/d_model` of the stream; `w/rank` is what an in-span control of width `w` retains of it. Reword.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -125,6 +153,7 @@ class HeadSVD:
     rank_report: List[RankReportRow]
     eps: float
     null_rtol: float
+    compatibility_mode: bool
```
jlarson4: Can we default this to `False`? `decompose_head` passes it explicitly and the test helper already defaults it.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,503 @@ def decompose_head(
...
+    if head_svd.which != "OV":
+        raise ValueError(
+            f"project_activations requires an OV HeadSVD, got which={head_svd.which!r}"
+        )
+    _validate_decomposition_matches_model(model, head_svd)
+    # A batched token tensor carries its batch dimension up front, so reject it before the
+    # forward instead of running the model on every row only to discard the result. A list of
+    # prompt strings only reveals its batch size once tokenized, so the post-cache check below
+    # still guards that path.
+    if isinstance(prompt, torch.Tensor) and (prompt.ndim != 2 or prompt.shape[0] != 1):
```
jlarson4: Requiring exactly two dimensions also rejects a 1-D `[pos]` tensor, which the previous revision accepted with identical results and which `patch_along_directions` still accepts. Reject only a tensor whose leading dimension is larger than one.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +483,744 @@ def test_decompose_head_rejects_bare_string_which():
...
+        assert math.isfinite(result.baseline_delta_metric)
+        assert isinstance(result.gated, bool)
+        assert stub.cfg.use_attn_result == initial
+
+
+def _span_aligned_stub(ov, weights):
+    """A patch stub whose single head writes ``ov.V @ weights`` at every position and whose
+    readout is that same in-span vector.
+
+    Because the head output lies entirely in ``span(ov.V)`` and the readout equals it, removing
```
jlarson4, line 1059: Because the readout equals the head output here, every control delta has the same sign, so the absolute signed mean equals the mean of magnitudes and no gate test on this stub can tell which one the gate uses; the same geometry lets a full-residual-stream control and a single draw pass every test. Add an assertion against the stub's analytic Haar expectation, `E[baseline_delta] = (width/rank − 1)·‖h‖²·pos`, with enough draws to separate in-span from full-stream, and one that threading a generator through single-draw calls reproduces the multi-draw average bit for bit.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +483,744 @@ def test_decompose_head_rejects_bare_string_which():
...
+    )
+    assert isinstance(result, PatchResult)
+    assert math.isfinite(result.delta_metric)
+    assert math.isfinite(result.baseline_delta_metric)
+
+
+def test_patch_along_directions_discriminates_causal_direction(tiny_bridge):
+    """Ablating the head's strongest OV direction should move a metric aligned with that
+    direction materially more than ablating its weakest, least load-bearing direction."""
+    layer, head = 0, 0
```
jlarson4: The sibling-head test calls `_make_subspace_hook` directly and every `patch_along_directions` test in this file uses head 0, so the control-draw call inside `patch_along_directions` can be hardcoded to head 0 without failing anything; the kept-projector call is caught by the integration test. Run this test on head 1 as well, with a check on `baseline_delta_metric` that a control drawn for the wrong head cannot pass.

jlarson4 (commented):
Thanks for working through all those comments. The compatibility-state binding and the dtype and device fixes are excellent. An additional batch of comments added below:

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,526 @@ def decompose_head(
...
+    projecting a rank-1 OV direction through ``W_U`` does not depend on the block-layout
+    assumptions that check exists for elsewhere.
+    """
+    # Lazy import - keeps the module importable without the bridge as a hard dependency.
+    from transformer_lens.model_bridge import TransformerBridge
+
+    if not isinstance(model, TransformerBridge):
+        return
+    processed = getattr(model, "_weights_processed", False)
+    fold_ln_requested = getattr(getattr(model, "adapter", None), "_fold_ln_requested", True)
```
jlarson4, line 477: Thirty-five adapters in this tree declare no fold support, OLMo 2, Phi-3, Qwen3, OPT, Nemotron, DeepSeek, GLM and Gemma among them, so on those `enable_compatibility_mode()` leaves the final LayerNorm in place, this check passes, and `vocab_readout` projects through an unfolded `W_U`. I pointed you in the wrong direction for this fix, apologies. Can you please add `supports_fold_ln` to the conjunction?

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,526 @@ def decompose_head(
...
+    ``enable_compatibility_mode`` folds ``ln1`` into ``W_V`` and centres ``W_O``/``W_U``,
+    so a :class:`HeadSVD` decomposed before the call describes a different OV map than the
+    model now computes: the cached ``U``/``S``/``V`` are stale, and the returned readout or
+    patch would be silently wrong rather than raise a shape error. This guard is orthogonal
+    to :func:`_validate_bridge_compatibility` (which checks that the bridge actually folded
+    LayerNorm into ``W_U`` via its recorded processing state): here the model may be in
+    either state, only mismatched from the decomposition's.
+    """
+    current = getattr(model, "compatibility_mode", False)
+    if current != head_svd.compatibility_mode:
```
jlarson4, line 500: Now that the fold state and `compatibility_mode` are decoupled, comparing the flags no longer detects a decomposition whose weights have moved. An example run following the advice would work as follows: 

`enable_compatibility_mode(no_processing=True)`, decompose, get the refusal, call `enable_compatibility_mode()`, retry — folds `W_V` and shifts the spectrum while both flags stay `True`, so the stale decomposition sails through and the readout is silently wrong. Record the same folded-state predicate on the `HeadSVD` that the readout guard now reads, and compare that instead of the flag; the field is new this round so repurposing it breaks nothing. The two public docstrings this guard backs need the same pass: `vocab_readout` and `logit_signature` still say "with compatibility mode enabled" and promise a `ValueError` for a bridge "without compatibility mode enabled", and both are now false.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,526 @@ def decompose_head(
...
+        raise ValueError(
+            f"project_activations requires an OV HeadSVD, got which={head_svd.which!r}"
+        )
+    _validate_decomposition_matches_model(model, head_svd)
+    # A batched token tensor carries a leading batch dimension larger than one, so reject it
+    # before the forward instead of running the model on every row only to discard the result.
+    # A 1-D [pos] tensor has no batch axis and runs as a single prompt, matching what
+    # patch_along_directions accepts. A list of prompt strings only reveals its batch size once
+    # tokenized, so the post-cache check below still guards that path.
+    if isinstance(prompt, torch.Tensor) and prompt.ndim >= 2 and prompt.shape[0] > 1:
```
jlarson4, line 674: Accepting a 1-D `[pos]` tensor was the ask, but replacing the old rank test with a leading-dimension test also lets a `[1, 1, pos]` tensor past the pre-forward guard, and makes a 0-D `torch.tensor(5)` raise `IndexError: tuple index out of range` where the `Raises` block promises `ValueError`. Keep the rank check alongside the batch check, `prompt.ndim not in (1, 2) or (prompt.ndim == 2 and prompt.shape[0] > 1)`, so both shapes are refused here rather than after the forward.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -62,6 +71,15 @@
...
 _SIGMA_FLOOR = 1e-12
 
+# Default seed for the causal gate's random baseline, so a bare patch_along_directions
+# call is reproducible instead of drawing from the global RNG. Matches estimate_occupancy,
+# which seeds its random controls by default.
+_DEFAULT_BASELINE_SEED = 0
+
+# Number of random in-span control subspaces the gate averages its baseline delta over,
+# so one lucky or unlucky draw does not decide the gate.
+_DEFAULT_N_BASELINE = 8
```
jlarson4, line 81: The threshold is now a mean of magnitudes, so its variance decides borderline verdicts, and the per-draw magnitudes are heavy-tailed on a real head (median 0.016 against a mean of 0.032 on gpt2 L9H9). At eight draws the threshold's 5–95% band spans 4.26× and 27 of that head's 64 directions gate differently depending on the seed; at 32 draws the band is 1.99× and 13 do. Raise the default or say in the `n_baseline` docs that borderline directions are seed-sensitive at the default, whichever you prefer given each draw costs a forward pass.

transformer_lens/tools/analysis/svd_circuits.py
```diff
@@ -400,9 +437,526 @@ def decompose_head(
...
+            exact basis the head writes in.
+        str_tokens: Tokenized prompt, aligned with the position axis, for display.
+    """
+
+    head_svd: HeadSVD
+    coefficients: Float[torch.Tensor, "pos rank"]
+    str_tokens: List[str]
+
+
+def project_activations(
```
jlarson4, line 635: The inclusion of the latest `dev` code brings `require_eval_mode` into this branch, and both other forward-pass analysis tools in this directory call it on entry, but this function and `patch_along_directions` at line 817 do not, so on a model left in train mode dropout perturbs each of the `n_baseline + 2` passes independently and the control draws stop being comparable to the patched run, which makes `gated` noise rather than a verdict. 

Please call `require_eval_mode(model, operation="project_activations()")` here and the equivalent in `patch_along_directions`, leaving the two weight-space readouts unguarded since they run no forward pass.

tests/unit/tools/test_svd_circuits.py
```diff
@@ -470,3 +485,904 @@ def test_decompose_head_rejects_bare_string_which():
...
+        *_factored_with_spectrum([8.0, 4.0, 2.0, 1.0]), which="QK", layer=0, head=0, eps=1e-2
+    )
+    with pytest.raises(ValueError, match="OV"):
+        vocab_readout(model, qk, k=1)
+
+    with pytest.raises(ValueError, match="k must be"):
+        vocab_readout(model, ov, k=D_HEAD + 1)
+
+
+def test_vocab_readout_raises_without_compatibility_mode(tiny_bridge):
```
jlarson4, line 646: Reverting the folded-state check to the old `compatibility_mode` read leaves all 94 tests passing, because this is the only test that touches the guard and it uses a bridge where the flag and the fold state are both false. Its name and docstring also still describe the flag contract. Add cases on this same no-download bridge for the states that separate them: `enable_compatibility_mode(no_processing=True)` and `enable_compatibility_mode(fold_ln=False)` must raise, and `process_weights(fold_ln=True)` without the flag must not.

jlarson4 (commented):
Thanks for the quick updates @janmenjayap! I am hopeful this will be my last review pass on this, sorry for the multitude of review requests. It is a big feature, and we are very close to shipping it!

jlarson4:
Great work on getting this tackled @janmenjayap, this is approved and I am merging it now

PR #1780, opened 2026-09-13
Reject fabricated MoE weights in SVDInterpreter

### LINKED ISSUE #1779: [Bug Report] SVDInterpreter returns plausible non-zero MLP directions from fabricated zero weights on MoE models
**Describe the bug**

`SVDInterpreter` returns plausible, non-zero MLP singular directions for sparse-MoE layers even though those results do not depend on the model's expert weights.

`get_bridge_params()` cannot extract a single dense `W_in` / `W_out` from `MoEBridge`, because a sparse-MoE layer has multiple expert-specific projections. It currently substitutes dense zero tensors for those missing weights. `SVDInterpreter.get_singular_vectors()` then runs SVD on the fabricated zero matrix and projects an arbitrary basis through `unembed.W_U`, producing a normal-looking non-zero result.

The warning emitted while constructing the TL-style parameter dictionary does not prevent downstream consumers from treating the returned tensor as a real interpretation.

**Code example**

This reproduces without downloading a checkpoint, using a seeded tiny `MixtralConfig`:

```python
import copy
import torch
from transformers import AutoModelForCausalLM, MixtralConfig

from transformer_lens import SVDInterpreter
from transformer_lens.model_bridge.sources._bridge_builder import build_bridge_from_module

cfg = MixtralConfig(
    vocab_size=128,
    hidden_size=64,
    intermediate_size=128,
    num_hidden_layers=2,
    num_attention_heads=4,
    num_key_value_heads=2,
    num_local_experts=4,
    num_experts_per_tok=2,
    max_position_embeddings=128,
    pad_token_id=0,
    bos_token_id=1,
    eos_token_id=2,
)
cfg._attn_implementation = "eager"

torch.manual_seed(42)
hf = AutoModelForCausalLM.from_config(cfg).eval()
    hf,
    "MixtralForCausalLM",
    hf_config=copy.deepcopy(cfg),
    tokenizer=None,
    device="cpu",
).eval()

params = bridge.tl_parameters()
result_before = SVDInterpreter(bridge).get_singular_vectors(
    "w_in", layer_index=0, num_vectors=4
)

with torch.no_grad():
    for name, parameter in bridge.original_model.named_parameters():
        if ".experts." in name:
            parameter.zero_()

result_after = SVDInterpreter(bridge).get_singular_vectors(
    "w_in", layer_index=0, num_vectors=4
)

print(torch.count_nonzero(params["blocks.0.mlp.W_in"]))  # tensor(0)
print(torch.count_nonzero(result_before))                 # tensor(512)
print(result_before.abs().sum())                          # tensor(8.4830)
print(torch.equal(result_before, result_after))           # True
```

Observed on current `dev` (`be40a9b4`):

```text
W_in shape: (64, 128), non-zero elements: 0
W_out shape: (128, 64), non-zero elements: 0
SVD output shape: (128, 1, 4), non-zero elements: 512
SVD output absolute sum: 8.482950210571289
SVD output unchanged after zeroing all real expert tensors: True
```

This means a user can receive a credible-looking scientific output that is completely independent of the expert weights it purports to analyze.

**Suggested scope**

For the safe short-term fix, I suggest that `get_bridge_params()` stop representing a real but structurally unrepresentable sparse-MoE MLP with fabricated dense zero tensors. `SVDInterpreter` should then fail clearly before `torch.linalg.svd`, identifying the unsupported layer and explaining that sparse MoE requires an expert-aware interpretation.

A real MoE implementation would need an explicit contract such as an `expert_index`, or a separately defined router-weighted quantity. Silently selecting or averaging experts would make a different scientific claim, so I do not think that belongs in the minimal bug fix.

Regression coverage should verify that:

- sparse Mixtral-style MoE fails before SVD;
- dense TransformerBridge and HookedTransformer MLP behavior remains unchanged;
- no unsupported real component is exposed as a fabricated zero tensor.

**System Info**

- OS: Windows 11
- Python: 3.12.10
- PyTorch: 2.10.0+cpu
- Transformers: 5.13.0
- TransformerLens base: `dev` at `be40a9b4`

**Additional context**

I searched open and closed Issues and PRs for combinations of `SVDInterpreter`, sparse MoE, Mixtral, zero MLP weights, and weight analysis, and did not find a matching report. Proposal #1767 concerns attention-head QK/OV SVD and does not cover sparse-MoE MLP parameters.

### Checklist

emerardd:
# Description

Sparse-MoE layers do not have a single dense `W_in` or `W_out`, but `get_bridge_params()` previously represented these structurally unavailable weights with zero tensors. `SVDInterpreter` then decomposed those zero matrices and projected an arbitrary singular basis through the unembedding, returning plausible non-zero directions that were independent of the real expert weights.

This PR:

- omits dense MLP weight keys when a real Bridge MLP cannot expose them;
- makes `SVDInterpreter` raise a layer-specific `NotImplementedError` before SVD for unavailable `w_in` or `w_out`;
- documents the unsupported sparse-MoE contract;
- adds unit coverage for parameter omission and a download-free tiny Mixtral regression that proves `torch.linalg.svd` is not reached.

The change deliberately does not select, average, or router-weight experts. Those behaviors need a separately defined expert-aware interpretation contract.

Closes #1779

## Type of change

## Validation

- Relevant parameter, SVD, Mixtral, hybrid-MoE, optimizer, and parameter-count tests: `103 passed, 3 skipped`
- Dense tiny GPT-2 control: both `w_in` and `w_out` still return non-zero SVD results
- `uv run --no-cache --no-sync mypy .`: no issues in 398 source files
- pycln, isort, Black, and `git diff --check`: passed

The complete test suite was not run locally.

# Checklist:


koriyoshi2041 (approved):
Reviewed at `1efd56e0`. Omitting structurally unavailable sparse-MoE weights closes the dangerous path where a fabricated zero matrix can produce plausible non-zero singular directions, while the dense/absent-MLP cases remain explicit. I also checked the partial-weight path: an available `W_in` is preserved while an unavailable `W_out` is omitted, so the interpreter fails at the requested boundary rather than erasing usable data. The focused parameter, Mixtral, and dense-dispatch tests pass locally (20 passed). The full hosted compatibility/quality matrix is green.

jlarson4:
Thanks for writing up the issue and for closing up this bug. Looks great, @emerardd, approved & merging as is!

PR #1781, opened 2026-09-13
feat(attribution_patching): edge attribution (EAP) scoring on TransformerBridge

janmenjayap:
---

### Summary

Extends `attribution_patching` with **edge-level** attribution (EAP): every residual-stream writer -> reader edge is scored from the same single gradient cache that node-level attribution already produces, with no extra forward/backward passes. `granularity="edge"` no longer raises `NotImplementedError`; `ig_steps>1` (EAP-IG) still does and lands in a later
PR.

An edge `u -> v` carries writer `u`'s contribution into reader `v`'s input. Its score is `(a_clean[u] - a_corrupt[u]) . d(metric)/d(input of v)`, evaluated at `v`'s read point. Sweeping every `(u, v)` pair from the clean and corrupt caches yields all edge scores in one pass.

---

### What's included

- **Per-head hook requirements.** Edges into/out of attention heads need `hook_result` (writer) and split `hook_q_input`/`hook_k_input`/`hook_v_input` (reader), which only fire once `cfg.use_attn_result` and `cfg.use_split_qkv_input` are on. These are now turned on before caching, and the missing-hook check is generalized so any granularity's cache is validated the same way rather than duplicating it per graph.
- **Edge enumeration in the typed graph.** `enumerate_edges(model, cache)` builds every writer -> reader pair position-by-position over the residual stream's running sum, sharing `(layer, position, head)` keys with the activation and gradient caches. `NodeKind` gains reader kinds (`q_input`/`k_input`/`v_input`/`mlp_in`); a writer feeding both a direct edge and a through-MLP edge yields two distinct pairs, enforced by a uniqueness guard rather than silently summed.
- **Edge scoring wired into `attribution_patch`.** `granularity="edge"` populates `AttributionResult.edge_scores`; `node_scores` becomes each writer's aggregate over its own outgoing edges. `top_edges()` ranks by absolute score instead of raising.
- **Exact-patch parity, mutation-checked reconstruction, and sum-to-total (edge form) tests** on a fully linear 2-layer toy Bridge: a genuine single-edge activation patch matches `_edge_effects`' sign and (on this linear toy) magnitude; perturbing one writer's captured contribution changes only the edges it feeds and nothing else; a reader's incoming edge scores sum to the same value as scoring its cached input-delta directly. `generic_activation_patch` itself isn't used for the parity check — its `model: HookedTransformer` parameter is enforced at runtime by this repo's jaxtyping/beartype test configuration, which rejects a `TransformerBridge` instance outright, so the check patches directly through the Bridge's own `hooks()`, the same mechanism `generic_activation_patch` uses internally.
- **Memory caveat documented:** per-head `[batch, seq, n_heads, d_model]` tensors are fine on `gpt2-small` and do not scale to large models.

---

### Files

- `transformer_lens/tools/analysis/attribution_patching.py` (extended)
- `tests/unit/tools/test_attribution_patching.py` (extended)

---

### Scope / follow-ups

Edge-level EAP only, `ig_steps` stays `1`. The following land in follow-on PRs: ablate-outside faithfulness, integrated gradients (EAP-IG, `ig_steps>1`), oracle parity, and the demo notebook.

Addresses part of #1742 — this PR covers edge-granularity attribution (EAP); faithfulness, EAP-IG, oracle parity, and the demo are tracked as follow-on work above.

---

### Checklist

- [x] `make unit-test` green
- [x] `uv run mypy .` clean
- [x] `make format` applied
- [x] `granularity="edge"` produces `edge_scores`/`top_edges`; `ig_steps>1` still raises `NotImplementedError`
- [x] Exact-patch parity (sign, and magnitude within tolerance on the 2-layer toy) and mutation-checked reconstruction both hold
- [x] Per-head hooks required and raised on when missing; memory caveat documented
- [x] No `HookedTransformer` reference added; all model interaction via the Bridge

---


koriyoshi2041 (approved):
Reviewed exact head `7a436b73586455681c46705c4a7b056453d45ecc`. I traced the graph in residual-stream order and checked the three core invariants: writers are only connected to downstream readers, attention writer deltas are taken in `hook_result` d_model space while Q/K/V reader gradients use the matching head/position index, and incoming edge scores reconstruct each reader input-delta dot gradient. The final-readout omission means edge-mode `node_scores` is intentionally an outgoing-edge aggregate rather than node-granularity attribution; the code and docs state that boundary explicitly.

Local focused result: 34/34 attribution-patching tests pass at this exact head. The hosted compatibility, type, format, full-coverage, benchmark, and notebook matrix is green. I did not find a correctness blocker for the scoped EAP implementation.

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -249,6 +331,116 @@ def enumerate_nodes(model: Any, cache: GradientCache) -> list[Node]:
...
+
+            available = available + [
+                Node(kind="attn_head_out", layer=layer, head=head, position=position)
+                for head in range(n_heads)
+            ]
+
+            mlp_reader = Node(kind="mlp_in", layer=layer, position=position)
+            edges.extend((writer, mlp_reader) for writer in available)
+
+            available = available + [Node(kind="mlp_out", layer=layer, position=position)]
```
jlarson4, line 438: The edge graph has no terminal reader for the logits, so no writer's direct path to the output is ever an edge. On IOI that path carries the name movers' main effect. As a result, a writer's edge aggregate can flip sign relative to its node score, and `top_edges` can never return a head->logits edge. The pinned `hannamw/eap-ig` graph has a `LogitNode` that every writer feeds, and faithfulness and oracle parity will both need it. The escape term in `test_attribution_patch_edge_aggregation_reconstructs_direct_node_scores` is already this edge. Could you add a logits reader (gradient at the final residual, e.g. `blocks.{n_layers - 1}.hook_resid_post`) fed by every writer, so an edge sweep's writer aggregate equals its node score?

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -435,19 +627,104 @@ def _node_effects(
...
+        writer_name = _writer_hook_name(writer)
+        reader_name = reader.hook_name
+        grad = corrupt_cache.gradients.get(reader_name)
+        if grad is None:
+            raise ValueError(
+                f"edge {(writer, reader)} reads its gradient at {reader_name!r}, but "
+                "the corrupt cache holds none there; cache with a names_filter that "
+                "retains this hook point."
+            )
+        delta = clean_cache.activations[writer_name] - corrupt_cache.activations[writer_name]
```
jlarson4, line 675: This recomputes the full `clean - corrupt` subtraction over the whole cached tensor for every edge, then scores each edge with a Python-level product and a `float()`. That's ~485k per gpt2-small pair, which makes the batched IOI averaging this slice needs impractical. Could you compute each hook's delta once and score each writer-family x reader-family block with one batched product, unpacking into the `dict` at the end?

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -498,8 +780,34 @@ def attribution_patch(
             "Attribution patching aligns activations position-by-position."
         )
 
-    node_hook_names = _required_hook_names(int(model.cfg.n_layers))
     batch = int(clean.shape[0])
+    n_layers = int(model.cfg.n_layers)
+
+    if config.granularity == "edge":
+        _ensure_edge_hook_flags(model)
```
jlarson4, line 787: `_ensure_edge_hook_flags` leaves `use_attn_result`, `use_split_qkv_input` and `use_hook_mlp_in` on after `attribution_patch` returns. Every later forward on the caller's model then materializes the per-head `[batch, seq, n_heads, d_model]` tensors. For a caller with `use_attn_in` on, the call raises instead, because that flag is mutually exclusive with `use_split_qkv_input`, and `use_attn_result` is still left switched on. Could you save the original flags (including `use_attn_in`), set what the sweep needs, and restore them all in a `try/finally`?

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -498,8 +780,34 @@ def attribution_patch(
...
+                compute_gradient=False,
+            )
+            corrupt_cache = cache_activation_and_gradient(
+                model, corrupt[index : index + 1], metric_fn, names_filter=hook_names
+            )
+            edges = enumerate_edges(model, corrupt_cache)
+            for edge, score in _edge_effects(clean_cache, corrupt_cache, edges).items():
+                edge_totals[edge] = edge_totals.get(edge, 0.0) + score
+
+        edge_scores = {edge: total / batch for edge, total in edge_totals.items()}
```
jlarson4, line 806: No edge-granularity test runs more than one pair. This loop's per-pair slicing, accumulation and `/ batch` averaging are all unguarded: a loop that scored only the first pair on every iteration would go unnoticed. Could you add an edge counterpart to `test_attribution_patch_averages_scores_across_the_batch` whose two pairs score differently?

tests/unit/tools/test_attribution_patching.py
```diff
@@ -736,3 +771,797 @@ def test_nonlinear_node_scores_read_the_corrupt_run_gradient() -> None:
...
+    clean_cache = cache_activation_and_gradient(
+        model, clean, metric, names_filter=edge_hook_names, compute_gradient=False
+    )
+    corrupt_cache = cache_activation_and_gradient(
+        model, corrupt, metric, names_filter=edge_hook_names
+    )
+    edges = enumerate_edges(model, corrupt_cache)
+    edge_scores = _edge_effects(clean_cache, corrupt_cache, edges)
+
+    reader = Node(kind="mlp_in", layer=1, position=1)
```
jlarson4, line 1552: The toy never mixes positions, `_metric_fn` reads only the last one, and clean and corrupt share the token at position 1. Every delta and gradient at position 1 is therefore zero. This parity test and `test_edge_effects_mutation_only_changes_the_perturbed_writers_edges` (writer at line 1454) both compare zeros with zeros, so they pass even if `_edge_effects` returns 0.0 for every edge. Could you move both to the last position and assert that the scores they check are nonzero?

transformer_lens/tools/analysis/attribution_patching.py
```diff
@@ -249,6 +331,116 @@ def enumerate_nodes(model: Any, cache: GradientCache) -> list[Node]:
...
+            for reader in attn_readers:
+                edges.extend((writer, reader) for writer in available)
+
+            available = available + [
+                Node(kind="attn_head_out", layer=layer, head=head, position=position)
+                for head in range(n_heads)
+            ]
+
+            mlp_reader = Node(kind="mlp_in", layer=layer, position=position)
+            edges.extend((writer, mlp_reader) for writer in available)
```
jlarson4, line 436: This connects every same-layer head to `mlp_in`, which is wrong for `cfg.parallel_attn_mlp` models (Pythia, GPT-J, Falcon, Phi, …). In those models the MLP reads the layer input, not the post-attention residual. The same-layer head->MLP edges get sizeable scores for a path that doesn't exist, and a reader's incoming edges no longer sum to its own input change. Could you drop the same-layer heads from the MLP reader's writers when `model.cfg.parallel_attn_mlp` is set, or raise on those models until they're supported?

jlarson4 (commented):
Thanks for picking up the edge portion of #1742, the typed edge graph and the reader-sum test are a solid base. A couple comments below, let me know if you have any questions

koriyoshi2041 (approved):

koriyoshi2041 (commented):
Exact-head review receipt for `9fb101b42b2acf9223ac2c7425dd098e320971e5` (the approval immediately above was submitted without its intended body):

- terminal `hook_resid_post` reader closes the graph, including the final MLP writer, and outgoing-edge aggregates reconstruct direct node scores;
- `parallel_attn_mlp` excludes same-layer head → MLP edges while preserving later-reader and terminal edges;
- reader-grouped batched matrix-vector scoring matches the per-edge reference;
- edge-only hook flags restore across normal and exceptional exits, including `use_attn_in`.

Local exact-head result: `uv run pytest tests/unit/tools/test_attribution_patching.py -q` → 50 passed. Hosted compatibility, coverage, type, format, benchmark, and notebook checks are green. I found no remaining correctness blocker for the scoped EAP implementation.

jlarson4:
@janmenjayap All requests resolved expertly, thank you for taking care of those. Approved and merging now

PR #1785, opened 2026-09-14
feat(model_bridge): shared LN/Identity/Half relevance-rule backend

janmenjayap:
---

### Summary

Lands the scoped custom-VJP backend shared between R-lens and the (separate) RelP issue: LN-,
Identity-, and Half-rule primitives, plus a `use_relevance_rules(model, rules)` context that
installs them on live `TransformerBridge` components by mount position. Every rule-active forward
is `torch.equal` to the rule-inactive forward — only backward semantics change. No public API is
added; the module is underscore-private and undocumented until the RelP issue is filed. A reviewer
can evaluate this backend in isolation, before any `RelevanceLens.fit` code exists.

Part of #1755 (R-lens). This is PR2 of the 3-PR plan on that issue; PR1
([#1764](https://github.com/TransformerLensOrg/TransformerLens/pull/1764)) landed the
estimator-independent J-lens fit driver this backend will later plug into, PR3 adds
`RelevanceLens.fit` + provenance + docs. **Note:** #1755 is currently closed
(`state_reason: completed`) even though only PR1 has landed so far. Would it be possible to
reopen it (or file a fresh tracking issue) so PR2 and PR3 stay linked to it? Happy to help with
that if useful.

---

### Motivation

R-lens and generic RelP both need forward-equivalent, backward-modified norm/activation/gate
semantics on live Bridge components. Landing the backend as its own PR, ahead of
`RelevanceLens.fit`, lets a reviewer evaluate the risky part — custom autograd wrapping
production `NormalizationBridge`/`GatedMLPBridge`/`JointGateUpMLPBridge` forward paths — in
isolation, before layering the estimator-specific fitting logic on top. This PR addresses five
explicit conditions raised during discussion of the 3-PR plan; each is called out against the
commit that satisfies it below.

---

1. **`test(relevance_rules): closed-form VJP + forward-identity primitives`**
   Analytic tests for LN-, Identity-, and Half-rule against their closed forms, covering zero,
   negative, mixed-sign, non-contiguous, and batched inputs at fp32/fp64. Red as expected — the
   primitive module doesn't exist yet.

2. **`feat(relevance_rules): forward-equivalent rule primitives`**
   Implements each rule as a `torch.autograd.Function`: forward reproduces the native value
   exactly, backward substitutes the rule's closed-form VJP (LN-rule treats the denominator as
   constant; Identity-rule uses `f(x)/x` with the `1/2` limit at zero; Half-rule splits the
   product-rule gradient evenly).

3. **`test(model_bridge): scoped rule context forward-identity + cleanup`**
   Tests for `use_relevance_rules`: forward-identity (standalone and nested), gradient restoration
   on exit, reference-counted nested scopes, no-op on exception, positional (not class-based)
   targeting, and "skipped" reporting for a mount that doesn't implement the rule protocol. Red —
   the context doesn't exist yet.

4. **`feat(model_bridge): scoped relevance-rule context + coverage report`**
   Adds `RelevanceRules` (frozen dataclass: `normalization`, `activation`, `multiplicative_gate`,
   `attention`), `RelevanceRuleCoverage` (`installed`/`skipped`), and `use_relevance_rules(...)`.
   Installs each requested rule only at that rule kind's canonical mount name — never by
   `isinstance` — with reference-counted nesting and exception-safe cleanup. No model config is
   mutated.
   *Satisfies reviewer condition 3 (positional, not class-based, targeting).*

5. **`feat(model_bridge): native-forward LN-rule on NormalizationBridge`**
   Wraps the existing native-autograd no-edit branch's own `original_component(x)` call in the
   custom `Function`, so the rule-active forward is bit-identical to today's **by construction** —
   no restructuring of the existing native-autograd / python-norm / hook-fallback dispatch. A
   backward hook on `hook_scale`/`hook_normalized`, or a forward hook editing either, now raises
   `RelevanceRuleConflictError` while the rule is active (checked on every forward call) instead of
   the previous warn-and-fallback, which would have silently composed the rule with the hook edit.
   `NormalizationBridge` implements the rule protocol so it's targeted at `ln1`/`ln2` positionally;
   any instance that never reaches the native-autograd branch reports its kind as empty, so it's
   "skipped" rather than falsely "installed."
   *Satisfies reviewer conditions 1 (strictly-additive integration) and 2 (fail-closed
   hook/rule precedence, checked per forward call).*

6. **`feat(model_bridge): gated-MLP rule via recompute-from-weights VJP`**
   For the raw `GatedMLPBridge` (Qwen2/Llama/Gemma family), wraps the opaque
   `original_component(x)` call and recomputes gate/up/down checkpointing-style from the
   TL-oriented `W_gate`/`W_in`/`W_out` properties (never the compatibility-mode-only
   `_processed_*` attributes) to apply Identity-/Half-rule; weight/bias gradients come back off the
   same recomputed graph via `torch.autograd.grad`. The recompute is allowlisted by the HF module
   class backing the weights (`nn.Linear` vs. `Conv1D`), not per adapter. `JointGateUpMLPBridge`
   (Phi-3/GLM) attaches the rule directly at its already-Python-visible `act_fn(gate) * up`
   multiplication. A gated-MLP node now answers to both "activation" and "multiplicative_gate"
   independently at one mount, so the protocol moved from a single `_relevance_rule_kind` to a
   tuple `_relevance_rule_kinds`, and refcounting is keyed per `(module, kind)`.
   *Satisfies reviewer condition 4 (gated-MLP allowlist keyed on HF module class, not adapter).*

7. **`feat(model_bridge): fail-closed coverage on unsupported paths`**
   Distinguishes benign non-applicability (no protocol, or a dispatch path a rule genuinely
   doesn't wrap) from a component that's expected to honor a rule but currently can't. Adds an
   optional `_relevance_rule_unsupported_kinds` attribute; `use_relevance_rules` now raises
   `RelevanceRuleUnsupportedError`, naming the component's dotted path, before any forward/backward
   runs. `GatedMLPBridge` uses it for an unrecognized weight-backing class (disqualifies both
   kinds) and for relu-family activations (disqualifies only "activation" — `f(x)/x` isn't the
   correct derivative for relu/relu2, but Half-rule is unaffected by activation form).

8. **`test(relevance_rules): tolerant parity vs pinned FarnoushRJ/RelP`**
   Compares all three primitives against formulas ported from `FarnoushRJ/RelP` pinned at
   `8219d6dc417c3fd7f318342cf61cd2a0c20b7250`, and explicitly asserts the one known discrepancy:
   this backend's Identity-rule uses the paper-defined `1/2` limit at `x == 0`, while the
   reference's epsilon stabilizer collapses to exactly `0` there. All other cases match to
   floating-point precision.

9. **`fix(model_bridge): find relevance-rule mounts shadowed by named_modules dedup`** *(not in
   the original plan)*
   `named_modules()` deduplicates by object identity and keeps only the first-visited path per
   object. On a real assembled `TransformerBridge`, the raw HF-attribute path
   (`blocks.N._original_component.input_layernorm`) is registered before the canonical alias
   (`blocks.N.ln1`), so the canonical name was never seen — the LN-rule never installed on any
   real model and was silently reported neither installed nor skipped. Fixed by walking with
   `remove_duplicate=False` and keeping the shortest-path alias per object. Found and fixed while
   building the Commit 10/11 integration fixtures below, before this backend had ever been
   exercised end-to-end on an assembled model.

10. **`test(integration): backend on tiny Qwen2 + JointGateUp fixtures`**
    End-to-end, fully offline (no network, no checkpoint download) coverage on a tiny random
    Qwen2 (opaque `GatedMLPBridge` recompute path) and a tiny random Phi-3
    (`JointGateUpMLPBridge`), with normalization, activation, and multiplicative_gate all active
    together: forward stays `torch.equal` to the rule-inactive baseline, every canonical mount on
    both fixtures reports installed with nothing skipped, and each rule-active node's upstream
    gradient matches its closed-form VJP in the real graph.

---

### Testing

- `tests/unit/tools/test_relevance_rules.py` +
  `tests/unit/model_bridge/test_relevance_rules.py` +
  `tests/unit/model_bridge/generalized_components/test_gated_mlp_relevance_rule.py` +
  `test_joint_gate_up_relevance_rule.py` + `test_normalization_relevance_rule.py` +
  `tests/unit/model_bridge/supported_architectures/test_relevance_rule_mount_placeholders.py` —
  **118 passed**.
- `tests/integration/test_relevance_lens.py` — **6 passed** (new backend integration fixtures).
- `tests/integration/test_backward_lens.py` +
  `tests/unit/model_bridge/generalized_components/test_normalization_hook_semantics.py` — **43
  passed** (defensive gate, untouched by this PR).
- `uv run mypy` on the four touched source files — clean. `black --check` / `isort --check-only`
  on all touched files — clean.

The `backward_lens` and hook-semantics suites are a defensive gate, not a functional one: this PR
adds a strictly-additive branch to `NormalizationBridge`/`GatedMLPBridge`/`JointGateUpMLPBridge`
without restructuring existing dispatch, and both suites stay green with the rule context
inactive, which is the state every existing caller runs in today.

---

### Out of scope (later PR on #1755)

`RelevanceLens.fit`, estimator/rule provenance in the lens artifact contract, merge/registry
guards, and docs/notebook integration — all PR3. Source-artifact conversion, pass@k evaluation,
CKA, probe-upper-bound, and MoE/mHC rules are tracked as separate follow-up issues.

---

### API

No public API added or changed. `transformer_lens/model_bridge/_relevance_rules.py` is
underscore-private, not exported from any `__init__.py`, and undocumented — per reviewer condition
5, it stays that way until the RelP issue is filed.

---

### Checklist

- [x] Base set to `dev`; working branch `feat/r-lens-backend`.
- [x] Rule-enabled and native forwards are bit-identical (`torch.equal`) on every claimed path.
- [x] LN-rule wraps `original_component(x)` itself; no restructuring of the existing three-way
  dispatch; rule-inactive behavior unchanged (regression-covered).
- [x] Rule-vs-hook precedence fails closed on every forward call, not just at context entry.
- [x] Rule targeting is positional (`ln1`/`ln2` mount name), never `isinstance`; `gemma3n`/`gemma4`
  placeholder mounts and `stablelm`'s `q_layernorm`/`k_layernorm` naming are covered and classified
  "not applicable."
- [x] Gated-MLP backward recomputes from `W_gate`/`W_in`/`W_out` properties, never `_processed_*`;
  allowlisted per HF module class, not per adapter.
- [x] Unsupported requested coverage (opaque gate backing, relu-family Identity-rule) fails closed
  with the component's dotted path.
- [x] Context cleanup verified after success and after exceptions; nested contexts deterministic
  and reference-counted per `(module, kind)`.
- [x] Tolerant parity vs. pinned `FarnoushRJ/RelP`, with the exact-zero stabilizer discrepancy
  recorded rather than tolerance-absorbed.
- [x] No `model.cfg` flags; no `.data`.
- [x] Backend module stays underscore-private, unexported, undocumented.
- [x] `tests/integration/test_backward_lens.py` and
  `tests/unit/model_bridge/generalized_components/test_normalization_hook_semantics.py` green,
  untouched.
- [x] `make format` (black/isort) and `uv run mypy` clean on touched files.
- [ ] Full `make test-pr` run (unit + docstring + acceptance + integration) attached before
  submit.
- [ ] Issue #1755 reopened (or a fresh tracking issue filed) before/alongside this PR.

---


transformer_lens/model_bridge/generalized_components/gated_mlp.py
```diff
@@ -46,6 +73,122 @@ def relu_squared(x: torch.Tensor) -> torch.Tensor:
...
+        extra_kwargs: Dict[str, Any],
+    ) -> torch.Tensor:
+        ctx.save_for_backward(x, w_gate, w_in, w_out)
+        ctx.b_gate = b_gate
+        ctx.b_in = b_in
+        ctx.b_out = b_out
+        ctx.act_fn = act_fn
+        ctx.activation_rule_active = activation_rule_active
+        ctx.gate_rule_active = gate_rule_active
+        result: torch.Tensor = component(x, *extra_args, **extra_kwargs)
```
jlarson4: This call runs inside the Function's forward, so hooks on `gate`/`in`/`out` still fire, but nothing they do reaches the backward. A forward edit on `in.hook_out` changes the output while the recomputed gradient ignores it, and a backward hook on `gate.hook_out` silently never fires. That's the composition `NormalizationBridge` now refuses. Could this path raise `RelevanceRuleConflictError` for backward hooks and forward edits at those points, as the normalization path does, while observe-only hooks keep working?

transformer_lens/model_bridge/generalized_components/gated_mlp.py
```diff
@@ -84,6 +227,88 @@ def __init__(
...
+        which occupy exactly this node's shape but cannot be honored correctly.
+        """
+        if self.original_component is None:
+            return False
+        gate_module = getattr(self, "gate", None)
+        in_module = getattr(self, "in", None)
+        out_module = getattr(self, "out", None)
+        return gate_module is not None and in_module is not None and out_module is not None
+
+    def _has_recognized_weight_backing(self) -> bool:
```
jlarson4: The recompute assumes the wrapped module computes `down(act(gate(x)) * up(x))`, but this check only looks at the projection classes. A HF MLP whose forward does more gets a wrong backward while reporting installed. Two do today: `FalconH1MLP` scales by `gate_multiplier` and `down_multiplier`, and `Gemma3nTextMLP` applies `_gaussian_topk` to the gate on its early layers. Could the allowlist key on the HF MLP class whose forward is being recomputed, with anything else reported unsupported?

transformer_lens/model_bridge/generalized_components/gated_mlp.py
```diff
@@ -140,7 +365,25 @@ def forward(self, *args, **kwargs) -> torch.Tensor:
         hidden_states = args[0]
         hidden_states = self.hook_in(hidden_states)
         new_args = (hidden_states,) + args[1:]
-        output = self.original_component(*new_args, **kwargs)
+        if self._relevance_rule_activation_active or self._relevance_rule_gate_active:
```
jlarson4: Under `enable_compatibility_mode()`, `forward` returns from the processed-weights branch above before reaching this check. `JointGateUpMLPBridge` delegates to that branch, so neither rule runs, yet `use_relevance_rules` still reports every `mlp` mount installed. Could `_relevance_rule_kinds` treat processed weights as unsupported so the context raises?

transformer_lens/model_bridge/generalized_components/normalization.py
```diff
@@ -216,13 +322,59 @@ def _hf_autograd_forward_with_hooks(self, x: torch.Tensor) -> torch.Tensor:
...
+            eps_value: float | torch.Tensor = getattr(self.config, "eps", 1e-05)
+        else:
+            eps_value = eps_tensor
+        variance = x_centered.pow(2).mean(-1, keepdim=True)
+        denom = (
+            (variance + eps_value).sqrt()
+            if isinstance(eps_value, torch.Tensor)
+            else (variance + float(eps_value)).sqrt()
+        )
+        weight = cast(torch.Tensor, self.weight)
```
jlarson4, line 370: OLMo's `OlmoLayerNorm` has no weight, but its `ln1`/`ln2` mounts report the LN-rule installed, and the first rule-active forward then fails with `AttributeError` on `self.weight`. Could a missing weight be treated as 1 here and in the backward?

transformer_lens/model_bridge/_relevance_rules.py
```diff
@@ -0,0 +1,305 @@
+"""Forward-equivalent relevance-rule primitives, plus the scoped context that installs them.
+
+Each primitive is a ``torch.autograd.Function`` that reproduces its native forward
+value exactly while replacing the backward pass with the rule's closed-form VJP.
+``use_relevance_rules`` installs these rules on a model's canonical mount points only
```
jlarson4, line 5: `RelevanceRules(attention=True)` has no entry in `_CANONICAL_MOUNTS`, so the context installs nothing, skips nothing and doesn't raise. A caller asking for attention rules gets ordinary gradients and an empty coverage report. Could requesting a kind with no canonical mount raise instead?

transformer_lens/model_bridge/_relevance_rules.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def _disable_relevance_rule(self, kind: str) -> None:
+        ...
+
+
+# Canonical mount name per rule kind. Targeting is positional: a component is only
+# considered for a kind when it sits at that kind's mount name, never by isinstance,
+# so a same-class component mounted elsewhere (for example a q_norm sharing
+# NormalizationBridge's class) is left untouched.
+_CANONICAL_MOUNTS: Mapping[str, Tuple[str, ...]] = {
```
jlarson4, line 195: The leaf-name match is exact, so the `ln1_post`/`ln2_post` norms that Gemma 2/3/4, GLM-4 and several others mount never get the LN-rule and appear nowhere in the coverage report. The pinned RelP reference applies the LN-rule to those post-norms. OLMo 2 already gets it on the same `post_attention_layernorm` role because its adapter mounts that norm as `ln1`. Could `ln1_post` and `ln2_post` join the normalization mounts?

jlarson4 (changes requested):
Thanks for splitting the backend out ahead of the fit code, and for the closed-form and RelP-parity tests.

I do have a larger change request for this one than I usually do, as it touches generalized components in a signficant way. I want to specifically focus in on the design of the gated-MLP rules. `_GatedMLPRecomputeRule` rebuilds `down(act(gate(x)) * up(x))` from the weights, so an HF MLP whose forward does anything more gets a wrong gradient while coverage reports the rule installed. Falcon-H1's multipliers and Gemma3n's gate sparsity both hit this today. Running HF's forward inside the Function also takes the MLP's own hooks out of the backward, and the compatibility-mode branch skips the rules entirely. Neither rule needs a recompute. The Half-rule is exactly ordinary autograd with the product's gradient halved, so it can be a gradient-scale hook on `mlp.out.hook_in`. The Identity-rule can sit on the HF MLP's own `act_fn` module, applied inline wherever the bridge computes the activation itself, as `JointGateUpMLPBridge` already does. Could the gated-MLP rules move to that shape, with the recompute dropped?

Generally, where possible, we try to avoid adding recompute if we can. Any areas where we recompute HF elements is a potential failure point on new models, if they have new unexpected behaviors.

jlarson4:
Looks great @janmenjayap, thank you for making these updates. Merging now

PR #1788, opened 2026-09-15
Require evaluation mode for attribution patching

### LINKED ISSUE #1787: [Bug Report] attribution_patch reports non-zero causal effects for identical inputs in train mode
**Describe the bug**

`attribution_patch()` can report non-zero causal attributions when the clean and corrupt inputs are identical if the model, or one of its nested stochastic submodules, is in training mode.

The function runs the clean and corrupt forwards separately. With active dropout, the two forwards use independent masks, so the activation difference contains dropout noise even when there is no input difference. The resulting scores are returned as clean-to-corrupt causal attributions without an error or warning.

This is a silent research-correctness problem: the API can identify apparently important nodes even though the requested intervention is an exact no-op.

**Code example**

I reproduced this without downloading model weights by constructing a tiny random GPT-2 Bridge from config:

```python
import torch
from torch import nn
from transformers import GPT2Config, GPT2LMHeadModel

from transformer_lens.tools.analysis.attribution_patching import attribution_patch

torch.manual_seed(0)
config = GPT2Config(
    n_layer=2,
    n_head=2,
    n_embd=32,
    n_positions=8,
    n_ctx=8,
    vocab_size=32,
    resid_pdrop=0.5,
    embd_pdrop=0.5,
    attn_pdrop=0.5,
)
hf_model = GPT2LMHeadModel(config)
    hf_model,
    "GPT2LMHeadModel",
    hf_config=config,
    tokenizer=None,
    dtype=torch.float32,
    device="cpu",
    model_name="tiny-gpt2-attribution-dropout-repro",
)
tokens = torch.tensor([[1, 2, 3, 4]], dtype=torch.long)

def metric(logits):
    return logits[0, -1, 5] - logits[0, -1, 6]

def summarize():
    result = attribution_patch(bridge, tokens, tokens.clone(), metric)
    scores = torch.tensor(list(result.node_scores.values()))
    return {
        "nonzero": int(torch.count_nonzero(scores)),
        "max_abs": scores.abs().max().item(),
        "sum_abs": scores.abs().sum().item(),
    }

bridge.train()
print("root train:", summarize())

bridge.eval()
print("all eval:", summarize())

bridge.eval()
nested_dropout = next(module for module in hf_model.modules() if isinstance(module, nn.Dropout))
nested_dropout.train()
print("root eval, nested dropout train:", summarize())
```

Output:

```text
root train: nonzero=15, max_abs=0.155229106545, sum_abs=0.456637024879
all eval: nonzero=0, max_abs=0.000000000000, sum_abs=0.000000000000
root eval, nested dropout train: nonzero=15, max_abs=0.089158877730, sum_abs=0.162343397737
```

The clean and corrupt tensors are identical. All scores are exactly zero when every module is in evaluation mode. Re-enabling only one nested dropout module is enough to restore the false non-zero scores even though `bridge.training` remains `False`.

The root cause is that `attribution_patch()` performs independent clean and corrupt forwards and then computes `(clean_activation - corrupt_activation) * corrupt_gradient`. It verifies global autograd state but does not verify the training state of the model or its nested modules.

**Expected behavior**

`attribution_patch()` should fail before either forward if the Bridge, its hidden `original_model`, or any nested module is in training mode. The error should identify the offending module and instruct the caller to run `model.eval()`.

The analysis function should not switch modes internally, because doing so would silently mutate caller-owned state and would make mixed per-module training states difficult to restore correctly.

The all-submodule check should follow or share the pattern already accepted for `JacobianLens.fit()` in #1762 / #1763. Checking only `model.training` is insufficient, as demonstrated by the nested-dropout reproduction.

Suggested regression coverage includes a training root, an eval root with a nested Bridge module in training mode, an eval root with a nested module inside `original_model` in training mode, unchanged training flags after failure, an all-eval success control, and exact zero scores for identical inputs in eval mode.

This is separate from the fully-frozen-model backward-cache issue: that issue concerns the absence of an autograd graph, while this one concerns silent stochastic corruption of the measured attribution.

**System Info**

- Installation: source checkout managed with `uv`
- OS: Windows 11
- Python: 3.12.10
- PyTorch: 2.10.0+cpu
- Transformers: 5.13.0
- TransformerLens: `upstream/dev` at `a4717b41ac09b7b9f21a3bcd086543964c1815eb`

**Additional context**

The implicated implementation is in `transformer_lens/tools/analysis/attribution_patching.py`, around `cache_activation_and_gradient()` and `attribution_patch()`.

Open and closed Issue/PR searches on 2026-09-14 for attribution patching, dropout, train mode, identical inputs, eval mode, and `cache_activation_and_gradient` found no matching report. Issue #1762 and PR #1763 address the same class of stochastic-state problem in `JacobianLens.fit()`, but do not cover this separate public API.

### Checklist

emerardd:
# Description

Require `attribution_patch()` to run with the model and every submodule in evaluation mode, preventing separate clean and corrupt forwards from mixing independent dropout realizations into false causal effects.

This extracts the all-submodule state validation already used by `JacobianLens.fit()` into a shared private helper. The helper also inspects a Bridge's deliberately unregistered `original_model`, reports the offending module paths, and rejects invalid state without mutating caller-owned training flags. Input-shape validation still runs first, while the state guard runs before either attribution forward.

Regression coverage includes root training mode, an eval root with only a nested dropout in training mode, hidden `original_model` state, preservation of training flags, and exact-zero attribution for identical inputs in evaluation mode.

Fixes #1787

## Type of change

# Checklist:

- [ ] New and existing unit tests pass locally with my changes; only the affected unit-test surface was run (`142 passed`)

## Local validation

```text
pytest tests/unit/tools/test_attribution_patching.py tests/unit/tools/test_jacobian_lens.py -q
142 passed, 3 warnings in 27.65s

pycln --check --all: passed
isort --check-only: passed
black --check: passed
mypy: Success: no issues found in 399 source files
git diff --check: passed
```


jlarson4:
Looks good, thanks @emerardd

Issue #1769, opened 2026-09-10
[Bug Report] stop_at_layer silently no-ops on RavenArchitectureAdapter (Huginn depth-recurrent decoder)


LightWork666:
**Describe the bug**

`TransformerBridge.forward(..., stop_at_layer=N)` (and anything built on it, e.g. `input_to_embed()`) silently does nothing on a Raven/Huginn bridge -- no error, no early stop. The forward pass runs to full completion as if `stop_at_layer` had never been passed.

Two independent gaps combine to cause this:

1. **Top-level wiring only knows about `self.blocks`.** In `TransformerBridge.forward()`, the `stop_at_layer` handling explicitly rejects a fixed list of known non-standard block-list names (`L_blocks`, `H_blocks`, `encoder_blocks`, `decoder_blocks`) with a clear `NotImplementedError`, then sets `_stop_at_layer_idx` on every block only `if hasattr(self, "blocks")`. `RavenArchitectureAdapter.component_mapping` has no `"blocks"` key at all -- Huginn's three block lists are `prelude`, `core_block`, and `coda` (matching its prelude / weight-tied recurrent core / coda structure). Raven isn't in the reject-list either, so the request silently falls through both checks: no error, and `_stop_at_layer_idx` never gets set on anything.
2. **Even if it were set, `OpaqueBlockBridge._check_stop_at_layer()`'s layer-index regex wouldn't recognize it anyway.** It only matches `\.layers\.(\d+)` or `blocks\.(\d+)`; Raven's per-block names (`transformer.core_block.0`, `transformer.prelude.0`, `transformer.coda.0`) match neither. The standard `BlockBridge` class already has a more general `_extract_layer_idx()` (`(?:^|\.)(?:blocks|h|layers)\.(\d+)`, covering GPT-2's `h.N` too) that `OpaqueBlockBridge` doesn't share or reuse -- and even that wouldn't match `core_block`/`prelude`/`coda` either way.

Surveyed every other architecture using `OpaqueBlockBridge`/`SSMBlockBridge` (mamba, mamba2, nemotron_h, rwkv7, zamba2): all name their list `backbone.layers` or `model.layers`, which matches fine. Raven is the only affected architecture currently in the adapter set.

**Reproduction (no model download needed -- pure component-level and adapter-level verification)**

```python
import torch
import torch.nn as nn
from transformer_lens.model_bridge.generalized_components.opaque_block import OpaqueBlockBridge
[...]

PR #1789, opened 2026-09-17
fix: raise on stop_at_layer when no 'blocks' stack is registered

Aurnawr:
# Description

`TransformerBridge.forward(..., stop_at_layer=N)` silently did nothing on architectures that don't register a `blocks` stack (e.g. Raven/Huginn, whose layers are `prelude` / `core_block` / `coda`). The old guard only rejected four known names (`L_blocks`, `H_blocks`, `encoder_blocks`, `decoder_blocks`) and only set `_stop_at_layer_idx` when `hasattr(self, "blocks")`, so Raven fell through both checks and the forward pass ran to completion.

Following @jlarson4's suggested approach in the issue:

- **Allowlist instead of reject-list:** `stop_at_layer` now raises `NotImplementedError("stop_at_layer requires a 'blocks' stack; ...")` whenever no `blocks` stack is registered. This covers Raven and any future non-standard adapter, and removes the four-name reject-list.
- **`_modules` instead of `hasattr`:** added a private helper `_has_registered_blocks()` that checks `"blocks"` in the bridge's registered submodules. `hasattr(self, "blocks")` can return `True` via `__getattr__` falling through to the wrapped HF model. The existing `start_at_layer` guard now uses the same helper.
- **Test:** added `tests/unit/model_bridge/test_stop_at_layer_guard.py` with a regression test that builds a bare bridge with Raven-style `prelude`/`core_block`/`coda` lists (no checkpoint download) and asserts `stop_at_layer=0` raises.

No new dependencies.

Addresses #1769. Per the maintainer's note, the issue should stay open to track actual `stop_at_layer` support for Raven via top-level wiring.

## Type of change

Note: architectures with `L_blocks`/`H_blocks`/`encoder_blocks`/`decoder_blocks` still raise `NotImplementedError`, but the message text changed.

# Checklist:

Tests run locally:
- `uv run pytest tests/unit/model_bridge/test_stop_at_layer_guard.py tests/unit/model_bridge/test_audio_start_at_layer_guard.py -v`: all pass; the new test fails on the pre-fix code.
- `black --check` / `isort --check-only` on the changed files: clean.


Aurnawr:
Retargeted this to dev-4.x (it was accidentally opened against main, which pulled in 92 unrelated commits). The new test_stop_at_layer_guard.py passes. The remaining failures (test_generate_stream_batch_decode.py, test_hf_generate_batch_padding.py with AttributeError: ... no attribute '_driver') are in generate_stream, which this PR doesn't touch, so they seem to come from dev-4.x itself. Happy to look into them separately if useful. cc @jlarson4

jlarson4:
@Aurnawr I pulled in the latest `dev-4.x` commits, I unintentionally introduced a bug in the tests for the version of 4.x you based this one, should be resolved now

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -2136,25 +2147,20 @@ def forward(
         if start_at_layer is not None:
             input = self._setup_start_at_layer(input, start_at_layer)
 
+        # change this to have an allowlist
```
jlarson4: `# change this to have an allowlist` reads as a note-to-self and contradicts the code directly beneath it, which already is the allowlist. The same applies to `# guard for the block stack` at line 2459, which restates the line under it. Please delete both.

transformer_lens/model_bridge/transformer_bridge.py
```diff
@@ -2136,25 +2147,20 @@ def forward(
...
+
         # Set stop_at_layer flag on all blocks if requested
         if stop_at_layer is not None:
-            if (
-                hasattr(self, "L_blocks")
-                or hasattr(self, "H_blocks")
-                or hasattr(self, "encoder_blocks")
-                or hasattr(self, "decoder_blocks")
-            ):
+            if not self._has_registered_blocks():
```
jlarson4, line 2154: `forward`'s `stop_at_layer` argument is still described only as "Layer to stop forward pass at", so nothing tells a caller that architectures without a registered `blocks` stack now raise instead of running to completion. Please note the `NotImplementedError` in that argument's docstring.

tests/unit/model_bridge/test_stop_at_layer_guard.py
```diff
@@ -0,0 +1,36 @@
...
+        d_mlp=16,
+        n_heads=2,
+        architecture="RavenForCausalLM",
+    )
+    for name, module in block_lists.items():
+        bridge.add_module(name, module)
+    return bridge
+
+
+def test_stop_at_layer_raises_without_blocks_stack() -> None:
```
jlarson4, line 30: This test passes just as well when `_has_registered_blocks()` is replaced by `hasattr(self, "blocks")`, because the bare bridge has no wrapped model, so both implementations return False. The case the helper exists for is a bridge whose wrapped model exposes its own `.blocks` while none is registered. Can you add that case? Set `original_model` to an object with a `blocks` attribute and assert `stop_at_layer` still raises.

jlarson4 (commented):
Clean implementation of both halves of the approach from the issue, and the regression test is solid. Two comments on the diff's leftover scratch comments and the guard's test coverage, plus one on the docstring.

transformer_lens/model_bridge/transformer_bridge.py
Aurnawr: Done. The entry now says stop_at_layer is only supported on the standard blocks stack and that other architectures raise NotImplementedError instead of running to completion.

tests/unit/model_bridge/test_stop_at_layer_guard.py
Aurnawr: Added test_blocks_guard_ignores_unregistered_blocks_attribute. It sets original_model to an object with a .blocks attribute, asserts hasattr(bridge, "blocks") is True while _has_registered_blocks() is False, and checks both stop_at_layer and start_at_layer still raise. I confirmed it fails if the guard is reverted to hasattr

transformer_lens/model_bridge/transformer_bridge.py
Aurnawr: Done

jlarson4:
This looks great! Thank you for quickly getting to those, I will merge this and make sure it gets into today's release

Aurnawr:
Thanks for the quick reviews! Happy to take on the follow-up for actual
Raven stop_at_layer support if you'd like it scoped as a separate issue

Issue #1776, opened 2026-09-13
[Proposal] Causal coordinate-swap benchmark for coordinate_patch_hooks


janmenjayap:
---

### Proposal

Add a benchmark that measures whether `JacobianLens.coordinate_patch_hooks(...)` (#1749) causes a directional change in model output, under three controls the codebase does not yet have for this mechanism: baseline capability filtering, norm-matched random-atom controls, and bootstrap uncertainty on every reported rate. Ship the harness as tested library code, a frozen per-trial artifact generated once against a published lens, and a notebook that consumes that artifact rather than recomputing it.

**Suggested labels:** `enhancement`, `complexity-moderate`, `TransformerBridge`

---

### Motivation

`coordinate_patch(...)` (#1741) and `coordinate_patch_hooks(...)` (#1749) establish that an anchored J-space coordinate edit can be computed and, now, installed live inside a forward pass. Neither establishes whether performing that edit changes the model's output in the direction the edited concept predicts. `proposal.md`'s own Alternative 5 deferred this deliberately when it was first raised against the offline primitive: "model capability and prompt design would obscure review of the algebraic primitive." That reasoning was carried forward unchanged into #1749's own "Out of scope" section, which named this exact benchmark as PR3.

---

### Relationship to the existing `swap_hooks` benchmark (`Jacobian_Lens_Demo.ipynb` section 4)

`demos/Jacobian_Lens_Demo.ipynb` already contains a 48-trial country-swap benchmark (source in the notebook's section "4. Quantifying causal-swap success across the country subset"), shipped with #1507/#1545 well before `coordinate_patch_hooks` existed. It is the nearest precedent for this issue and this issue reuses several of its conventions directly, but it is a different benchmark, not an earlier version of this one:

- It exercises `swap_hooks` (a closed-form pseudoinverse exchange over a fixed 2-vector basis, re-applied at every position across a 15-layer band, `layers=range(10, 25)`), not `coordinate_patch_hooks` (a full sparse-solve anchored edit, one `(batch, position)` pair at a time). The two mechanisms have different failure modes under multi-layer composition -- see "Layer scope" below -- and are not interchangeable in a shared harness without conflating what each measures.
[...]

PR #1793, opened 2026-09-19
feat(analysis): causal-swap benchmark for coordinate_patch_hooks

janmenjayap:
---

### Summary

Adds the causal coordinate-swap benchmark for `JacobianLens.coordinate_patch_hooks(...)` (#1749):
corpus/metric primitives, a baseline-capability filter, a norm-matched control-token selector, a
per-trial runner that shares one `decomposition_cache` between the real and control conditions, a
bootstrap-CI / fingerprinted-artifact schema, a generation entry point, a checked-in frozen artifact
against GPT-2-small, one cached-model integration test, and a new demo notebook that only reads the
frozen artifact. The library layers are model-free or tested against the existing `_ToyBridge`
fixture; only the integration test and artifact generation touch a real model, and GPT-2-small needs
no `HF_TOKEN`.

Implements #1776. `Jacobian_Lens_Demo.ipynb` is untouched.

---

### Motivation

#1749 shipped `coordinate_patch_hooks` but explicitly deferred measuring whether it causes a
directional change in model output, under controls the codebase doesn't otherwise have for this
mechanism: baseline capability filtering, a norm-matched random-atom control, and bootstrap
uncertainty on every reported rate. #1776 is the tracking issue for that deferred work (named "PR3"
in both #1749's own "Out of scope" section and this repo's roadmap notes); this PR ships it.

---

### What ships

New module `transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py`, six commits,
layered bottom-up:

1. **`feat(analysis): causal-swap benchmark corpus and answer metrics`** — `FunctionSpec` /
   `BenchmarkCorpus` / `PromptTrialSpec` / `iter_prompt_trials`, and `AnswerMetrics` /
   `compute_answer_metrics` (rank/margin/tie scoring against a target token, ported from
   `Jacobian_Lens_Demo.ipynb`'s existing `_target_metrics` cell as tested library code).
2. **`feat(analysis): baseline-capability filter`** — `BaselineRecord` / `filter_baseline_capable`,
   splitting prompts into capable/excluded by whether the source's own answer is already the
   deterministic argmax, so a benchmark trial is never run against a prompt the model can't answer to
   begin with.
3. **`feat(analysis): norm-matched control-token selection`** — `select_norm_matched_control_token`,
   picking a token whose `lens_vector_dictionary` atom norm is within a relative tolerance of the real
   target's, deterministically given a seed, so a trial can distinguish "swapping toward this concept
   mattered" from "any similarly sized coordinate edit would have moved the logits."
4. **`feat(analysis): coordinate-patch causal-swap trial runner`** — `TrialResult` /
   `run_causal_swap_trial` / `run_causal_swap_benchmark`, which calls `coordinate_patch_hooks` for the
   real and control conditions against a **shared** `decomposition_cache`, and catches `ValueError`
   per condition (recording `status="skipped_source_inactive"`) rather than aborting the sweep or
   touching `coordinate_patch_hooks`'s own fail-fast contract.
5. **`feat(analysis): bootstrap CI and frozen artifact schema`** — `BootstrapResult` /
   `bootstrap_success_rate_ci` (percentile bootstrap, seeded), and `build_protocol_manifest` /
   `fingerprint_manifest` / `serialize_artifact` / `load_artifact` (the same manifest-plus-SHA256
   recipe `Jacobian_Lens_Demo.ipynb` already uses, with `load_artifact` re-verifying the fingerprint on
   read so a hand-edited artifact is rejected rather than silently trusted).
6. **`docs(notebook): causal-swap benchmark for coordinate_patch_hooks`** — the generation entry point
   (`python -m transformer_lens.tools.analysis.jacobian_lens_causal_swap_benchmark`, no `HF_TOKEN`
   required), the checked-in frozen artifact
   (`demos/data/jacobian_lens_causal_swap_benchmark_gpt2.json`), the read-only demo notebook
   (`demos/Jacobian_Lens_Coordinate_Patch_Benchmark_Demo.ipynb`), the docs subsection in
   `jacobian_lens_fitting.md`, notebook registration (`docs/make_docs.py`, `docs/source/index.md`,
   `.github/workflows/checks.yml`), and the public-symbol exports in
   `transformer_lens/tools/analysis/__init__.py` (`BenchmarkCorpus`, `FunctionSpec`,
   `run_causal_swap_benchmark`, `bootstrap_success_rate_ci`, `load_artifact`).

---

### The frozen artifact

`demos/data/jacobian_lens_causal_swap_benchmark_gpt2.json` (schema version 1) is generated against the
published GPT-2-small Jacobian lens (`neuronpedia/jacobian-lens`,
`gpt2-small/jlens/Salesforce-wikitext/gpt2_jacobian_lens.pt`, revision `a4114d77`) over the
`countries` corpus and layers 0-10, with `alpha=1.0`, `k=25`, `control_tolerance=0.1`,
`control_seed=0`. It records 198 trials and 10 baseline-excluded prompts. On this model/lens/corpus
the current run reports a **null effect**: both the real-swap and the norm-matched control success
rates are 0.0, each with a 95% percentile-bootstrap CI of [0.0, 0.0] over 10,000 resamples. That is a
legitimate, honestly reported outcome, not a bug: the benchmark's contribution is the controlled
methodology and the uncertainty on every rate, not a headline success number. The notebook renders
the artifact and never calls the model.

---

### Design decisions (from the tracking issue #1776)

- **Single layer per trial, swept as a trial dimension, never a fixed multi-layer band** —
  `coordinate_patch_hooks` re-decomposes the live activation at each hooked layer, and stacking layers
  is order-dependent and can raise (confirmed on GPT-2: `[6,7]`/`[6,9]`/`[6,10]` raise, `[6,8]`
  survives). Every hook installation uses exactly one layer and `positions=[-1]`.
- **Harness-level skip-and-record fail policy** — the trial runner's own `try/except ValueError`,
  not a change to `coordinate_patch_hooks`'s existing fail-fast contract.
- **Shared `decomposition_cache` between the real and control conditions** — both conditions patch the
  same pre-hook activation at the same `(layer, batch, position)`; only the first call performs the
  vocabulary-scale scan, verified by an explicit test.
- **GPT-2-small as the reference model** for the frozen artifact and integration test, to avoid the
  `HF_TOKEN` gate the gemma family needs.

---

### Non-goals

A successful swap in the artifact would show a directional causal effect under the stated controls on
GPT-2-small, not proof of unique causal mediation, exhaustive concept coverage, or a result that
transfers to closed-weight models. No new prompt corpora beyond the reused country/function set, no
multi-layer band protocol, and no change to `Jacobian_Lens_Demo.ipynb`.

---

### Testing

- `tests/unit/tools/test_jacobian_lens_causal_swap_benchmark.py` — model-free: prompt-trial
  enumeration, answer-metric rank/margin/tie arithmetic and its validation errors, baseline-capability
  splitting, norm-matched control-token selection (determinism, tolerance, exclusion, self-exclusion),
  bootstrap CI bounds/determinism/empty-input rejection, manifest fingerprinting, and
  serialize/load round-trip including a tampered-fingerprint rejection.
- `tests/unit/tools/test_jacobian_lens_causal_swap_benchmark_trials.py` — `_ToyBridge`-backed: shared
  `decomposition_cache` reuse between real/control conditions, skip-and-record on an inactive source,
  baseline-incapable prompts excluded from the trial set, and seeded determinism of a full benchmark
  run.
- `tests/integration/test_jacobian_lens_causal_swap_benchmark.py::test_causal_swap_benchmark_gpt2_smoke`
  — cached GPT-2-small, structural assertions only (no specific success-rate claim), including that
  each surviving trial's control token is norm-matched to its target within tolerance.

Local gates verified:
- `uv run pytest tests/unit/tools/test_jacobian_lens_causal_swap_benchmark.py tests/unit/tools/test_jacobian_lens_causal_swap_benchmark_trials.py -q` — 27 passed.
- `uv run mypy .` — `Success: no issues found in 399 source files`.

---

### Related work

- **#1776 (open, tracking issue)** — the proposal this PR implements.
- **#1749 (merged)** — `coordinate_patch_hooks`, the mechanism under benchmark here; named this
  benchmark as its own deferred "PR3."
- **#1741 (merged)** — the offline `coordinate_patch(...)` primitive `coordinate_patch_hooks` wraps;
  no overlap.
- **#1747 (merged) / #1746 (closed)** — `swap_hooks`-specific clean-coordinate clamping; different
  mechanism, no functional conflict with this PR.

---

### Checklist

- [x] `uv run mypy .` clean.
- [x] Unit tests (`test_jacobian_lens_causal_swap_benchmark.py` + `_trials.py`) pass.
- [ ] `source .env && uv run pytest tests/integration/test_jacobian_lens_causal_swap_benchmark.py` run.
- [ ] `uv run build-docs` and nbval on `Jacobian_Lens_Coordinate_Patch_Benchmark_Demo.ipynb` run.
- [ ] `make test-pr` run (unit + docstring + acceptance + integration).
- [x] Maintainer sign-off on single-layer-per-trial scope, skip-and-record fail policy, and
      GPT-2-small model/lens choice (carried over from #1776).

---


transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+    if len(successes) == 0:
+        raise ValueError("successes must be a non-empty sequence")
+    values = np.asarray([bool(success) for success in successes], dtype=np.float64)
+    n = values.shape[0]
+    point_estimate = float(values.mean())
+    rng = np.random.default_rng(seed)
+    resample_indices = rng.integers(0, n, size=(n_resamples, n))
+    resample_rates = values[resample_indices].mean(axis=1)
+    tail = (1.0 - confidence) / 2.0
+    ci_low = float(np.quantile(resample_rates, tail))
```
jlarson4, line 457: A percentile bootstrap on an all-failure sample returns `[0.0, 0.0]` by construction, since every resample of an all-zero vector is all zeros, so the shipped artifact reports the same interval for its six executed trials that it would for one or for a thousand, where the exact 95% upper bound for 0 of 6 is 0.46. The block also carries `n_resamples: 10000` and no sample size, so the notebook's `0% [0%, 0%]` row, its error bars and the artifact all assert a certainty the data does not support. #1776 specified the seeded percentile bootstrap, which I approved, but looking at the data I think we need to adjust. 

Please move the interval to Clopper-Pearson, carry the executed-trial count on `BootstrapResult` into the serialized block, and add a test pinning the 0-of-6 upper bound at 0.459 so that deleting the interval math fails.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+        and bool(within_tolerance[token_id])
+    ]
+    if not candidates:
+        raise ValueError(
+            "no candidate token within relative tolerance "
+            f"{tolerance} of target_token_id={target_token_id}'s atom norm "
+            f"({float(target_norm):.4f})"
+        )
+    generator = torch.Generator(device=dictionary.device).manual_seed(seed)
+    pick = int(torch.randint(len(candidates), (1,), generator=generator).item())
```
jlarson4, line 206: The generator is built on the dictionary's device, but `torch.randint` allocates its output on CPU, and torch requires the two to match. On a CUDA or MPS model the selector raises before the `try` that records skips and takes the whole sweep down on its first trial; `main()` selects CUDA whenever one is present, so the documented regeneration command cannot run on a GPU machine. Drawing the index on the dictionary's device instead would not do because the same seed then selects a different control token per device and the artifact stops being reproducible. 

Please build the generator with `device="cpu"`, with a case on a non-CPU dictionary asserting the same seed selects the same token as on CPU.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+    token. Raises ``ValueError`` if no candidate qualifies; the tolerance is never silently
+    widened and selection never falls back to the globally nearest atom.
+    """
+    if dictionary.ndim != 2:
+        raise ValueError(
+            f"dictionary must be 2-D [num_atoms, d_model], got shape {tuple(dictionary.shape)}"
+        )
+    atom_norms = dictionary.float().norm(dim=1)
+    target_norm = atom_norms[target_token_id]
+    within_tolerance = (atom_norms - target_norm).abs() <= tolerance * target_norm
```
jlarson4, line 191: The perturbation is `c_src · (a_target − a_source)`, so its size is set by `||a_target − a_source||`; matching atom norms leaves that free, and across the shipped pools the control edit runs from half to twice the real one, which breaks the same-magnitude contract stated at line 257.

Please match on `||a_c − a_source|| against `||a_target − a_source||`, exclude atoms already in the active support, and record an empty pool as its own skip status instead of raising.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+    model = TransformerBridge.boot_transformers("gpt2", dtype=torch.float32, device=device)
+    lens = JacobianLens.from_pretrained(
+        _GPT2_LENS_REPO,
+        filename=_GPT2_LENS_FILE,
+        revision=_GPT2_LENS_REVISION,
+        model=model,
+    )
+    layers = list(lens.source_layers)
```
jlarson4, line 648: #1776 asks for the swept layers to be chosen for a non-degenerate trial set and recorded as that choice. This records all eleven, nine of which contribute 162 trials that can never execute, and the notebook shows the executed count but never the skipped one. Please record the layers that execute and show the skipped count beside the executed count wherever the rates appear.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+    def _condition_metrics(condition_target_id: int) -> AnswerMetrics:
+        hooks = lens.coordinate_patch_hooks(
+            model,
+            source_id,
+            condition_target_id,
+            layers=[layer],
+            positions=[-1],
+            decomposition_cache=decomposition_cache,
+            k=k,
+            alpha=alpha,
```
jlarson4, line 312: Hardcoding `alpha=0.0` here leaves every test green, and so does swapping the real and control assignments; a comparison against the trial's baseline would catch neither, since the baseline is scored on the source answer. 

Please add a toy-fixture case that installs `coordinate_patch_hooks` itself and asserts each condition's metrics equal that independently patched pass.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+        layers=layers,
+        alpha=args.alpha,
+        k=args.k,
+        control_tolerance=args.control_tolerance,
+        control_seed=args.control_seed,
+        success_definition="target token id equals deterministic argmax token id",
+        baseline_definition="source answer token id equals deterministic argmax token id",
+        rank_definition="1 + count(logits strictly greater than target logit)",
+    )
+    real_ci = bootstrap_success_rate_ci(real_successes)
```
jlarson4, line 688: All six executed trials are conditions on one prompt, so pooling them as independent draws understates the uncertainty about twofold even once the interval is fixed. Please report the rate per prompt, or state the number of independent prompts beside the pooled rate.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+            alpha=alpha,
+        )
+        with model.hooks(fwd_hooks=hooks), torch.no_grad():
+            condition_logits = model(tokens)[0, -1].float()
+        return compute_answer_metrics(condition_logits, target_answer_id)
+
+    try:
+        real_metrics = _condition_metrics(target_id)
+        control_metrics = _condition_metrics(control_token_id)
+    except ValueError as exc:
```
jlarson4, line 321: This also catches unrelated `ValueError`s (a same-id source and target, an out-of-vocabulary answer, non-finite logits) and files them as `skipped_source_inactive`, which `main()` then drops from both denominators, while an unfitted layer escapes uncaught from line 294. The raising modules define no subclass to narrow on. 

Please set the skip status from an up-front check that the source is in the decomposition's support, seeded into the shared cache under the normalized position, and let other errors surface under their own status.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+)
+
+
+def load_artifact(path: Path) -> Dict[str, Any]:
+    """Reads and validates a frozen artifact produced by :func:`serialize_artifact`.
+
+    Validates that every required top-level key is present, that ``schema_version``
+    matches :data:`SCHEMA_VERSION`, and that ``protocol_fingerprint`` matches a fresh
+    :func:`fingerprint_manifest` of the loaded ``protocol_manifest`` -- catching a
+    hand-edited or corrupted artifact rather than trusting the stored fingerprint blindly.
```
jlarson4, line 548: Only `protocol_manifest` is hashed, so an artifact with rewritten trials or CI blocks loads clean and prints the same fingerprint the notebook shows as provenance. 

Please have `load_artifact` also verify a hash over the trial, exclusion and CI blocks, stored beside `protocol_fingerprint` so the existing recipe and value stay as they are.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+    )
+
+
+_REQUIRED_MANIFEST_FIELDS = (
+    "model_id",
+    "model_revision",
+    "lens_repo",
+    "lens_file",
+    "lens_revision",
+    "corpus_name",
```
jlarson4, line 474: `corpus_name` is a bare string, so the prompts and answers that decide capability and scoring never enter the fingerprint; the precedent notebook embeds the config with its repo, path and revision, which #1776 asks to pin. 

Please add the corpus definition and that triple to the manifest, regenerating the artifact and re-executing the notebook.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+) -> Tuple[List[BaselineRecord], List[BaselineRecord]]:
+    """Splits baseline records into (capable, excluded) prompts.
+
+    A prompt is baseline-capable when the model's own deterministic argmax already matches
+    the source's answer (``metrics.target_is_top1``); only such prompts are eligible for
+    later intervention trials, so an edit's effect is never measured against a prompt the
+    unperturbed model already gets wrong. Order-preserving in both outputs; never mutates
+    ``baselines``.
+    """
+    capable = [record for record in baselines if record.metrics.target_is_top1]
```
jlarson4, line 163: On an exact top tie `argmax` returns the lowest index, so token-id order decides whether a tied prompt enters the trial set, and the `target_tied_for_top` flag computed for this case is read nowhere. Please have the filter treat a tied-for-top baseline as not capable.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+        baseline_logits = model(tokens)[0, -1].float()
+    baseline_metrics = compute_answer_metrics(baseline_logits, source_answer_id)
+
+    dictionary = lens.lens_vector_dictionary(model, layer)
+    control_token_id = select_norm_matched_control_token(
+        dictionary,
+        target_id,
+        excluded_ids={source_id, source_answer_id, target_answer_id},
+        tolerance=control_tolerance,
+        seed=control_seed,
```
jlarson4, line 300: One seeded control draw per trial leaves the control arm's own variance unmeasured; at layer 10 Egypt&#8209;>France, six of 150 seeded draws beat the real condition's margin, so a reader cannot tell the shipped gap from draw variation. Please draw each trial's control under several seeds and carry the spread in the artifact.

transformer_lens/tools/analysis/jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,701 @@
...
+    / "jacobian_lens_causal_swap_benchmark_gpt2.json"
+)
+
+
+def main(argv: Optional[Sequence[str]] = None) -> None:
+    """Generates the frozen GPT-2 causal-swap benchmark artifact.
+
+    Reproduce with (no ``HF_TOKEN`` needed -- GPT-2 is not gated)::
+
+        uv run python -m transformer_lens.tools.analysis.jacobian_lens_causal_swap_benchmark
```
jlarson4, line 619: The package `__init__` imports this module, so the documented `python -m` form executes it twice and runpy warns; the `__main__` guard is already here, and dropping the re-export would break the notebook's imports. Please point the documented command at a small CLI module that the package `__init__` does not import.

tests/integration/test_jacobian_lens_causal_swap_benchmark.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+    )
+    trials, excluded = run_causal_swap_benchmark(
+        published_gpt2_lens,
+        gpt2_bridge,
+        corpus,
+        layers=[6],
+        control_seed=0,
+    )
+    assert len(trials) + len(excluded) >= 1
```
jlarson4, line 66: With this corpus and layer nothing executes: `ok_trials` is empty, the norm-matching loop never runs, and four selector breakages pass the suite. In the shipped corpus only the continent template with source Egypt executes, at layers 9 and 10. Please use that fixture, Egypt and France at layer 9, and assert a trial reaches `status="ok"`.

demos/Jacobian_Lens_Coordinate_Patch_Benchmark_Demo.ipynb
```diff
@@ -0,0 +1,639 @@
...
+   "source": [
+    "from pathlib import Path\n",
+    "\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "\n",
+    "import transformer_lens\n",
+    "from transformer_lens.tools.analysis import bootstrap_success_rate_ci, load_artifact\n",
+    "\n",
+    "REPO_ROOT = Path(transformer_lens.__file__).resolve().parents[1]\n",
```
jlarson4, line 123: `parents[1]` is the repo only under a checkout install; the wheel ships no `demos/`, so the Colab badge's `%pip install` path ends in a bare `FileNotFoundError`. Please state that the notebook needs a checkout, fail with that message when the artifact is absent, and drop the badge.

jlarson4 (commented):
This is a very well thought out benchmark system. I do have a few structural requests below, let me know if you have any questions.

PR #1798, opened 2026-09-22
docs: align analysis tools guide with v4 capabilities

emerardd:
# Description

Update the analysis-tools overview to match the current v4 APIs. The guide no longer suggests using removed HookedTransformer classes for DLA or direct path patching, describes the supported edge-granularity attribution workflow and its `ig_steps=1` limit, and broadens Backward Lens guidance to the dense-MLP decoder-only Bridge models now supported by the implementation.

This is documentation-only and does not change runtime behavior. It does not overlap the hook-alias documentation work in #1797.

## Type of change

## Validation

- [x] The updated page builds with a scoped Sphinx HTML build.
- [x] `git diff --check` passes.
- [x] `uv run --frozen mypy .` passes (383 source files).
- [ ] The full `uv run --frozen build-docs` build passes locally. It is blocked on this Windows machine by a missing Pandoc executable while converting `generated/demos/Exploratory_Analysis_Demo.ipynb`; the failure is outside this page. The docs CI job will run because the branch name contains `docs`.

## Checklist

No code or tests were changed, so the code-comment and test checkboxes are not applicable.


docs/source/content/analysis_tools.md
```diff
@@ -33,10 +33,10 @@ available hooks, and the meaning of the selected tensor axes.
...
+| DLA | Targets TransformerBridge in 4.0. Enable compatibility mode and use the standard attention/MLP residual decomposition; Mamba/SSM/Mixer/LinearAttention hybrid layouts are rejected. | One cached forward pass, or reuse of a suitable cache. Head decomposition can require additional per-head results. |
+| Attribution Patching | Targets TransformerBridge. Clean/corrupt inputs must be `[batch, seq]` token tensors with matching shapes and aligned positions. The required embedding, head, and MLP hook aliases must exist; edge granularity temporarily enables additional writer and reader hooks, then restores the caller's hook flags. Run the model and all nested stochastic modules in evaluation mode. The metric must return a differentiable scalar, and the current backward-cache path requires an active autograd graph; fully frozen models are not yet supported. | Two forwards and one backward **per prompt pair**; activation and gradient caches. Edge sweeps capture extra per-head tensors and use more memory. Pair scores are averaged across the batch. |
```
jlarson4: **For Attribution Patching, generally:**
The hook aliases this row names are present in both cases where an edge sweep actually fails. It also needs an attention bridge that can fork before Q/K/V, and `n_key_value_heads == n_heads`. **opt-125m** raises from `set_use_attn_result`, and **Qwen2.5-0.5B** dies with a bare `IndexError` that gives no hint the architecture is at fault. Could the row state that precondition, the way Direct Path Patching row states its GQA precondition?

**Specifically for the Edge sweeps section:**
The edge count grows with sequence length times layers x heads squared. Can we state that more explicitly? "more memory" is vague enough that it could be misconstrued.

docs/source/content/analysis_tools.md
```diff
@@ -33,10 +33,10 @@ available hooks, and the meaning of the selected tensor axes.
...
+| Direct Path Patching | Requires a TransformerBridge exposing the expected attention weights, Q/K/V hook aliases, source `hook_z`, and destination LayerNorm scales. Folded LayerNorm parameters improve the approximation. On GQA/MQA models, direct Q-path patching remains available, but `component="k"` and `"v"` currently require `n_key_value_heads == n_heads`; per-query-head K/V semantics are not yet defined. | Clean/corrupt caching, then repeated forwards over destination heads. Sweeping all source heads adds another sweep dimension. |
+| Backward Lens | Requires a raw dense-MLP decoder-only TransformerBridge without compatibility mode, such as GPT-2 or Pythia/GPT-NeoX. Use one target token and check the restrictions in the [tool guide](backward_lens.md). | Gradient computation plus vocabulary projections for selected layers and positions; retaining full logits increases memory. |
```
jlarson4: Backward Lens also requires a *non-gated* MLP. A Llama, Mistral or Qwen user satisfies "dense-MLP decoder-only", but gets a `NotImplementedError` at construction. Could the row carry the non-gated qualifier, as `backward_lens.md` does?

jlarson4 (commented):
Thanks for updating these docs, I missed them in my pre-release sweep. A couple small comments below

docs/source/content/analysis_tools.md
emerardd: Thanks for catching both gaps. I updated the Attribution Patching row to state the edge-specific requirements: per-head results, a pre-Q/K/V residual fork, and `n_key_value_heads == n_heads` (so GQA/MQA is currently unsupported). It also now distinguishes the `O(seq_len × (n_layers × n_heads)²)` edge-score count from the additional `[batch, seq, n_heads, d_model]` tensors held in the cache. This is in b1f0b4c8.

docs/source/content/analysis_tools.md
emerardd: Good point. I added the non-gated requirement to the Backward Lens row in b1f0b4c8, matching the detailed tool guide and its construction check.

jlarson4:
Looks great @emerardd, as long as it passes CI I will get this merged

PR #1800, opened 2026-09-22
Remove dead gemma2/gemma3 eager_attention_forward monkey-patch

### LINKED ISSUE #1795: [Bug Report] `PositionEmbeddingsAttentionBridge` replaces `eager_attention_forward` in transformers' gemma-2 and gemma-3 modules for the rest of the process

MdSadiqMd:
# Description

`PositionEmbeddingsAttentionBridge` monkey-patched transformers' module-level `eager_attention_forward` for `gemma2` and `gemma3` the first time any such bridge was constructed, to fire `hook_rot_q`/`hook_rot_k`. The wrapper was installed once per process, never removed (deleting the bridge did not restore it), and carried no `functools.wraps` metadata.

This is dead code. `_setup_eager_attention_hook_wrapper` was added so those hooks would fire when HF's eager attention ran, but the bridge's `forward` now reimplements attention and fires `hook_rot_q`/`hook_rot_k` itself — it never calls the HF attention module, so HF's eager path never executes for a bridged module.

The patch also leaked outside TransformerLens. The component backs ~25 architectures (llama, qwen2/3, gemma 1/2/3, olmo, phi, mixtral, ...), so building a bridge for *any* of them replaced gemma's `eager_attention_forward` process-wide. Consequences:

- Anything inspecting the function bound at gemma's attention call site saw TransformerLens's wrapper instead of transformers' function. nnsight's recursive `.source` tracing (used by circuit-tracer to locate the attention pattern for gemma models) parses that source; the wrapper has no `nn.functional.dropout(...)` call, so the expected node vanished and circuit-tracer's nnsight backend failed with `AttributeError: 'SourceEnvoy' object has no attribute 'nn_functional_dropout_0'` — but only in processes where a bridge was built first, so it looked like a flaky test.
- The wrapper ran for every gemma-2/3 forward in the process, bridged or not (passing through by module id), so plain HF gemma models paid a lookup on every attention call after any bridge had existed.

This PR deletes the patch entirely: `_setup_eager_attention_hook_wrapper`, its module globals (`_ATTENTION_BRIDGE_REGISTRY`, `_EAGER_ATTENTION_WRAPPED`, `_ORIGINAL_EAGER_ATTENTION_FORWARD`), the two `set_original_component` calls, and the now-unused `weakref` / `Callable` / `gemma2_module` imports. Rotary hooks still fire from the bridge's own `forward`, and building a bridge no longer changes what other libraries see in transformers' gemma modules.

A network-free guard test builds a tiny Gemma3 bridge from `Gemma3TextConfig`, runs `run_with_cache`, and asserts (1) `blocks.0.attn.hook_rot_q` is in the cache and (2) `eager_attention_forward.__module__` on both `modeling_gemma2` and `modeling_gemma3` still starts with `transformers.`. It fails on current `dev` and passes with this change.

Fixes https://github.com/TransformerLensOrg/TransformerLens/issues/1795

## Type of change

### Screenshots
N/A

# Checklist:


MdSadiqMd:
Hey @jlarson4 i based the branch on `dev` if it's not the right approach let me know, will base it on `main`

On a side note, the pipeline failing is not because of my change to the best of my knowledge, please have a look at it once, and if it creates a new issue, feel free to assign it to me

jlarson4:
Hi @MdSadiqMd! Your PR is failing a CI check it is not meant to run against. MPS Checks on runs on release PRs to `main`, PRs to `dev` do not run that check. I switched your base to `dev`, and am attempting to clear the failed check now. Full review of the PR coming within the next couple hours

transformer_lens/model_bridge/generalized_components/position_embeddings_attention.py
```diff
@@ -9,11 +9,9 @@
 """
 from __future__ import annotations
 
-import weakref
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Dict, Optional
 
 import torch
-import transformers.models.gemma2.modeling_gemma2 as gemma2_module
```
jlarson4, line 16: This import had an interesting, unintended side effect: it forced transformers' deep import early enough that `model_bridge/sources/transformers/helpers.py` binds the same `transformers` module object that ends up in `sys.modules`. transformers rebinds that entry to a second lazy module on its first deep import, so without this line `helpers.py` holds the stale one and `tests/unit/model_bridge/sources/test_modality_helpers.py` fails 10 of its 17 tests. The `monkeypatch.setattr(f"transformers.{class_name}", ...)` at line 45 no longer reaches the object `helpers.py` reads. 

This error only reproduces on macOS, which is why only the MPS Checks CI run caught it. Rather than keeping the import, please point that patch site at the module the consumer actually holds.

tests/unit/model_bridge/generalized_components/test_gemma_eager_no_patch.py
```diff
@@ -0,0 +1,62 @@
...
+        hf, "Gemma3ForCausalLM", hf_config=copy.deepcopy(cfg), tokenizer=None, device="cpu"
+    ).eval()
+
+
+def test_rotary_hook_fires_without_patching_hf_eager_attention():
+    bridge = _tiny_gemma3_bridge()
+    tokens = torch.randint(3, 200, (1, 6))
+    _, cache = bridge.run_with_cache(tokens)
+
+    assert "blocks.0.attn.hook_rot_q" in cache
```
jlarson4, line 67: The wrapper being deleted fired both `hook_rot_q` and `hook_rot_k`, but only `hook_rot_q` is asserted here. If you suppressed the `hook_rot_k` call in the bridge's forward, this test would still pass. Can you please assert both?

tests/unit/model_bridge/generalized_components/test_gemma_eager_no_patch.py
```diff
@@ -0,0 +1,62 @@
...
+
+    import transformers.models.gemma2.modeling_gemma2 as gemma2_module
+    import transformers.models.gemma3.modeling_gemma3 as gemma3_module
+
+    assert gemma2_module.eager_attention_forward.__module__.startswith("transformers.")
```
jlarson4: Reinstating the deleted patch with `functools.wraps` on the wrapper leaves both of these assertions green while the monkey-patch is fully live and installed on both gemma modules, so the guard passes. Please assert on something a wrapper cannot carry over, such as `__code__.co_filename` or identity against transformers' own function.

jlarson4 (commented):
Full review below! I found the source of the CI errors caught by the MPS Check, and have spelled those out below, the are related to an interesting side effect of the removed code.

MdSadiqMd:
hey @jlarson4 did the fixes and pipeline passed, please have a look
feel free to let me know if there is anything required from my side

jlarson4:
Looks great, thank you! I am merging now

